PR #16, opened 2017-06-15
Avoid unwanted spaces

didierfranc:
A little change to avoid that 
screen shot 2017-06-15 at 16 54 39


didierfranc:
Yes good idea ! I've just copied what I've done  https://github.com/didierfranc/react-sc/blob/master/src/Styled.js#L10-L14

PR #20, opened 2017-06-23
Add tests for fragment and other misc changes

mitchellhamilton:
This is adds some tests for `fragment`(all the babel stuff is at ❤️ 😍 🎉 100%🎉😍 ❤️  coverage) and a few other things...

- ~~make the function in the `styled` call anonymous~~
- make the function in the `fragment` call ~~an anonymous function~~ instead of an arrow function so it's the same as `styled`
- remove the call to `styled` if there are no styles(technically it's only if the hash is 0 so if there's whitespace the call to `styled` won't be removed(would it be a good idea to remove whitespace before hashing?))
- run prettier on test, example and jest-utils
- there were two dashes when fragments didn't have a name(e.g.`frag--somehash`), this fixes that
- [x] Make `inline` and `fragment` the same function [(this broke a test)](https://github.com/tkh44/emotion/pull/20/files#r123715058)

[Should all classes have the `css-` prefix?](https://github.com/threepointone/glam/blob/master/src/babel.js#L75)

Also, now this doesn't convert calls to `css`, was that intentional or is it cool if I add that?

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -92,7 +92,7 @@ export default function (babel) {
           let arrayValues = parseDynamicValues(rules, t)
 
           const inlineContentExpr = t.functionExpression(
-            t.identifier(''),
+            t.identifier('createEmotionStyledRules'),
             built.expressions.map((x, i) => t.identifier(`x${i}`)),
             t.blockStatement([
               t.returnStatement(t.arrayExpression(arrayValues))
@@ -138,10 +138,13 @@ export default function (babel) {
           t.isIdentifier(path.node.tag) &&
           path.node.tag.name === 'fragment'
         ) {
+          const parent = path.findParent(p => p.isVariableDeclarator())
+          const identifierName = parent && t.isIdentifier(parent.node.id)
+            ? parent.node.id.name
+            : ''
           const { hash, name, rules } = inline(
             path.node.quasi,
-            undefined,
-            'frag',
+            identifierName,
             'frag'
           )
           path.replaceWith(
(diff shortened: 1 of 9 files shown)


test/styled.test.js
```diff
@@ -137,7 +137,7 @@ describe('styled', () => {
     expect(tree).toMatchSnapshotWithEmotion()
   })
 
-  test('fragments', () => {
+  test.skip('fragments', () => {
```
mitchellhamilton: I skipped this test because it's output was this(note the function in the css)

```css
.H2-1t8i2zo-1ci3vp8 {
  font-size: 40;
}

.H1-1qd8mfb-n3glby {
  font-size: 20px;
  height: ({ height }) => heightpx;
  color: ({ color }) => color;
}
```
```jsx
<h1
  className="legacy__class H2-1t8i2zo-1ci3vp8 H1-1qd8mfb-n3glby"
  scale={2}
>
  hello world
</h1>
```
I also broke it by changing some stuff but it shouldn't have worked in the first place.(I love the syntax highlighting)

src/babel.js
```diff
@@ -82,13 +82,18 @@ export default function (babel) {
...
+          if (hash === '0') {
+            return tag
+          }
 
           let arrayValues = parseDynamicValues(rules, t)
 
           const inlineContentExpr = t.functionExpression(
-            t.identifier('inlineCss'),
-            stubs.map((x, i) => t.identifier(`x${i}`)),
+            t.identifier(''),
```
tkh44: Please put this back. If something bombs in here we want to have a proper call stack. Granted it needs a better name like `createEmotionRules`

test/babel.test.js
```diff
@@ -7,30 +7,52 @@ describe('emotion/babel', () => {
...
-      const {code} = babel.transform(basic, {
+      const { code } = babel.transform(basic, {
         plugins: [plugin]
       })
       expect(code).toMatchSnapshot()
     })
 
     test('basic', () => {
-      const basic = 'styled.h1`font-size: ${fontSize + \'px\'};`'
+      const basic = "const H1 = styled.h1`font-size: ${fontSize + 'px'};`"
```
tkh44: Wow! good catch on adding these.

test/styled.test.js
tkh44: Yeah, I knew these were failing but we are going to rework fragments so I decided to let them go.

threepointone:
yeah all classes should have the same prefix, this will make SSR hydration simpler when we get to it. 

please add back the code for css calls :)

tkh44:
@threepointone totally my fault `css` isn't being transformed. I think @mitchellhamilton can just copy and paste the `fragment` code and add another branch to the if/else to transform this. Especially since he has generalized the inline function.

test/babel.test.js
mitchellhamilton: That was all Prettier 😁

PR #23, opened 2017-06-26
Add static extraction

mitchellhamilton:
I based this on the keyframes branch since it's where you've been committing.

This goes through the PostCSS AST and turns it into two sets of rules, one set of rules that's static and can be put into a css file and another set of rules that's dynamic and must be in js(this is really inefficient right now but can be improved later). It then writes all the static rules in a file to `filename.css`(just like glam) and puts the dynamic rules in the js just like before.

Currently this adds an import declaration for the css, I think this is a good default(there isn't an option to disable it yet but it would be trivial to add) so that users don't have to deal with the underlying css files and remembering to add imports(and remove imports since if there is static css in a file, then there isn't static css the css file will still be there) as long as they have the appropriate loader/some method for importing css

**Currently this only extracts static css in `styled` and `css`.**

I made this opt-out with an extract option that if it's set to false will disable it(it is also disabled if the filename is unknown so it works fine in the example site), I think this is needed for tests because without it `jest-utils`(And hopefully `jest-glamor-react` soon!) won't take into account static css.

Tests
- For the non babel tests I set extract to false so `.toMatchSnapshotWithEmotion` would work with static css
- For the babel tests, most of them have no filename set so they don't extract css but I added some that do and they use a mocked `fs`
- I copied some other tests I wrote for glam to test writing to the filesystem(with a mock)

Also random idea, maybe instead of the css filename being `filename.css` it should be `filenamewithoutextension.emotion.css` e.g. `main.emotion.css` vs `main.js.css`.

[unrelated to this PR but it seems that `lib/` isn't published in emotion 3](https://unpkg.com/emotion@3.0.1/)


tkh44:
Sorry about me commiting to that random branch. Master is up to date now.

tkh44:
> unrelated to this PR but it seems that `lib/` isn't published in emotion 3

Fixed in `v3.0.2`

tkh44:
> Also random idea, maybe instead of the css filename being `filename.css` it should be `filenamewithoutextension.emotion.css` e.g. `main.emotion.css` vs `main.js.css`.

This is a great idea.

mitchellhamilton:
What I've done:
- the name of css files are now `filenamewithoutextension.emotion.css` e.g. `main.emotion.css` for a file named `main.js`
- css is only extracted if there are no `@apply`s in it and css vars are used for dynamic bits
- the extract option is removed and is replaced with an inline option which when true causes the babel plugin to not extract anything

src/babel.js
```diff
@@ -63,7 +65,41 @@ export default function (babel) {
     name: 'emotion', // not required
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
-      TaggedTemplateExpression (path) {
+      Program: {
+        enter (path, state) {
+          state.inline =
+            path.hub.file.opts.filename === 'unknown' || state.opts.inline
```
tkh44: Does this mean it will automatically switch to inline mode in situations where there is no file?

mitchellhamilton: Yep😁

PR #28, opened 2017-06-27
Add styled.js to files in package.json

mitchellhamilton:
`styled.js` wasn't being published, this adds it to files in package.json.

Related to this, the module field in the package.json goes to src/index.js and since it has flow types it throws errors. We should probably run babel with modules false to generate an es folder or something like that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "babel.js",
     "src",
     "dist",
-    "lib"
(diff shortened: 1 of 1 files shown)


tkh44:
Ah yeah it should be pointing to `lib` which is created when you run `build`

PR #29, opened 2017-06-27
CSS Syntax highlighting with react-live

mitchellhamilton:
This is just a proof of concept that uses `react-live` with Prism.

~~It's not working perfectly, the main thing that's broken is that syntax highlighting gets messed up when you paste into it until you edit it manually.~~

I think I broke the actual CSS syntax highlighting.

~~DEMO: https://emotion-css-syntax-highlighting.netlify.com~~


example/src/playground/index.js
```diff
@@ -1,106 +1,92 @@
...
 `
 
-class ReactPlayground extends Component {
-  static defaultProps = {
-    theme: 'dracula',
-    noRender: true,
-    context: {},
-    initiallyExpanded: false
-  }
+const Content = styled(LiveProvider)`
```
tkh44: This is pretty cool isn't it 😀

tkh44:
Do you know how much lighter/faster this version would be than what we have now? 

Also see https://github.com/tkh44/emotion/pull/30/files

PR #31, opened 2017-06-28
Add SSR/Critical CSS for inline mode

mitchellhamilton:
This ~~isn't done yet~~ is done, I just copied glamor's implementation. The tests are really simple ~~and I haven't tested it outside of jest yet~~ but it's a start and it ~~should theoretically~~ does work.

This will only work with inline mode since there will be imports to css files and node won't know what to do with that.

~~This also fixes the `fontFace` test.~~ Reverted to avoid merge conflicts with #39


tkh44:
closes https://github.com/tkh44/emotion/issues/32

tkh44:
Can you add `server.js` to the main dir so that we can have a shortcut like `babel.js` and `styled.js`?

src/server.js
```diff
@@ -0,0 +1,51 @@
...
+
+export function renderStaticOptimized (fn) {
+  // parse out ids from html
+  // reconstruct css/rules/cache to pass
+  let html = fn()
+  if (html === undefined) {
+    throw new Error('did you forget to return from renderToString?')
+  }
+  let o = { html, ids: [], css: '', rules: [] }
+  let regex = /css-([a-zA-Z0-9]+)/gm
```
tkh44: I think we need to pull out `vars` too right?

mitchellhamilton: This is just for inline mode right now so no

mitchellhamilton:
UPDATE:
I tested it with next and it works 🎉  https://emotion-next.now.sh (once this PR is merged I'll submit a PR to next.js and add it as an example)
Everything has tests.
I added a `server.js` file to the root so people can just import `emotion/server`

PR #37, opened 2017-06-28
WIP: Add typescript typings

### LINKED ISSUE #40: Add TypeScript definitions
I'll do this when I have some free time (probably right after #EuroPython, I have a talk to prepare :) ).

Should be pretty straightforward to port styled-components' definitions.

What are the main API differences with it?

marvinhagemeister:
This PR adds typing so that the `styled` function will type-check when using typescript.

Still marked as `wip` because somehow the template tags are not recognized yet

Fixes #40

### DIFF
--- a/styled.d.ts
+++ b/styled.d.ts
@@ -1,14 +1,48 @@
-declare module "emotion/styled" {
-  import { Component } from "react";
+declare module "emotion" {
+  export type InputValue = string | number;
 
-  export type Styler = (
+  export function flush(): void;
+  export function css(
+    cls: string,
     content: TemplateStringsArray,
-    vars?: string[] | number[],
-  ) => Component;
+    ...vars: InputValue[],
+  ): string;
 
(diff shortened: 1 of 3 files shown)


styled.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module "emotion/styled" {
+  import { Component } from "react";
+
+  export type Styler = (
```
tkh44: This may need a name change. I plan on adding a `Style` component in the future and this might get confusing. If this is how styled-components looks though then I say leave it.

tkh44:
I'm not really familiar with typescript so I'm going to have to rely on external reviewers on this. I trust you, @patrick91, and others can work together on this 👍

A-gambit:
@marvinhagemeister Hey can you update your pull with the latest version of emotion?

tkh44:
I'm going to close this for #379

PR #39, opened 2017-06-28
Add injectGlobal and other stuff

mitchellhamilton:
#### `injectGlobal`
This adds `injectGlobal` which as the name suggests inserts global styles. It works like this:
```jsx
injectGlobal`
  body {
    margin: 0;
    padding: 0;
  }
`
```
It's basically the same as [`styled-components`' `injectGlobal`](https://www.styled-components.com/docs/api#injectglobal). Also, this allows people to use `@font-face`, so should we remove `fontFace`?(it doesn't really matter that much because `fontFace` is pretty small `export const fontFace = injectGlobal`)

##### Allow interpolation in:
- [x] `injectGlobal`
- [x] `fontFace`
- [x] `keyframes`

##### Statically extract if there are no vars or `@apply`s:
- [x] `injectGlobal`
- [x] `fontFace`
- [x] `keyframes`

##### Misc
- [x] split `babel.test.js` into multiple files(it's 550 lines)


tkh44:
released in v4.1.1

PR #44, opened 2017-06-28
Remove babel-plugin-syntax-jsx requirement

CompuIves:
**What**: Add `babel-plugin-syntax-jsx` as official dependency to `package.json`.

**Why**: This dependency is required by the babel plugin to work.

**How**: Adding the dependency to package.json and doing `npm install`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

I did a clean npm install with latest npm after adding the library to package.json, it seems like the diff doesn't only contain my addition.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,6 +40,20 @@
         }
       }
     },
+    "acorn-object-spread": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/acorn-object-spread/-/acorn-object-spread-1.0.0.tgz",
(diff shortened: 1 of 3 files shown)


tkh44:
You can actually just remove the require statement from the babel.js 😅 Much simpler fix.

Can you drop the package-lock.json from this as well?

CompuIves:
Hah! Even better, I removed the `require` statement.

tkh44:
It is leftover from the css prop parsing (which will go back in tonight)

PR #48, opened 2017-06-29
css prop

tkh44:
**What**:
Adding css prop

```
<p css={`color: red;`}>
  hello world
</p>
```

**Why**:
dynamic styling

**How**:
adding a babel plugin.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -3,6 +3,7 @@ import { basename } from 'path'
 import { touchSync } from 'touch'
 import { inline, keyframes, fontFace, injectGlobal } from './inline'
 import findAndReplaceAttrs from './attrs'
+import cssProps from './css-prop'
 
 function parseDynamicValues (rules, t) {
   return rules.map(rule => {
@@ -63,6 +64,7 @@ export default function (babel) {
 
   return {
     name: 'emotion', // not required
+    inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
       Program: {
         enter (path, state) {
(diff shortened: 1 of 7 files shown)


test/babel/css-prop.test.js
```diff
@@ -0,0 +1,126 @@
...
+fs.existsSync.mockReturnValue(true)
+
+describe('babel css prop', () => {
+  test('basic', () => {
+    const basic = '(<div className="a" css={`color: brown;`}></div>)'
+    const { code } = babel.transform(basic, {
+      plugins: [plugin],
+      filename: __filename,
+      babelrc: false
+    })
```
mitchellhamilton: You need to add this to test that the css files were written correctly. 
```js
expect(fs.writeFileSync).toHaveBeenCalledTimes(1)
expect(fs.writeFileSync.mock.calls[0][1]).toMatchSnapshot()
```
You need to increment the `toHaveBeenCalledTimes` by one each time and for this `fs.writeFileSync.mock.calls[0][1]` increment the first number(the 0)

src/babel.js
```diff
@@ -98,6 +100,9 @@ export default function (babel) {
           }
         }
       },
+      JSXOpeningElement (path, state) {
+        cssProps(path, t)
```
mitchellhamilton: This should probably be called with state too so you can insert static rules

PR #54, opened 2017-06-29
Restyles api

tkh44:
**What**:
Add the restyles api

```jsx
const Profile = props => (
  <Style
    css={`
      color: blue;
      font-size: ${props.fontSize}px;

      &:hover {
        color: green;
      }

      & profile {
        font-size: 20px;
      }
    `}
    render={cls => (
      <div className={cls}>
        This will be blue until hovered.
        <div className="profile">
          This font size will be 20px
        </div>
      </div>
    )}
  />
)
```

**Why**:
It will allow for very quick porting of css files to css-in-js

**How**:
Add the component

```jsx
export function Style ({children, css, render}) {
  let fn = typeof render === 'function'
    ? render
    : Array.isArray(children) ? children[0] : children
  return fn(magic(css))
}
```

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I am not sure about the components name. I like `Style` but it might be confusing for some. I'm open to ideas here.


tkh44:
cc @threepointone @mitchellhamilton

mitchellhamilton:
I'm wondering if this is really needed since we already have the css prop. For example the code could be expressed like this with the css prop

```jsx
const Profile = props => (
  <div css={`
    color: blue;
    font-size: ${props.fontSize}px;

    &:hover {
      color: green;
    }

    & profile {
      font-size: 20px;
    }
  `}>
    This will be blue until hovered.
    <div className="profile">
      This font size will be 20px
    </div>
  </div>
)

```

tkh44:
~In a project I'm working on now we use the css prop.... a lot. This can get out of control quickly. I like the declarative api for the fact that it encourages the use of child selectors and...~

Yeah you are right 🙂

PR #58, opened 2017-06-30
Add size badges to README and size check in CI

mitchellhamilton:
**What**:

This adds size badges to the README and a check in CI that `emotion/react`(I chose `emotion/react` since most people will likely use the react version) is less than 3Kb.

I also removed some unused devDependencies and removed the babel config that was specific to tests in the root `.babelrc`.

**Why**:

So we can keep track of the size of emotion when PRs are submitted and so people who want to use emotion can see how small it is.

**How**:

I made the rollup config build a umd build to `dist/DO-NOT-USE.min.js`. I also added the `bundlesize` module which checks if a file's gzip size is below a certain threshold and integrates with GitHub statuses.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

We have to set up the GitHub status integration.(We don't NEED to do this but it might be nice to show us in the GitHub UI)
https://www.npmjs.com/package/bundlesize#build-status

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,20 +9,5 @@
     "stage-0",
     "react",
     "flow"
-  ],
-  "env": {
-    "test": {
-      "presets": [
-        [
-          "env",
-          {
-            "loose": true
-          }
-        ],
-        "stage-2",
-        "react",
-        "flow"
-      ]
-    }
-  }
+  ]
 }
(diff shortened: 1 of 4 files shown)


tkh44:
giphy-6

siddharthkp:
❤️ 

@mitchellhamilton Where did you come to know about this?

@tkh44 The configuration is not complete. You won't be able to get build status.
Ref: https://www.npmjs.com/package/bundlesize#build-status

mitchellhamilton:
@siddharthkp Your PR to preact

tkh44:
I followed the guide. It didn't work but I didn't have time to look into it in detail.

siddharthkp:
@tkh44 no worries, I'll look into it

siddharthkp:
@tkh44 Got it, you don't have push (branch updates) enabled for travis. That's why the values for master were never set.

PR #67, opened 2017-07-01
Add composes property and remove fragment

### LINKED ISSUE #61: Add support for css modules composes: so we can support css modules
```jsx
import foo from './foo.css'

const H1 = styled('h1')`
   composes: ${foo.heading};
   color: blue;
`
```

mitchellhamilton:
**What**:

Removing fragment/`@apply` and adding the composes property.

The composes property lets you compose classes like this.
```jsx
const cls1 = css`font-size: 20px`
const cls2 = css`display: flex;`

  composes: ${cls1} ${cls2};
`
```

This is slightly different from how it was originally going to work, mainly it only allows classes to be interpolated **not** arrays of classes. This is so we can do optimisations, for example if there are no interpolations in a `css` call except for composes we can remove the `css` call completely so it's just a string.  

This also removes `renderStatic` and renames `renderStaticOptimized` to `extractCritical`.(I did this now so we could do it in one major release)

**Why**:

Closes #61 

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


src/utils.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+
+export const PLACEHOLDER_REGEX = /xxx(\S)xxx/gm
```
tkh44: This is the only thing I'm concerned about because patterns with a `g` modifier are stateful. I'm not 100% it matters for us here but I want to be sure.

See here (comments as well)
https://ponyfoo.com/articles/stateful-regular-expressions

tkh44:
> This is slightly different from how it was originally going to work, mainly it only allows classes to be interpolated **not** arrays of classes. This is so we can do optimisations, for example if there are no interpolations in a `css` call except for composes we can remove the `css` call completely so it's just a string.

Great observation.

src/utils.js
tkh44: @threepointone can you chime in on this. I'm not 100% and don't want to hold this up due to a hunch.

threepointone: Yeah just define it in place each time. It's simple enough. Alternately, export a function that returns the regex

The rest looks great!

PR #76, opened 2017-07-03
Support Dynamic Values in attr

### LINKED ISSUE #5: support dynamic values inside the attr function
I don't know if this is possible, but we can try.

### LINKED ISSUE #7: support deg, grad, rad value types

mitchellhamilton:
**What**:

Allow dynamic values in attr. e.g.
```jsx
const Header = styled.div`
  background-color: attr(backgroundColor, ${defaultBackgroundColor})
`
```

**Why**:

#5 
#7

**How**:

Parse attr in `parseDynamicValues`.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

Closes #7 
Closes #5


src/babel.js
```diff
@@ -13,10 +12,136 @@ function joinExpressionsWithSpaces (expressions, t) {
...
     quasis.push(t.templateElement({ cooked: ' ', raw: ' ' }, true))
   })
-  return t.templateLiteral(
-    quasis,
-    expressions
-  )
+  return t.templateLiteral(quasis, expressions)
+}
+
+function parseStyledDynamicValues (rules, t, composes = 0, vars) {
```
tkh44: My only concern here is maintainability. I'm not sure if we will need to support more than `attr` in the future.

PR #77, opened 2017-07-03
components as selectors example block

peterpme:
**What**:

Added documentation for components as selectors

**Checklist**:
- [ x ] Documentation


example/src/main.js
```diff
@@ -125,6 +126,21 @@ class App extends React.Component {
             }}
           />
 
+          <Markdown markdown={require('../../docs/components-as-selectors.md')} />
```
tkh44: should be `styling-any-component.md`

PR #85, opened 2017-07-04
Add preact support

tkh44:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/src/preact.js
+++ b/src/preact.js
@@ -1,0 +1,35 @@
+import { h } from 'preact'
+import { css } from './index'
+
+export default function (tag, cls, vars = [], content) {
+  if (!tag) {
+    throw new Error(
+      'You are trying to create a styled element with an undefined component.\nYou may have forgotten to import it.'
+    )
+  }
+
+  function Styled (props) {
+    const getValue = v => (v && typeof v === 'function' ? v.cls || v(props) : v)
+    const className = css(
+      cls,
+      vars.map(getValue),
+      content
+    )
+
+    return h(
+      tag,
+      Object.assign({}, props, {
+        ref: props.innerRef,
+        class: props.class
+          ? className + ' ' + props.class
+          : className
+      })
+    )
+  }
+
+  const name = cls[0].split('-')[1]
+  const componentTag = tag.displayName || tag.name || 'Component'
+  Styled.displayName = `styled(${componentTag}${name})`
+  Styled.cls = '.' + cls
+  return Styled
+}
--- a/src/preact.js
+++ b/src/preact.js
@@ -20,9 +20,7 @@ export default function (tag, cls, vars = [], content) {
       tag,
       Object.assign({}, props, {
         ref: props.innerRef,
-        class: props.class
-          ? className + ' ' + props.class
-          : className
+        classname: `${className} ${props.classname || props.class || ''}`
       })
     )
   }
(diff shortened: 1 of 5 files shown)


src/preact.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default function (tag, cls, vars = [], content) {
+  if (!tag) {
+    throw new Error(
+      'You are trying to create a styled element with an undefined component.\nYou may have forgotten to import it.'
+    )
+  }
+
+  function Styled (props) {
+    const getValue = v => (v && typeof v === 'function' ? v.cls || v(props) : v)
```
lukeed: I'd hoist this, redefined per `Styled` tag.

src/preact.js
```diff
@@ -0,0 +1,35 @@
...
+      cls,
+      vars.map(getValue),
+      content
+    )
+
+    return h(
+      tag,
+      Object.assign({}, props, {
+        ref: props.innerRef,
+        class: props.class
```
lukeed: Should use `classname` key. Also can be this:

```js
classname: `${className} ${props.class || ''}`
```

src/preact.js
tkh44: Can't because of props.

lukeed: Can still hoist and pass, but looking at it again, is just a style thing.

ooade:
Dropping this?

tkh44:
It's too much to maintain. I wouldn't mind someone creating an external version of `styled` for preact.

ooade:
Okay. Will try taking this when i'm free.

Connorelsea:
Sad to see this closed. What would the best alternative be for use with Preact?

mitchellhamilton:
@Connorelsea emotion should work fine with preact-compat or _maybe_ even just aliased to preact.

developit:
We can add some guides on how to do the latter. I don't think this even uses any part of compat, it's more just the package name.

mitchellhamilton:
The package we use for theming uses `React.Children.only` so I think we might need compat. AFAIK emotion should work fine without compat if you don't use theming

https://github.com/iamstarkov/theming/blob/master/src/create-theme-provider.js#L96

Connorelsea:
Seems to have mismatched class names when using the `emotion/babel` plugin with react-cli, unless I am doing something irrelevant incorrectly. See the below picture for example:

image
image

mitchellhamilton:
I'm guessing you're using `preact-cli`, IIRC in the components folder css files are imported as css modules

developit:
yup yup. Something like this would work:

```js
export default (config, env, helpers) => {
  let babelConfig - helpers.getLoadersByName(config, 'babel-loader')[0].rule.options;
  babelConfig.plugins.push('emotion/babel');

  // turn off CSS Modules:
  helpers.getRulesByMatchingFile('src/components/x.css').rule.test = () => false;
  delete helpers.getRulesByMatchingFile('x.css').rule.exclude;
};







```

tkh44:
I would recommend putting `emotion/babel` first in your list. It needs to come before a couple of babel plugins.

developit:
If the above works, it can be published to npm as `preact-cli-plugin-emotion`. The next version of the CLI will support plugins via package.json instead of requiring they be invoked from config.

Connorelsea:
Nice, sounds good. Also, enabling inline-mode works as well, dunno what the benefits/cons to this are though.

```javascript
  let { rule } = helpers.getLoadersByName(config, "babel-loader")[0]
  let babelConfig = rule.options

  babelConfig.plugins.push(["emotion/babel", { inline: true }])
```

tkh44:
I would run extract mode unless you need ie11 support. I'm not sure you can extract the critical css needed to run inline mode in a server rendered env.

PR #100, opened 2017-07-06
Theming

### LINKED ISSUE #17: Theme support
Hello,

do you intend to support theming with emotion ? ✨

### LINKED ISSUE #35: theming support

tkh44:
**What**:
Add support for themes

**Why**:
They are nice to work with

**How**:
Adding ThemeProvider component and withTheme hoc. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

closes #17 
closes #35


k15a:
Hey, you may want to checkout https://www.npmjs.com/package/theming which is a unified theming solution for CSSinJS solutions. We at styled-components want to eventually swap to theming as well.

tkh44:
For some reason theming is huge.
```
Approximate weight of theming:
  Uncompressed: 170 kB
  Minified (uglify): 41.8 kB
  Minified and gzipped (level: default): 12.4 kB
```

tkh44:
It looks like they are bundling prop-types in their bundle.

I think it would be a good idea to use that if we can. I don't want to maintain a theming solution.

k15a:
Where did you get this stats from? I think it's 10kb Uncompressed.

tkh44:
the tool `weigh` gets the entire cost of using a library. The cost of using theming includes prop-types. Since we are already using prop-types this doesn't matter.

I'm going to put npmjs.com/package/theming in tonight

PR #101, opened 2017-07-06
Prefix css objects

nishp1:
Fixes: https://github.com/tkh44/emotion/issues/92

**What**:
autoprefix object styles at compile time.

**Why**:
Cross browser styles

**How**:
Update babel plugin to parse css array of objects

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Code complete


test/react.test.js
```diff
@@ -29,8 +29,7 @@ describe('styled', () => {
   })
 
   test('basic render with object as style', () => {
-    const fontSize = 20
-    const H1 = styled.h1({ fontSize })
+    const H1 = styled.h1({ fontSize: 20 })
```
nishp1: Wanted to call this change out: little hard to optimize object properties whose value is identifier as the value can be defined in same file or another. Maybe https://github.com/kentcdodds/css-in-js-precompiler would help down the road. Thoughts?

tkh44: We have to support those. Are the array values what are tripping you up? If so I'd rather not support those and do this.

nishp1: No, array values are not the issue at all, just happened to be the last thing I fixed. 

I build an object from AST here: https://github.com/tkh44/emotion/pull/101/files#diff-bb35c33015964b91f5c83deb2a33a814R128, so need an actual scalar value. I think it should be easier to resolve the literal value if defined in the same module like done in that test. But imported ones would be little harder since we'd have to resolve imported val via reading/parsing modules.

I will look at resolving literals defined in same module later today.

nishp1:
@tkh44 This is good to go. Take a look!

tkh44:
I could only take a quick look but im really excited.  I'm gonna look over it tonight

tkh44:
I'm going to merge this into my glamor branch.

andrewsuzuki:
Prefixing isn't the same. For example. `display: flex` yields:

for template literal:
```
display: -webkit-box;
display: -moz-box;
display: -ms-flexbox;
display: -webkit-flex;
display: flex;
```

for style object:
```
display: -webkit-box;
display: -ms-flexbox;
display: flex;
```

thangngoc89:
@andrewdelprete AFAIK, 7.0 beta fixed this. You can install the beta tag and try it

sovietspaceship:
The last line in docs/objects.md should be updated.

PR #107, opened 2017-07-07
update docs

tkh44:
This pull request is for the reorganization and expansion of the docs. 

**If you would like to help please open PRs**

We need to move/merge the docs that are inlined in the readme and put them into their relevant file in the `docs` folder. When moving the files over it may be tempting to just delete the entire code block. Keep in mind we will have these linked to in the readme so they still need some context.

If you need to expand a live example. 

1. find or add the example in `example/src/blocks/[your-example].example`
2. update the code to demonstrate what you need to. Please keep these demos focused and remember you can have more than one example per section. 

The currently available variables are
```
avatarUrl,
css,
React,
styled,
render
```

3. Once you have your docs and your examples ready go into `main.js` and copy/paste a similar group of Markdown and Playground and modify however needed.

4. open pr


hawkins:
I'd like to touch just a little bit on the `composes` section in order to better cover some pitfalls, like we discussed in slack earlier this afternoon (i.e., since `composes` with a pseudo selector like `&:hover` does not work, here's the suggested pattern...). I think addressing this here will help readers adopt the right patterns for quickly working with emotion effectively

tkh44:
@hawkins that would be great

tkh44:
I'm going to go ahead and merge this portion of the readme so that we can get it up in front of users.

PR #115, opened 2017-07-09
Apply `Array.map` and `Array.forEach` alts

lukeed:
I would run the new code in your benchmark suite first, for real-world advantages.

The only "perf testing" I did here was compare the before-after for tests for complete.

Before: `12.4s` to `13.3s`

After: `10.8s` to `11.2s`

Applied the changes to `index.js` and `react.js`... as per @tkh44's request. Not entirely sure why tbh, but can apply to the rest of the files if welcomed.

Lastly, if you really wanted to, you could probably use `@arr/map` for all `forEach` loops & remove `@arr/foreach`. This would save 73 bytes; not sure if it matters much. 😆


tkh44:
They are the only files that run on the client. I wanted to benchmark the results and see what it nets before having you do more work.

lukeed:
Did the benchmark suite benefit?

tkh44:
I didn't have time to run them last night when I was getting the site ready.

lukeed:
Ah ok, when you do get the chance, lemme know. Curious!

PR #121, opened 2017-07-10
[WIP] Extract Static CSS

hamlim:
**What**: Extract static css declarations. Initial POC to get some feedback on idea/implementation.

**Why**: Emotion's inline mode is great, but sometimes I want to pull out static css from being inlined in JS and also support older browsers (i.e. not use the default mode with css custom property replacements).

**How**: Parsed the input css twice, once for dynamic styles and again for static css, returned the static css from inline and pushed to the state variable storing those declarations.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

This is really early on, still don't have any tests, and I am sure there are a few things I can do to clean up/optimize the code. A few lingering questions I want to get some feedback are:

1. Should this be another mode like `["emotion/babel", {"inline": "dynamic-only"}]`?
2. Are there quick and easy ways to also handle things like:
```Javascript
const linkColor = `#f00`;

const Link = styled.a`
  border: solid 1px ${'red'}; // this should be extracted as it is static
  color: ${linkColor}; // Should this be extracted? linkColor could change out of scope of react however.
`;
```

As always if you have questions, comments, or concerns, feel free to add a comment!


hamlim:
Turns out this actually breaks SSR with extractCritical. I can investigate what would be a good workaround for that, this is one of the things I remember @mitchellhamilton talking about in Slack.

mitchellhamilton:
Static extraction was going to work in a similar way to this but we changed it because otherwise specificity gets messed up.

```jsx
const someClass = css`
  display: ${'flex'}
`

const StyledDiv = styled.div`
  composes: ${someClass};
  display: block;
`
```
In this example the `display: flex;` would take priority over the `display: block;` even though it would make more sense for the `display: block;` to take priority.

If we can solve the specificity issues (I have no idea how we would) we should splitting static and dynamic rules in the PostCSS AST instead of splitting them by semicolons. This is how I originally did static extraction, look at the first few commits in #23 if you're curious how to do it.

hamlim:
I wonder if you could get the generated css to append another class selector on the end of the class that contains the `composes`:

```Javascript
const someClass = css`
  display: ${'flex'}
`

const StyledDiv = styled.div`
  composes: ${someClass};
  display: block;
`
```
Makes:
```CSS
.css-someClass-[hash] { display: flex; } // specificity of 1
.css-StyledDiv-[hash].css-StyledDiv-[hash] { display: block; } // specificity of 2
```

faceyspacey:
@hamlim awesome stuff!

Thinking about creating multiple css chunks, I think all we would need to do is create multiple css files--one for each file entered, rather than one `main` for the entire `Program`. So on `exit` of `File`, one css file per javascript file is created. Than [https://github.com/faceyspacey/extract-css-chunks-webpack-plugin](extract-css-chunks-webpack-plugin) or Webpack in the future when it supports it, will automatically generate multiple css chunks just like if developers manually created css modules.

I could be mistaken though--does `Program` correspond to the entire set of files passed to babel, or is it also "file level"?

hamlim:
A quick note, sorry this has been sitting here with no changes recently. Needed to focus some time and effort into other things. 

I decided to scrap this work and go in the direction of handling the split inside css parsing. Have that mostly structured out. Still a few more edge cases and bugs I know exist that need to get squashed.

I also made a change to bump the specificity of the classes that call compose, however it will still need to be more dynamic because you could compose a class that composes another class:

```Javascript
const classA = css` // -> .css-classA-[hash] { color: red; }
  color: red;
`
const classB = css` // -> .css-classB-[hash].css-classB-[hash] { color: purple; }
  composes: ${classA};
  color: purple;
`
const classC = css` // -> (currently) .css-classC-[hash].css-classC-[hash] { color: green; }
  composes: ${classB};
  color: green;
`
```
`classC` should actually be 3x the regular selector if we want to properly scale this pattern, but we have lost scope of if `classB` contains a `composes` afaik.


Either way my current plan is to squash a few more bugs, go through the tests to see which snapshots should updated (with the increasing specificity for composes) and then write some more tests for this extraction mode.

tkh44:
This is looking good, but I may have a solution for `composes` that's outside the scope of this PR.

hamlim:
Going to close this PR (decided to tackle this in a different way that @mitchellhamilton suggested in another branch). I can make sure to link back here to keep some of the context of what people shared here in that PR as well.

PR #123, opened 2017-07-10
Optimize Server: `extractCritical`

lukeed:
Made some optimizations to the `server.js` code, as per @tkh44's request. This has been tested against the included suite, and I used those as a poor-man's benchmark 😅 I'd highly suggest/encourage these changes are run thru the _real_ benchmark suite before considering.

**Changes**

* Hoisted the RegExp pattern. It was redeclared twice & both times were at run-time
* Installed `@arr/filter` for fastest possible filtering. This function mutates(!), so it should be (and was) used carefully.
* Used a `forEach` loop instead of `map().join('')` because raw string concatenation is faster than joining an array.
* Removed the string-casting for `ids` keys. To me, it looks like they are/will always be strings. I may be wrong on this -- please correct me!


src/server.js
```diff
@@ -1,31 +1,32 @@
+import filter from '@arr/filter'
```
tkh44: Let's drop the filter. I don't want anyone to accidentally use it when mutations matter.

lukeed: You sure? This was responsible for largest drop in test suite time.

tkh44: Oh really?

tkh44: Keep it.

lukeed: Lol ok. IMO isn't too bad to keep an eye out on future changes/PRs 👻

thangngoc89: Eh. Add a bunch of comments about this "feature" of @arr/filter :D

PR #126, opened 2017-07-11
Optimize: `index.js`

lukeed:
**What**:

* Performance updates: 
  * `join()`s are slower than string concatenation
  * fixed `new RegExp(classes[0])` -- was re-creating per loop
  * replace `RegExp.test` with 38-52X faster alternatives
  * remove problematic implicit casting -- de-opts function

* Bugfix: Long/multi-`@media` blocks were not caught by RegExp pattern. Resulted in inverted (ergo, invalid) CSS rule.
  - _Updated Jest snapshot as a result of this!_

**Why**:

* Requested // it's fun
* Fix the bug 😅

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -122,7 +122,7 @@ function deconstruct (selector, styles, media) {
       decs.push(createDec(key, value))
     } else if (Array.isArray(value)) {
       forEach(value, val => decs.push(createDec(key, val)))
-    } else if (/^:/.test(key)) {
+    } else if (key.charCodeAt(0) === 58) {
       forEach(deconstruct(selector + key, value, media), r => rules.push(r))
     } else if (/^@media/.test(key)) {
       forEach(deconstruct(selector, value, key), r => rules.push(r))
--- a/src/index.js
+++ b/src/index.js
@@ -124,7 +124,7 @@ function deconstruct (selector, styles, media) {
       forEach(value, val => decs.push(createDec(key, val)))
     } else if (key.charCodeAt(0) === 58) {
       forEach(deconstruct(selector + key, value, media), r => rules.push(r))
-    } else if (/^@media/.test(key)) {
+    } else if (key.indexOf('@media') !== -1) {
       forEach(deconstruct(selector, value, key), r => rules.push(r))
     } else {
       forEach(deconstruct(selector + ' ' + key, value, media), r => rules.push(r))
--- a/src/index.js
+++ b/src/index.js
@@ -152,7 +152,7 @@ function hyphenate (str) {
 }
 
 function addPx (prop, value) {
-  if (typeof value !== 'number' || unitlessProps[prop]) return value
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -115,23 +120,14 @@ function deconstruct (selector, styles, media) {
...
-        decs.push(createDec(key, val))
-      })
-      continue
-    } else if (/^:/.test(key)) {
+      forEach(value, val => decs.push(createDec(key, val)))
+    } else if (key.charCodeAt(0) === 58) {
       forEach(deconstruct(selector + key, value, media), r => rules.push(r))
-      continue
-    } else if (/^@media/.test(key)) {
+    } else if (key.indexOf('@media') !== -1) {
```
tkh44: shouldn't this be
`key.indexOf('@media') === 0`

lukeed: I thought so too, but it can't. The bug happened because your media rules come in like this:

```js
console.log(`WAS MEDIA: ${key}`);
//=> WAS MEDIA:
          @media only screen and (-webkit-min-device-pixel-ratio: 1.5),
          only screen and (min--moz-device-pixel-ratio: 1.5),
          only screen and (-o-min-device-pixel-ratio: 1.5/1),
          only screen and (min-resolution: 144dpi),
          only screen and (min-resolution: 1.5dppx)
```

There's leading whitespace / multi-line, so your regex of `/^@media/` would always fail for large MQs.

tkh44: Good point. There may be something I'm missing here. @threepointone might know if this is safe.

src/index.js
```diff
@@ -156,7 +152,7 @@ function hyphenate (str) {
 }
 
 function addPx (prop, value) {
-  if (typeof value !== 'number' || unitlessProps[prop]) return value
+  if (typeof value !== 'number' || unitlessProps[prop] !== void 0) return value
```
kurtextrem: imo `void 0` should be `undefined` as it's easier to understand what the condition is for. Minifiers will change it to void anyway, so no need to save space manually.

src/index.js
```diff
@@ -48,8 +54,9 @@ export function css (classes: string[], vars: vars, content: () => string[]) {
 
     if (!inserted[hash]) {
       inserted[hash] = true
+      const rgx = new RegExp(classes[0], 'gm')
       forEach(
-        map(src, r => r.replace(new RegExp(classes[0], 'gm'), `${classes[0]}-${hash}`)),
+        map(src, r => r.replace(rgx, `${classes[0]}-${hash}`)),
         r => sheet.insert(r)
```
mitchellhamilton: I think we can remove the map here.
```js
forEach(r => sheet.insert(r.replace(rgx, `${classes[0]}-${hash}`)))
```

src/index.js
lukeed: No real preference here. _Hopefully_ people are running Uglify 😉

src/index.js
lukeed: Ah, good call 👍

PR #127, opened 2017-07-11
Added documentation for injectGlobal function

nwaywood:
**What**:
Documentation for the injectGlobal function

**Why**:
So people that are not coming from `styled-components` know about the function

**How**:
Just documentation

### DIFF
--- a/docs/inject-global.md
+++ b/docs/inject-global.md
@@ -1,0 +1,10 @@
+## Inject Global
+
+```jsx
+import { injectGlobal } from 'emotion'
+
+injectGlobal`
+	* {
+		box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


docs/inject-global.md
```diff
@@ -0,0 +1,10 @@
+## Inject Global
+
+```jsx
+import { injectGlobal } from 'emotion'
+
+injectGlobal`
+	* {
+		box-sizing: border-box;
+	}
+```
```
thangngoc89: Hey, you're missiong a backtick here ```

nwaywood: Woops thanks! Fixed :)

PR #130, opened 2017-07-11
Move theming into its own file

mitchellhamilton:
⚠️ Breaking Change ⚠️ 

**What**:

Move `ThemeProvider` and `withTheme` exports to `emotion/react/theming`.

**Why**:

Because not everyone wants theming and this doesn't force people to include it if they don't want it.

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


thangngoc89:
Yes. Yes. Yes. I'm upset about the size of emotion after adding theming feature

tkh44:
@mitchellhamilton I'm sitting on this until I figure out the glamor and className merging.

PR #131, opened 2017-07-11
Optimize hash.js

thangngoc89:
**What**:

Optimize hash.js

**Why**:

Speed.

**How**:
Since anything in `for(...)` will be executed in each loop, extract and cache them outside the loop can improve perfomance alot.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete

---

Update: Gonna use `@arr/forEach` to avoid duplicate

### DIFF
--- a/src/hash.js
+++ b/src/hash.js
@@ -8,7 +8,9 @@ export function hashArray (arr: string[]): string {
 
 export function hashObject (...objs: Array<{ [string]: any }>): string {
   let str = ''
-  for (let i = 0; i < objs.length; i++) {
+  let i = 0
+  let length = objs.length
+  for (; i < length; i++) {
     str += JSON.stringify(objs[i])
   }
   return hashString(str, str.length).toString(36)


thangngoc89:
@tkh44 At first, I want to use `@arr/forEach` to save some bytes but with the cost of invoking a function.

I have [some benchmark here](https://esbench.com/bench/59653b8499634800a03489b4)

What is you opinion here? 

image

tkh44:
@thangngoc89 I don't think that's enough of a speed difference to justify the reduction in readability.

k15a:
AFAIK the engines do this kind of optimization automatically.

lukeed:
It wasn't losing because of the function call, but rather the extra set of spread arguments. Removing those, and casting all use cases to ES5 produces these results:

screen shot 2017-07-11 at 2 53 47 pm

That said, it's not consistent. The three are close enough that any one of them can "win" the results chart. In fact, it's pretty even.

tkh44:
I would rather just use @arr/forEach because we use that everywhere else. What do you think @thangngoc89? Fair compromise?

lukeed:
@tkh44 Looks like you forgot a word somewhere 👻 Use what?

tkh44:
Whoops! It looks like github doesn't like the namespace `@arr/forEach` and just deleted it.

lukeed:
Gotcha. IMO, I would remove the spreads & do

```js
var objs = [].slice.call(arguments)
```

It's not _ideal_, but it's a whole lot better than the helper Babel will give you.

thangngoc89:
@lukeed I don't understand this point:

> but rather the extra set of spread arguments

All functions in the bench are the same, only different in the iterator code.

@tkh44 yeah. I also prefer using forEach to save bytes. Bytes matter :p 

@k15a my benchmark result was executed on Chrome.

Issue #128, opened 2017-07-11
Class name concatenation causes problems with showing the correct style


xymostech:
- `emotion` version: 5.2.0
- `react` version: 15.6.1

Relevant code:

```javascript
import {css} from "emotion";
import React, {Component} from "react";
import ReactDOM from "react-dom";

const fakeBlue = css([
  {
    color: "blue"
  }
]);

const red = css([
  {
    color: "red"
  }
]);

const blue = css([
  red,
  {
    color: "blue"
  }
]);

ReactDOM.render(
  <div className={blue}>Hello</div>,
  document.getElementById("main")
);
```

What you did:
I added some styles which I thought would make some text blue

What happened:
I saw that the text was red, instead! 😮

Reproduction:
Webpackbin example of this happening: https://www.webpackbin.com/bins/-KokR2IRsfMsEaoiVBTX

Problem description:
Hey there, co-creator of Aphrodite here. Just looking at the class names that you generate from calls to `css()` I spotted a bug in emotion. I'm pretty sure what's happening here is that you're inserting styles from the first `fakeBlue` `css()` call into a stylesheet first, then from the `red` `css()` call, and then when you get to the `blue` `css()` call you assume that both of those objects have already been added to the stylesheet, and return a class name that looks like `red-something blue-something`. Unfortunately, because CSS precedence rules say that for styles of equal specificity (which all of yours are, they only have one class name) the styles that come later in the stylesheet win! So since you inserted the `red` styles later, they're going to beat out the `blue` styles no matter what you do.

Suggested solution:
Something other than concatenating class names together with spaces! There are a bunch of solutions that other libraries are using, like merging the styles and using a single class name, using a single class name per style value, or duplicating the class name selector to ensure that you get the correct specificity. All of those are fine. :)

I wrote up a blurb about what we do in Aphrodite here (and why concatenating the class names is bad): https://github.com/khan/aphrodite#overriding-styles

Hope this is helpful!

tkh44:
Thank you for taking the time to write this up. 👍

tkh44:
Funny thing... It never occurred to me that you could use emotion in a repl site now. 🤦‍♂️

thangngoc89:
@tkh44 which one?

tkh44:
> https://www.webpackbin.com/bins/-KokR2IRsfMsEaoiVBTX

Look at his link. @xymostech is using object styles so it works fine.

thangngoc89:
[...]

Issue #104, opened 2017-07-07
CSS priority when composing multiple Styled Component


thangngoc89:
- `emotion` version:
- `react` version:

Relevant code.

```javascript
Component = styled.div``

ComponentDecorator = styled(Component)``
```

What you did:

Composing multiple styled component

What happened:

The css order is not guaranteed. 

Reproduction:

Problem description:

Suggested solution:

Concatenate css string instead of generating new class

tkh44:
This might be related, we should detect if `tag.cls` exists and do something with that in `styled`.

tkh44:
I put the difficult tag because of the fact that selectors get crazy with edge cases fast.

thangngoc89:
How React Native solves this: 

![image](https://user-images.githubusercontent.com/3049054/27966043-fb44ad40-6367-11e7-8ad0-502336fe9f3d.png)

https://github.com/necolas/react-native-web/blob/master/docs/guides/style.md#how-styles-are-resolved

nishp1:
@tkh44 what are thoughts here? I think it makes sense to stick with and support browser behavior here.

tkh44:
fixed in v7

PR #138, opened 2017-07-12
Initial working version of using glamor core

tkh44:
**What**:
Use glamor's core algorithm instead of cxs'

**Why**:
CSS object support is much cleaner. This will lead to fixes related to #128, #104 eventually as well.

**How**:
I ported glamor's core to work in our setup

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I did this on @threepointone's advice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@arr/foreach": "^0.1.0",
     "@arr/map": "^0.1.0",
     "babel-plugin-syntax-jsx": "^6.18.0",
+    "fbjs": "^0.8.12",
     "styled-components": "2.0.0",
     "theming": "^1.0.1",
     "touch": "^1.0.0"
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "format": "prettier-eslint --write \"src/**/*.js\" \"test/**/*.js\" \"example/**/*.js\" \"jest-utils/**/*.js\""
   },
   "dependencies": {
+    "@arr/filter": "^0.1.0",
     "@arr/foreach": "^0.1.0",
     "@arr/map": "^0.1.0",
     "babel-plugin-syntax-jsx": "^6.18.0",
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
   "bundlesize": [
     {
       "path": "./dist/DO-NOT-USE.min.js",
-      "threshold": "3 Kb"
+      "threshold": "6 Kb"
     }
   ]
 }
--- a/package.json
+++ b/package.json
@@ -29,9 +29,9 @@
     "format": "prettier-eslint --write \"src/**/*.js\" \"test/**/*.js\" \"example/**/*.js\" \"jest-utils/**/*.js\""
   },
   "dependencies": {
-    "@arr/filter": "^0.1.0",
-    "@arr/foreach": "^0.1.0",
-    "@arr/map": "^0.1.0",
+    "@arr/filter.mutate": "^1.0.0",
+    "@arr/foreach": "^1.0.0",
+    "@arr/map": "^1.0.0",
     "babel-plugin-syntax-jsx": "^6.18.0",
     "fbjs": "^0.8.12",
     "styled-components": "2.0.0",
--- a/src/glamor/CSSPropertyOperations/CSSProperty.js
+++ b/src/glamor/CSSPropertyOperations/CSSProperty.js
@@ -1,0 +1,146 @@
+/**
+ * Copyright 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule CSSProperty
+ */
+
+/**
+ * CSS properties which accept numbers but are not in units of "px".
+ */
+
+let isUnitlessNumber = {
+  animationIterationCount: true,
+  borderImageOutset: true,
+  borderImageSlice: true,
+  borderImageWidth: true,
+  boxFlex: true,
+  boxFlexGroup: true,
+  boxOrdinalGroup: true,
+  columnCount: true,
+  flex: true,
+  flexGrow: true,
+  flexPositive: true,
+  flexShrink: true,
+  flexNegative: true,
+  flexOrder: true,
+  gridRow: true,
+  gridColumn: true,
+  fontWeight: true,
+  lineClamp: true,
+  lineHeight: true,
+  opacity: true,
+  order: true,
+  orphans: true,
+  tabSize: true,
+  widows: true,
+  zIndex: true,
+  zoom: true,
+
+  // SVG-related properties
+  fillOpacity: true,
+  floodOpacity: true,
+  stopOpacity: true,
+  strokeDasharray: true,
+  strokeDashoffset: true,
+  strokeMiterlimit: true,
+  strokeOpacity: true,
+  strokeWidth: true
+}
+
+/**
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -112,7 +113,7 @@
   "bundlesize": [
     {
       "path": "./dist/DO-NOT-USE.min.js",
-      "threshold": "3 Kb"
+      "threshold": "6 Kb"
```
thangngoc89: Wow. This looks scarryyyyyyyyyyyy

tkh44: The react bundle went from 4k to 5k

thangngoc89: Well, that's acceptable to me. Probably need a seperate preact bundle again since preact don't need to filter props.

tkh44: Keep in mind this is my first pass at getting glamor's core algo working here. I keep finding code from features that I don't need and removing it.

src/glamor/CSSPropertyOperations/CSSProperty.js
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Support style names that may come passed in prefixed by adding permutations
+ * of vendor prefixes.
+ */
+let prefixes = ['Webkit', 'ms', 'Moz', 'O']
+
+// Using Object.keys here, or else the vanilla for-in loop makes IE8 go into an
+// infinite loop, because it iterates over the newly added props too.
+Object.keys(isUnitlessNumber).forEach(function (prop) {
```
lukeed: Change `isUnitlessNumber` to an array of names, iterated to create the final object with standard & prefixed names.

10-30x faster & less code.

Also, capitalize `prefixes` already. It's only used here (`prefixKey`) and `ms` is the only lowercase-leading entry.

src/glamor/CSSPropertyOperations/dangerousStyleValue.js
```diff
@@ -0,0 +1,92 @@
...
+  if (
+    isNonNumeric ||
+    value === 0 ||
+    (isUnitlessNumber.hasOwnProperty(name) && isUnitlessNumber[name])
+  ) {
+    return '' + value // cast to string
+  }
+
+  if (typeof value === 'string') {
+    if (process.env.NODE_ENV !== 'production') {
```
lukeed: Cache `process.env.NODE_ENV` to a top-level variable. Accessing this is ~1ms (credit Inferno).

src/glamor/CSSPropertyOperations/index.js
```diff
@@ -0,0 +1,170 @@
...
+  let warnedStyleValues = {}
+  let warnedForNaNValue = false
+
+  let warnHyphenatedStyleName = function (name, owner) {
+    if (warnedStyleNames.hasOwnProperty(name) && warnedStyleNames[name]) {
+      return
+    }
+
+    warnedStyleNames[name] = true
+    process.env.NODE_ENV !== 'production'
```
lukeed: I'd move all these `!isProd ? warning(...) : void 0` to a helper function -- probably a small wrapper around existing `warning`.

nishp1:
> This will lead to fixes related to #128, #104 eventually as well.

#128 also applies to tagged template literals as well. Would glamor help with that too? Just want to make sure, we are solving both variants if we are going to bring extra bytes.

nishp1:
Would also recommend changing for loops to use `@arr/forEach`.

src/glamor/CSSPropertyOperations/CSSProperty.js
```diff
@@ -0,0 +1,146 @@
...
+  outline: {
+    outlineWidth: true,
+    outlineStyle: true,
+    outlineColor: true
+  }
+}
+
+const CSSProperty = {
+  isUnitlessNumber: isUnitlessNumber,
+  shorthandPropertyExpansions: shorthandPropertyExpansions
```
nishp1: This doesn't seem to used elsewhere, so can be removed.

src/index.js
```diff
@@ -93,94 +98,400 @@ export function hydrate (ids: string[]) {
...
 
-  return className
+// define some constants
+
+const isDev = process.env.NODE_ENV === 'development' || !process.env.NODE_ENV
+const isTest = process.env.NODE_ENV === 'test'
+
+// takes a string, converts to lowercase, strips out nonalphanumeric.
+function simple (str) {
+  return str.toLowerCase().replace(/[^a-z0-9]/g, '')
```
nishp1: Regex should be moved out of func as a const.

src/index.js
```diff
@@ -93,94 +98,400 @@ export function hydrate (ids: string[]) {
...
-function deconstruct (selector, styles, media) {
-  const decs = []
-  const rules = []
+// of shape { 'data-css-<id>': '' }
+export function isLikeRule (rule: GlamorRule) {
+  let keys = Object.keys(rule).filter(x => x !== 'toString')
+  if (keys.length !== 1) {
+    return false
+  }
+  return !!/css\-([a-zA-Z0-9]+)/.exec(keys[0])
```
nishp1: Same as https://github.com/tkh44/emotion/pull/138/files#r127111313

src/index.js
```diff
@@ -93,94 +98,400 @@ export function hydrate (ids: string[]) {
...
+}
 
-  for (let key in styles) {
-    const value = styles[key]
-    const type = typeof value
+// extracts id from a { 'css-<id>': ''} like object
+export function idFor (rule: GlamorRule) {
+  let keys = Object.keys(rule).filter(x => x !== 'toString')
+  if (keys.length !== 1) throw new Error('not a rule')
+  let regex = /css\-([a-zA-Z0-9]+)/
```
nishp1: https://github.com/tkh44/emotion/pull/138/files#r127111313

src/index.js
```diff
@@ -35,17 +38,66 @@ export function flush () {
...
+  return rule
+}
+
+// The idea on how to merge object class names come from glamorous
+// 💄
+// https://github.com/paypal/glamorous/blob/master/src/get-glamor-classname.js
+function getGlamorStylesFromClassName (className) {
+  const id = className.trim().slice('css-obj-'.length)
+  if (sheet.registered[id]) {
+    return sheet.registered[id].style
```
mitchellhamilton: Should this be `registered` or should the other references to `registered` be `sheet.registered` or am I totally wrong?

src/glamor/CSSPropertyOperations/CSSProperty.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export default CSSProperty
```
lukeed: At least in context of this PR, `isUnitlessNumber` is the only thing used from this file. 

Regardless, should use `export default { isUnitlessNumber }` as opposed to exporting a created Object. This way you can actually import what you need/use & shake away the rest. This alone would probably drop a 1KB and make @thangngoc89 happy lol.

thangngoc89: @lukeed lol. I'm sensitive about size :D . Everything must be small and fast. 

P/s: Actually, my target platform is 512MB Android phone so you know :D

tkh44: I'm going to use the version from cxs so I can just avoid using facebook code.

https://github.com/tkh44/emotion/blob/master/src/index.js#L157-L186

thangngoc89: @tkh44 that sounds better. But perhaps just an array instead? `indexOf` is fast

lukeed: I checked that snippet earlier. It's faster to do `obj[key] !== undefined` than it is to do `arr.indexOf(key) !== -1`

PR #150, opened 2017-07-15
Composition Fixes

tkh44:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 // @flow
 import forEach from '@arr/foreach'
 import { StyleSheet } from './sheet'
-import { hashArray, hashObject } from './hash'
+import { hashString, hashArray, hashObject } from './hash'
 
 export const sheet = new StyleSheet()
 sheet.inject()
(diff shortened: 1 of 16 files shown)


src/index.js
```diff
@@ -40,10 +40,13 @@ export function css (classes: string[], vars: vars, content: () => string[]) {
...
-  forEach(classes, (cls): string => {
-    computedClassName && (computedClassName += ' ')
-    computedClassName += typeof cls === 'string' ? cls : objStyle(cls)
+  let computedClassNames = []
+  forEach(classes, (cls): void => {
+    if (typeof cls === 'string') {
+      forEach(cls.split(' '), c => computedClassNames.push(c))
+    } else {
+      computedClassNames.push(objStyle(cls))
+    }
```
mitchellhamilton: I think we should do this stuff with string concatenation like before.

tkh44: I have a proper fix for objects incoming with the glamor branch.

PR #152, opened 2017-07-15
Static CSS Extraction (take 2)

hamlim:
**What**:
This is a continuation of some of the discussions and work that took place over in this PR: https://github.com/tkh44/emotion/pull/121. This adds another mode to the babel plugin that acts a bit like the extract mode and a bit like the inline mode.

```JSON
{
  "plugins": [
    ["emotion/babel", {"extractStatic": true }]
  ]
}
```

**Why**:

I want the ability to cache static css and offload as much as possible from being done at run time, while also supporting older browsers than IE11.

**How**:

Clones the css root inside `parser.js`, removes dynamic content from it and removes static content from the original root, then returns two arrays (one dynamic, the other all static).

The static css then gets send to `state.insertStaticRules` to be inserted into the written out css file.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete

Follow the original discussion about this here: https://github.com/tkh44/emotion/pull/121

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -8,10 +8,10 @@ type CSSDecl = {
   parent: { selector: string, nodes: Array<mixed> },
   prop: string,
   value: string,
-  remove: () => {},
+  remove: () => {}
 }
 
-export function parseCSS(
+export function parseCSS (
   css: string,
   options: {
     nested?: boolean,
@@ -20,7 +20,7 @@ export function parseCSS(
     name: string,
     hash: string,
     canCompose?: boolean,
-    extractStatic?: boolean,
+    extractStatic?: boolean
   } = {
     nested: true,
     inlineMode: true,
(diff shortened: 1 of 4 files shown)


src/parser.js
```diff
@@ -74,6 +77,11 @@ export function parseCSS (
         vars++
       }
     }
+    if (options.extractStatic) {
+      if (typeof decl.value === 'string' && !/xxx(\d+)xxx/gm.exec(decl.value)) {
```
hamlim: This typeof check might not be needed. I think it only needs to check that the value does not match the replacement pattern.

tkh44:
We got this working in #154

hamlim:
That's awesome! This sort of felt like a long shot when looking at the initial work on the glamor re-write. Too excited right now.

PR #154, opened 2017-07-16
Te Fiti

tkh44:
**What**:

- Fork glamor core code and use it to support objects
- Turn string styles into objects during compilation
- **Inline mode is default**
- **Extract mode removed**
- **New `extractStatic` option** - Inline mode but if the block has **0** interpolations it is extracted into a file. If users choose, they can write all static styles and use inline styles for dynamic values. Any static style blocks are replaced with a string class name by the emotion plugin. 

```jsx
const a = css`color:blue`;
```
is replaced with a string like 

```jsx
const a = 'css-12345'
```

**Why**:
Composition is awesome and predictable. 

image

## TODO:
- [ ] **code review** I need as many people looking at this as possible.
- [ ] remove the facebook code in the `/glamor` folder and use some smaller code for that (maybe from cxs or postcss-js)
- [ ] remove code for features that aren't used
- [x] remove the code [here](https://github.com/tkh44/emotion/pull/154#discussion-diff-128493505R41) and update the examples & tests
- [ ] replace `[].forEach|map|reduce` with their equivilents from `./utils` for consistancy
- [ ] match the glamor code code style to our code style (stuff needs to be renamed and moved)
- [x] prefixer should not use the older spec for flexbox
- [x] prefixer should not use array values `display: flex` is converted to `display: -webkit-box,-moz-box,-ms-flexbox,-webkit-flex,flex`
- [ ] update documentation about how it works internally
- [ ] udpate injectGlobal docs
- [ ] update composition docs

If you want to help please feel free to make a PR against this branch `glamor-extended`


tkh44:
image

tkh44:
The only thing I can't figure out is how to do nested composes.

example/src/blocks/css-prop.example
```diff
@@ -12,7 +12,7 @@ render(
       background-color: #8c81d8;
       border-radius: 4px;
 
-      img {
+      & img {
```
mitchellhamilton: Just curious, is the `&` required now or is it just stylistic?

src/parser.js
```diff
@@ -13,91 +25,50 @@ type CSSDecl = {
...
-    if (decl.prop === 'name') decl.remove()
-    if (decl.value.match(/attr/)) {
-      hasCssFunction = true
+
+  postcssNested(root)
+
+  root.walkRules((rule: Rule) => {
+    // TODO: do this everywhere except `,xxx9xxx`
+    if (/\bxxx\d+xxx/.exec(rule.selector)) {
+      rule.selector = `.${rule.selector}`
```
mitchellhamilton: I don't think it's a good idea to add the `.`, I think it makes more sense that when you interpolate a class, you have to add the `.`. This would also stop people from interpolating ids and other selectors.

tkh44: Yeah I agree with you. I keep thinking of weird edge cases.

example/src/blocks/css-prop.example
tkh44: It is required now because it's how glamor knows to expand the class name

example/src/markdown/index.js
```diff
@@ -6,6 +6,7 @@ import styles from './index.css'
 
 const MarkdownContainer = styled('div')`
   composes: ${styles.markdownContainer};
+  
   h1, h2, h3, h4, h5 {
```
lukeed: How come these don't lead with a `&`?

src/glamor/CSSPropertyOperations/CSSProperty.js
```diff
@@ -0,0 +1,146 @@
...
+  outline: {
+    outlineWidth: true,
+    outlineStyle: true,
+    outlineColor: true
+  }
+}
+
+const CSSProperty = {
+  isUnitlessNumber: isUnitlessNumber,
+  shorthandPropertyExpansions: shorthandPropertyExpansions
```
lukeed: These `shorthandPropertyExpansions` still unused. But because of how `CSSProperty` is defined and exported, the `shorthandPropertiesExpansions` are still in the production bundle.

src/react/index.js
```diff
@@ -37,25 +49,62 @@ export default function (tag, cls, vars = [], content) {
...
-      const getValue = v =>
-        v && typeof v === 'function' ? v.cls || v(mergedProps) : v
+      const getValue = v => {
+        if (v && typeof v === 'function') {
+          if (v.__emotion_spec) {
+            return css(
+              map(v.__emotion_spec.objs, getValue),
+              map(v.__emotion_spec.vars, getValue),
+              v.__emotion_spec.content
+            )
```
mitchellhamilton: I don't understand this, wouldn't this break with components that have function interpolations? e.g. in this example the class generated to target the child selector would be based on the `fontSize` prop of `AnotherComponent`
```jsx
  font-size: ${props => props.fontSize || 20};
  display: flex;
`

const AnotherComponent = styled.div`
  ${SomeComponent} {
    border: 0;
  }
`
```

tkh44: You are right. This needs to be fixed.

src/index.js
```diff
@@ -1,186 +1,523 @@
...
+            mq,
+            supp,
+            src: _src[key]
+          })
+          build(dest, {
+            selector: joinSelectors(selector, '::-ms-input-placeholder'),
+            mq,
+            supp,
+            src: _src[key]
+          })
```
mitchellhamilton: Should we remove this? I added tests for this and we add the `webkit` and `ms` prefixes in the babel plugin ([not the `moz` one since the usage of firefox versions that require the prefix is under 1%](http://caniuse.com/#feat=css-placeholder)) for tagged template literals but not for objects, we could _probably_ do it for objects if we wanted to.

tkh44: Yes remove this whole thing

example/src/markdown/index.js
tkh44: cause bug

src/glamor/CSSPropertyOperations/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+    warnedForNaNValue = true
+    process.env.NODE_ENV !== 'production'
+      ? warning(
+          false,
+          '`NaN` is an invalid value for the `%s` css style property.%s',
+          name,
+          checkRenderMessage(owner)
+        )
+      : void 0
```
mitchellhamilton: Do we want to keep all these warnings?

test/babel/__snapshots__/font-face.test.js.snap
```diff
@@ -3,50 +3,60 @@
...
-          font-weight: bold;}"
+    font-weight: bold
+}"
 `;
 
 exports[`fontFace babel extract basic assign to variable 1`] = `
 "import \\"./font-face.test.emotion.css\\";
 
-const thisWillBeUndefined = undefined;"
+const thisWillBeUndefined;"
```
mitchellhamilton: We need to fix this.

PR #157, opened 2017-07-17
[WIP] Add tests for bugs in `extractCritical`

mattfysh:
**What**: Adding tests for bugs found

**Why**: Bug squashing

**How**: Without fixes (sorry - not enough time!)

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete

### DIFF
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -20,6 +20,7 @@ const Image = styled.img`
 `
 
 // this will not be included since it's not used
+
 css`
   display: none;
   name: unused-class;
@@ -31,6 +32,9 @@ injectGlobal`
   .no-prefix {
     display: flex;
     justify-content: center;
+  }
+  .should-be-included {
+    content: 'css-foo matches regex and this rule is swallowed (current bug)'
   }
 `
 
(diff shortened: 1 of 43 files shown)


test/server.test.js
```diff
@@ -32,6 +33,9 @@ injectGlobal`
     display: flex;
     justify-content: center;
   }
+  .should-be-included {
+    content: 'css-foo matches regex and this rule is swallowed (current bug)'
+  }
 `
```
mitchellhamilton: This isn't a bug, this rule isn't included since it'll expand to have the hashed class name which isn't used in the html

mattfysh: I'm sure it is, right? Change the `css-foo` to be `cs-foo` and it will now appear in the output

tkh44: the css call returns a class name. Any nested selectors are prefixed with that class name for specificity.

Lets just pretend that that `css`...`` call returns `css-12345`. 

`.should-be-include { ... }` becomes `.css-12345 .should-be-included { ... }`

mattfysh: This is a call to `injectGlobal`... not `css`. If you're viewing on GitHub, you might be having a few lines between getting clipped and that's why you're seeing it as a call to `css`

But ignore that for now - please just watch the bug in motion by changing `css-foo` in the content value to `cs-foo` and you'll now see this rule coming through in `extractCritical`.

tkh44: Oh man. My bad on that. It looks like you are correct. @mitchellhamilton you mind taking one more look at this?

mitchellhamilton: I'm so sorry, you're right @mattfysh

mattfysh: No worries, just a minor edge case. I guess you could leave the test in pending (`xtest()`?) and fix it if/when you want to support

PR #158, opened 2017-07-17
Add webpack.config.js example to install doc

bjudson:
**What**: Update to the installation documentation

**Why**: Clarify adding emotion via webpack.config.js vs .babelrc. If users have babel-loader in webpack config, putting emotion plugin into .babelrc can cause subtle bugs, depending on which other plugins are being used – seems that plugins in webpack.config.js run before those in .babelrc.

**How**:

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A


docs/install.md
```diff
@@ -17,6 +17,24 @@ npm install -S emotion
 }
 ```
 
+**webpack.config.js**
+
+If you keep babel-loader configuration in your Webpack config, add Emotion as the first item in the plugins section:
```
silvenon: Why is it important that it's the first item? Why is this specific to Babel config in `webpack.config.js`? Maybe that explanation could be added too? 🤔

bjudson: @silvenon thanks for the review. i have added a separate note about plugin order, since it's not specific to webpack.config.js, and quickly explained why it's significant.

Issue #147, opened 2017-07-14
Initialize css-vars if possible


kylpo:
- `emotion` 5.2.0:
- `react` 15.5.4:

### Relevant code.

```jsx
render() {
    const grow = 1

    return (
      <div
        css={`
          display: flex;
          flex-direction: column;
          position: relative;
          flex-shrink: 0;
          align-items: center;
          justify-content: center;
          flex-grow: ${grow};
        `}
      >
```

### What you did:
Set to `extract` mode.

### What happened:
![emotion-flash](https://user-images.githubusercontent.com/720177/28230546-4ccf7328-689c-11e7-9ecd-60c343d538b3.gif)

### Problem description:
If a variable is used in `css`, it is converted to a css-var. With SSR (in `external` mode), the page loads with all other styles shown, then javascript loads and injects a value for the associated css-var. This results in a style flash.

Above is a contrived example, but I could see it being a bigger issue for sharing constants.

For example,
```jsx
import COLOR from 'constants/color'

//...

<div
  css={`
    background-color: ${COLOR.RED};
  `}
/>
```

will see a paint flash on load.

### Suggested solution:
Initialize the generated css-var to `1` so that the style flash does not occur.

@hamlim said it best: "SSR ideally should evaluate that at first though right? It won’t change but should compile out initially"

*Note, this was briefly discussed in the slack channel from 8:00am on July 14, 2017*

mitchellhamilton:
Just to confirm, you're doing SSR with extract mode? If so, how? (We've never done SSR with extract mode)

hamlim:
Not actually sure if this is possible with just babel, I could be wrong but babel won't evaluate the code, so even if you could keep a map of identifiers and their initial values, and if you could inject those initial values in the second argument of the `var()` function, from a babel standpoint you will have no idea what is valid css.

For example:
```Javascript
const initValue = Math.pow(3, 2);
///
<div
  css={`
    height: ${initValue};
  `}
/>
/// ->
.selector { height: var(--css-[hash], Math.pow(3, 2)); }
```

kylpo:
Hey, @mitchellhamilton. Yes, this is SSR with extract mode. I just followed Emotion's guide for SSR, but also enabled `extract`.

You're welcome to pull down this project's branch and tinker with it. Here are the relevant lines:

enable extract - https://github.com/constelation/static-starter/blob/emotion-flash/.babelrc#L7

[...]

PR #159, opened 2017-07-17
use extractCritical to set css-vars inline

kylpo:
**What**:
This PR adds css-vars to extractCritical so their initial values may be set inline (when in `extract` mode). Discussed in #147.

**How**:
Started with @mitchellhamilton's regex substitute, but needed to account for `vars-` strings a bit differently. Since the regex was greedy, for a string like

```css
.vars-176qc2r {--css-Home-1wjhkv7-0: 0; --css-Home-1wjhkv7-1: 1}
```

, the matcher would grab a `css-` string instead of the desired `vars-`. We could change our regex to be lazy with something like `/(?:css|vars)(?:[a-zA-Z0-9-]*?)-([a-zA-Z0-9]+?)/gm`, but that seems brittle. Surely there'll be a time when `css-` will be shorter than the `vars-`.

So, this implementation checks for `vars-` first, then moves on to the normal `css-` check.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -3,7 +3,8 @@ import filter from '@arr/filter.mutate'
 import { sheet, inserted } from './index'
 import { keys } from './utils'
 
-const RGX = /css(?:[a-zA-Z0-9-]*)-([a-zA-Z0-9]+)/gm
+const RGX = /(?:css|vars)(?:[a-zA-Z0-9-]*)-([a-zA-Z0-9]+)/gm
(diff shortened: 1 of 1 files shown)


kylpo:
Happy to! Thanks for pointing me to the correct file. 

I'm on vacation until Monday. When I get back, I'll work on it.

kylpo:
Hey, sorry. I haven't forgotten this. Just busy.

tkh44:
We aren't using vars in v7 so further work on this shouldn't be necessary.

PR #177, opened 2017-07-25
fresh regex every time extractCritical is called

threepointone:
**What**: Use a fresh regex every time `extractCritical` is called

**Why**: regexes are stateful, and the current method would break critical css extraction unexpectedly because of previous runs

**How**: the regex declaration was brought inside the function

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1,8 +1,6 @@
 import { sheet, inserted } from './index'
 import { keys, forEach } from './utils'
 
-const RGX = /css(?:[a-zA-Z0-9-]*)-([a-zA-Z0-9]+)/gm
-
 export {
   flush,
   css,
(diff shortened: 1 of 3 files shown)


src/server.js
```diff
@@ -16,6 +14,8 @@ export {
 export function extractCritical (html) {
   // parse out ids from html
   // reconstruct css/rules/cache to pass
+  const RGX = /css(?:[a-zA-Z0-9-]*)-([a-zA-Z0-9]+)/gm
+
```
tkh44: Do we need to do `RGX.lastIndex = 0` after the while loop?

threepointone: nyah, it gets GCed and lost anyway after it's done

tkh44: What about its usage here though. https://github.com/tkh44/emotion/pull/177/files/737e2b32325571d1dec2da461222aa995981e091#diff-e6a5b42b2f7a26c840607370aed5301aR29

PR #184, opened 2017-07-26
Accept passed in "class" attribute for preact support

ajoslin:
For example:

```js
const H1 = styled`color: red;`

<H1 class='some-class' />
```

**What**: Allow passed in `class` attr

**Why**: So preact support "just works" using `config.resolve.alias.react = "preact"` 

**Checklist**:

- [ ] Documentation (N/A)
- [x] Tests
- [x] Code complete


tkh44:
Great idea. Thanks!

PR #190, opened 2017-07-27
[Update] import `styled` from 'emotion/react'

bogas04:
**What**: Docs are updated to use 'styled' from 'emotion/react' and not 'emotion'. 

**Why**: I guess docs weren't updated when styled was moved to 'emotion/react'

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete


bogas04:
Thanks for the merge !

PR #191, opened 2017-07-27
More tests, use fns from utils and rework core

mitchellhamilton:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


src/index.js
```diff
@@ -236,21 +221,22 @@ function selector (id: string, path: string = '') {
...
-    .split(',')
-    .map(
-      x =>
-        x.indexOf('&') >= 0 ? x.replace(/&/gm, `.css-${id}`) : `.css-${id}${x}`
-    )
-    .join(',')
+  let x = map(
+    path.split(','),
+    x =>
+      x.indexOf('&') >= 0 ? x.replace(/&/gm, `.css-${id}`) : `.css-${id}${x}`
```
tkh44: We should probably hoist this callback and the regex inside.

src/index.js
```diff
@@ -342,9 +328,11 @@ function joinSelectors (a, b) {
   let as = map(a.split(','), a => (!(a.indexOf('&') >= 0) ? '&' + a : a))
   let bs = map(b.split(','), b => (!(b.indexOf('&') >= 0) ? '&' + b : b))
 
-  return bs
-    .reduce((arr, b) => arr.concat(as.map(a => b.replace(/&/g, a))), [])
-    .join(',')
+  return reduce(
+    bs,
+    (arr, b) => arr.concat(map(as, a => b.replace(/&/g, a))),
```
tkh44: We could hoist this up as well.

src/index.js
mitchellhamilton: We can't because it uses `id`

src/index.js
mitchellhamilton: We can't because it uses `a`

mitchellhamilton: wait, i might be wrong

mitchellhamilton: We can't hoist it because it uses `as`

PR #192, opened 2017-07-27
Use postcss-nested and postcss-safe-parser instead of styled-components fork and pass filename to postcss

### LINKED ISSUE #188: Pass from option to PostCSS
- `emotion` version: github master
- `react` version: irrelevant

What you did:

Have `browserslist` config in my project root.

What happened:

This config didn’t apply.

Problem description:

Here you parse CSS by PostCSS: https://github.com/tkh44/emotion/blob/89cb7956f667e71226f2bb589bbc2bb574cb38b7/src/parser.js#L39

But you didn’t pass `from` option to parser. As result nodes in AST don’t have origin file name.

As result, Autoprefixer didn’t load correct `browserslist` config.

Suggested solution:

We need to pass `parser(css, { from: 'path/to/file.js' })`

mitchellhamilton:
**What**:

Use `postcss-nested` and `postcss-safe-parser` instead of `styled-components` fork and pass the filename to postcss with the `from` option

I also moved the object parser into the src since we need to change it for #189

**Why**:

#188 

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I'm pretty sure I'm passing the filename to postcss incorrectly since I don't think the `.browserslistrc` file is being read

Closes #188


tkh44:
Does this break the example site? 😄 I don't really care but we will have to figure out something new for the site.

ai:
@tkh44 also you can clean up code a little:

```diff
function stringifyCSSRoot (root) {
  return root.nodes.map((node, i) => {
-    let str = ''
-    stringify(node, x => {
-      str += x
-    })
-    return str
+   node.toString()
  })
}
```

In this case you even don’t need to import `postcss/lib/stringify`.

PR #193, opened 2017-07-27
Use spread over Object.assign

bogas04:
**What**: Object.assign is being replaced by Object spread syntax.

**Why**: Object spread doesn't require you to also include polyfill for Object.assign, hence it's better for users trying it out on older browsers.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/src/react/index.js
+++ b/src/react/index.js
@@ -102,10 +102,11 @@ export default function (tag, objs, vars = [], content) {
       return h(
         tag,
         omit(
-          Object.assign({}, mergedProps, {
+          {
+            ...mergedProps
             ref: mergedProps.innerRef,
             className
-          }),
+          },
           ['innerRef', 'theme']
         )
       )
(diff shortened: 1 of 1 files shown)


thangngoc89:
>  Object spread doesn't require you to also include polyfill for Object.assign, hence it's better for users trying it out on older browsers

Actually, babel includes Object.assign polyfill for object spread.

bogas04:
@thangngoc89 it includes it for spread, but if you directly use Object.assign, then the user library will need to add polyfill.

bogas04:
Thank you so much for accepting the PR ! I feel I didn't quite explain myself well. Either way, kudos to Emotion for being awesome !

thangngoc89:
Yeah. I didn't think about manually providing Object.assign polyfill

bogas04:
@thangngoc89 Sorry for the rush, but when can I expect this patch to get released to npm ? Thanks!

thangngoc89:
@bogas04 No idea. Ask @tkh44 . Current he's focusing on shaping version 7.0 which is a major rewrote of emotion core

PR #203, opened 2017-07-29
Component as selector fix

### LINKED ISSUE #178: Using a component as a selector breaks when it has function interpolations
- `emotion` version: 7.0.0
- `react` version:

Relevant code.

```javascript
  font-size: ${props => props.fontSize || 20};
  display: flex;
`
// Also the . is required now, I think we should change it back so people don't have to add it themselves.
const AnotherComponent = styled.div`
  & .${SomeComponent} {
    border: 0;
  }
`
```

Problem description:

A change in #154 made it so the class used to target components as selectors are generated in the component that has the interpolation. The main problem with this is that function interpolations are called with the props of the component that has the selector interpolation so it will likely be different to the one generated by the component with the function interpolation.

In the example above the class generated to target the child selector would be based on the fontSize prop of AnotherComponent.

https://github.com/tkh44/emotion/blob/master/src/react/index.js#L60-L66

https://github.com/tkh44/emotion/pull/154/files#r128906394

Suggested solution:

We could call function interpolations with an empty object, this would mean assuming that function interpolations are deterministic. While they should generally be deterministic I don't think we should enforce this.

Another option would be generating the hash at build time. This would be more complex, we can't just hash the styles like we did before. (we could but it wouldn't work very well, for example if the entire contents of a components' styles were a single composes it would have the same hash as another component with a single composes) We would likely have to do something like what [`styled-components` does](https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/visitors/displayNameAndId.js#L66-L93). The problem with this is that it would break for components built in a certain way (e.g. `rebass-emotion` and `grid-emotion`) where the babel plugin sees them all as a single component so they would all have the same hash

Another option could be casting the functions to strings, I'm guessing this is a bad idea since the function could be different on the server vs on the client because of minification, etc.

Another idea I thought of as I was finishing this issue would be maybe just ignoring function interpolations and replacing them with some value in the hashing. This would still have the aforementioned issue in some cases though, for example if two components both have a single composes value that is a function they would have the same hash.

mitchellhamilton:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

Closes #178

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -1,11 +1,12 @@
 // @flow weak
 import fs from 'fs'
-import { basename } from 'path'
+import { basename, dirname, join as pathJoin, sep as pathSep, relative } from 'path'
 import { touchSync } from 'touch'
 import { inline } from './inline'
 import { parseCSS } from './parser'
 import { getIdentifierName } from './babel-utils'
 import { map } from './utils'
+import { hashArray } from './hash'
 import cssProps from './css-prop'
 import ASTObject from './ast-object'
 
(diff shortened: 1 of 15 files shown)


src/babel.js
```diff
@@ -79,6 +80,63 @@ export function replaceCssWithCallExpression (
...
+    const packageJsonString = fs.readFileSync(pathJoin(moduleRoot, 'package.json'))
+    if (packageJsonString) {
+      try {
+        moduleName = JSON.parse(packageJsonString).name
+      } catch (e) {}
+    }
+  }
+  const code = file.code
+
+  const fileHash = hashArray([moduleName, filePath, code])
```
mitchellhamilton: Since this hashes the contents of the whole file it means that the hashes change whenever any code in that file changes. This is only really a problem for tests because it will result in unrelated snapshots failing. @tkh44 what do you think about this? Maybe we should try and hash some different stuff like the identifier name?

mitchellhamilton: This isn't a problem anymore. The classes now have the `css-` prefix so they get replaced with `jest-glamor-react`.

src/babel.js
```diff
@@ -79,6 +80,64 @@ export function replaceCssWithCallExpression (
...
+const getNextId = (state) => {
+  const id = state.file.get(COMPONENT_POSITION) || 0
+  state.file.set(COMPONENT_POSITION, id + 1)
+  return id
+}
+
+const getComponentId = (state) => {
+  // Prefix the identifier with css- because CSS classes cannot start with a number
+  // Also in snapshots with jest-glamor-react the hash will be replaced with an index
+  return `css-${getFileHash(state)}${getNextId(state)}`
```
tkh44: Should we make these prefixed with `emotion` because if they start with `css` then we are going to try to look up that id from the rule cache.

mitchellhamilton: `jest-glamor-react` only replaces classes that have the `css-` prefix. I made it so if a class that starts with `css-` isn't in the rule cache it's just concatenated.

src/react/index.js
```diff
@@ -16,7 +16,7 @@ export {
 
 const push = (obj, items) => Array.prototype.push.apply(obj, items)
 
-export default function (tag, objs, vars = [], content) {
+export default function (tag, cls, objs, vars = [], content) {
```
tkh44: Could we just insert the `cls` into the obj array

mitchellhamilton: We could but in the future I think we should remove `objs` and move all the things from composes into content

tkh44: I am in 100% agreement on moving it all into content. I'm going to merge this in as is but I think we should focus on that next.

PR #213, opened 2017-07-31
Allow random expressions as long as they result in a className that begins with css

### LINKED ISSUE #179: Better Composition
Because of #154 we can make composition way more powerful. We _should_ be able to support interpolations anywhere.

This is what I'm thinking should be possible
```jsx
const flexCenter = css`
  display: flex;
  justify-content: center;
  align-items: center;
`

const CenteredComponent = styled.div`
  display: block;
  ${flexCenter}
`
```
In this example `display: flex;` should take priority over `display: block;` since the interpolation is after `display: block;`.

This should be possible since we can split the output object at each interpolation. We will likely have to use a custom version of `postcss-js`'s objectify to do this. We'll also have to make postcss think interpolations are at rules so they be parsed.

### LINKED ISSUE #217: Missing style definitions when using polished like mixin functions
- `emotion` version: 7.0.4
- `react` version: 15.6.1

**What you did:**
Created a mixin called `responsiveGutter()`, whose goal is to do the following:
At the various breakpoints in my application, setting different left/right paddings:

```javascript
// mixins.js
import { above } from './breakpoints';
import breakpoints from '../breakpoints';
import spacing from '../spacing';

export const spacingX = (type, size) => (
  `
  ${type}-left: ${size};
  ${type}-right: ${size};`
);

export const spacingY = (type, size) => (
  `
  ${type}-top: ${size};
  ${type}-bottom: ${size};`
);

export const responsiveGutter = () => (
  `
  ${spacingX('padding', spacing.sm)}

  ${above(breakpoints.sm)} {
    ${spacingX('padding', spacing.md)}
  }

  ${above(breakpoints.md)} {
    ${spacingX('padding', spacing.lg)}
  }`
);
```

**What happened:**

Now in my component...

```javascript
...
import { responsiveGutter } from '../../theme/mixins/spacing';
...

const styles = css`
  padding-top: 3.2rem;
  margin: 0 auto;
  max-width: ${breakpoints.lg};
  ${responsiveGutter()}
`;

```

If I log the template literal I am passing into css, I get the following (see below), however, the media queries that contain the padding declarations are never applied to the element.

```
padding-left: 1.6rem;
  padding-right: 1.6rem;

  @media (min-width: 375px) {
    
  padding-left: 2.4rem;
  padding-right: 2.4rem;
  }

  @media (min-width: 719px) {
    
  padding-left: 3.6rem;
  padding-right: 3.6rem;
  }
  padding-top: 3.2rem;
  margin: 0 auto;
  max-width: 960px;

  @media (min-width: 375px) {
    padding-top: 3.2rem;
  }
```

emotion

**Problem description:**
Styles declared in helper functions are not being applied to class.

tkh44:
random interpolations

closes #217 
closes #179

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -77,7 +77,7 @@ export function css (objs: any, vars: Array<any>, content: () => Array<any>) {
 
 function insertRawRule (css: string) {
   let spec = {
-    id: hash(css, css.length),
+    id: hash(css),
     css,
     type: 'raw'
   }
--- a/src/index.js
+++ b/src/index.js
@@ -234,27 +234,25 @@ function deconstruct (style) {
   let selects
   let medias
   let supports
+
   forEach(keys(style), key => {
     if (key.indexOf('&') >= 0) {
       selects = selects || {}
       selects[key] = style[key]
     } else if (key.indexOf('@media') === 0) {
       medias = medias || {}
-      const result = deconstruct(style[key])
-      console.log('result', result)
-      medias[key] = result
+      medias[key] = deconstruct(style[key])
     } else if (key.indexOf('@supports') === 0) {
       supports = supports || {}
       supports[key] = deconstruct(style[key])
     } else if (key.indexOf('css-') === 0) {
-      console.log('key', key)
-      console.log('style', style)
-      console.log('plain', plain)
-      console.log('medias', medias)
+      // replace fragments
       plain = plain || {}
-      plain = { ...plain, ...style[key] }
+      const registeredStyles = registered[key.split('css-')[1]].style
+      assign(plain, registeredStyles)
     } else {
       plain = plain || {}
+
       plain[key] = style[key]
     }
   })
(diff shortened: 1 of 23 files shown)


src/index.js
```diff
@@ -244,6 +244,8 @@ function deconstruct (style) {
     } else if (key.indexOf('@supports') === 0) {
       supports = supports || {}
       supports[key] = deconstruct(style[key])
+    } else if (key.indexOf('css-') === 0) {
+      plain = { ...plain, ...style[key]}
```
tkh44: silly way to get babels object assign

src/parser.js
```diff
@@ -89,3 +89,88 @@ export function expandCSSFallbacks (style: { [string]: any }) {
...
+function atRule (node) {
+  if (typeof node.nodes === 'undefined') {
+    return true
+  } else {
+    return objectifyPostcssRules(node)
+  }
+}
+
+// postcss -> js obj
+function objectifyPostcssRules (node) {
```
tkh44: This function needs to be removed.

src/parser.js
```diff
@@ -89,3 +89,88 @@ export function expandCSSFallbacks (style: { [string]: any }) {
...
+  parser.parse()
+
+  return parser.root
+}
+
+export class EmotionSafeParser extends SafeParser {
+  unknownWord (tokens) {
+    if (tokens[0][0] === 'word') {
+      if (/xxx(\d+)xxx/gm.exec(tokens[0][1])) {
+        this.decl(tokens)
```
tkh44: giphy 6

src/index.js
mitchellhamilton: We have an `assign` util

Issue #196, opened 2017-07-28
injectGlobal


Geczy:
- `emotion` version: 6.0.3
- `react` version: 15.6.1

Relevant code.

```javascript
import { injectGlobal } from 'emotion'

injectGlobal`
  ...
`
```

What happened:
```
ERROR in ./src/components/App.js                
Module build failed: SyntaxError: C:/Users/Matt/Documents/GitHub/frontend/src/components/App.js: NodePath has been removed so is read-only. (This is an error on an internal node. Probably an internal error)
    at File.buildCodeFrameError (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-core\lib\transformation\file\index.js:427:15)
    at NodePath.buildCodeFrameError (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\index.js:140:26)
    at NodePath._assertUnremoved (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\removal.js:70:16)
    at NodePath.remove (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\removal.js:20:8)
    at replaceGlobalWithCallExpression (C:\Users\Matt\Documents\GitHub\frontend\node_modules\emotion\lib\babel.js:264:23)
    at PluginPass.TaggedTemplateExpression (C:\Users\Matt\Documents\GitHub\frontend\node_modules\emotion\lib\babel.js:91:13)
    at newFn (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\visitors.js:276:21)
    at NodePath._call (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\context.js:76:18)
    at NodePath.call (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\context.js:48:17)
    at NodePath.visit (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\path\context.js:105:12)
    at TraversalContext.visitQueue (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\context.js:150:16)
    at TraversalContext.visitSingle (C:\Users\Matt\Documents\GitHub\frontend\node_modules\babel-traverse\lib\context.js:108:19)
[...]

PR #214, opened 2017-07-31
[Fix] #196

bogas04:
As per @mitchellhamilton's suggestion, removing this part should be good enough.

**What**: As per @mitchellhamilton  https://github.com/tkh44/emotion/issues/196#issuecomment-318642670, the if and else block is now just changed the contents of else block.

**Why**: This was causing an error on importing `injectGlobal` from 'emotion'.

**How**: As per @mitchellhamilton  "The optimisation ... isn't needed since uglify will remove the undefined expression anyway."

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/src/babel.js
+++ b/src/babel.js
@@ -44,11 +44,9 @@ export function replaceCssWithCallExpression (
       if (!removePath) {
         return path.replaceWith(t.stringLiteral(`${name}-${hash}`))
       }
-      if (t.isExpressionStatement(path.parent)) {
-        path.parentPath.remove()
-      } else {
-        path.replaceWith(t.identifier('undefined'))
-      }
+
+      path.replaceWith(t.identifier('undefined'))
+
       return
     }
 
--- a/test/babel/__snapshots__/font-face.test.js.snap
+++ b/test/babel/__snapshots__/font-face.test.js.snap
@@ -1,9 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`fontFace babel extract basic 1`] = `
-"import \\"./font-face.test.emotion.css\\";"
+"import \\"./font-face.test.emotion.css\\";
 
-undefined;
+undefined;"
 `;
 
 exports[`fontFace babel extract basic 2`] = `
--- a/test/babel/__snapshots__/font-face.test.js.snap
+++ b/test/babel/__snapshots__/font-face.test.js.snap
@@ -1,6 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`fontFace babel extract basic 1`] = `"import \\"./font-face.test.emotion.css\\";"`;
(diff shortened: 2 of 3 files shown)


tkh44:
@bogas04 can you please fill out basic information of this PR. It has a very vague title and none of the fields are filled out. I am much more likely to merge things that I can get the entire context of the situation. 

Thanks!

thangngoc89:
@tkh44 I think it's per @mitchellhamilton request here https://github.com/tkh44/emotion/issues/196#issuecomment-318642670

tkh44:
I understand that @thangngoc89, but I'm just asking nicely to use the provided template so that @mitchellhamilton and I can quickly scan the PR to get some context.

bogas04:
@tkh44 Sorry for being lazy! I've added more details to the description. I often use commit messages like '[Fix] #issuenumber' as they close the issue and also carry information in the `issuenumber`. But you're right, I should've filled the basic information.

bogas04:
@mitchellhamilton I think the fix isn't working as intended. We output `undefined;` even now. I'll try to investigate further.

mitchellhamilton:
@bogas04 it should output undefined. As I said the optimisation I was trying to do by removing the undefined isn't needed.

bogas04:
@mitchellhamilton In that case, are we supposed to update the tests ?

mitchellhamilton:
Yes

bogas04:
@mitchellhamilton 

Something like this in [font-face.test.js.snap#L3](https://github.com/tkh44/emotion/blob/master/test/babel/__snapshots__/font-face.test.js.snap#L3)

```diff
// Jest Snapshot v1, https://goo.gl/fbAQLP

- exports[`fontFace babel extract basic 1`] = `"import \\"./font-face.test.emotion.css\\";"`;
+ exports[`fontFace babel extract basic 1`] = `
"import \\"./font-face.test.emotion.css\\";"

undefined;
`; 
```

And in [inject-global.test.js.snap#L22](https://github.com/tkh44/emotion/blob/master/test/babel/__snapshots__/inject-global.test.js.snap#L22) as well.

```diff
- exports[`babel injectGlobal extract injectGlobal basic 1`] = `"import \\"./inject-global.test.emotion.css\\";"`;
+ exports[`babel injectGlobal extract injectGlobal basic 1`] = `
"import \\"./inject-global.test.emotion.css\\";"

undefined;
`;
```

mitchellhamilton:
Yes, that's exactly what it should be.

bogas04:
perfect, onto it :+1:

mitchellhamilton:
It looks like you updated the snapshots manually and they aren't don't match. I would recommend reading [this](http://facebook.github.io/jest/docs/en/snapshot-testing.html) to understand how snapshot tests work but basically press `u` when you're in watch mode and the snapshots will update.

bogas04:
Thank you for supporting @mitchellhamilton

PR #218, opened 2017-08-01
Fix SSR hydration

### LINKED ISSUE #216: SSR + hyrdate causes two sets of <style> tags to be created
See example repo and follow the steps to install: https://github.com/zeit/next.js/tree/master/examples/with-emotion

- `emotion` version (same as that in the example repo, but also tried `6.0.3` and `7.0.4`)
- `react` version (`15.6.1` per the example repo)

What happened:

You can see below that there are two sets of `<style>` tags injected and it appears emotion is not respecting the IDs that were passed in with `hydrate`.

6.0.3:

image

7.0.4:

image

I've also had this with my project using emotion but thought it would be easier to replicate by referring to the example repo.

mitchellhamilton:
**What**:

Fix SSR hydration for `keyframes`, `fontFace` and `injectGlobal`

I also added some tests for hydration to make sure styles don't get reinserted

Also, raw rules had hashes that were longer numbers so I changed the hashString export to call `.toString(36)` on the hash.

**Why**:

Because otherwise rules get reinserted on the client.

**How**:

Don't filter out rules with the type `raw` or `keyframes` in `extractCritical`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

Closes #216 

I found an interesting problem while I was doing this which I'm not entirely sure how to solve.

Let's say you have a component which composes a class.

```jsx
const cls1 = css`
  color: hotpink;
`
  composes: ${cls1};
  background-color: yellow;
`
```

`cls1` won't be in the html so it won't be included in the `css` from `extractCritical` so it's inserted on the client.

Maybe we could solve this by inserting rules in `toString` so if they're not used as strings they're not inserted?

I'm not sure if this really matters that much since `insertRule` is so fast but in a large app with lots of `composes` lots of rules will be inserted that are never actually used.

@threepointone Would this also be a problem in glamor? if not how does glamor solve it?
```jsx
const cls1 = css({ color: 'hotpink' })
const classThatWillBeInHtml = css({ backgroundColor: 'yellow' }, cls1)
```


tkh44:
I think that problem might exist in glamorous as well. 

cc @kentcdodds

kentcdodds:
Hmmm... Seems like most changes are glamor-based files. cc @threepointone

threepointone:
this is a *known* issue, and it's fine, it's one of the tradeoffs of injecting as soon as a rule is created. it's also one of the reasons we have a separate insertion and hash cache. shouldn't cause problems.

tkh44:
Sorry for the ping. I was just letting you know that this issue would appear in your apps if you are using glamorous because of how the component style objects are composed.

kentcdodds:
Sorry, the problem is a little unclear to me as is what a glamorous-level solution would look like 🤔

sassanh:
@mitchellhamilton I think I'm affected by the problem you found while solving this issue. Is it solved? If yes is there a sample repo showing how to use emotion hydration? If no is there any workaround?

Issue #212, opened 2017-07-30
RFC: remove theming


tkh44:
I'm starting to think we should remove built in theming and just let users use the `theming` library themselves. Having theming built in adds overhead and code complexity to `styled` that I don't think is necessary. I prefer to just import my theme file but I may be an outlier.

Some notes.
If we remove theming...

- `Styled` becomes a pure functional component and we can create a factory with [createEagerFactory](https://github.com/acdlite/recompose/blob/master/docs/API.md#createeagerfactory) to call it 🔥
- Bundle size will drop
- Performance will be better
- Users that want theming can just wrap styled with withTheme
- We **may** be able to add theming conditionally with babel option (we could just wrap styled with `withTheme` and import it if the user wants theming)

Connorelsea:
I'd be happy with the decrease in bundle size when not using theming. Emotion currently sitting around 3.4kb in my app

![image](https://user-images.githubusercontent.com/1844458/28756157-7f487694-752e-11e7-92e0-075ac92bee1c.png)

thangngoc89:
@Connorelsea since version 6.0 , you have to opt-in for themeing to use it.

tkh44:
closed in #225

ericclemmons:
Personally, I wouldn't mind if `theming` could be automatically like so:

```js
import styled from "emotion/react/themed";

export default styled.div`...`;
```

Which would be the equivalent of:

```js
import styled from "emotion/react";
import { withTheme } from "theming";

export default withTheme(styled.div`...`);
```

Having converted a large `styled-components` repo to `emotion`, everything was largely find/replace _besides_ wrapping all of the components with `withTheme`.

PR #225, opened 2017-08-02
#212 - remove theming

tkh44:
Putting this up early to get more eyes on it.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/test/__snapshots__/react.test.js.snap
+++ b/test/__snapshots__/react.test.js.snap
@@ -193,13 +193,13 @@ exports[`styled component as selector function interpolation 1`] = `
 `;
 
 exports[`styled composes 1`] = `
-.glamor-1 {
+.glamor-2 {
   color: blue;
   font-size: 32px;
 }
 
 <h1
-  className="glamor-0 legacy__class glamor-1"
+  className="glamor-0 glamor-1 legacy__class glamor-2"
   scale={2}
 >
   hello world
@@ -232,7 +232,7 @@ exports[`styled composes based on props 2`] = `
 `;
 
 exports[`styled composes with objects 1`] = `
-.glamor-1 {
+.glamor-2 {
   color: #333;
   font-size: 32px;
   height: 64px;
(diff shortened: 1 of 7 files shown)


test/__snapshots__/react.test.js.snap
```diff
@@ -520,15 +537,15 @@ exports[`styled random object expression 1`] = `
...
+.glamor-3 {
   background-color: #ffd43b;
   color: #8c81d8;
   height: 64px;
   font-size: 32px;
 }
 
 <span
-  className="glamor-0 glamor-1"
+  className="glamor-0 glamor-1 glamor-2 glamor-3"
```
tkh44: @mitchellhamilton I'm not sure how to interpret these snapshot diffs. Do they look right to you?

mitchellhamilton: They don't, there should only be two classes on each component.

tkh44: This is the actual output on a web page. Maybe this will have a clue.

image

tkh44: Notice the difference in length of the hash.

tkh44: This is due to wrapping the components with `withTheme` I believe. I tested this out on the browser and it works as expected.

mitchellhamilton: Yes, that's correct because they're wrapped in `withTheme` so more components are rendered and the class is added for using components as selectors.

Now that I think about it, we might want to always add the class for using components as selectors because in this example  `OneComponent` will have `color: yellow;` when rendered inside of `SomeComponent` but `AnotherComponent` will not.
```jsx
const OneComponent = styled.div`
  display: flex;
`
const AnotherComponent = styled(OneComponent)`
  background-color: hotpink;
`
  ${OneComponent} {
    color: yellow;
  }
`

tkh44: Do you think I should hold up merging this for that fix or is it something we can separate?

mitchellhamilton: I already pushed a fix to this branch 😀

PR #232, opened 2017-08-05
Allow Styled tags to be updated via withComponent

ifyoumakeit:
**What**:
Added `withComponent` method to allow `styled` components to update tag name. Components can also be extended via this method.

**Why**:
Tags may need to be changed to comply with accessibility and HTML standards. One use case is headings (h1, h2, etc...) that may need to be swapped depending on document depth.

The method na

**How**:
Added a new method to the `Styled` component that re-calls `createStyled` with the new tag/component.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


ifyoumakeit:
Fixing tests today.

ifyoumakeit:
Running into some errors due to updates. Adding some documentation for `sites` too.

ifyoumakeit:
Fixed up errors, but running into an issue where components created with `withComponent` lose their named css class. So `css-Avatar-{xxxxx} css-{xxxxx}` becomes `css-{x} css-{xxxxx}`

See Avatar class name here: https://deploy-preview-232--emotion.netlify.com/#withcomponent

Example.
```html
<span data-reactroot="" class="css-Circle-wrfx8o0 css-mzfoqi">
  <img src="/a76dfa0d18a0536af9e917cdb8f873b9.png" class="css-1 css-mzfoqi">
</span>
```

@tkh44, could this cause collisions? I was able to get it to use the same class as Circle `css-Circle-wrfx8o0` but that seemed more confusing than helpful.

mitchellhamilton:
@ifyoumakeit This is because we create those unique classes with the identifier in the babel plugin. We could create a unique class for `withComponent` calls but that wouldn't work in the babel macro. We could make an export named `withComponent` or something instead of a method that accepts a styled component and a new tag which could work in the babel macro. I'm not really sure what we should do. @tkh44 what do you think?

tkh44:
I need to think about this a bit more. We want to get this in version 8, but I want to do it right.

ifyoumakeit:
Makes sense, definitely want to make sure the solution is performant/useful and fits in well with the future of emotion.

packages/react-emotion/src/index.js
```diff
@@ -90,6 +90,12 @@ export default function(tag, options: { e: string }) {
     Styled.__emotion_styles = styles
     Styled.__emotion_base = baseTag
 
+    Styled.withComponent = nextTag => {
+      return createStyled(nextTag, options)(styles)
```
ifyoumakeit: 👍 this is much cleaner and more obvious from first glance. Thanks @mitchellhamilton .

PR #236, opened 2017-08-05
Filter props on composed components

mitchellhamilton:
**What**:

Filter props on components that are styled components that have string tags. For example the props on `Link` will now be filtered.
```jsx
const BaseLink = styled.a`background-color: hotpink;`
const Link = styled(BaseLink)`color: green;`
```

**Why**:

Because otherwise there will be unknown prop warnings on composed styled components.

**How**:

Move the declaration of `spec` to be before `Styled` is defined and check wether the first spec's tag is a string instead of just the tag.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

This is also a bit smaller now. 🎉

### DIFF
--- a/src/react/index.js
+++ b/src/react/index.js
@@ -27,14 +27,25 @@ export default function(tag, cls, objs, vars = [], content) {
   }
 
   const componentTag = tag.displayName || tag.name || 'Component'
+  const spec = {
+    vars,
+    content,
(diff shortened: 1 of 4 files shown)


tkh44:
This should fix factories as well! Awesome work.

PR #243, opened 2017-08-08
Pass filename to PostCSS correctly

### LINKED ISSUE #242: Autoprefixer ignores configuration
- `emotion` version: 7.0.11
- `react` version: 16.0.0-alpha.13

Relevant code.

```javascript
const Styled = styled.div`
  display: flex;
`
```

Problem description:
I tried to configure Autoprefixer by setting the Browserslist config in a few ways

 - Using the browserslist key in the package json:
```json
"browserslist": [
  "last 1 Chrome version"
],
``` 

 - Having a .browserslistrc
```
[production]
last 1 Chrome version

[development]
last 1 Chrome version
```

- Setting the BROWSERSLIST environment variable. `BROWSERSLIST="last 1 Chrome version"`

What happened:

Only when setting the env variable did I get the expected result:
```css
.css-1fe3owl {
    display: flex;
}
```

In the other two cases I got:
```css
.css-71uu79 {
    display: -webkit-box;
    display: -ms-flexbox;
    display: flex;
}
```

mitchellhamilton:
Closes #242

**What**:

Pass filename to PostCSS correctly so browserslist can read configs correctly.

**Why**:

#242 

**How**:

For tagged template literals: don't objectify the root node before passing it to the prefixer so that source nodes remain.
For objects: create an `Input` object and set it on the root node before passing it to the prefixer. (it's pretty hacky)

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/src/ast-object.js
+++ b/src/ast-object.js
@@ -1,7 +1,10 @@
+import postcssJs from 'postcss-js'
+import Input from 'postcss/lib/input'
 import { expandCSSFallbacks, prefixer } from './parser'
 import { forEach, reduce } from './utils'
+import { getFilename } from './babel'
 
-function prefixAst(args, t) {
+function prefixAst(args, t, path) {
   function isLiteral(value) {
     return (
       t.isStringLiteral(value) ||
@@ -11,7 +14,7 @@ function prefixAst(args, t) {
   }
 
   if (Array.isArray(args)) {
-    return args.map(element => prefixAst(element, t))
+    return args.map(element => prefixAst(element, t, path))
   }
 
   if (t.isObjectExpression(args)) {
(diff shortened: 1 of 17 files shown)


src/ast-object.js
```diff
@@ -75,7 +78,12 @@ function prefixAst(args, t) {
           : property.value.value
 
         const style = { [property.key.name]: propertyValue }
-        const prefixedStyle = expandCSSFallbacks(prefixer(style))
+        const parsedStyle = postcssJs.parse(style)
+        parsedStyle.source = {}
+        parsedStyle.source.input = new Input(parsedStyle.toString(), {
```
tkh44: Clever!

Issue #219, opened 2017-08-02
Doc cleanup for v7


tkh44:
Problem description:
We are close to releasing v7 and need help going over the docs to make sure they are correct.

Suggested solution:
Pick a section proof read it 😄 

If you find something wrong either describe it here or better yet open a PR and just reference this issue.

Thanks!

Geczy:
https://github.com/tkh44/emotion/blob/master/example/webpack.config.js

Still references extract, when extract was removed

PR #252, opened 2017-08-10
Add documentation for nested selectors

silvenon:
The documentation for nested selectors was missing, so I read the relevant tests and comments to figure out how it works exactly and added the documentation. I'm differentiating between three types of nested selectors:

1. regular selectors
2. class selectors from `css`
3. component selectors from `styled`

The first two require a leading `&`, but component selectors don't work that way, the `&` must be omitted. This inconsistency should probably be fixed at some point.

This is a part of #219.

### DIFF
--- a/docs/nested.md
+++ b/docs/nested.md
@@ -1,1 +1,65 @@
 ## Nested Selectors
+
+Sometimes you will want to nest selectors to target only elements inside the current class or React component. Here is an example of a simple element selector nested in the class generated with `css`:
+
+```jsx
+import { css } from 'emotion';
+
+const paragraph = css`
+  color: gray;
+
+  & a {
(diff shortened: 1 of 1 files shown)


docs/nested.md
```diff
@@ -1 +1,63 @@
...
+
+```jsx
+const link = css`
+  color: hotpink;
+`;
+
+const paragraph = css`
+  color: gray;
+
+  & ${link} {
```
tkh44: Does this really work? I think it would but you would need the `.` before the interpolation.

silvenon:
Ok, I added a sentence explicitly explaining why to add the dot `.` in case someone misses it when reading the code snippet. I also squashed everything into a single commit, I hope that's a desired practice.

Issue #258, opened 2017-08-12
Module parse failed in emotion/src/index.js (12:19) and emotion/src/utils.js (3:5), unexpected token


granmoe:
I'm getting the following error output from webpack when attempting to build a simple component that uses emotion:

```
WARNING in ./~/emotion/src/react/index.js
81:11-12 "export 'createElement' (imported as 'h') was not found in 'react'

ERROR in ./~/emotion/src/utils.js
Module parse failed: /mnt/c/Users/Granmoe/projects/boilerplates/react-redux-saga-boilerplate/node_modules/emotion/src/utils.js Unexpected token (3:5)
You may need an appropriate loader to handle this file type.
| // @flow weak
| export function omit(
|   obj: { [string]: any },
|   testFn: (key: string, obj: any) => boolean
| ) {
@ ./~/emotion/src/react/index.js 3:0-52
@ ./~/emotion/lib/react/index.js
@ ./~/emotion/react/index.js
@ ./src/components/app.jsx
@ ./src/main.jsx
@ multi (webpack)-dev-server/client?http://localhost:3000 webpack/hot/dev-server ./src/main.jsx

ERROR in ./~/emotion/src/index.js
Module parse failed: /mnt/c/Users/Granmoe/projects/boilerplates/react-redux-saga-boilerplate/node_modules/emotion/src/index.js Unexpected token (12:19)
You may need an appropriate loader to handle this file type.
| sheet.inject()
|
| export let inserted: { [string | number]: boolean | void } = {}
|
| export function flush() {
@ ./~/emotion/src/react/index.js 2:0-30 6:0-14:17
@ ./~/emotion/lib/react/index.js
@ ./~/emotion/react/index.js
@ ./src/components/app.jsx
@ ./src/main.jsx
@ multi (webpack)-dev-server/client?http://localhost:3000 webpack/hot/dev-server ./src/main.jsx
```

I read something in another issue about problems being caused by having your webpack entry file in a source dir named "src," which mine is, as you can see below.
 
I do have the emotion babel plugin in my babelrc:

 ```{
  "env": {
    "production": {
      "presets": [["es2015", { "modules": false }], "react", "stage-1"],
      "plugins": ["emotion/babel"]
    },
    "test": {
      "presets": [
        ["env", {
          "targets": {
            "chrome": 60
          }
        }],
        "react",
        "stage-1"
      ],
[...]

PR #259, opened 2017-08-14
Lerna

mitchellhamilton:
**What**:

Move everything into their own packages

- emotion (core + aliases to the other packages for backwards compat until we do a major release)
- babel-plugin-emotion
- emotion-server 
- emotion-utils (internal utils/stuff that's used across packages)
- react-emotion

emotion, emotion-utils and react-emotion have rollup builds with all their dependencies external so they're reused across the packages.
babel-plugin-emotion and emotion-server have babel builds

**Why**:

- ES Module builds
- Fixes issues like #258 

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Code complete

[I wrote a codemod](http://astexplorer.net/#/gist/df53254b550cd46f9c1ae2ae75b0383f/a681af8dae247fe8edc426df444a200cb90c6818) to change the imports but I'm not sure where we should put it. Should we link to it? put it in the emotion package? another package? a folder in the repo?


babel-plugin-emotion-test.js
```diff
@@ -0,0 +1,7 @@
+require('babel-register')
```
tkh44: Does this have to be in the base directory?

PR #266, opened 2017-08-18
chore: fix install README babel instructions

probablyup:


probablyup:
Ah. Is that the recommended way to use emotion going forward? It was kind
of nice having things coalesced in the main package.

mitchellhamilton:
@probablyup It is. Currently, everything is aliased so all the old imports still work but we'll likely remove them in the next major version.

probablyup:
@mitchellhamilton updated

probablyup:
@mitchellhamilton Why'd you add `emotion` as well? It's already a hard dependency for both of those packages.

mitchellhamilton:
@probablyup Some people don't want react-emotion.

probablyup:
Perhaps, but I wouldn't want both of those installed at the same time. That'd be confusing.

PR #276, opened 2017-08-20
chore: update packages to deeplink into their respective directories

probablyup:
I think this makes slightly more sense than just linking to root. The visitor can always back up if they want to.

This might also be a good time to start thinking about individual package READMEs. The NPM pages are a little bare without them: https://www.npmjs.com/package/react-emotion

### DIFF
--- a/packages/babel-plugin-emotion/package.json
+++ b/packages/babel-plugin-emotion/package.json
@@ -29,10 +29,7 @@
   "author": "Kye Hohenberger",
   "homepage": "https://github.com/tkh44/emotion#readme",
   "license": "MIT",
-  "repository": {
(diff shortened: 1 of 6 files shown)


probablyup:
I'm not sure about pulls, but I think issues uses a different property: https://github.com/tkh44/emotion/pull/276/files#diff-54043be09c6d011d84adfabf6e46686aR34

tkh44:
If we are removing the git reference we should just use this format:

https://github.com/babel/babel/blob/7.0/packages/babel-plugin-transform-decorators/package.json#L7

tkh44:
Thanks!

PR #279, opened 2017-08-22
Add dynamic value support for extract static in the styled function.

tkh44:
**What**:
Add dynamic value support for `styled` (react components) when using extractStatic mode. This intentionally handicapped to just `styled` because of the inline style requirement. 

Todos:
- [x] the `styled` component is one ugly hack and needs to be cleaned up.
- [ ] css var support detection (use whatever postcss uses to determine if css var support is expected)
- [ ] documentation on the limitations. 

This will work
```javascript
const H1 = styled.h1`
  font-size: ${p => p.fontSize};
`
```

**Why**:
It's possible.

Make @ai happy.

**How**:
This was actually a very straight forward process.
1. Always create the static css 
2. replace any placeholders with the className and index `${name}-${hash}-${index}`
3. detect if there are `vars` but no content function in the higher order component
4. add the inline style defs for the css vars to the elements props

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


packages/babel-plugin-emotion/src/index.js
```diff
@@ -144,21 +144,40 @@ const getComponentId = (state, prefix: string = 'css') => {
...
-    const { staticCSSRules } = parseCSS(cssText, true, getFilename(path))
+    const { staticCSSRules, composesCount } = parseCSS(
+      cssText,
+      true,
+      getFilename(path)
+    )
+
+    // Help Needed:
+    // We should read the browser preferences for postcss
+    // and turn this on only if css vars are supported in their browser targets.
```
mitchellhamilton: https://github.com/ai/browserslist https://github.com/ben-eb/caniuse-lite

ai: The best way to get browsers:

```js
browsers = browserslist(undefined, { path: cssRoot.source && cssRoot.source.input.file })
```

Autoprefixer and Stylelint uses the same way.

tkh44:
@mitchellhamilton I have a PR in for https://github.com/developit/preact/issues/666 here https://github.com/developit/preact/pull/827

packages/emotion/src/index.js
```diff
@@ -82,7 +83,38 @@ export function css(objs: any, vars: Array<any>, content: () => Array<any>) {
   return computedClassName.trim()
 }
 
-function insertRawRule(css: string) {
+type inputVar = string | number
+
+export function customProperties(baseClassName: string, vars: Array<inputVar>) {
+  const hash = hashArray([baseClassName, ...vars])
+  const varCls = `css-vars-${hash}`
```
tkh44: I did this so that users could SSR and extract the custom property values, but I'm not sure if it will trip up in other places.

mitchellhamilton: I have to write a test for it but I think we have to do something different to get SSR to work.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -144,21 +144,40 @@ const getComponentId = (state, prefix: string = 'css') => {
...
+    const objs = path.node.quasi.expressions.slice(0, composesCount)
+
+    if (objs.length) {
+      throw new Error('Cannot use composes in extractStatic mode.')
```
mitchellhamilton: @tkh44 Do you think we should just go to inline mode instead of throwing an error. We also need to check if there are any interpolations in other places before going to extract.

packages/emotion/src/index.js
mitchellhamilton: I was wrong! it works right now! 🎉

mitchellhamilton:
Just a thought: extracted `css` calls won't work in random interpolations.

PR #285, opened 2017-08-24
Upgrade bundlesize to 0.13.2

siddharthkp:
**What**:
Upgrading bundlesize to 0.13.2

**Why**:
This update adds a better UI for multiple files: https://github.com/siddharthkp/bundlesize/issues/55#issuecomment-315567994

**How**:
Update version in `package.json`

**Checklist**:
- [ ] Documentation: NA
- [ ] Tests: NA
- [ ] Code complete: NA

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "babel-preset-flow": "^6.23.0",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-0": "^6.24.1",
-    "bundlesize": "^0.5.7",
+    "bundlesize": "^0.13.2",
     "css": "^2.2.1",
     "enzyme": "^2.9.1",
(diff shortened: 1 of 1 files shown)


siddharthkp:
@mitchellhamilton bundlesize only works on travis-ci/push though 😢 

Should I make that configuration? 🤔

mitchellhamilton:
I think it only works on push because the environment variable is secret so it's not exposed on forks. I think you could do this without having to have the API key be public with [this](https://docs.travis-ci.com/user/jwt) or something else, like what codecov does how it doesn't require an environment variable at all(I'm not sure how that works though).

siddharthkp:
That's super helpful! I'll check what codecov does

tkh44:
Thank you @siddharthkp

Issue #267, opened 2017-08-18
Babel plugin doesn't work if you rename the export


probablyup:
e.g.

```js
import emotion from 'emotion/react'; // instead of import styled

const Thing = emotion.div`
    color: red;
`;
```

It'd be nice if the plugin detected whatever the default export was being called and handled it.

mitchellhamilton:
This is supported in [the babel macro](https://github.com/tkh44/emotion/blob/master/docs/babel-macros.md). I personally don't think we should do this in the regular babel plugin because it would mean it wouldn't be possible to use other styled runtimes like [emotion-vue](https://github.com/egoist/emotion-vue) and it would make the playground on the website much more difficult.

probablyup:
It's very confusing. At the very least, a warning message would be helpful,
because otherwise it looks like a random garbage error during compilation.

mitchellhamilton:
I think a warning could be a good idea. @tkh44 What do you think?

probablyup:
To be clear though, I do strongly think that the babel plugin should detect the left side of the import and not make assumptions about its naming. Even if there's a warning, it's an unexpected behavior that is not difficult to assuage.

I'd even be willing to try and do it myself if you want.

tkh44:
Short term: I want to put some kind of warning. 

I've been thinking about this a while. I want to make the whole thing customizable from the babel options.

```javascript
[
  'emotion/babel', 
  { 
    styledName: 'emotion', 
    cssName: 'c', 
    extractStatic: false 
  }
]
```

Something like that. What do you think @mitchellhamilton? This should be fairly simple to do, but I may be missing something.

mitchellhamilton:
@tkh44 I totally agree with making it customizable, probably something more like this though.
```js
[
  'emotion/babel', 
  { 
    imports: {
      css: 'c',
      styled: 'emotion'
    },
    extractStatic: false 
  }
]

probablyup:
What's the benefit of making it a config instead of autodetecting it in the code?

mitchellhamilton:
Autodetecting would mean we would we're forced to make assumptions about the way people are loading the runtime and what runtime they're using.

probablyup:
Hmm, I get what you're saying but if I make a variable named "emotion" that isn't an import, it'd get caught by the simple config right?

mitchellhamilton:
[...]

PR #290, opened 2017-09-01
Configurable imports

ChristopherBiscardi:
~~Built on top of lint-fix changes (#289) -- [compare between these two branches](https://github.com/ChristopherBiscardi/emotion/compare/lint-fix...ChristopherBiscardi:configurable-imports?expand=1)~~

**What**:

Solves #267 by implementing a solution that will use the default import name instead of sticking only to `styled`.

in short, this works:

```js
import lol from 'emotion';

lol`color: blue`;
```

**Why**: See #267 

**How**: 

Modified the visitor to also visit ImportDeclarations. Then cache the values of any `import whatever from 'emotion'` in an object for later use when looking for `styled` calls.

**Checklist**:

These need work. This is just a POC. There definitely needs to be more tests.

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/docs/configurable-imports.md
+++ b/docs/configurable-imports.md
@@ -32,6 +32,10 @@ a prop other than `css` for processing.
 }
 ```
 
+Beware that if you use the babel configuration, you must import as the
+same name. In the previous example, we would have to `import { css as
+cows } from 'emotion';` then use `cows` to construct the template
+literals. 
 
 # Use Case
 
@@ -40,7 +44,7 @@ styled-jsx application. When compiling the following file with emotion
 and styled-jsx.
 
 ```js
-import styled from "react-emotion";
+import styled, { css } from "react-emotion";
 
 export default () => (
   <div>
@@ -61,8 +65,7 @@ outputs.
 
 ```js
 import _JSXStyle from "styled-jsx/style";
-import { css as _css } from "emotion";
-import styled from "react-emotion";
+import styled, { css } from "react-emotion";
 
 export default (() => <div data-jsx={2648947580}>
     <p data-jsx={2648947580}>only this paragraph will get the style :)</p>
@@ -88,11 +91,12 @@ By adding the babel opt config rename as such.
 }
 ```
 
-We can avoid re-compiling the `css` props from styled-jsx.
+We can avoid re-compiling the `css` props and instead use `cows` for
+our template literals, etc.
 
 ```js
 import _JSXStyle from "styled-jsx/style";
-import styled, { css } from "react-emotion";
+import styled, { css as cows } from "react-emotion";
 
 export default (() => <div data-jsx={2648947580}>
     <p data-jsx={2648947580}>only this paragraph will get the style :)</p>
@@ -100,3 +104,4 @@ export default (() => <div data-jsx={2648947580}>
     <_JSXStyle styleId={2648947580} css={"p[data-jsx=\"2648947580\"]{color:red}"} />
   </div>);
 ```
+
--- a/docs/configurable-imports.md
+++ b/docs/configurable-imports.md
@@ -1,12 +1,13 @@
 # Configurable Imports
 
-If you are using ES Module imports (`import styled from 'emotion'`)
-the emotion babel plugin can handle two types of import renaming.
+If you are using ES Module imports (`import styled from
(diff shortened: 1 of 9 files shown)


packages/babel-plugin-emotion/src/index.js
```diff
@@ -228,13 +228,50 @@ export function replaceCssObjectCallExpression(path, identifier, t) {
...
+    css: 'css',
+    keyframes: 'keyframes',
+    injectGlobal: 'injectGlobal'
+  }
+
   return {
     name: 'emotion', // not required
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
+      ImportDeclaration: {
```
tkh44: Can you move this below `Program`?

packages/babel-plugin-emotion/src/index.js
```diff
@@ -228,13 +228,50 @@ export function replaceCssObjectCallExpression(path, identifier, t) {
...
+      names.default = singleImport.local.name
+    }
+  })
+
+  return names
+}
 export default function(babel) {
   const { types: t } = babel
 
+  let importedNames = {
```
tkh44: The `importedNames` arg values come from `state.opts`. We might also want to set the imported names to `state.importedNames`

packages/babel-plugin-emotion/src/index.js
```diff
@@ -228,13 +228,50 @@ export function replaceCssObjectCallExpression(path, identifier, t) {
...
+    injectGlobal: 'injectGlobal'
+  }
+
   return {
     name: 'emotion', // not required
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
+      ImportDeclaration: {
+        enter({ node }) {
+          if (node.source.value === 'emotion') {
```
mitchellhamilton: I think this should be `node.source.value.indexOf('emotion') !== -1` so it includes emotion, react-emotion, preact-emotion and any other runtimes people create.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -228,6 +228,34 @@ export function replaceCssObjectCallExpression(path, identifier, t) {
...
+      // Is default import
+      names.default = singleImport.local.name
+    }
+  })
+
+  return names
+}
+
+const defaultImportedNames = {
+  default: 'styled',
```
tkh44: Can you change this from `default` to `styled`

packages/babel-plugin-emotion/src/index.js
```diff
@@ -228,6 +228,34 @@ export function replaceCssObjectCallExpression(path, identifier, t) {
...
+
+  imports.forEach(singleImport => {
+    if (singleImport.imported) {
+      // Is helper method
+      names[singleImport.imported.name] = singleImport.local.name
+    } else {
+      // Is default import
+      names.default = singleImport.local.name
+    }
+  })
```
mitchellhamilton: This can be done with `state.file.metadata.modules.imports` in the `Program` enter visitor.

http://astexplorer.net/#/gist/39f0d575f432d2a93f29c0aebe34956c/36992fb4d485246b651b79d3579756ef016c9c5c

ChristopherBiscardi: Thanks for the suggestion. Was able to remove `parseImports` and the `ImportDeclaration` visitor.

docs/configurable-imports.md
```diff
@@ -0,0 +1,101 @@
+# Configurable Imports
+
+If you are using ES Module imports (`import styled from 'emotion'`)
```
mitchellhamilton: `styled` doesn't exist in emotion. It's only in react-emotion and preact-emotion.

docs/configurable-imports.md
```diff
@@ -0,0 +1,102 @@
...
+    ["emotion", { "importedNames": { "css": 'cows' }}]
+  ]
+}
+```
+
+We can avoid re-compiling the `css` props from styled-jsx.
+
+```js
+import _JSXStyle from "styled-jsx/style";
+import styled, { css } from "react-emotion";
```
tkh44: should this say `css as cows`?

ChristopherBiscardi: yes

Issue #274, opened 2017-08-20
withTheme(Component) cannot be targeted as a component selector


probablyup:
I'm writing up some more benchmarks and I noticed a bug:

```js
import styled from 'emotion/react';
import { ThemeProvider, withTheme } from 'theming';

const Ancestor = styled.div`
    color: red;
`;

const Child = withTheme(styled.div`
    background-color: ${p => p.theme.backgroundColor};
    font-size: ${p => p.fontSize}px;
`);

const Parent = withTheme(styled.div`
    border: 1px solid ${p => p.theme.borderColor};
    color: ${p => p.color};

    ${Child} {
        font-family: monospace;
    }
`);

const theme = {
    backgroundColor: 'cornflowerblue',
    borderColor: 'antiquewhite',
};

ReactDOM.render(
    <ThemeProvider theme={theme}>
        <Ancestor>
            <Parent color="red">
                <Child fontSize={12} />
            </Parent>
        </Ancestor>
    </ThemeProvider>, MOUNT_NODE
);
```

results in an error due to the static properties emotion relies on not being hoisted by the `withTheme` ~constructor~ hoc. I made a PR against theming to fix it: https://github.com/iamstarkov/theming/pull/33

PR #292, opened 2017-09-05
Migrate "theming" library

probablyup:
Fixes: #274 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

~~The only thing left I think is to update the existing docs to point to `emotion-theming` instead.~~


evan-scott-zocdoc:
I'll finish this up tonight

probablyup:
I'll finish this up tonight

packages/emotion-theming/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "module": "dist/esm",
+  "files": [
+    "src",
+    "dist"
+  ],
+  "scripts": {
+    "clean": "rimraf dist",
+    "build": "run-s clean babel:*",
+    "babel:cjs": "cross-env BABEL_ENV=cjs babel src -d dist/cjs",
+    "babel:esm": "cross-env BABEL_ENV=esm babel src -d dist/esm"
```
mitchellhamilton: Could you change this to use the rollup setup we have for the other packages?

https://github.com/emotion-js/emotion/blob/master/packages/preact-emotion/package.json#L12-L14

packages/emotion-theming/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "devDependencies": {
+    "babel-cli": "^6.24.1",
+    "cross-env": "^5.0.1",
+    "npm-run-all": "^4.0.2",
+    "prop-types": "^15.5.8",
+    "react": "^15.5.4",
+    "react-dom": "^15.5.4",
+    "react-test-renderer": "^15.5.4",
+    "rimraf": "^2.6.1"
+  },
```
mitchellhamilton: Could you remove all the devDependencies except rollup, rimraf, npm-run-all and I think prop-types since it's not in the root package.json. It's better to have devDependencies in the root package.json because otherwise they have to be installed multiple times except for packages where we use binaries since binaries aren't linked.

package.json
```diff
@@ -8,7 +8,7 @@
     "clean": "lerna run clean --parallel",
     "test": "npm-run-all -p lint:check coverage test:size",
     "coverage": "jest --coverage --no-cache --ci --runInBand",
-    "lint:check": "eslint .",
+    "lint:check": "eslint . --fix",
```
mitchellhamilton: Please change this back. This is for CI to check that there are no linting errors.

probablyup: I will, debugging sometime else

probablyup:
Ok so I came up with an interesting bug while trying to finish this... looking for feedback.

Basically if you have a situation where you're using `styled()` on top of another emotion primitive wrapped with `withTheme()` it throws on render because I think emotion is processing all the css upfront before `withTheme` can inject the theme prop deeper in the stack.

probablyup:
We could get away with doing this:

```js
hoist(WithTheme, Component)
delete WithTheme.__emotion_spec
```

any objections?

packages/babel-plugin-emotion/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
+    "emotion-theming": "^1.0.0",
```
mitchellhamilton: This doesn't need to be a devDependency here since it's in jest's `moduleNameMapper`.

probablyup: I’m on vacation until Tuesday, so if you want to get this in sooner feel free to push that change on.

tkh44: No rush. Enjoy your vacation!

probablyup:
@tkh44 @mitchellhamilton all comments should be addressed

mitchellhamilton:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

probablyup:
@mitchellhamilton I’ve been thinking about this PR a lot and I think we should reintegrate this into react-emotion, rather than a separate library.

The use cases for it are extemely common and since we need to pay attention to the certain static class properties, it might make more sense to couple the implementations.

tkh44:
@probablyup I do not want to do this (right now). I think theming is overrated and there is a better way that we haven't thought of.

I know people coming from styled-components expect it though...

This is also possible
```javascript
import _styled from 'emotion'
export default (...args) => withTheme(_styled(...args))
```

tkh44:
@probablyup give it a shot. I want to slim it down though. Also, disable the `theme` check, please.

probablyup:
Wait why is component selector being removed? It’s quite hard to achieve the same functionality without first class support, and the majority of the reason I even opened this PR :/

mitchellhamilton:
@probablyup #329

tkh44:
Thank you for putting all this work in @probablyup 🍻

PR #293, opened 2017-09-05
Fix flowtype errors in babel-plugin-emotion

bigslycat:
**What**:

I received this errors using Flow version 0.54.0:

```
Error: packages/babel-plugin-emotion/src/index.js:121
121:         moduleName = JSON.parse(packageJsonContent).name
                                     ^^^^^^^^^^^^^^^^^^ Buffer. This type is incompatible with the expected param type of
485:     static parse(text: string, reviver?: (key: any, value: any) => any): any;
                            ^^^^^^ string. See lib: /tmp/flow/flowlib_2d62ca57/core.js:485

Error: packages/babel-plugin-emotion/src/parser.js:97
 97: export function parse(css, opts) {
                           ^^^ parameter `css`. Missing annotation

Error: packages/babel-plugin-emotion/src/parser.js:97
 97: export function parse(css, opts) {
                                ^^^^ parameter `opts`. Missing annotation
```

**Why**:

If we use Flow, we must not ignore this error reports.

**How**:

The type of the variable is explicitly converted to string. Function parse in parser.js is annotated.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -113,12 +113,12 @@ const getFileHash = state => {
     moduleRoot && relative(moduleRoot, filename).replace(pathSep, '/')
   let moduleName = ''
   if (moduleRoot) {
-    const packageJsonString = fs.readFileSync(
+    const packageJsonContent = fs.readFileSync(
       pathJoin(moduleRoot, 'package.json')
     )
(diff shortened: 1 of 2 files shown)


tkh44:
Released in 7.2.2

PR #296, opened 2017-09-06
Allow to set custom type for Element

RafalFilipek:
**What**: feature

**Why**: Right now there is no way to change tag / Component for styled component. For example with bootstrap you can use `btn` class for `button`, `a`, `span` etc. Lets say I have:

```js
const Text = styled('span')`/* styles */`;
```

Now in order to style `<p>` tag I have to move styles definition into separated variable and create new `styled('p')` or create completely new component. 

But lets talk about `flex-box` :) Some time ago I've created [just-box](https://github.com/RafalFilipek/just-box). You can find there `Box` component with `as` prop. This property allows you to render any element with magic flex-box styles. For example:

```js
<Box f1 column /> // <div>
<Box as="p" flexInline alignCenter /> // <p>
<Box as={Header} f2 /> // <Header />
<Box as="button" justifyEnd /> // <button>
```

**How**:  This PR simply adds `as` prop. if *falsy* value is provided `localTag` is used.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

Thank you!

### DIFF
--- a/packages/react-emotion/src/index.js
+++ b/packages/react-emotion/src/index.js
@@ -30,10 +30,6 @@ export default function(tag, cls, objs, vars = [], content) {
     tag.__emotion_spec !== undefined ? tag.__emotion_spec.concat(spec) : [spec]
   const localTag = newSpec[0].tag
 
-  const omitFn =
-    typeof localTag === 'string'
-      ? testOmitPropsOnStringTag
-      : testOmitPropsOnComponent
   function Styled(props, context) {
     const getValue = v => {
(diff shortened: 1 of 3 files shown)


b2whats:
++

tkh44:
@RafalFilipek I'm waiting on #299 before I move on this. I do not like the idea of an arbitrary prop name on an element. You could argue that `innerRef` is a made up prop and I agree, but it is more "standard" across the css-in-js landscape.

PR #298, opened 2017-09-08
upgrade to babel-macros 1.0.0

kentcdodds:
**What**: upgrade to babel-macros 1.0.0

**Why**: So folks get nice error messages

**How**: Add `babel-macros` dep so we can get the `createMacro` function, and use it.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete


tkh44:
Thanks @kentcdodds

PR #299, opened 2017-09-09
Use Stylis

mitchellhamilton:
**What**:
Use [Stylis](https://stylis.js.org).

**Why**:
- It's faster than using objects
- The runtime is slightly bigger but the prefixes are in the runtime instead of each time they're needed so bigger apps should be smaller
- Better prefixing, e.g. all objects will be prefixed.
- Some features will work without the babel plugin
- Random string interpolations

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


packages/emotion/src/index.js
```diff
@@ -1,531 +1,134 @@
...
 
-export let inserted: { [string | number]: boolean | void } = {}
+export const sheet = new StyleSheet()
 
-export function flush() {
-  sheet.flush()
-  inserted = {}
-  sheet.inject()
-}
+sheet.inject()
```
tkh44: `// 🚀` is good luck 😄

packages/emotion/src/index.js
```diff
@@ -1,531 +1,141 @@
...
-function insertRawRule(css: string) {
-  let spec = {
-    id: hash(css),
-    css,
-    type: 'raw'
-  }
-
-  register(spec)
+stylis.use([compositionPlugin, insertionPlugin])
+registerCacheStylis.use(compositionPlugin)
```
tkh44: `shadowStylis`?

tkh44:
image

Just sayin...

mitchellhamilton:
Just a note for anyone looking at that chart: new-css-in-js is what is in this PR.

probablyup:
Damn! Awesome!

docs/composition.md
```diff
@@ -0,0 +1,53 @@
+## Composition
```
tkh44: Nice!

docs/composition.md
```diff
@@ -0,0 +1,53 @@
...
+  align-items: center;
+`
+
+// interpolate it where you want to apply the styles
+const flexCenterClass = css`
+  ${flexCenter};
+  flex-direction: column;
+`
+
+// You can also it in in styled and the css prop too
```
greggb: typo?
> You can also it in in styled [...]

mitchellhamilton: Yep, thanks!

docs/composition.md
```diff
@@ -0,0 +1,53 @@
...
+const FlexCenterComponent = styled.div`
+  ${flexCenter};
+`
+
+
+const flexWrap = props => css`
+  flex-wrap: ${props.wrap};
+`
+
+// You can compose with use multiple classes
```
greggb: Suggestion:

>Compose with multiple classes.

docs/composition.md
```diff
@@ -0,0 +1,53 @@
...
+const ColumnCenteredComponent = styled.div`
+  ${flexCenter};
+  ${flexWrap};
+`
+
+// composition can get very powerful because the styles are expanded where you interpolate
+// for example, this class still has flex-direction: column; since
+// the style is interpolated after the flex-direction: row;
```
greggb: Suggestion:

>Composition can be very powerful. For example, styles are expanded where you interpolate, so the following class has `flex-direction: column` because `${flexCenter}` is interpolated after `flex-direction: row`

docs/composition.md
```diff
@@ -0,0 +1,53 @@
...
+const stillColumn = css`
+  flex-direction: row;
+  ${flexCenter}
```
greggb: Add `;` for consistency.

docs/composition.md
greggb: Suggestion:

>Interpolate in `styled` and css props too.

docs/composition.md
tkh44: great catch

docs/composition.md
```diff
@@ -0,0 +1,53 @@
...
+`
+
+// You can also use it in styled and the css prop
+const FlexCenterComponent = styled.div`
+  ${flexCenter};
+`
+
+
+const flexWrap = props => css`
+  flex-wrap: ${props.wrap};
```
mrmartineau: Suggestion:

Could there be an example with conditional styling based on props? e.g. 

```js
const ButtonCSS = (props) => css`
  display: ${props.block ? 'block' : 'inline-block'};
`
```

I'm sure there are other examples, but this is a start

mitchellhamilton:
WIP changelog

#### :boom: Breaking Change
- composes is gone and replaced with random interpolations (should we call them mixins like s-c or something else?) ([there's a codemod](https://github.com/emotion-js/emotion-composes-codemod))
- nested class selectors default to `& .class` instead of `&.class` if the `&` is omitted

#### :rocket: Enhancement
- Better performance

#### :bug: Bug Fix
- All edge cases with prefixing are gone

#### :house: Internal
- Don't convert tagged template literals to objects

PR #309, opened 2017-09-18
Docs: use a more decent alternative to 'cows'

rstacruz:
**What**:

- This changes the example `cows` to a more-likely-to-be-used `ecss`.

**Why**:

- The alias `cows` is arbitrary, but people are likely to copy-paste this code from documentation. This gives them a less silly name to start with.

**Checklist**:

- [x] Documentation
- [x] Tests - N/A
- [x] Code complete - N/A

### DIFF
--- a/docs/configurable-imports.md
+++ b/docs/configurable-imports.md
@@ -7,9 +7,9 @@ import renaming.
 ## Dynamic
 
 ```js
-import something, { css as ecss } from 'react-emotion';
+import something, { css as emotion } from 'react-emotion';
 
-const classes = ecss`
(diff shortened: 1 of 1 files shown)


tkh44:
Can you change `ecss` to `emotion`?

rstacruz:
okay. 

FWIW I prefer `ecss` over `emotion` because:

- it stands for `emotion css`
- it's close to `css`
- `emotion` seems to imply it's the default export of `'emotion'`, which it isn't

tkh44:
Thanks!

PR #320, opened 2017-09-21
Add Source Maps 🗺 

### LINKED ISSUE #15: sourcemaps?
Does it support css source maps like https://github.com/zeit/styled-jsx ?

tkh44:
**What**:
Add sourcemaps like styled-jsx has.

Only one sourcemap is allowed per style tag so we make each insert use a new tag. The rough plan is to insert the sourcemap for the user with the babel plugin in dev mode. I am not sure how to do this process for dev and the current process for production. I've never seen a babel plugin that checks NODE_ENV. Is this a thing?
  

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

closes #15


tkh44:
I still need to document this, but I do want to get it into the `next` release.

tkh44:
This doesn't work with object styles. I'm not sure how to handle them yet. I don't think this should block the merge though.

package.json
```diff
@@ -20,7 +20,10 @@
     "watch": "lerna run --parallel watch",
     "bootstrap": "lerna bootstrap"
   },
-  "dependencies": {},
+  "dependencies": {
+    "convert-source-map": "^1.5.0",
+    "source-map": "^0.5.7"
+  },
```
mitchellhamilton: Should these be a babel-plugin-emotion's package.json?

packages/emotion/src/index.js
```diff
@@ -170,9 +173,27 @@ function createStyles(strings, ...interpolations) {
...
+  } else {
+    const index = styles.indexOf('/*# sourceMappingURL')
+    if (index > -1) {
+      currentSourceMap = styles.substr(index)
+    }
+    parser(selector, styles)
+    currentSourceMap = ''
+  }
+}
+
```
mitchellhamilton: Maybe instead of putting the sourcemap inside of the styles we could make an object the first argument to createStyles if there is a sourceMap and do something like this.
```jsx
let sourceMap = ''

function createStyles(strings, ...interpolations) {
    if (strings !== undefined && strings.sourceMap !== undefined) {
      sourceMap = strings
      return createStyles(...interpolations)
    }
  }
  let stringMode = true
  let styles = ''
  if (
    (strings !== undefined && strings.raw === undefined) ||
    strings === undefined
  ) {
    stringMode = false
    styles = handleInterpolation(strings, false)
  } else {
    styles = strings[0]
  }
  interpolations.forEach((interpolation, i) => {
    styles += handleInterpolation(interpolation, isLastCharDot(styles))
    if (stringMode === true && strings[i + 1] !== undefined) {
      styles += strings[i + 1]
    }
  })
  return styles
}
```

tkh44: This is close to what I did.

tkh44:
emotion-source-maps-first-look

tkh44:
source-map-demo

tkh44:
source-map-demo

PR #326, opened 2017-09-22
At rule fixes

mitchellhamilton:
**What**:
Change the way at rules are joined

**Why**:
We had two plugins for insertion and it fixes bugs with non-nested at rules 

**How**:
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L339-L356

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

cc @giuseppeg


giuseppeg:
@mitchellhamilton cool, I fixed a few more bugs.. eg `@import` `@charset` are ignored in your patch. https://github.com/zeit/styled-jsx/blob/d4e21c38797da2cd5529953a617190abbcc0447b/src/lib/style-transform.js#L92

There are still [a few edge cases](https://github.com/zeit/styled-jsx/blob/d4e21c38797da2cd5529953a617190abbcc0447b/test/styles.js#L151-L154) where it fails but I might ignore them for now.

mitchellhamilton:
@giuseppeg Just had a quick look at your fixes, I'm pretty sure they would break if there was more than just the `@import` or `@charset` in the css because the -2 context is when the css is done processing and the content has all of the css. I think it would work if it was like this 
```js
if (context === 1) {
  if (content.charAt(0) === '@') {
    splitRulesQueue.push(content)
    return ''
  }
}
```

giuseppeg:
oh right silly me, thanks for the heads up

PR #327, opened 2017-09-22
Use cross-env to pass env vars across platforms

apostolos:
**What**: Fix test & build scripts in Windows

**Why**: rollup UMD builds need UMD=true to be set as an environment variable. This breaks test & build scripts on Windows machines making it hard for people to contribute.

**How**: Adds [cross-env](https://github.com/kentcdodds/cross-env) which takes care of differences to setting and using environment variables across platforms.

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

Only tested this in Windows and Linux. I don't currently have access to an OSX machine.


tkh44:
I can test this more tonight. Thank you for taking the time to help others @apostolos

PR #338, opened 2017-09-25
change "DO-NOT-USE" to "emotion.umd" in umd file names

tkh44:
I think we need to wait until 8 release because it will break readme.


richardanaya:
Woot!

richardanaya:
Would love to see this one merged in!

PR #340, opened 2017-09-25
Emotion system

tkh44:
**What**:

**Why**:
i want to see how fast this is

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


packages/emotion-system/src/index.js
```diff
@@ -0,0 +1,26 @@
...
+  ${fontSize};
+  ${color};
+  ${flex};
+  ${order};
+  ${wrap};
+  ${column};
+  ${direction};
+  ${align};
+  ${justify};
+`
```
mitchellhamilton: You can just have all these as arguments `styled('div')(display, ...otherArgs)`

tkh44: I'm just sketching. I want to show some patterns that devs can steal.

PR #343, opened 2017-09-25
Move babel-plugin-emotion to devDependencies

Andarist:
**What**:

Dependencies

**Why**:

`babel-plugin-emotion` is opt-in. Shouldnt be downloaded by each `emotion` consumer.

**How**:

Made a `package.json` tweak.

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/packages/emotion/package.json
+++ b/packages/emotion/package.json
@@ -20,11 +20,11 @@
     "rollup:umd": "UMD=true rollup -c ../../rollup.config.js"
   },
   "dependencies": {
-    "babel-plugin-emotion": "^8.0.2-6",
     "emotion-utils": "^8.0.2-3"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


tkh44:
This needs documentation in the install section.

Andarist:
@tkh44 Sure thing! [Installation section](https://github.com/emotion-js/emotion/blob/master/docs/install.md#install) already suggests separate installs, is anything else needed to be added?

mitchellhamilton:
@Andarist `babel-plugin-emotion` is used by `emotion/macro` so it at least needs to be a peerDependency and having it as a dependency means people don't have to worry about it.

 @tkh44 do you agree?

tkh44:
I think just having it as a dependency is fine.

PR #346, opened 2017-09-26
Function return values

tkh44:
**What**:
Allow the return value of dynamic values to be a function. This function will be called with zero arguments.
```javascript
const fontSize = () => 20;
const H1 = styled('h1')`font-size: ${theme.fontSize}px;`
```

**Why**:
@CompuIves has some really cool stuff in his theme. 
https://github.com/CompuIves/codesandbox-client/blob/master/src/common/theme.js#L23-L30

**How**:
Just call the interpolation if it is a function

**Checklist**:

- [ ] Documentation
- [X] Tests
- [X] Code complete


packages/emotion/src/index.js
```diff
@@ -116,6 +116,10 @@ function handleInterpolation(
     return ''
   }
 
+  if (typeof interpolation === 'function') {
+    return interpolation()
```
mitchellhamilton: I think this should be `handleInterpolation(interpolation())`

PR #347, opened 2017-09-26
Allow external stylis plugins

mitchellhamilton:
**What**:

Allow external stylis plugins.

**Why**:

So people can do things like RTL.

**How**:

Add a usePlugin export (I'm not sure about the name) which pushes the plugin to an array of external plugins, removes all stylis plugins, inserts the external ones and then the insertion plugin.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete


tkh44:
cc @thysultan 👍

PR #360, opened 2017-09-30
Handle boolean values

tkh44:
**What**:
Literal false values were tripping up the parser.

**Why**:
We were checking undefined and null but not false.

**How**:
Add explicit false check.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion/src/index.js
+++ b/packages/emotion/src/index.js
@@ -120,8 +120,7 @@ function handleInterpolation(
   if (
     interpolation === undefined ||
     interpolation === null ||
-    typeof interpolation === 'boolean' ||
-    typeof value === 'boolean'
+    typeof interpolation === 'boolean'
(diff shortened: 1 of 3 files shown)


packages/emotion/src/index.js
```diff
@@ -120,6 +120,7 @@ function handleInterpolation(
   if (
     interpolation === undefined ||
     interpolation === null ||
+    interpolation === false ||
     typeof value === 'boolean'
```
mitchellhamilton: I think this should be `typeof interpolation === 'boolean'`.

mitchellhamilton: @tkh44 The `typeof value === 'boolean'` can also be removed, I must've wrote value instead of interpolation by mistake.

PR #361, opened 2017-09-30
Add more context to the styled component displayName

### LINKED ISSUE #358: Better displayNames for React DevTools
Love this library and I am currently working on a POC to persuade my team and organization to use it in production. Thanks!

- `emotion` version: 7.3.2
- `react` version: 15.6.1

Problem description:
Currently, React DevTools shows emotion components like so: `styled(Component)`. Is there an option in the babel plugin to transform the displayName into something more debuggable/readable. 

Suggested solution:
I would prefer if the displayName was transformed to whatever the name of the component is when the variable is declared. Cheers!

tkh44:
**What**:
Add the base_tag displayName to the styled component

**Why**:
So they show up in React dev tools.
#358

**How**:
Append the property

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

closes #358

### DIFF
--- a/packages/react-emotion/src/index.js
+++ b/packages/react-emotion/src/index.js
@@ -116,11 +116,16 @@ const createStyled = (tag, options: { e: string }) => {
         )
       }
     }
+
     Styled.contextTypes = contextTypes
 
     Styled.__emotion_styles = styles
     Styled.__emotion_base = baseTag
     Styled.__emotion_real = Styled
+
+    Styled.displayName = `Styled(${typeof baseTag === 'string'
+      ? baseTag
+      : baseTag.displayName || 'Component'})`
 
     Styled.withComponent = nextTag => {
       return createStyled(nextTag, options)(styles)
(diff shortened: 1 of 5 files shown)


packages/react-emotion/src/index.js
```diff
@@ -116,12 +116,17 @@ const createStyled = (tag, options: { e: string }) => {
...
+
     Styled.contextTypes = contextTypes
 
     Styled.__emotion_styles = styles
     Styled.__emotion_base = baseTag
     Styled.__emotion_real = Styled
 
+    Styled.displayName = `Styled(${typeof baseTag === 'string'
+      ? baseTag
+      : baseTag.displayName || 'Component'})`
```
mitchellhamilton: Should `baseTag.name` also be here?

tkh44: Not sure where `name` comes from.

mitchellhamilton: screen shot 2017-09-30 at 1 31 59 pm

PR #375, opened 2017-10-03
Add label property and autoLabel option for babel-plugin-emotion

tkh44:
**What**:
This should help devs figure out where css code is written or the
component that it comes from. This should also open the doors for
`name` css property for devs to customize.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


packages/babel-plugin-emotion/src/babel-utils.js
```diff
@@ -1,10 +1,25 @@
 import { hashArray } from './index'
 
-export function getIdentifierName(path, t) {
+function getDeclaratorName(path, t) {
   const parent = path.findParent(p => p.isVariableDeclarator())
   return parent && t.isIdentifier(parent.node.id) ? parent.node.id.name : ''
 }
 
+export function getIdentifierName(path, t) {
+  const classParent = path.findParent(p => t.isClass(p))
```
tkh44: This is so we can include the name of the component the inline css call is in.

tkh44:
@mitchellhamilton can you take a look at my updates. The emotion-server updates could use another look.

PR #380, opened 2017-10-05
Create webpack.md

greggb:
**What**: Adding documentation for bundling extracted css with webpack. May also be handy for other webpack topics in the future.

**Why**: tkh44 shared this snippet on Slack and it seems useful

**How**: Wrote a markdown file.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

Could use a once over for correctness. I left the `modules: true` option as that was in the example, but may need clarification on that.

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -1,0 +1,26 @@
+[github](https://github.com/emotion-js/emotion)
+[npm](https://npm.im/emotion)
+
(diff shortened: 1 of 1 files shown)


tkh44:
Thanks Gregg 🍻

PR #382, opened 2017-10-05
Modified typography, layout and button styles

norbertdragan:
I have modified some CSS to make on the index page to make it look closer than the original design. Mostly typography related, but I also added some margin here and there, plus changed the buttons a bit.

This is my first PR, I am not sure if I am doing this right.


tkh44:
I'll fix it @norbertdragan. I know it can be a pain.

PR #386, opened 2017-10-06
Insertion plugin with tree

### LINKED ISSUE #384: css prop should override styles of styled component
- `emotion` version: 8.0.2
- `react` version: 16

Relevant code.

https://codesandbox.io/s/kwq667x997

What you did:

Tried to override pseudo selectors with the css prop

What happened:

The styles defined in the styled component override those in the css prop.

Reproduction:

https://codesandbox.io/s/kwq667x997

Problem description:

Somehow the insertion order is off it seems.

tkh44:
**What**:
Stylis processes styles from the inside out so we need to build some kind of structure to ensure proper insertion order of the rules. 

See #384 for more details.

**Why**:
Nested pseudo selectors were breaking.

**How**:
Simple tree structure is used. The idea came from the stylis AST plugin.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

closes #384

### DIFF
--- a/packages/emotion/src/index.js
+++ b/packages/emotion/src/index.js
@@ -34,69 +34,173 @@ export function flush() {
 
 let currentSourceMap = ''
 let queue = []
+let orphans = {}
 
-function insertRule(rule) {
-  sheet.insert(rule, currentSourceMap)
+function insertNode(node) {
+  console.log('node', node)
+  if (node.children) {
+    // console.log('children', node.children)
+  }
+  // sheet.insert(rule, currentSourceMap)
+}
+
+function Node(id, name, value, rule, parent, ruleText) {
+  this.name = name
+  this.value = value
+  this.rule = rule
+  this.parent = parent
+  this.id = id
+  this.ruleText = ruleText
+  this.children = []
 }
 
 function insertionPlugin(
   context,
   content,
   selectors,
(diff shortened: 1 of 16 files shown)


packages/emotion/src/index.js
```diff
@@ -34,38 +34,76 @@ export function flush() {
...
+let orphans = Object.create({})
 
-function insertRule(rule) {
-  sheet.insert(rule, currentSourceMap)
+function insertNode(node) {
+  sheet.insert(node.ruleText, currentSourceMap)
+}
+
+function Node(id, selector, parent, ruleText) {
+  this.id = id
```
mitchellhamilton: Is `id` or `parent` used here?

tkh44: No, they were but not anymore.

thysultan:
@tkh44 The current AST plugin in the stylis repo could benefit from a lot of optimizations and improvements. For example [this gist](https://gist.github.com/thysultan/2189897d4f3e0461b1ed98fabdd05e2f) demos one that preserves nested children within the AST.

Did you find any pain points that could be used to improve the stylis plugin system? I'm thinking of passing a depth argument to have data about what depth a rule in a nested selector is in.

PR #393, opened 2017-10-10
Clean up extractCritical code

mattfysh:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/emotion-server/src/index.js
+++ b/packages/emotion-server/src/index.js
@@ -15,12 +15,10 @@ export function extractCritical(html) {
       ids[match[1]] = true
     }
   }
-
-  o.rules = sheet.sheet.slice().filter(x => {
-    RGX.lastIndex = 0
-    let match = RGX.exec(x)
-    const ret = match == null || ids[match[1]] || false
-    return ret
+  
+  o.rules = sheet.sheet.filter(x => {
+    const match = x.match(RGX)
+    return !match || ids[match[0]]
   })
 
   o.ids = Object.keys(inserted).filter(id => {


packages/emotion-server/src/index.js
```diff
@@ -15,12 +15,10 @@ export function extractCritical(html) {
...
   }
-
-  o.rules = sheet.sheet.slice().filter(x => {
-    RGX.lastIndex = 0
-    let match = RGX.exec(x)
-    const ret = match == null || ids[match[1]] || false
-    return ret
+  
+  o.rules = sheet.sheet.filter(x => {
+    const match = x.match(RGX)
```
tkh44: Are you sure we want to remove the `RGX.lastIndex = 0` line? This is a global regex and the index needs to reset after every run from my understanding.

mattfysh: that's right for the way it was used before `RGX.exec(x)`, however it's not applied when using `String.prototype.match`

tkh44: 👍 missed that part 😅

packages/emotion-server/src/index.js
```diff
@@ -15,12 +15,10 @@ export function extractCritical(html) {
...
+    return !match || ids[match[0]]
```
mitchellhamilton: Could you change `!match` to `match === null`, it's a bit more explicit and there will be a (extremely small and probably unimportant) performance improvement.

mattfysh:
sorry all, I'm getting slammed with work and won't have time to update. As this is purely cosmetic feel free to close the PR

Issue #180, opened 2017-07-25
Add outputDir config option to emotion/babel


ChristopherBiscardi:
As per #167, I'm using babel-cli to compile packages down into targeted browser ranges and es* versions. Since we can not access the --out-dir config option from babel-cli in babel plugins, the next best option would be to allow an option that could be used as such:

```js
{
  "env": {
    "es6": {
      "plugins": [
        ["emotion/babel", {outputDir: "es6"}]
      ]
    }
  }
}
```

We could then prepend this directory to the output filename when the [file is written](https://github.com/tkh44/emotion/blob/02ab1b140542092d8bfd5b42c8745f1c59563d50/src/babel.js#L347-L366) and npm scripts could use it as such, with all files being put in the right place:

```js
  "scripts": {
    "build:es6": "BABEL_ENV=es6 babel src --out-dir es6"
  },
```

vladnicula:
Any status on this? 

I stumbled upon emotion when researching css in js frameworks and emotion looks the most promising (in terms of package size, dependencies and features). 

This static extraction is crucial to my production build process.

tkh44:
@vladnicula are you doing server side rendering or client only rendering?

vladnicula:
@tkh44 I'm looking to do both. The current project I'm working on is only using client side rendering.

Would be nice to have support for both in the end. 

Btw, opened this https://github.com/emotion-js/emotion/pull/396, it works for my client side project and for my server side bundling in combination with weback extract text plugin. Could not write a proper test because the fs tests are fishy.

tkh44:
I left some comments that may help you out.

mitchellhamilton:
Closed via #583

PR #396, opened 2017-10-12
Output dir support wip #180

vladnicula:
**What**:
#180 

**Why**:
Prevents polluting src tree when preparing production / static builds. 

**How**:
By saving the optional key `outputDir` in the babel plugin state and doing extra stuff if that key is present at the end of the babel transformation. 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/test/fs.test.js
+++ b/packages/babel-plugin-emotion/test/fs.test.js
@@ -1,9 +1,15 @@
 import { transform } from 'babel-core'
 import fs from 'fs'
+// import { basename } from 'path'
+import mkdirp from 'mkdirp'
 import touch from 'touch'
 import plugin from 'babel-plugin-emotion'
 
-jest.mock('fs').mock('touch')
+jest
+  .mock('fs')
+  .mock('mkdirp')
+  .mock('touch')
+
 fs.statSync.mockReturnValue({ isFile: () => false })
 
 const basic = `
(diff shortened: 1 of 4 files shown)


vladnicula:
Cannot add test. With the current master, duplicating the first fs test breaks the following tests. Need help with that. Not sure what code complete is.

packages/babel-plugin-emotion/test/fs.test.js
```diff
@@ -32,13 +38,34 @@ describe('babel plugin fs', () => {
       babelrc: false
     })
     expect(fs.existsSync).toBeCalledWith(cssFilename)
+    expect(mkdirp.sync).not.toBeCalledWith()
     expect(touch.sync).toBeCalledWith(cssFilename)
     expect(fs.writeFileSync).toHaveBeenCalled()
     expect(fs.writeFileSync.mock.calls[0][0]).toBe(cssFilename)
```
tkh44: In order to use the fs mock, you need to increment the array index like these.

tkh44: More information. https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockimplementationfn

vladnicula: That's a mistake. I wanted to check it wasn't called with anything, because the mkdirp should only be used when the `outputDir` property is provided.

vladnicula:
Test are running fine on my local machine, not sure what's with the failing checks.

Issue #163, opened 2017-07-18
How to use emotion without Babel?


screendriver:
This is not a bug report but a general question: is it possible to use `emotion` without Babel and the corresponding `emotion/babel` plugin? I'm not using Babel, webpack and React. Instead I use TypeScript, fuse-box and Preact.

It seems a little bit overblown to add Babel, the emotion Babel plugin and expand my build chain only "just" for emotion.

I would be grateful for any help.

thangngoc89:
@screendriver hi there, AFAIK emotion is integrated deeply into babel toolchain. You need babel to use emotion. If you don't want to add babel, you can use others alternative CSS-in-JS styling solution like styled-component/glamarous

screendriver:
This is a pity 😞 The only answer that I didn't want to hear. The `AFAIK` in your answer gives me a little hope that there is someone out there that could find another solution ☺️

tkh44:
You can use the raw api that babel converts the strings into. It would be verbose but it is very fast. You don't get extraction either.

```
import { css } from 'emotion'

const H1 = css(['css-H1-duiy4a'], [colorVariable], function createEmotionStyles(colorVariable) {
  return [`.css-H1-duiy4a { color:${colorVariable} }`]
})
```

tkh44:
@screendriver You do not need babel to run emotion as version 8.

https://codesandbox.io/s/5v42zojqll

Andarist:
> Are there any plans to port the babel plugin to typescript?

IMHO that would be a duplicated effort

> I realize the core API is available without the plugin, but I can't migrate without the additional features, which the plugin offers. :(

Upcoming emotion@10 is going to support all features out of the box - babel plugin will be used merely to optimize (like bundle size) the output.

elektronik2k5:
That's great news! Thanks for the update :)

Brooooooklyn:
https://github.com/LeetCode-OpenSource/emotion-ts-plugin

elektronik2k5:
@Brooooooklyn that awesome!!! Are you planning to get your transformer to feature parity with the Babel plugin?

Brooooooklyn:
@elektronik2k5  yes, we are planning to use emotion in production, and I will keep developing this plugin.

DominicTobias:
@Andarist it's not exatly clear to me if the babel plugin just minifies, or is it optimizing something more important like the parsing of css string literals into AST?

Andarist:
[...]

PR #398, opened 2017-10-13
Add types to react-emotion

renatorib:
**What**:
TypeScript declarations for react-emotion package

**Why**:
Support people who use TypeScript and need typings. 

**How**:
* Created `react-emotion.d.ts` and linked in "types" package.json field
* Followed same structure/names from #379 

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

---

### Supports

**Props Inference by Tag Name:**

image

---

**Supports template string and object as styles:**

image

---

**Passing Custom Props:**

image
*Note1: needs a second type parameter passing tag name because it stop to infer when you pass props at first parameter*
*Note2: glamorous do not support custom props with shorthands, like Link2 :stuck_out_tongue:. It was a little tricky, but works...* 

---

**Create styled with another component:**

image

---

**Pass custom props to style another component:**

image
*Note3: as well as Note1 it's not infer 'MyComponent's Props' when you pass props as first type parameter, so you need to pass 'MyComponent' props together*
```tsx
type Props = MyNewProps & MyComponentProps
```
---

**Theming props:**

By default props.theme is `any`, but you can create another styled instance passing theme props.
Basically the same approach from styled-components 

image

---

**withComponent:**

image

Also works passing another component instead of tag name


packages/react-emotion/typescript_tests/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
luke-john: For glamorous we've added declaration: true to help make sure we're exporting everything required for use in libraries Microsoft/TypeScript#5938.

luke-john:
Awesome work doing these types. They're looking great 👍 .

packages/react-emotion/typings/react-emotion.d.ts
```diff
@@ -0,0 +1,103 @@
+import { StatelessComponent, Component as ClassComponent, CSSProperties } from 'react'
+import { Interpolation as EmotionInterpolation } from 'emotion'
+
+export type InterpolationFn<Props = {}> = (
+  props: Props
+) => EmotionInterpolation
+
+export type Interpolation<Props = {}> =
+  | InterpolationFn<Props>
+  | EmotionInterpolation
```
mitchellhamilton: I don't know much about typescript but I don't think this handles the fact that there can be nested function interpolations and they will receive props.

```jsx
  display: ${(p) => props => props.display};
  color: ${[props => 'hotpink']};
`
```

screendriver:
But you still need Babel in the build pipeline as I already asked here #163?

packages/react-emotion/typings/react-emotion.d.ts
renatorib: It works on emotion?
props => props_ => ...

It doesn't make any sense to me. What's this use case for?

renatorib:
@screendriver This is only typings, do not deal with build, or any emotion code. I don't even know the internals of emotion yet.

screendriver:
Ah ok. But TypeScript definitions for emotion are a little but useless when you can't use emotion without Babel 😉 (emotion only works as Babel plugin so Babel is mandatory) 

Apart from that: thank you for the typings 😎👍

renatorib:
If you don't want to use babel (or can't) just don't use it. Pick another one: styled-components, glamorous, etc.

Anyway, this is another issue. Don't have any relation with typescript or this pull request.

packages/react-emotion/typings/react-emotion.d.ts
renatorib: I don't know exactly all edge cases of emotion API. I'll try to study it to improve the coverage support.

packages/react-emotion/typescript_tests/tsconfig.json
renatorib: Done! Added.
No warnings
:smile:

packages/react-emotion/typings/react-emotion.d.ts
renatorib: @mitchellhamilton I've added support to these two cases you mentioned. 

```tsx
export type InterpolationFn<Props = {}> =
  (props: Props) =>
    | EmotionInterpolation
    | InterpolationFn<Props>

export type InterpolationTypes<Props = {}> =
  | InterpolationFn<Props>
  | EmotionInterpolation

export type Interpolation<Props = {}> =
  | InterpolationTypes<Props>
  | InterpolationTypes<Props>[]
```

If you know more cases, just tell me, please.

renatorib:
Someone can help me with documentation? I think it's the only thing missing to this pr.

My suggestion is to write a section "TypeScript" in emotion website, but you are creating a new website, right? @tkh44

tkh44:
@renatorib any new markdown file in the `/docs` folder is added to the website. Just add a new file there called `typescript.md` and do your thing 👍

renatorib:
That's great! I'll try to write something. :)

hemlok:
This is awesome, I just started using typescript as a replacement for PropTypes, and this will go nicely with that 👌 Thanks @renatorib

screendriver:
> If you don't want to use babel (or can't) just don't use emotion. Pick another one: styled-components, glamor/ous, fela, etc.

I am not complaining about that. It is much more a question how to use emotion with TypeScript 😉 Because you found obviously a way that I am not aware of at the moment and I really want to use emotion as well.

renatorib:
Today I'm 'blocked' from using emotion in my company codebase because we use typescript.
What I did:
- [ ] Ask for a improve on every issue even non-related about ts.
- [x] Sent a PR so I can use it.

If you really needs use emotion and can't use Babel, you can:
1. Use without babel, abdicating the shorthands (using `styled('div')`).
2. Sent a PR that makes emotion shorthands work without babel, so we can work together in that.

I still do not think we should be talking about it here. This has no connection to the PR thread.

tkh44:
@renatorib Is this ready except for documentation?

renatorib:
@tkh44 yep! 

I did not have time to write it yet, but it's the only thing missing.

renatorib:
If you want, you can merge it and I send another PR with documentation.

renatorib:
I've added a typescript documentation

Unfortunately my englis  is not as good as I wish it were.
Someone please give a review so nothing's going wrong :smile: 
Thanks!

tkh44:
Great work @renatorib! I did a quick edit but I'd like someone who knows more about typescript do a review.

renatorib:
Looks great @tkh44, thanks!

PR #407, opened 2017-10-14
Clarify Babel plugin usage in install docs

wKovacs64:
**What**:

Update install docs with regards to the Babel plugin to remove any confusion surrounding `transform-react-constant-elements` being related to emotion.

**Why**:

People new to emotion and CSS-in-JS in general may not know if `transform-react-constant-elements` is related to emotion and whether or not it's required to use emotion's Babel plugin.

**How**:

Remove reference to `transform-react-constant-elements` by substituting a fictitious plugin to show it is only there to illustrate order.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -18,12 +18,12 @@ npm install --save emotion react-emotion babel-plugin-emotion
 
 **Notes:**
 - Make sure `"emotion"` is the first plugin.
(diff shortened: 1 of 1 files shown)


tkh44:
Thank you!

PR #408, opened 2017-10-14
readme and doc consolidation updates

greggb:
**What**: Added some quick start info on main readme and consolidated some documentation.

**Why**: The docs need updating

Still a work in progress. 
Todo:
- [x] mark all features that require babel in their respective doc files
- [ ] N/A notes for each api item (future PR)
- [x] further doc consolidation


tkh44:
Thank you so much Gregg. This is going to have a huge impact on our adoption.

greggb:
I have some more WIP for another PR, but I think this is a good update on its own.

greggb:
Mind giving it a look @tkh44 ?

docs/babel.md
```diff
@@ -0,0 +1,36 @@
+## Usage with Babel
```
mrmartineau: @greggb @tkh44 there are quite a few babel options (e.g. sourcemaps, inline etc etc).. I think they should be documented here..

docs/babel.md
```diff
@@ -0,0 +1,36 @@
...
+
+The Babel Plugin is highly recommended, but not required in version 8 and above.
+
+See the [installation instructions](install.md).
+
+### Features which are enabled with the babel plugin
+
+### styled.element Syntax
+`styled('div')` will work without the plugin
```
mrmartineau: I'd like to see an explanation of the two syntax types and the reason for using one over the other. It probably should go in the 'getting started' section (if there is one..) though

tkh44:
I will look at this when I get some time today

docs/babel.md
greggb: Absolutely - I'm not sure of all the options

docs/babel.md
greggb: I have a getting started doc on deck for my next PR. I'll need some help understanding the big differences between syntaxes

PR #414, opened 2017-10-17
Automatic hoisting.

tkh44:
**What**:
Hoist any arguments supplied to css or styled if they are not template literals (not sure if this matters).

**Why**:
Hoisting should increase the performance of inline css calls the most, but will have a secondary benefit for anyone using object styles.

**How**:
Use babel's `isPure` and `hoist` methods on the `path`.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -76,7 +76,10 @@ export function replaceCssWithCallExpression(
       )
     )
 
-    hoistPureArgs(path)
+    if (state.opts.hoist) {
+      hoistPureArgs(path)
+    }
+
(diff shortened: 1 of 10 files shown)


packages/babel-plugin-emotion/src/index.js
```diff
@@ -257,7 +271,7 @@ export default function(babel) {
         } catch (e) {
           throw path.buildCodeFrameError(e)
         }
-
+        hoistPureArgs(path)
         path[visited] = true
```
mitchellhamilton: This will hoist every single `CallExpression`

PR #416, opened 2017-10-18
Added TypeScript definitions for css prop.

cameron-martin:
**What**:
Added TypeScript definitions for the `css` prop which is compiled to `className` by the babel plugin.

**Why**:
Because otherwise TypeScript users will get type errors when using this feature.

**How**:
Used module augmentation/declaration merging to add the `css` property to the `HTMLAttributes` inferface.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion/package.json
+++ b/packages/emotion/package.json
@@ -30,6 +30,7 @@
     "stylis-rule-sheet": "^0.0.5"
   },
   "devDependencies": {
+    "@types/react": "^16.0.15",
     "babel-cli": "^6.24.1",
     "babel-plugin-transform-define": "^1.3.0",
     "cross-env": "^5.0.5",
--- a/packages/emotion/typescript_tests/tsconfig.json
+++ b/packages/emotion/typescript_tests/tsconfig.json
@@ -4,7 +4,8 @@
     "module": "es2015",
     "strict": true,
     "allowSyntheticDefaultImports": true,
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "jsx": "react"
   },
   "include": [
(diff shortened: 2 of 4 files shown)


cameron-martin:
On seconds thoughts, maybe this should go into the dts files in the `emotion` package instead, since the css prop can be used without using the `react-emotion` package. Or even in the types for the `babel-plugin-emotion` package - although I have no idea if this would work.

What do you think?

tkh44:
I should go in the react-emotion package I think.

packages/react-emotion/typings/react-emotion.d.ts
```diff
@@ -3,6 +3,12 @@ import { Interpolation as EmotionInterpolation } from 'emotion'
 
 export * from 'emotion'
 
+declare module 'react' {
+  interface HTMLAttributes<T> {
+    css?: string;
```
tkh44: This can be any type of value.

cameron-martin: Anything of type `Interpolation`?

cameron-martin:
I'm using react + emotion without react-emotion so it would be useful to be in the emotion package.

packages/react-emotion/typings/react-emotion.d.ts
tkh44: Correct

tkh44:
I agree. @cameron-martin. I misread your original comment. Sorry about that.

cameron-martin:
I've made the css prop take anything of type `Interpolation` and moved it to the emotion package.

PR #418, opened 2017-10-18
Use existing style tag if nonce attribute is present

dstaley:
**What**:
Default to using an existing `style` tag if one is present with a `nonce` attribute.

**Why**:
In locked-down CSP environments, adding inline styles are usually blocked. To allow this, you can provide a `style` tag with a nonce attribute. However, emotion will always attempt to create a new `style` tag, which would trigger a CSP violation.

**How**:
When `StyleSheet.inject` is called, we check to see if there's already a `style` tag with a `nonce` attribute. If so, we use that tag, otherwise we make a new one.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete


tkh44:
I have no problem with this implementation. What do you think @mitchellhamilton?

packages/emotion/src/sheet.js
```diff
@@ -56,7 +56,9 @@ export default class StyleSheet {
       throw new Error('already injected!')
     }
     if (this.isBrowser) {
-      this.tags[0] = makeStyleTag()
+      // If an existing style tag with a nonce attribute is present, use that before attempting to create a new tag.
+      const tagWithNonce = document.querySelector('style[nonce]')
```
mitchellhamilton: I'm not sure about the `style[nonce]` selector since it'll target any style element with the nonce attribute, maybe `style[nonce][data-emotion]`?

tkh44: Good idea @mitchellhamilton

dstaley: @mitchellhamilton That's a *much* better idea!

mitchellhamilton:
Closing as we're providing a [`nonce` option in create-emotion](https://github.com/emotion-js/emotion/tree/master/packages/create-emotion#nonce-string).

PR #419, opened 2017-10-19
Typescript typings: Added emotion-theming typings and fixed react-emotion typings

asvetliakov:
**What**: First commit in PR fixes the problem with union type in TS typings, seconds commit adds TS typings for ```emotion-theming``` package

**Why**:
* emotion-theming doesn't have TS typings
* There is a problem with union type ```Tag | Component<Props>``` in ```react-emotion``` typings: The second overload won't be taken into account when passing ```styled(MyComp)``` and thus ```MyComp``` will infer all Intrinsic properties from all tags. This is wrong and slows down the TS service. First commit in PR fixes this.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion-theming/package.json
+++ b/packages/emotion-theming/package.json
@@ -4,12 +4,15 @@
   "description": "A CSS-in-JS theming solution, inspired by styled-components",
   "main": "dist/index.cjs.js",
   "module": "dist/index.es.js",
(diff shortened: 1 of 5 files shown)


tkh44:
Thank you @asvetliakov !

PR #421, opened 2017-10-19
Switched to using dtslint from just running tsc --noEmit for typescript tests

cameron-martin:
**What**:
Changed the TypeScript tests to using dtslint.

**Why**:
* Most importantly it lets you use `$ExpectError` to assert that a certain statement *does* type error, which the previous type tests did not.
* It also tests against multiple TypeScript versions.

**How**:
By changing from running `tsc --noEmit` to running `dtslint` for running the TypeScript tests.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

I'm having trouble running the dtslint tests for the react-emotion package - some of the features used in the dts files require a new version of TypeScript but I keep getting an error (https://github.com/Microsoft/dtslint/issues/78) when trying to use new versions of TypeScript.

This PR is just to make sure you agree with this direction before I pursue it even more and to collaborate if anyone else is working on it.

A few related changes, some as a side-effect of moving to dtslint:
* Made some changes to the dts files to satisfy the default dtslint rules.
* Renamed typings directory to types because this seems to be more standard.
* Renamed emotion.d.ts, etc to index.d.ts because dtslint seems to require this.
* Stopped "build" running before "test:typescript" as this is not actually necessary.

### DIFF
--- a/packages/emotion-theming/package.json
+++ b/packages/emotion-theming/package.json
@@ -4,15 +4,14 @@
   "description": "A CSS-in-JS theming solution, inspired by styled-components",
   "main": "dist/index.cjs.js",
   "module": "dist/index.es.js",
-  "types": "typings/emotion-theming.d.ts",
+  "types": "types/index.d.ts",
   "files": [
     "src",
     "dist"
   ],
   "scripts": {
     "build": "npm-run-all clean rollup rollup:umd",
-    "test:typescript": "tsc --noEmit -p typescript_tests/tsconfig.json",
-    "pretest:typescript": "npm run build",
+    "test:typescript": "dtslint types",
     "clean": "rimraf dist",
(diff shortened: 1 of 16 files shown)


renatorib:
That's great!

I saw that you renamed some names, like `typings` => `types`, `react-emotion.d.ts` => `index.d.ts`, etc.
I think this is not a good idea since all other emotion packages follows this names/structure.

Or, if you think this names/structure is better, you can change all other packages also to maintain a consistency.

cameron-martin:
I think dtslint requires that the dts file is named index.d.ts (although I could be wrong about this, I'll check). But yeah I'll make sure everything is consistent.

cameron-martin:
The above problem is now fixed.

PR #426, opened 2017-10-21
Added TypeScript definitions for innerRef.

### LINKED ISSUE #424: innerRef is missing from react-emotion typings
From twitter.

https://twitter.com/rossipedia/status/921138316677976064

I don't know much about TS so I need outside help on this one.

cc @cameron-martin @renatorib

cameron-martin:
Closes #424.

**What**:
Added innerRef prop to styled component typescript definitions.

**Why**:
To fix this issue: https://github.com/emotion-js/emotion/issues/424

**How**:
By adding the `StyledComponentProps` type, which has all the normal element props and also `innerRef` that is the same type as `ref`.

**Checklist**:

- [ ] Documentation - N/a
- [x] Tests
- [x] Code complete

I'm not sure what to do about `withComponent(AnotherComponent)`. What even is the behaviour of `withComponent` when called with another component?


renatorib:
Hey, this looks great. 

But I think `StyledComponentProps` isn't a good name for this. I think it as something like `ElementProps`. 
I consider that to be a `StyledComponentProps` it should have all props including `Props` and `Theme`

cameron-martin:
Sure, I renamed it.

PR #428, opened 2017-10-22
Fix editing styles in devtools with source maps

mitchellhamilton:
**What**:
Fix editing styles in devtools with source maps

**Why**:
Because it broke other styles on the page.

**How**:
Remove the `@ sourceURL=some/path` comment, I'm not sure if this is a good idea but it seems to work fine and the path is already inside the source map data so ¯\\\_(ツ)_/¯

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

Before: 
before fix

After:
after fix


greggb:
It was emotion doing this? I was debugging it as a Storybook issue as it was driving me crazy. Thanks @mitchellhamilton!

PR #440, opened 2017-10-31
Babel 7 compatibility and change tests

### LINKED ISSUE #169: Improve Babel tests
Currently the tests for the Babel plugin aren't great and have a lot of boilerplate. We should switch to [`babel-plugin-tester`](https://github.com/babel-utils/babel-plugin-tester).

Also the current way of testing extraction isn't very good since you can't use `.only` since the assertions rely on other tests. This can be fixed with [`.mockClear()`](http://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockclear)

### LINKED ISSUE #437: Babel 7 compatibility for babel-plugin-emotion
Hello,

- `emotion` version: latest (8.0.9)
- `babel-plugin-emotion` version:  latest (8.0.9)
- `babel` version: 7.0.0-beta3 

I'm having an issue when trying to compile using babel 7.
image

Here's where the error is coming from:
image

After console.logging I realized that `metadata` property on `state.file` is an empty object 

The problem doesn't come from any plugin incompatibility. It doesn't work even with the simplest .babelrc / .babelrc.js file 
```
{
  "plugins": ["emotion"]
}
```

Babel 7 is still in beta, but it has a lot of benefits (most notably babel-preset-env that can include only polyfills for features that are actually used (!) by your code).

Please, look into it. Thanks!

mitchellhamilton:
**What**:
Babel 7 compatibility and change tests

**Why**:
Closes #437 
Closes #169 

**How**:
`jest-in-case`, writing some helpers and [a babel plugin i wrote to convert the boilerplately cases into objects](http://astexplorer.net/#/gist/280e316734f3c21b2de75d90e7cd8b99/68438eb8a82c3728d994877de5a47e357f753063) .

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

This isn't done yet, I'm having some problems with macro tests.

Once this is done, I'm thinking of switching to babel 7 for bundling so we get the smaller helpers and stuff.


packages/babel-plugin-emotion/test/css-prop.test.js
```diff
@@ -1,128 +1,86 @@
...
-
-  test('className as expression', () => {
-    const basic = '(<div className={variable} css={`color: brown;`}></div>)'
-    const { code } = babel.transform(basic, { plugins: [plugin] })
-    expect(code).toMatchSnapshot()
-  })
-
-  test('className as expression string', () => {
-    const basic =
+import { createInlineTests, createExtractTests } from './util'
```
tkh44: ❤️

tkh44: I tried doing this once and stopped at 2 files. Awesome work.

packages/babel-plugin-emotion/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "babel-generator": "^6.26.0",
+    "babel-helper-module-imports": "^7.0.0-beta.3",
```
Andarist: this is also published already as scoped - `@babel/helper-module-imports`

you probably should update other babel dependencies to their respective` @babel/*@7`

packages/babel-plugin-emotion/package.json
```diff
@@ -14,15 +14,16 @@
     "clean": "rimraf lib"
   },
   "dependencies": {
-    "babel-generator": "^6.26.0",
-    "babel-macros": "^1.0.2",
+    "@babel/helper-module-imports": "7.0.0-beta.31",
+    "babel-macros": "^1.2.0",
     "babel-plugin-syntax-jsx": "^6.18.0",
```
Andarist: is there a reason this plugin is left here with version 6? mixing babel 7 and babel 6 packages together might not work and is considered highly risky

mitchellhamilton: `@babel/plugin-syntax-jsx` has a peerDependency on `@babel/core` whereas `babel-plugin-syntax-jsx` doesn't and we want to maintain babel 6 compatibility for now. I understand it's not ideal.

PR #448, opened 2017-11-08
Add SSR streaming API and stuff

### LINKED ISSUE #341: extractCritical and html stream SSR
- `emotion` version: 7.2.2
- `react` version: 15.6.1

Relevant code.

```javascript
import {extractCritical} from 'emotion-server';
import ReactDOMStream from 'react-dom-stream/server';

const {html, ids, css} = extractCritical(ReactDOMStream.renderToString(app));
```

What you did:
Trying to make emotion critical css working with SSR and streamable html.

What happened:
Snake biting his tail.

Reproduction:

Problem description:

Hey there, I wanted to make my app use html stream by using react-dom-stream : https://github.com/aickin/react-dom-stream

But the emotion critical css rendering need to first get all the rendered html from ReactDOM...
Am I right?

Do you think if there is a way to achieve critical css in emotion with dom stream?

Suggested solution:

mitchellhamilton:
**What**:
Add two new exports from `emotion-server`.

**renderStylesToString**
This returns a string of html that inlines the critical css required right before it's used.

```jsx
import { renderStylesToString } from 'emotion-server'
import App from './App'

const html = renderStylesToString(renderToString(<App />))
```

**renderStylesToNodeStream**
This returns [Node Stream Writable](https://nodejs.org/api/stream.html#stream_class_stream_writable) that can be used to insert critical css right before it's required. It can be used with [React's streaming API](https://reactjs.org/docs/react-dom-server.html#rendertonodestream).

```jsx
import { renderToNodeStream } from 'react-dom/server'
import { renderStylesToNodeStream } from 'emotion-server'
import App from './App'

const stream = renderToNodeStream(<App />).pipe(renderStylesToNodeStream())
```
I'm not totally sure about the names of these.

There are also some internal changes to `extractCritical` that should make it faster.

This also fixes some stuff with SSR and labels.

**Why**:
Closes #341 

**How**:
`glamor-stream`, https://github.com/threepointone/glamor/blob/master/src/inline.js and some changes in a few places.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


packages/emotion-server/src/inline.js
```diff
@@ -0,0 +1,65 @@
+import { inserted, registered, names } from 'emotion'
+
+function toTag(ids, thing) {
```
tkh44: Should this just accept an array as the second argument and have the function return an object with the updated `keys` and tag?

mitchellhamilton: Yeah, that's probably a better idea.

packages/emotion-server/src/stream.js
```diff
@@ -0,0 +1,53 @@
...
+import tokenize from 'html-tokenize'
+import pipe from 'multipipe'
+import { inserted, registered, names } from 'emotion'
+
+export default function renderStylesToNodeStream() {
+  let insed = {}
+  const tokenStream = tokenize()
+
+  const inlineStream = through(
+    function write(thing) {
```
tkh44: Could this be

mitchellhamilton: Yup, it was originally like that but I was debugging something so I changed it.

PR #451, opened 2017-11-09
Styles on the css prop should override styles found in the className

### LINKED ISSUE #450: `css` does not override/add-last to `className`
I expect [this](https://codesandbox.io/s/1yj28q3m1l) to be red

```js
const fooStyles = css({ background: 'blue' })

<div
  className={fooStyles}
  css={{ background: 'red' }}
/>
```

tkh44:
**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

closes #450


kirkstrobeck:
@mitchellhamilton that's an interesting point, but the point would be to either 

append a className as a prop 

```js
<div className={cx(localStyles, stylesProp)} />
```

or add an attribute, like a color, as a prop and apply the style via `css`

```js
<div 
  className={localStyles}
  css={{ background: colorProp }}
/>
```

I think the use-case you mention could be addressed as 

```js
<div 
  css={[
    { background: colorProp },
    externalStylesProp,
  ]}
/>
```

tkh44:
@kirkstrobeck I've been going back and forth on this all day and I think what you are saying makes sense. I'm leaning this way as I think the css prop should be treated, specificity wise, like the style attribute in that it should be the last thing applied.

cameron-martin:
Another option would be to make the one that appears last take precedence. So

```
className={cls}
css={{ color: 'red' }}
```

gets transformed to

```
className={cx(cls, { color: 'red' })}
```

And


```
css={{ color: 'red' }}
className={cls}
```

gets transformed to

```
className={cx({color: 'red' }, cls)}
```

tkh44:
That is a great idea on the surface @cameron-martin, but there are several eslint rules for ordering prop names and I think the ambiguity would cause confusion for devs in the long run.

cameron-martin:
Yeah that would be an issue. It's also weird because the order that props are specified normally doesn't change behaviour, and this could break people's expectations.

Another option would be to error if both the className and css prop are specified, forcing the user to explicitly specify the precedence.

kirkstrobeck:
- I haven’t seen “ordering of props as an API,” nor would I recommend it. I believe it would result in unintended side-effects for anyone unfamiliar with it. 
- The recommendation of having `css` append last has precedence with the `class` and `style` tags in `html`, which these closely resemble in behavior.

kirkstrobeck:
Any update on this? Running into it again

tkh44:
@kirkstrobeck tbh we've actually talked about this between ourselves (@mitchellhamilton and I) to exhaustion. I think we might be moving to a pattern that uses `cx` or `css` calls direction on the className.

kirkstrobeck:
Cool. Not sure I understand this tho

> cx or css calls direction on the className.

cameron-martin:
I didn't understand that last sentence either.

tkh44:
Whoops! Was not done with that comment 😄

What I mean is manually following a pattern similar to what @cameron-martin proposed doing automatically.

```javascript
<div className={css({ color: 'blue' })} />
// or
<div className={cx(someClassName1, css({ color: 'blue' }))} />
```

kirkstrobeck:
I guess I’m only confused because the `css` prop is still there. Does it come before or after `className`?

cameron-martin:
Would this mean getting rid of the `css` prop altogether, or just forbidding the use of the `css` and `className` prop together?

daphnesmit:
Can someone please fix this override issue!:D

mitchellhamilton:
I'm gonna close this since it's a breaking change and in the next major version the css prop is going to be implemented differently

I just opened #753 to discuss this more.

PR #453, opened 2017-11-09
Proposal: export emotion-theming from react-emotion

evan-scott-zocdoc:
**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

While refactoring for a proof of concept, it quickly became cumbersome to have to split up the import just to pull in theming HOCs.

Since theming functionality is partially built-in by default in emotion v8 via supporting `p => p.theme` etc, I think it makes sense to go the rest of the way and re-export `emotion-theming` for ease of use.

~~It might make sense to get rid of `emotion-theming` entirely and just have it be part of `react-emotion`?~~ in v9 maybe


evan-scott-zocdoc:
Side note, I don't use typescript so no idea what the errors mean. I think think I have everything else working...

evan-scott-zocdoc:
@tkh44 What do you think? I'm @probablyup's work account btw :)

tkh44:
I like it on the surface, but what does this do the bundle size? Previously we were just importing one file from emotion-theming.

evan-scott-zocdoc:
It increases by ~.76kB.

evan-scott-zocdoc:
I guess the question here is how often do emotion users also use emotion-theming? If the majority do not also use theming, then this could not make sense.

That being said, treeshaking and DCE should take care of pieces of the library that aren't used right?

tkh44:
I think the majority of users are using theming in some way. I'm leaning towards merging this.

packages/react-emotion/types/index.d.ts
```diff
@@ -4,6 +4,7 @@ import { StatelessComponent, ComponentClass, CSSProperties } from 'react';
 import { Interpolation as EmotionInterpolation } from 'emotion';
 
 export * from 'emotion';
+export * from 'emotion-theming';
```
tkh44: This should be updated to match @mitchellhamilton's work in #457

mitchellhamilton:
I agree, I'm also going to try some stuff with babel to see if I can get the size down a bit as well because I checked and right now tree shaking won't work with what babel 6 outputs but with babel 7 it should work

evan-scott-zocdoc:
Kk I'll rebase

packages/react-emotion/types/index.d.ts
mitchellhamilton: I don't understand, #457 shouldn't require any changes to this PR.

evan-scott-zocdoc:
Do you folks prefer re-exporting `emotion-theming`, or simply merging it into `react-emotion`?

evan-scott-zocdoc:
I suppose merging it would be a breaking change, nvm.

evan-scott-zocdoc:
@mitchellhamilton Could also add this in preparation for webpack 4: https://github.com/webpack/webpack/tree/next/examples/side-effects

packages/react-emotion/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "babel-plugin-emotion": "^8.0.10",
+    "emotion-theming": "^8.0.10",
```
mitchellhamilton: This also needs to be added to `preact-emotion`. Also, in the rollup config, in the `preact-emotion` bit, `prop-types` and `emotion-theming` need to added to `external`.

packages/react-emotion/types/index.d.ts
tkh44: Oops, I thought you manually changed the export in that PR. Didn't realize the update fixed it.

packages/react-emotion/package.json
evan-scott-zocdoc: done

probablyup:
Redoing this under my personal account

PR #464, opened 2017-11-15
Add Instances

### LINKED ISSUE #349: extractCritical does not seem to work when `npm link`ing
- `emotion` version: 7.3.2
- `react` version: 16 (but also observed in latest 15.x)

Relevant code.

### What you did:

```js
import React from 'react';
import { renderToString } from 'react-dom/server';
import { extractCritical } from 'emotion/server';

// ... some more setup and usual stuff here ...

const markup = renderToString(
    <StaticRouter context={context} location={req.url}>
      <App />
    </StaticRouter>
);

const { css, ids } = extractCritical(markup);

console.log("css", css);
console.log("ids", ids);

// css and ids are blank/empty
```

### What happened:

`npm link`ing a shared component repo (styled with `emotion`) into a real app causes `extractCritical` to not work (e.g. `ids` and `css` are blank/empty)

### Reproduction: 

https://github.com/tizmagik/emotion-link-repro

### Problem description:

`npm link` appears to break extractCritical

### Suggested solution:

I'm not sure if it's actually a problem with `extractCritical` or just with my setup (likely the latter), but any pointers welcome!

### LINKED ISSUE #403: Supporting locked down CSP environments.
We run a pretty locked down content security policy in production, which doesn't allow for "unsafe-inline" style tags. This blocks both inserted `<style>` tags or the `style` attribute on elements.

I think the solution for this that would work for us (and others like us) would be to use the nonce directive. In this setup, the server supplies a nonce in the CSP header, and any inline `<style>` tags that use it will be allowed to run, e.g: `<style nonce="some-nonce">`.

I think there are two options for this:

1. Do what styled-components did: https://github.com/styled-components/styled-components/issues/887 and https://github.com/styled-components/styled-components/pull/1022/files here the containing page sets window.__webpack_nonce__ (somehow) to match the nonce in the header. And emotion would set the attribute on the tags it inserts.

2. The alternative would be to have the containing page create a style tag with the nonce already set, and emotion could reuse that as mentioned: https://twitter.com/tkh44/status/918901965924397057

I think either is roughly equivalent from an end user point of view. When deployed to production the http server will have to inject the nonce either in a js script, or into the style tag into index.html which is roughly equivalent.

I am _tempted_ to suggest that option 1. would help spread awareness/usage of webpack_nonce and allow the same setup work in either lib, but I personally am not too strongly wedded to either (and obviously I don't know the internals of emotion.

NB to cover two potential concerns:
1. The issue raised [here](https://github.com/threepointone/glamor/issues/323) about IE11 is a red-herring - IE11 will ignore all the csp rules regardless, so it doesn't need to be worked around.
2. I made a test page to confirm that CSP rules won't block the modification of existing stylesheet rules, looks like they do not: https://gist.github.com/latentflip/61b3102a90c31999664b6e4933a55d0e

### LINKED ISSUE #430: Allow instances of Emotion
- `emotion` version: 8
- `react` version: n/a

I'm currently evaluating different CSS-in-JS solutions and one thing that seems missing from most is a mode where you create an instance of the sheet/registry and manage styles using this instance instead of relying on a global singleton.

Why do I need this? Aside from "global shared mutable state is bad" etc. I maintain [Catalog](https://www.catalog.style), a style guide builder that can render components from your codebase (similar to Storybook). Catalog itself is a React app and currently uses Radium which I want to replace for several reasons. Because users can include _any_ code, I need to make sure that the replacement styling library doesn't conflict with anything else, including other versions of the same library.

Something like this could work, probably even without breaking existing API:

```js
import Emotion from 'emotion/class';

const emo = new Emotion({key: 'my-special-key'});

emo.css`
  /* ... css etc */
`
```

The `Emotion` instance basically would expose the same methods like the [emotion/index](https://github.com/emotion-js/emotion/blob/8eb316b785fd8bcfad0211564c2bdf0abd0f912a/packages/emotion/src/index.js)  module does.

Alternatively, `css` et al. could take the instance as an additional argument.

The `key` option has two purposes:

- Identify the style tags with the instance key (`<style data-emotion="my-key">` instead of `<style data-emotion>`)
- Prefix the generated classes with this key instead of generic `css-` (although this could be a distinct option)

Other aspects:

- An instance could have a `destroy` method which would clean up styles managed by it.
- Extracting styles for SSR would be safer (see #349)
- Setting options (e.g. for vendor prefixing) for a global singleton is weird (must set globals _before_ the singleton is created) but simple with instances

For reference, [styletron](https://github.com/rtsao/styletron) and [fela](http://fela.js.org/docs/basics/Renderer.html) already work like this.

mitchellhamilton:
**What**:

Add three new packages: `create-emotion`, `create-emotion-styled` and `create-emotion-server`.

`create-emotion` exports a function which accepts a `context`, this should generally be Node's `global` or the browser's `window` and other options like `nonce` and maybe `stylisPlugins` (I'm not sure if we should remove `useStylisPlugin` and use an option, keep both or keep using `useStylisPlugin`, I'm leaning towards an option since stylis plugins are behind the caches so it could be confusing because if you add a plugin with `useStylisPlugin` and call `css` with something that is cached, it won't run the plugin). It returns an object that contains all the current exports of emotion.

`create-emotion-styled` exports a function which accepts an emotion instance (from `create-emotion`) and options for the `channel` used for theming, the `contextTypes` for theming, `createElement` (a React-like `createElement`) and `Component` (a React-like `Component`). It returns a `styled` API, like `react-emotion` exports.

`create-emotion-server` (I haven't done this yet as I'm waiting to finish #448 first)

There will also be some new options to `babel-plugin-emotion` to specify the paths to emotion instances and the primary emotion path that will be used to add imports for the css prop (and potentially other things in the future).

This will also store the caches on globals for the `emotion` package so that #349 is solved and loading the same instance twice will use the same caches. This mainly affects SSR where there are often multiple instances of `emotion` in different dependencies and they're not deduplicated by webpack.

This does not affect people who currently use `emotion`, `react-emotion`, `preact-emotion` and `emotion-server` and don't need instances. (except that loading the modules twice will use the same caches so #349 will be fixed)

**Why**:
Closes #430 
Closes #403 
Closes #349 

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Code complete


packages/create-emotion/src/sheet.js
```diff
@@ -35,49 +36,61 @@ function sheetForTag(tag) {
...
 export default class StyleSheet {
-  constructor() {
+  injected: boolean
+  isSpeedy: boolean
+  isBrowser: boolean
+  ctr: number
+  sheet: string[]
+  tags: HTMLStyleElement[]
+  nonce: string | void
+  constructor(nonce?: string) {
```
tkh44: Should we make this an object in case we need more options later.

mitchellhamilton: I don't think we need to since it's an internal API.

mitchellhamilton:
I'm just going to start documenting the (short) list of breaking changes this causes here:

- `fontFace` is gone, it can be replaced with `injectGlobal` with a regular css `@font-face` rule.
-  The `registered` and `inserted` caches are not exports of emotion (they're now on the `caches` object which is exported by emotion) but these aren't documented and should never be relied on externally so this shouldn't cause a problem.
- `extractCritical` no longer returns a rules property, this was also never documented so shouldn't cause any problems.
- `StyleSheet` is no longer used for SSR, this is completely internal so shouldn't cause any problems.

rodrigopivi:
what is missing for this to get merged? i cant wait to start using renderToNodeStrean with emotion

jest.dist.json
```diff
@@ -9,10 +9,15 @@
     "^emotion-utils$": "<rootDir>/packages/emotion-utils/dist/index.cjs.js",
     "^emotion-server$": "<rootDir>/packages/emotion-server/lib",
     "^emotion-theming$": "<rootDir>/packages/emotion-theming/dist/index.cjs.js",
-    "^babel-plugin-emotion": "<rootDir>/packages/babel-plugin-emotion/lib"
+    "^babel-plugin-emotion": "<rootDir>/packages/babel-plugin-emotion/lib",
```
Andarist: @mitchellhamilton would you be interested in using a package that would create those for you? I've created some time ago https://github.com/Andarist/lerna-alias for my use case, but can easily add an option to use `main` field / standard node resolve algo

mitchellhamilton: I think we might be able to remove these mappings since it should just work with lerna/yarn linking but I just haven't tried. I looked at lerna-alias before and I tried to use it for the rollup config but it wasn't working, I can't remember exactly why though. I think it would be really good for the main jest config though.

packages/emotion-theming/src/theme-provider.js
```diff
@@ -1,19 +1,27 @@
...
 const isPlainObject = test =>
   Object.prototype.toString.call(test) === '[object Object]'
 
-class ThemeProvider extends Component {
+type Props = {
+  theme: Theme
+}
+
+class ThemeProvider extends Component<Props> {
+  getTheme: ThemeProvider.prototype.getTheme
```
Andarist: had no idea a type can be specified like this, is there reason why it's defined like this and why it's defined at all? i mean - methods should be inferrable from the class definition, right?

packages/emotion-theming/src/theme-provider.js
```diff
@@ -1,19 +1,27 @@
...
   constructor() {
     super()
     this.getTheme = this.getTheme.bind(this)
   }
-
+  outerTheme: Object
+  broadcast: *
+  unsubscribeToOuterId: number
```
Andarist: this is not entirely true, the correct type would be `?number` although making it so would probably cause flow complain before using this and would force a runtime check 🤔

packages/emotion-theming/src/theme-provider.js
mitchellhamilton: Flow complains without it because of the bind in the constructor https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjGBDAzrsAFQAsBLAOwHMwBvVMMDOc3AFwCcBXDVudgCgCUtemFZlcAOlxwAtgFMAsvPFwAJmAC8YidLlKVxdZIBGFNf3Glcg0QF9RMhctUXhdBvdR2gA

Andarist: wouldnt `transform-class-properties` fix this?

PR #466, opened 2017-11-18
Prefixing

### LINKED ISSUE #400: Disable Vendor Prefixing?
- `emotion` version: 8.0.5

https://github.com/emotion-js/emotion/blob/master/packages/emotion/src/index.js#L9

Does the current API provide a way to disable vendor prefixing? I noticed that stylis can take a `prefix: false` option, but I can't find any way to set it.

greggb:
**What**: Add prefix option to babel plugin

**Why**: fixes #400 

**How**: Lovingly

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

I'd like to integrate [Browserslist](https://github.com/ai/browserslist) support, but not yet sure if that should be a Stylis plugin or part of babel-plugin-emotion.

### DIFF
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -22,6 +22,16 @@ export function hashArray(arr) {
 
 const staticStylis = new Stylis({ keyframe: false })
 
+const setIsPrefixed = ({ prefix = true }) => {
+  if (typeof prefix === 'function') {
+    staticStylis.use({
+      prefix
+    })
+  } else {
+    staticStylis.set({ prefix })
+  }
+}
+
 export function hoistPureArgs(path) {
   const args = path.get('arguments')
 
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-emotion/src/index.js
```diff
@@ -21,6 +21,20 @@ export function hashArray(arr) {
...
 
 const staticStylis = new Stylis({ keyframe: false })
+/**
+ * Sets the prefix option for Stylis.
+ * @param {Object} opts - babel options
+ * @param {boolean|function} opts.prefix - should the rule be prefixed
+ */
+const setIsPrefixed = ({ prefix = true }) => {
+  if (typeof prefix === 'function') {
+    staticStylis.use({
```
thysultan: The [documentation](https://github.com/thysultan/stylis.js#vendor-prefixing) for this had a mistake, this should use `.set`

greggb:
@mitchellhamilton You're right. I'll update

PR #468, opened 2017-11-21
Allow components to be used outside of `ThemeProvider` context

### LINKED ISSUE #467: Cannot use react-emotion without emotion-theming
- `emotion` version: 8.0.10
- `react` version: 16.0.0

### Problem description:
When seeing up a new project using react-emotion (without also setting up emotion-theming with a `ThemeProvider`, I get the following error in my console:

```term
TypeError: Cannot read property 'theme' of undefined
    at Styled.render (/Users/craigcavalier/workspace/ezcater/store/node_modules/react-emotion/dist/index.cjs.js:159:42)
```

### What you did:
I can avoid the error by including a theme provider, but this seems unnecessary as I'm not using themes.

```javascript

  <ThemeProvider theme={{}}>
    <app />
  </ThemeProvider>

```

### Suggested solution:
It seems like it should be possible to use react-emotion without emotion-theming. I __think__ if the following null check is update to check for `undefined`,  the error is avoided: https://github.com/emotion-js/emotion/blob/6992084aa91c824cc723db0e66dbb50a0b59f443/packages/react-emotion/src/index.js#L88

CraigCav:
**What**:
Fixes issue #467.

**Why**:
Currently react-emotion throws an error if a component is used outside the context of a `ThemeProvider`.

**How**:
If `state` is `undefined` or `null`, then we do not read from `state.theme`.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/react-emotion/src/index.js
+++ b/packages/react-emotion/src/index.js
@@ -85,7 +85,7 @@ const createStyled = (tag, options: { e: string, label: string }) => {
       render() {
         const { props, state } = this
         this.mergedProps = omitAssign(testAlwaysTrue, {}, props, {
-          theme: (state !== null && state.theme) || props.theme || {}
+          theme: (state && state.theme) || props.theme || {}
         })
 
         let className = ''


packages/react-emotion/src/index.js
```diff
@@ -85,7 +85,7 @@ const createStyled = (tag, options: { e: string, label: string }) => {
       render() {
         const { props, state } = this
         this.mergedProps = omitAssign(testAlwaysTrue, {}, props, {
-          theme: (state !== null && state.theme) || props.theme || {}
+          theme: (state && state.theme) || props.theme || {}
```
tkh44: Could this be `Object.assign({}, props.theme, state.theme)`?

tkh44: Not sure on the implications of using Object.assign as far as browser support goes.

CraigCav: The `undefined`/`null` check would still be necessary with object.assign. I don't think Object.assign has any support in IE.

CraigCav:
Closing this PR as issue #467 was resolved by updating `react-apollo` from `2.0.0` to `2.0.1`.

PR #474, opened 2017-11-24
reimplement component selectors

### LINKED ISSUE #461: Reimplement component selectors
Based on this convo: https://twitter.com/probablyup/status/928552573057097728

Action items:

1. Generate a second, stable hashed `className` for each `styled` component and tag it as a static property like `__emotion_target` (it should not change based on the style content of the component, but using `withComponent` should generate a new target)

2. When interpolating a component, `__emotion_target` should be used to create the selector.

The end result should look like this:

```js

const StyledThing = styled.div`
    color: blue;
`;

const StyledOtherThing = styled.div`
    color: red;

    ${StyledThing} {
        color: green;
    }
`;
```

and emits CSS like:

```css
.{dynamicHash}-StyledThing { color: blue }
.{dynamicHash}-StyledOtherThing { color: red }
.{dynamicHash}-StyledOtherThing .{stableHash}-StyledThing { color: green }
```

probablyup:
Fixes #461

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Code complete


packages/babel-plugin-emotion/src/index.js
```diff
@@ -97,6 +97,17 @@ export function replaceCssWithCallExpression(
 export function buildStyledCallExpression(identifier, tag, path, state, t) {
   const identifierName = getIdentifierName(path, t)
 
+  let stableClassName = `el-${hashString(state.file.opts.filename)}`
```
probablyup: I think filename + identifier name should be unique enough to generate a stable hash per emotion element instance.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -94,9 +94,26 @@ export function replaceCssWithCallExpression(
...
+
+  if (identifierName) {
+    stableClassName += `-${identifierName}`
+  }
+
+  const targetExpression = t.objectProperty(
+    t.identifier('target'),
+    t.stringLiteral(stableClassName)
+  )
+
```
mitchellhamilton: Could you use this code to generate the class name? (especially with the `css-` prefix)

https://github.com/emotion-js/emotion/blob/54f054a51e72ade3673f3d5a8fd0b0bab6e35643/packages/babel-plugin-emotion/src/index.js#L81-L134

packages/emotion-utils/src/index.js
```diff
@@ -7,6 +7,13 @@ export function memoize(fn) {
   }
 }
 
+export const KEY_STYLES = '__emotion_styles'
+export const KEY_TARGET = '__emotion_target'
```
mitchellhamilton: Could you rename these to `STYLES_KEY` and `TARGET_KEY` respectively so it's easier to immediately understand what they're for?

packages/emotion-utils/src/index.js
```diff
@@ -7,6 +7,13 @@ export function memoize(fn) {
...
+
+export function isEmotionElement(fn) {
+  return KEY_STYLES in fn
+}
```
mitchellhamilton: Could you inline this function and use `fn[STYLES_KEY] !== undefined` instead of `in`?

packages/emotion/src/index.js
```diff
@@ -49,6 +56,19 @@ function handleInterpolation(
     case 'boolean':
       return ''
     case 'function':
+      if (isEmotionElement(interpolation)) {
+        if (
+          process.env.NODE_ENV !== 'production' &&
+          !(KEY_TARGET in interpolation)
```
mitchellhamilton: Use `!== undefined` as above.

mitchellhamilton:
You may want to use some of the tests and stuff that was removed in #334

packages/babel-plugin-emotion/src/index.js
probablyup: Happy to change `el-` to `css-`. Why the more complicated hash function though? styled-components has a lot of bugs around hash generation... I'd rather go with as simple a solution as possible to avoid them and noticed styled uses a lot similar code to what was removed.

mitchellhamilton: The more complicated hash function is required so that it's unique, deterministic and won't change even if the project has a different full path (hence using the path to the nearest package.json).

mitchellhamilton: What bugs around hash generation are you referring to?

probablyup: https://github.com/styled-components/babel-plugin-styled-components/issues/66 Is the big one

probablyup: Backing up to package JSON definitely makes sense though. I’ll implement that.

probablyup:
Side question - has it ever been confirmed that the tests in react-emotion using the babel plugin work?

I added a few and copied the structure from the auto-label subtree but keep getting an error that indicates the babel plugin isn’t being used.

packages/babel-plugin-emotion/src/index.js
mitchellhamilton: @probablyup styled-components/babel-plugin-styled-components#66 would already happen with your code since you're using the `identifierName` in the hash.

mitchellhamilton: Thinking about it more, using the identifierName is the real problem in the hash so using the module name, path and the original code will work fine and styled-components/babel-plugin-styled-components#66 won't occur.

probablyup: The old code appears to use identifierName as well: https://github.com/emotion-js/emotion/blob/54f054a51e72ade3673f3d5a8fd0b0bab6e35643/packages/babel-plugin-emotion/src/index.js#L162

mitchellhamilton: I'm purely referring to the code above where the identifier name is not used.

probablyup: Would tracking position not cause the same issue if babel is rewriting the variable assignment and potentially introducing more lines?

probablyup: Nvm I think I get it now. It's using an incrementor to just refer vaguely to the nth component in the file, right?

mitchellhamilton: @probablyup yes.

probablyup:
@mitchellhamilton any further comments? I think I addressed what you asked for. The only thing I can't figure out is how to make this work with `extractStatic`, if that matters.

packages/react-emotion/src/index.js
```diff
@@ -123,12 +129,16 @@ const createStyled = (tag, options: { e: string, label: string }) => {
...
 
-    Styled.withComponent = nextTag => {
-      return createStyled(nextTag, options)(styles)
+    if (stableClassName) {
+      Styled[TARGET_KEY] = stableClassName
+    }
+
+    Styled.withComponent = (nextTag, nextOptions: { target: string } = {}) => {
+      return createStyled(nextTag, { ...options, ...nextOptions })(styles)
     }
```
mitchellhamilton: Could you change this to
``` jsx
Styled.withComponent = (nextTag, nextOptions?: { target: string }) => {
  return createStyled(
    nextTag,
    nextOptions !== undefined
      ? omitAssign(testAlwaysTrue, {}, options, nextOptions)
      : options
  )(styles)
}
```

packages/react-emotion/src/index.js
```diff
@@ -123,12 +129,16 @@ const createStyled = (tag, options: { e: string, label: string }) => {
...
     Styled.contextTypes = contextTypes
-    Styled.__emotion_styles = styles
+    Styled[STYLES_KEY] = styles
     Styled.__emotion_base = baseTag
     Styled.__emotion_real = Styled
 
-    Styled.withComponent = nextTag => {
-      return createStyled(nextTag, options)(styles)
+    if (stableClassName) {
+      Styled[TARGET_KEY] = stableClassName
```
mitchellhamilton: There's no need to wrap this in an `if`.

packages/react-emotion/src/index.js
```diff
@@ -41,20 +42,21 @@ const omitAssign = function(testFn, target) {
   return target
 }
 
-const createStyled = (tag, options: { e: string, label: string }) => {
+const createStyled = (
+  tag,
+  options: { e: string, label: string, target: string } = {}
```
mitchellhamilton: Please don't use default parameters as the ES5 output will be larger.

packages/emotion/test/component-selector/component-selector.test.js
```diff
@@ -0,0 +1,46 @@
...
+
+    delete FakeComponent[TARGET_KEY]
+
+    expect(() => {
+      css`
+        ${FakeComponent} {
+          color: red;
+        }
+      `
+    }).toThrow()
```
mitchellhamilton: Could you change this to `.toThrowErrorMatchingSnapshot()`

Also, these tests don't need to be in a separate folder, we only have tests in a separate folder if they need a custom babel config.

packages/emotion-utils/src/index.js
```diff
@@ -7,6 +7,13 @@ export function memoize(fn) {
   }
 }
 
+export const STYLES_KEY = '__emotion_styles'
+export const TARGET_KEY = '__emotion_target'
+
+export function isEmotionElement(fn) {
```
mitchellhamilton: Could you please inline this function. I'd rather not expose it, especially in a place where people can import it since it means people could start relying on it. I'm okay with exporting the keys since I think it's clear that they're implementation details.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -328,62 +366,83 @@ export default function(babel) {
...
+              path.node.callee &&
+              path.node.callee.property &&
+              path.node.callee.property.name === 'withComponent'
+            ) {
+              if (path.node.arguments.length === 1) {
+                path.node.arguments.push(
+                  t.objectExpression([
+                    buildTargetObjectProperty(path, state, t)
+                  ])
+                )
```
mitchellhamilton: Is there a reason this is done on the exit?

docs/babel.md
```diff
@@ -34,3 +34,6 @@ import { css, keyframes, fontFace, injectGlobal, flush, hydrate } from 'emotion/
 ```
 
 [create-react-app issue discussing macros](https://github.com/facebookincubator/create-react-app/issues/2730).
+
+### Components as selectors
+The ability to refer to another component to apply override styles depending on nesting context. Learn more in the [react-emotion docs](./styled.md#targeting-another-emotion-component).
```
mitchellhamilton: Could you add this to the table in the `babel-plugin-emotion` README?

packages/babel-plugin-emotion/src/index.js
probablyup: It seemed like a safe thing to do, since it's tagging on some additional stuff once the other processing is complete. I can move it to `enter` though if you'd prefer.

mitchellhamilton: >it's tagging on some additional stuff once the other processing is complete

I'm not sure what you mean by this though it shouldn't affect anything in this case so you can just leave it there.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -94,24 +95,62 @@ export function replaceCssWithCallExpression(
...
+function buildTargetObjectProperty(path, state, t) {
+  if (state.count === undefined) {
+    state.count = 0
+  }
+
+  const positionInFile = state.count++
+  const stableClassName = getName(
+    `${hashString(
+      normalizeFilename(state.file.opts.filename)
+    )}-${positionInFile}`,
```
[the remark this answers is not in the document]
probablyup: I could, but it would mean that changing the contents of a test file would cause the hashes to change and blow the test snapshots.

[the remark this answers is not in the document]
mitchellhamilton: For tests that don't use jest-glamor-react, that's correct. We need to use the code so that we have a fallback in case we're in a browser or something else where we don't have a filesystem/a filename and in general, it's another way to ensure that the hash is unique. It'll only affect the internal emotion tests so I don't really think it's a problem to break the snapshots.

probablyup:
Thanks Mitchell for all the reviews!

Andarist:
Great work @probablyup 👏 Have your figured out `extractStatic` support? Cannot find a note in the docs that it might not work with this feature, but you have mentioned having problems with this.

tkh44:
`extractStatic` could not work because by definition you would have expressions in your style blocks.

Andarist:
Isnt Component as Selector transformed to a static class name?

mitchellhamilton:
The class used for targeting components is added to the styled call, To make the component that targets another component be extracted we would have to know an interpolation is a styled component and know where it's defined which would be very hard to do.

You can use component selectors to target a styled component that is extracted but the one that targets the other will not be extracted.

Andarist:
Ok, gotcha - having runtime info would solve the problem, right? That's of course tougher to achieve with babel.

mitchellhamilton:
What do you mean by "runtime info"?

Andarist:
By statically analyzing a single file we don't know what imported things are, right? If we could trace the imports and require them to inspect their type etc it should get easy to use that 'runtime' info to extract more css when possible.

Btw - does it mean that interpolated 'static partials' cannot be extracted at the moment too?

mitchellhamilton:
> By statically analyzing a single file we don't know what imported things are, right? If we could trace the imports and require them to inspect their type etc it should get easy to use that 'runtime' info to extract more css when possible.

That's correct but tracing imports would get very annoying very quickly since we would have to deal with aliasing and other things like that and in general it would be complex.

> Btw - does it mean that interpolated 'static partials' cannot be extracted at the moment too?

Yes, the only things that are extracted right now are things with no interpolations whatsoever.

To be perfectly honest, we don't care about extraction very much since it dramatically limits the power of css-in-js.

PR #475, opened 2017-11-27
export emotion-theming from react-emotion

probablyup:
While refactoring for a proof of concept, it quickly became cumbersome to have to split up the import just to pull in theming HOCs.

Since theming functionality is partially built-in by default in emotion v8 via supporting `p => p.theme` etc, I think it makes sense to go the rest of the way and re-export emotion-theming for ease of use.

The package itself should probably go away for emotion v9.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


probablyup:
@tkh44 @mitchellhamilton hey folks, any comments on this?

probablyup:
From a somewhat selfish standpoint, this is a soft blocker on an internal migration away from styled-components at my company... trying to make the switch as painless as possible.

Out of curiosity, why the hesitation to merge this based on our previous twitter convo @tkh44?

tkh44:
Sorry, I've been on vacation for the last few weeks. 

@mitchellhamilton and I have talked about this about a week ago and we both were leaning on not merging. After some thought I think this might make sense just from a developer experience pov. Let's remove as much friction as possible for devs onboarding with emotion.

@mitchellhamilton when we talked you mentioned some concerns with the preact build. Can you elaborate here? What are the other cons?

probablyup:
> After some thought I think this might make sense just from a developer experience pov

Very much agree. When I was refactoring it felt weird to consume theming from a separate package — like it was disconnected in some way. But that really isn’t the case since it relies on certain static variables being available, etc.

mitchellhamilton:
`emotion-theming` requires `React.Children.only` so it requires `preact-compat` whereas `preact-emotion` currently only requires `preact` so exporting `emotion-theming` from `preact-emotion` would mean it would require an alias from `react` to `preact-compat`.

I think what we could do is make `emotion-theming` export two functions, `createWithTheme` and `createThemeProvider` which accept a React-like API and a channel, they would return what their names imply. We would then export the created versions from `react-emotion` and `preact-emotion`. For the `React.Children.only` thing, we could check if `Children` exists and return `props.children[0]` if it doesn't exist.

probablyup:
I could just revise emotion-theming. Children.only is essentially just a
check to prevent an array of children, so shouldn’t be challenging to
replace that check with a simpler alternative.

packages/emotion-theming/src/theme-provider.js
```diff
@@ -77,8 +77,18 @@ class ThemeProvider extends Component {
   render() {
     if (!this.props.children) {
       return null
+    } else if (Array.isArray(this.props.children)) {
```
tkh44: One trick I stole from preact-compat is to coerce it into an array.

tkh44: https://github.com/tkh44/holen/blob/master/src/index.js#L4

tkh44: Another option is to just not worry about it except for a warning. For React > 16 it wouldn't matter anyway, right?

probablyup:
Yeah for 16+ it wouldn’t matter.

packages/emotion-theming/src/theme-provider.js
probablyup: Yeah for 16+ it wouldn’t matter.

I went ahead and updated the impl to be simpler :) along the lines of your suggestion

probablyup:
Any more notes?

mitchellhamilton:
In my comment before, I should have made it more clear that this would mean preact-emotion requires an alias from `react` to `preact`. While most preact apps have an alias from `react` to `preact-compat`, it would mean that preact wouldn't be a first class citizen in preact-emotion, this was why I was suggesting having factories. The other option would be to make emotion-theming a private module and bundle it in preact-emotion and react-emotion with react aliased to preact in the preact-emotion build. IMO factories make more sense because they'll fit with #464.

probablyup:
Gotcha. Okay I’ll work on this over the next few days 👍

probablyup:
@mitchellhamilton okay, I converted withTheme and ThemeProvider into factory functions... but I don't really understand how babel-macro works. Mind giving some guidance on how to get the Preact part done?

mitchellhamilton:
@probablyup babel-macros isn't related to this. First, could you remove `createTheming` and change it to be separate factories, having them as separate factories means we can add the pure flag so they can be individually removed if they're not used. I should have said this before but could you base this on the `instances` branch and only export the factories from `emotion-theming` and then in `react-emotion` and `preact-emotion` create and export `withTheme` and `emotionTheming`. Also, I know before I said to have a channel option but I think let's just make the only option be the React-like library, especially if we change things in the future because of `createContext` and etc.

probablyup:
I'm not going to be able to work on this for a while, so just going to close.

PR #476, opened 2017-11-27
omit file contents from hashing function

probablyup:
1. Module name + file path is enough information to create a unique hash

2. Hashing file content means snapshots are constantly changing whenever
   a new test is added to a suite that doesn't use the jest-glamor plugin,
   which looks like a false positive and defeats the purpose of snapshots

3. I'd wager most people using emotion + jest aren't using that plugin and
   will see undesired snapshot churn


probablyup:
> so that we have something to hash in environments where we don't have a filename

When would that happen?

> Is there a reason you think most people wouldn't be using jest-glamor-react?

For one, my company isn't. I tried doing a rough code search to verify my assumption:

https://github.com/search?utf8=%E2%9C%93&q=%22emotion%22+%2B+%22jest%22+extension%3Ajson+path%3A%2F&type=Code

vs

https://github.com/search?utf8=%E2%9C%93&q=%22emotion%22+%2B+%22jest-glamor-react%22+%2B+%22jest%22+extension%3Ajson+path%3A%2F&type=Code

It looks like very few public projects using emotion and jest also use that jest plugin.

mitchellhamilton:
> When would that happen?

In a browser and any time where a filename is not passed to babel.

> For one, my company isn't.

Is there a reason for this? Is there a downside to using jest-glamor-react?

> It looks like very few public projects using emotion and jest also use that jest plugin.

That search doesn't seem very representative, I looked through some of them and couldn't find any actual tests with snapshots that had emotion classes in them, there may well be some but it certainly isn't common.

probablyup:
> Is there a reason for this? Is there a downside to using jest-glamor-react?

No, just isn't part of the config and I think most people don't know about the plugin. Might be worth adding a testing section to the docs with the recommendation to install it if you folks want to make it a best practice.

> In a browser and any time where a filename is not passed to babel.

What if we adjust the hasher to only include `file.code` if a filename is not passed? That'd handle both cases, but optimize for the 99% case.

mitchellhamilton:
[We do include it in our docs](https://github.com/emotion-js/emotion/blob/master/docs/testing.md).

I would prefer to keep all environments consistent and encourage people to use jest-glamor-react.

tkh44:
Maybe the proper solution is to fork jest-glamor-react.  We can make it much easier for new users of emotion to set up and get going. Just an idea, what do you guys think? The only downside is that it is one more thing to maintain.

Issue #479, opened 2017-11-29
Emotion adds px to unitless custom properties


TrySound:
- `emotion` version: 8.0.11
- `react` version: 16.2

Relevant code.

```javascript
const Icon = styled('div')({
  ':hover': {
    '--opacity': 0.4
  }
});
```

What happened:
```css
.css-14h7is1:hover {
    --opacity: 0.4px;
}
```

Problem description:
Custom properties should not add units

Suggested solution:
...

mitchellhamilton:
Closed via #480

PR #480, opened 2017-11-29
Do not add px to css custom-properties

TrySound:
Ref #479 

**What**:

Bug

**Why**:

To support css custom properties which can be used to simplify cascade

**How**:

Just added special case with checking keys started with `--`

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion/src/index.js
+++ b/packages/emotion/src/index.js
@@ -96,7 +96,12 @@ const processStyleValue = (key, value) => {
   if (value === undefined || value === null || typeof value === 'boolean')
     return ''
 
-  if (unitless[key] !== 1 && !isNaN(value) && value !== 0) {
+  if (
+    unitless[key] !== 1 &&
+    key.indexOf('--') !== 0 &&
+    !isNaN(value) &&
+    value !== 0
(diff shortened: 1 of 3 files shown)


packages/emotion/src/index.js
```diff
@@ -96,7 +96,12 @@ const processStyleValue = (key, value) => {
   if (value === undefined || value === null || typeof value === 'boolean')
     return ''
 
-  if (unitless[key] !== 1 && !isNaN(value) && value !== 0) {
+  if (
+    unitless[key] !== 1 &&
+    key.indexOf('--') !== 0 &&
```
mitchellhamilton: Could you change this to `key.charCodeAt(1) !== 45`(45 is the char code for -) since this is a very hot path and `charCodeAt` is much faster.

TrySound: Do you mean checking both first two chat codes?

mitchellhamilton: No, there's no need to since there are no regular css properties with a dash as the second character.

PR #495, opened 2017-12-19
Add jest-emotion

mitchellhamilton:
**What**:

Add jest-emotion which is just like jest-glamor-react except it's only for emotion and the other differences described below.

**Why**:

- We might want to add an option that changes the class name prefix from `css` to something else and this package will be able to support it.
- We can add more utilities like [this](https://github.com/emotion-js/emotion/blob/67e781d636a258885d2c00f30f41fa0edd80b87f/test/pretty-css.js) to the package.
- jest-emotion-react uses different emotion internals so it works without a DOM.
- We can deviate from the current StyleSheet implementation if we want to

**How**:

Use most of the implementation from `jest-glamor-react` except get the styles from the `inserted` cache so it works without a DOM and fix the ordering of media queries so they don't show up as being inserted after a rule even if the media query was actually inserted before the rule.

**Checklist**:

- [x] Documentation
- [x] Tests (it's tested in nearly every single test for emotion)
- [x] Code complete

~~I named it jest-emotion-react since @tkh44 has already published a module with that name but maybe it should just be jest-emotion if we add other utilities that aren't react specific?~~ It's called jest-emotion now.


greggb:
Awesome, fixing peer-dep warning would be reason enough.

mitchellhamilton:
I'm gonna merge this into instances because I want to work on something that depends on it.

PR #497, opened 2017-12-20
Fixed nested themes not being republished on outer theme changes

Andarist:
**What**:

This fixes a bug when having nested `ThemeProvider`s and updating outer theme the change is not reaching the `StyledComponent`
```jsx
<ThemeProvider theme={{}}>
	<ThemeProvider theme={{}}>
		<StyledComponent>
	</ThemeProvider>
</ThemeProvider>
```

**Why**:

This is a bug fix.

**How**:
Using publish in the nested `ThemeProvider`s subscriptions.

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests 
- [x] Code complete

### DIFF
--- a/packages/emotion-theming/src/theme-provider.js
+++ b/packages/emotion-theming/src/theme-provider.js
@@ -20,8 +20,13 @@ class ThemeProvider extends Component {
     if (this.context[channel] !== undefined) {
       this.unsubscribeToOuterId = this.context[channel].subscribe(theme => {
         this.outerTheme = theme
+
+        if (this.broadcast !== undefined) {
+          this.publish(this.props.theme)
+        }
       })
     }
+
     this.broadcast = createBroadcast(this.getTheme(this.props.theme))
   }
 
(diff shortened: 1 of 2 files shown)


mitchellhamilton:
`emotion-theming` is basically the same as styled-components' theming so I think this bug might also exist in styled-components.

I also checked and it seems like this would also happen in glamorous too.

cc @kentcdodds @mxstbr

Andarist:
Nice catch! I've sent out PRs to both, still need to sit down to write a test and port it to all 3 repos. 

Those ThemeProviders are basically the same in all packages, wouldn't it make sense to join forces and have a single ThemeProvider package to maintain?

mitchellhamilton:
@Andarist There's been an effort to do that in https://github.com/cssinjs/theming but it isn't the exact API we need and no one has been particularly motivated to change it.

Andarist:
@mitchellhamilton I've added a test so the PR is ready to merge (I think 😄 )

> There's been an effort to do that in https://github.com/cssinjs/theming but it isn't the exact API we need, it's not as performant and no one has been particularly motivated to change it.

That's a bummer, seems really sad having to reimplement this fix in every lib when their codebases around this component is 90+% the same.

PR #499, opened 2017-12-21
Add key and container options to create-emotion

mitchellhamilton:
**What**:
- Add `key` option to create-emotion which changes the class name prefix, gets set on data-emotion and is used in the attribute name for the new SSR methods, this defaults to `css`
- Add `container` option to create-emotion which is the DOM Node that emotion will insert styles into, this defaults to `document.head`
- Change component selector prefix to `e` from `css-` (we can do this now because of jest-emotion!)

**Why**:

- `key` - since stylis plugins and the prefix options run after hashing, there could be collisions
- `container` - running emotion in an iframe

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

The only qualm that I have with this is that it's more configuration but I think it's okay since it's totally optional and if someone's using `create-emotion` they're configuring stuff anyway.

cc @stephenjwatkins @herrstucki


stephenjwatkins:
Love it!

herrstucki:
Awesome 👏!

PR #502, opened 2017-12-21
Pass converted arguments as array directly to the underlaying classnames helper

Andarist:
**What**:

Passing already casted to an array arguments to next function helper.

**Why**:
This avoids `arguments` access in the underlaying function which is known to be deopt in many engines.

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

This added literally few bytes to the gzipped size of the bundle, so it requires bumping up bundlesize's limits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
   "bundlesize": [
     {
       "path": "./packages/emotion/dist/emotion.umd.min.js",
-      "threshold": "6.35 Kb"
+      "threshold": "6.36 Kb"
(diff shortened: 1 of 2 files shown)


Andarist:
This will use `arguments` but only in `cx` where the conversion happens, no need to spread it to `classnames` which causes second `arguments` conversion in the `classnames`

Andarist:
Done, should pass now.

Issue #501, opened 2017-12-21
Component as Selectors for Object Syntax


smlmrkhlms:
- `emotion` version: 8.0.12
- `react` version: 16.2.0

What you did:

We use object syntax to create our emotion components. There are times when we would like to use Components as selectors, just like for template literals (https://github.com/emotion-js/emotion/blob/master/docs/styled.md#targeting-another-emotion-component). I attempted this code:
```js
const One = styled('div')({
  display: 'block'
});

const Two = styled('div')({
  [One]: {
    display: 'none'
  }
});
```
However, the markup did not reflect the style hierarchy. The elements have classes, but no associated styles:

<img width="248" alt="screenshot 2017-12-21 14 32 58" src="https://user-images.githubusercontent.com/5240571/34275766-f04111aa-e65b-11e7-9e1c-367d95b9b53e.png">

<img width="108" alt="screenshot 2017-12-21 14 32 41" src="https://user-images.githubusercontent.com/5240571/34275767-f05924ca-e65b-11e7-8c1d-934decdd5b3b.png">

<img width="242" alt="screenshot 2017-12-21 14 32 32" src="https://user-images.githubusercontent.com/5240571/34275768-f070c512-e65b-11e7-9ce9-e4fc7c4a3fc6.png">

If I swap the code out for template strings, like so:
```js
const One = styled.div`
  display: block;
`;

const Two = styled.div`
  ${One} {
    display: none;
  }
`;
```
It works as expected. Is there a reason there is not enabled for object styles syntax?

Suggested solution:
I would prefer that there were a consistent api across object style syntax.

mitchellhamilton:
I think we definitely need to support this! There are a few changes that we need to make but it would be great if you could submit a PR for this!

To add the properties to styled object calls, we need to add the target property to the options object in the styled call, it would happen in a way something like below.
```diff
diff --git a/packages/babel-plugin-emotion/src/index.js b/packages/babel-plugin-emotion/src/index.js
index 5d4699d..1b7003a 100644
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -193,6 +193,8 @@ export function buildStyledCallExpression(identifier, tag, path, state, t) {
 }
 
 export function buildStyledObjectCallExpression(path, state, identifier, t) {
+  const targetProperty = buildTargetObjectProperty(path, state, t)
+
   const identifierName = getIdentifierName(path, t)
[...]

PR #503, opened 2017-12-23
[babel-plugin-emotion][react-emotion] Adds Support for Components as Selectors Using the Object Syntax #501

smlmrkhlms:
Issue #501 

**What**: Adding the ability to use components as selectors when object syntax is being used.

**Why**: To achieve api parity regardless of the syntax being used.

**How**: Under direction of @mitchellhamilton via comments on issue #501.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/react-emotion/src/index.js
+++ b/packages/react-emotion/src/index.js
@@ -147,7 +147,9 @@ const createStyled = (
           process.env.NODE_ENV !== 'production' &&
           stableClassName === undefined
         ) {
-          throw new Error('')
+          throw new Error(
+            'Component selectors can only be used in conjunction with babel-plugin-emotion.'
+          )
         }
         return `.${stableClassName}`
       }
--- a/packages/react-emotion/src/index.js
+++ b/packages/react-emotion/src/index.js
@@ -140,6 +140,19 @@ const createStyled = (
     Styled.__emotion_real = Styled
     Styled[TARGET_KEY] = stableClassName
 
+    Object.defineProperty(Styled, 'toString', {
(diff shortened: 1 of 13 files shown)


packages/react-emotion/src/index.js
```diff
@@ -140,6 +140,19 @@ const createStyled = (
...
     Styled[TARGET_KEY] = stableClassName
 
+    Object.defineProperty(Styled, 'toString', {
+      enumerable: false,
+      value() {
+        if (
+          process.env.NODE_ENV !== 'production' &&
+          stableClassName === undefined
+        ) {
+          throw new Error('')
```
mitchellhamilton: Could you add make this error be 'Component selectors can only be used in conjunction with babel-plugin-emotion.'

smlmrkhlms:
@mitchellhamilton Absolutely! I will add the documentation right away. It looks like CI is failing because of the size? Is this going to be an issue?

mitchellhamilton:
@smlmrkhlms In the root package.json, in the bundlesize section, increase the react-emotion limit to 9.1kb

mitchellhamilton:
Could you add a test [here](https://github.com/emotion-js/emotion/blob/master/packages/emotion/test/no-babel/index.test.js) to with something like this
```jsx
expect(() => {
// try to use a component selector with the object syntax
}).toThrowErrorMatchingSnapshot()
```

smlmrkhlms:
@mitchellhamilton Thanks for your help and guidance!

PR #507, opened 2017-12-24
Use lerna-alias for rollup config + fixed rollup plugins order

Andarist:
**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,6 +5,7 @@ import babel from 'rollup-plugin-babel'
 import alias from 'rollup-plugin-alias'
 import cjs from 'rollup-plugin-commonjs'
 import path from 'path'
+import { rollup as lernaAliases } from 'lerna-alias'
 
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -53,22 +54,7 @@ const mainConfig = Object.assign({}, baseConfig, {
...
-      'create-emotion': path.resolve(
-        __dirname,
-        './packages/create-emotion/src/index.js'
-      ),
-      'create-emotion-styled': path.resolve(
-        __dirname,
-        './packages/create-emotion-styled/src/index.js'
-      )
-    }),
+  plugins: [alias(lernaAliases())].concat(basePlugins).concat(
```
Andarist: alias has to come first otherwise if we build a single package and others have their built dist files ready `rollup-plugin-node-resolve` will kick in first to resolve the package and the dist files will be used

putting alias first we make sure that it always "wins" and source files will be used

Issue #478, opened 2017-11-28
Bug on documentation for Styled


lstrrs:
- `emotion` version: 8.0.2
- `react` version: 16.0.0

Relevant code.

```javascript
import styled from 'react-emotion'
// simple element
const H1 = styled('h1')`
  color: blue;
  font-size: 48px;
  transform: scale(${props => props.scale});
`

function Greeting ({ name }) {
  // blue, 48px, and scaled 2x text
  return <H1 scale={2}>Hello {name}</H1>
}

// Component
const H2 = styled(H1)`
  font-size: ${fontSize * 2/3}px;
  color: red;
`

function Greeting ({ name }) {
  return <H2>Hello {name}</H2> // red, 32px, and scaled 2x text
}
```

What you did: Read and followed example in documentation site https://emotion.sh/

What happened: The H2 component is not scaled 2x

Suggested solution: The example needs to either remove the `and scaled 2x text` part of the comment or add the `scale={2}` prop to the example. In addition it might be helpful to rename the `H2` to something different. The reason is that as a new comer to emotion, one could get confused and expect the Greeting component to be an `h2` and not an `h1`. 

One additional suggestion is to edit the example to have a conditional on the `props.scale`

```javascript
const H1 = styled('h1')`
  color: blue;
  font-size: 48px;
  ${props => props.scale ? `
    transform : scale(${props.scale });
  ` : ''}
`
```

What do you think? I can make a pull request and edit the `docs/styled.md`

greggb:
I agree this example is a little confusing (and a little wrong). Looks like the comment should say `red, 32px` and drop the scaled part entirely.

As for the `const H2 = styled(H1)`,  the intent was probably to show how you could create a pseudo `h2` with an `h1` tag, but maybe `const smallerHeading` would get the point across better.

Definitely PR this so the docs make sense for everyone.

ipedro:
I took a stab at it @greggb on PR #508

Issue #366, opened 2017-10-01
Composing Static Styles inside of Dynamic styles does not work. 


kristenmills:
- `emotion` version: 8.0.2-12
- `react` version: 16.0.0

With `{ extractStatic: true }` on the following code does not work

```javascript
const link = css`
  display: inline-block;
  line-height: 0;
  margin-right: 2.4rem;
`;

const More = styled(Link)`
  ${link};
  color: ${props => props.color};
`;
```

In this case none of the css from `link` is applied. I think it is because `link` is treated as a static style even when a dynamic style composes it.  I noticed that the output of the css looks something like this (assuming `props.color` is `#ffffff`;
```css
.css-1lylu6r{ css-link-1r92hdh; color: #ffffff; }
```

Also if I replace link with 
```javascript
const link = css`
  display: ${() => 'inline-block'};
  line-height: 0;
  margin-right: 2.4rem;
`;
```

it works fine as the style is now dynamic.

tkh44:
This is a limitation to extracting. You need to manually apply `link` to the className wherever you use `More`

```javascript
<More className={link}/>
```

Another option is 
```javascript
const More = styled((props) => <Link {...props} className={`${props.className} ${link}} />)`
  color: ${props => props.color};
`;
```

kristenmills:
Got it. Might want to make that explicit in the docs somewhere I think.

tkh44:
Would make for a great first PR 👍

ipedro:
@tkh44 I had that happen to me as well, but only came across this issue after losing a couple hours, so I made a PR pointing this issue. #508

PR #508, opened 2017-12-26
Improving `styled` examples

ipedro:
**What**:

Improvements to [`styled`](https://github.com/emotion-js/emotion/blob/master/docs/styled.md) documentation.

**Why**:

Composite styles can fail to work on custom components if they can't properly receive the appropriate props as highlighted on issue #366 and tweaked the sample code for correctness and easier understanding as suggested on issue #478.

**How**:

Split the sample code in two and made it more legible

**Checklist**:

- [X] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/docs/styled.md
+++ b/docs/styled.md
@@ -3,9 +3,12 @@
 `styled` accepts styles as a template literal, object, or function that returns an object.
 
 ### Styling elements and components
+
+Simple element:
+
 ```jsx
 import styled from 'react-emotion'
-// simple element
+
 const H1 = styled('h1')`
   color: blue;
   font-size: 48px;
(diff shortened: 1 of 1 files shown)


docs/styled.md
```diff
@@ -16,18 +19,35 @@ function Greeting ({ name }) {
...
+    <span>
+      <Heading>Hello {name}</Heading> // blue 48px
+      <SmallHeading>I'm an explainer</SmallHeading> // black 36px
+    </span>
+  )
 }
-
 ```
+
+When styling your custom components, be sure it inherits the appropriate props otherwise it won't propagate the changes. For more info check [issue #366](https://github.com/emotion-js/emotion/issues/366).
```
mitchellhamilton: I don't understand this statement. I think it makes more sense to mention this in the extract static doc and explain that composition is not supported in extract static with an example of how to do composition as is mentioned in that issue but also explain that it is extremely limited because the styles aren't actually merged so problems with specificity may happen.

greggb: I think this is talking about checking props inside the `styled` calls in general, not just for extract. Doesn't really fit here in either case though.

ipedro: I guess the wording wasn't clear enough. Let me explain where i'm trying to get at:

I just started out a typescript project with emotion and still getting a feel for it, and the first custom component I tried to style simply didn't get the new css applied to it.

```jsx
const PaddedComponent = styled(Component)`
  padding: 10px;
`
// padding was being ignored
```

Eventually I figured that was because `Component` wasn't set up to pass through props. 
(It's a typescript project, so we have to deal with interfaces and such.)

Reading the docs there was no reference to that, so this is what this PR is all about. That and making the example a little tidier for good measure.

docs/styled.md
```diff
@@ -16,18 +19,35 @@ function Greeting ({ name }) {
   // blue, 48px, and scaled 2x text
   return <H1 scale={2}>Hello {name}</H1>
 }
+```
 
-// Component
-const H2 = styled(H1)`
-  font-size: ${fontSize * 2/3}px;
-  color: red;
+Component:
```
tkh44: Can you make these headings so they are linkable?

docs/styled.md
tkh44: I think the confusion from us comes from the fact that that particular issue is dealing with static extraction and has nothing to do with the example you presented. I'm trying to figure out what context I'm missing but falling short. If you would just remove this section and place it in another PR for discussion we can merge the other work you've done.

PR #514, opened 2018-01-02
Better example for hoist option

### LINKED ISSUE #494: [docs] better example for babel plugin's hoist option
https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion#hoist

From the given example I'm not sure from where the performance gains are coming from. I guess you mean that it can help if css/styled are used "dynamically" in functions, but in such case it seems that most commonly there would be some dynamic dependency used for object creation and thus making it impossible to hoist, like in this example:
```js
const getClass = props => css({ backgroundColor: '#fff', color: props.color })
```

Andarist:
fixes #494 

**Checklist**:
- [x] Documentation
- [ ] Tests
- [ ] Code complete


tkh44:
Thank you @Andarist.

PR #516, opened 2018-01-04
Throw a nice error when using the styled shorthand without babel-plugin-emotion and remove duplication in component selector code

mitchellhamilton:
**What**:
Throw a nice error when using the styled shorthand without babel-plugin-emotion and remove duplication in component selector code.

**Why**:

So that people know what to do when they use the shorthand and it doesn't work.

**How**:

Use a `Proxy` if it's available and we're in dev. While this won't work in browsers without Proxies I think it's okay since it's just in dev and most people will be using modern browsers for dev.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/create-emotion-styled/src/index.js
+++ b/packages/create-emotion-styled/src/index.js
@@ -1,5 +1,5 @@
 // @flow
-import { STYLES_KEY, TARGET_KEY } from 'emotion-utils'
+import { STYLES_KEY } from 'emotion-utils'
 import type { Emotion, Interpolations } from 'create-emotion'
 import { channel, contextTypes } from '../../emotion-theming/src/utils'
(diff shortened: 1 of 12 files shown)


packages/create-emotion-styled/src/index.js
```diff
@@ -165,6 +164,16 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
       return Styled
     }
   }
+  if (process.env.NODE_ENV !== 'production' && typeof Proxy !== 'undefined') {
+    createStyled = new Proxy(createStyled, {
```
tkh44: Very cool! Great idea.

PR #519, opened 2018-01-05
Site homepage layout niggle

oliverturner:
**What**:
- Fixes a layout issue on the site home page
- Additionally, defines the styles in such a way as to remove the potential for discrepancies between displayed code sample and actual running code

**Why**:
Because this is the first view a prospective user has of the product

**How**:
Changes were restricted to the affected files
- `packages/site/src/utils/demo-buttons.js`
- `packages/site/src/pages/index.js`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

---

screenshot-2018-1-5 emotion

screen shot 2018-01-05 at 17 42 13

### DIFF
--- a/packages/site/src/utils/demo-buttons.js
+++ b/packages/site/src/utils/demo-buttons.js
@@ -10,7 +10,6 @@ const textCenter = css`
 `
 
 const stringLinks = `
-
   display: flex;
   flex-direction: column;
   justify-content: center;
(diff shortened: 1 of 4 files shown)


tkh44:
Thank you!

packages/site/src/utils/demo-buttons.js
```diff
@@ -5,86 +5,128 @@ import styled, { css } from 'react-emotion'
...
+    }
   }
 }))`
 
 export const GatsbyLink = ({ primary, ...props }: *) => (
   <_GatsbyLink {...props} />
 )
 
+export const PrecompiledLinks = styled.div`
+  ${stringLinks};
```
mitchellhamilton: This breaks the site when it's built since the function interpolations are escaped so they're just strings. Also I feel like we don't need to make `Links` editable since the main point of having them editable is to let people try the API and having two components to edit doesn't show more of the API and it makes the homepage more cluttered/most people won't be able to see all the code without scrolling.

tkh44: My bad! I should have waited to merge this.

PR #533, opened 2018-01-11
Reusable Media Queries with String Styles wasn't working

ARChilton:
I found an issue with the Reusable Media Queries with String Styles example on the website, however when I add min-width: as shown in the pull request it began to work.

I am not sure if the changes I have made are a sufficient fix to update the website but I have updated media-queries.md

Thanks,
Adam


docs/media-queries.md
```diff
@@ -95,7 +95,7 @@ const mq = Object.keys(breakpoints).reduce(
       typeof breakpoints[label] === 'string' ? '' : 'px'
     accumulator[label] = cls =>
       css`
-        @media (${breakpoints[label] + suffix}) {
+        @media (min-width:${breakpoints[label] + suffix}) {
```
mitchellhamilton: Thanks for the PR! I rewrote this while changing the docs and should've tested it better, this change breaks for the string case though, could you change it to something like this?
```jsx
let prefix = typeof breakpoints[label] === 'string' ? '' : 'min-width:'
// in the media query interpolation
prefix + breakpoints[label] + suffix

ARChilton:
I'm new to contributing to other projects so I hope I have done the right things GitHub wise. But it looks like the changes made with reference to your comment @mitchellhamilton are there. 
Thank you for your work.

PR #540, opened 2018-01-17
Fix create-emotion-styled proxy with react-hot-loader and add appearance: none to search input

mitchellhamilton:
**What**:
Fix create-emotion-styled proxy with react-hot-loader and add appearance: none to search input

**Why**:
react-hot-loader tries to access some properties on styled and because of the proxy it throws.

**How**:
Return the actual values when it's those properties

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete


theKashey:
Hi! If React-Hot-Loader would like to access another properties, to be more concrete - cast this component to string - should I just handle these additional keys here?

Sorry, it's just not quite clear why one need this Proxy.

PR #544, opened 2018-01-20
Change hash implementation

### LINKED ISSUE #536: Hash for css class names is slow
Good day. We use glamor, however when amount of css classes is huge, it starts to slow down performance, because it's murmur hash function is very strong, but very slow
Before switching to emotion we ask , if you use the same one. Seems like yes.
https://github.com/emotion-js/emotion/blob/master/packages/emotion-utils/src/hash.js

mitchellhamilton:
**What**:
Use a different implementation of murmurhash2 https://github.com/garycourt/murmurhash-js/blob/master/murmurhash2_gc.js ([react-native-web also uses this](https://github.com/necolas/react-native-web/blob/86263a2fa0cd591d3fec65b66668eca53a596cff/packages/react-native-web/src/vendor/hash/index.js))

**Why**:
Closes #536 
[It's generally faster and smaller than that the one we use currently but it's the same algorithm so none of the hashes will change](https://esbench.com/bench/5a62eb829f9721009fe9c85c).

Chrome results
Chrome results

Firefox results
Firefox results

Safari Results
Safari Results

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/emotion-utils/src/hash.js
+++ b/packages/emotion-utils/src/hash.js
@@ -1,74 +1,51 @@
 // @flow
-// murmurhash2 via https://gist.github.com/raycmorgan/588423
+/* eslint-disable */
+// murmurhash2 via https://github.com/garycourt/murmurhash-js/blob/master/murmurhash2_gc.js
 
 export function hashString(str: string) {
-  return hash(str, str.length).toString(36)
+  return murmurhash2_32_gc(str, str.length).toString(36)
 }
 
-function hash(str: string, seed: number) {
-  let m = 0x5bd1e995
-  let r = 24
-  let h = seed ^ str.length
-  let length = str.length
-  let currentIndex = 0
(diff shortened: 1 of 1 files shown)


greggb:
Perf results on iOS are fairly poor. Any worry about this being a bottleneck?

mitchellhamilton:
@greggb I didn't test these on iOS but I'm guessing you're referring to Safari? While the hashmumur2 functions are slower than `string-hash` in Safari, they're still faster than all the hash functions are in Chrome so I don't think it'll be a bottleneck and in general it's more likely that userland code will be a bottleneck rather than a hash function.

greggb:
@mitchellhamilton That's what I had figured (user code as bottleneck), but I thought it was worth mentioning as older phones may suffer more.

Unitless graph on an X for posterity (couldn't scroll to get numbers)

image uploaded from ios

Issue #563, opened 2018-02-09
Label format option in babel-plugin-emotion


sgal:
- `emotion` version: 9.0.0
- `react` version: 15.6.1

**Feature description**
It would be nice to have the possibility to configure the resulting class names that are generated by emotion. Currently in `babel-plugin-emotion` there is an option `autoLabel` that does part of it. But I would like to have the option to format the class name and for example, add component name to it.

Before:
`<div class="css-fg76s-container"></div>`

After:
`<div class="css-fg76s-ComponentName-container"></div>`
or
`<div class="ComponentName-container-fg76s"></div>`

**Use case**
We recently migrated our whole project from CSS modules + Less to emotion. It's going nice so far, but one thing that makes debugging of styles not so nice is those auto-generated class names. With CSS modules we could tweak the resulting class name to include component name as part of it (through `css-loader` in webpack).

**Suggested solution**
I think adding new option `labelFormat` to `babel-plugin-emotion` would be sufficient. Ideally, that option should accept function that takes the path and label that is extracted from the style declaration and returns the new class name.

Alternatively, the option could take a template string similar to `localIdentName` from [webpack's css-loader](https://github.com/webpack-contrib/css-loader).
```
labelFormat: "my-class-[filename]-[hash:base64:5]-[local]"
```

I'm not sure exactly how the class names are generated now, but it would be nice to be able to customize them.

tkh44:
This would be a great feature and I think your solution may work.

The code for the label transformation, if you are using the `css` function, is here

https://github.com/emotion-js/emotion/blob/master/packages/babel-plugin-emotion/src/index.js#L114-L116

You could create a utility fn that takes in `state.opts.labelFormat` with whatever other information you need and apply it to the label. Once you get the tests running locally you can isolate them to the babel plugin test and the results should be obvious. Once you figure out that location any other instances of `state.opts.autoLabel` in the code should just be copy and paste work.

Don't be afraid to put in a PR even if you get stuck. Someone will help you out 👍

sgal:
@tkh44 Thanks for the info! I did a little digging in the codebase and figured out that hashes for class names are generated in 2 places - in `create-emotion` (for dynamic styles) and in `babel-plugin-emotion` (if extracting static styles). I guess that should not be changed, since babel plugin is optional. But that also makes `[hash:base64:5]` option in label format unnecessary cause there's no use for 2 hash strings.

I'll create PR with 2 possible vars in the label format - `[filename]` and `[local]`.

[...]

PR #566, opened 2018-02-11
Add option for label formatting

sgal:
**What**:
New option for `babel-plugin-emotion` that allows to configure label format in generated class names. Implements https://github.com/emotion-js/emotion/issues/563

**Why**:
Details are in https://github.com/emotion-js/emotion/issues/563, but in general when debugging it is nice to have the file name in the class names (especially, for composed styles). The potential of the format is quite big. The feature and solution itself are inspired by `localIdentName` option in [webpack's css-loader](https://github.com/webpack-contrib/css-loader).

**How**:
New function that handles label generation for all cases (autoLabel, labelFormat, nothing).

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/README.md
+++ b/packages/babel-plugin-emotion/README.md
@@ -214,7 +214,7 @@ For example `labelFormat: "my-classname--[local]"`, where `[local]` will be repl
 with the name of the variable the result is assigned to.
 
 Allowed values:
-* `[local]` - the name of the variable the restul of the `css` or `styled` expression is assigned to.
+* `[local]` - the name of the variable the result of the `css` or `styled` expression is assigned to.
 * `[filename]` - name of the file (without extension) where `css` or `styled` expression is located.
 
 This format only affects the label property of the expression, meaning that `css` prefix and hash will
--- a/packages/babel-plugin-emotion/README.md
+++ b/packages/babel-plugin-emotion/README.md
@@ -207,7 +207,7 @@ const brownStyles = /*#__PURE__*/ css({ color: 'brown' }, 'label:brownStyles;')
 
 `string`, defaults to `null`.
 
-This option automatically adds the `label` property (same as `autoLabel`), but allows to
+This option automatically adds the `label` property (same as `autoLabel`), but allows you to
 define the format of the resulting `label`. The format is defined via string where
 variable parts are enclosed in square brackets `[]`.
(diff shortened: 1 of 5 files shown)


tkh44:
Thank you for your contribution! I took a quick look and it looks fine, but I'd like to have @mitchellhamilton look at it as well.

sgal:
@tkh44 Thanks for the feedback! I'm currently thinking about how to handle cases when label format does not include `[local]` cause it might result in colliding class names.

One way would be to print warning and append `[local]` value automatically.
Another way is to throw an error with explanation that `[local]` is mandatory. But something in this one rubs me the wrong way.

What do you think?

mitchellhamilton:
@sgal Collisions won't matter since there will be a hash as well and if the hash is the same then the style will be the same.

packages/babel-plugin-emotion/README.md
```diff
@@ -197,11 +198,46 @@ const brownStyles = css({ color: 'brown' })
...
+const brownStyles = /*#__PURE__*/ css({ color: 'brown' }, 'label:brownStyles;')
 ```
 
 `brownStyles`'s value would be `css-1q8eu9e-brownStyles`
 
+### `labelFormat`
+
+`string`, defaults to `null`.
+
+This option automatically adds the `label` property (same as `autoLabel`), but allows to
```
mitchellhamilton: Should "but allows to" be "but allows you to"?

sgal:
@mitchellhamilton Ah, thanks for clarifying the hashing. I fixed the phrasing in the docs.

mitchellhamilton:
I've been thinking about this more and maybe the `labelFormat` option should default to `[local]` and only insert labels when the `autoLabel` option is true. So rather than having two options that enable labels where one of them also lets you configure the labels, just one to enable labels and one to configure the labels.

sgal:
@mitchellhamilton That does sound reasonable and I think it is easier to understand as well. I'll update the PR tonight.

matpaul:
@sgal, Hi! maybe also add path/folder or getLocalIdent function? In our project we use component/index.js, so filename - doesn't help here

sgal:
@matpaul Which arguments do you think `getLocalIdent` function should take?

matpaul:
@sgal 
full path to file (from project root) and local for example ?

Issue #561, opened 2018-02-07
Object styles with line clamp adding 'px'


jschr:
`emotion` version: 8.0.12
`react` version: 16.0.0

Relevant code.

```javascript
css({ 
  WebkitLineClamp: 4 
})
```

What happened:

```css
{
  -webkit-line-clamp: 4px;
}
```

Expected:
```css
{
  -webkit-line-clamp: 4;
}
```

Reproducible:
https://codesandbox.io/s/430pkxw90x

Workarounds:
```js
css`
  -webkit-line-clamp: 4;
`
```
```js
css({
  ' -webkit-line-clamp': 4
})
```
^ note the first space before the `-`.

TrySound:
@jschr You may use string literal `'0'`. Emotion just adds px to numbers.

jschr:
@TrySound That unfortunately doesn't work either: https://codesandbox.io/s/302vqy3w6p.

Looks like emotion makes an [exception for lineClamp](https://github.com/emotion-js/emotion/blob/master/packages/emotion-utils/src/index.js#L38) but this needs to be 'WebkitLineClamp'. Happy to submit a PR if that's the correct fix.

tkh44:
I'd be happy to accept a PR 👍 Please make sure to add a small test for this.

jennyvallon:
For anyone still experiencing this issue if you are not using the most up to date emotion package, wrap the value in `calc` The above workarounds did not work for me.

steevsachs:
I am still running into this with @emotion/css 11.1.3.  Passing as a string works around the behavior but fails typecheck.  Is this a regression, or are types potentially just wrong?

Andarist:
Please always try to share a repro case in a runnable form - either by providing a git repository to clone or a codesandbox. OSS maintainers usually can't afford the time to set up the repro, even if exact steps are given.

steevsachs:
Apologies, this was my misunderstanding, I was attempting to use regular `line-clamp` which I think is probably in too early support to be considered here; `WebkitLineClamp` as described in this issue functions as expected.

PR #567, opened 2018-02-12
Line clamp property needs webkit vendor prefix

jschr:
**What**:
The lineClamp property needs to be vendor prefixed to WebkitLineClamp. Resolves https://github.com/emotion-js/emotion/issues/561.

**Why**:
The lineClamp property doesn't work without the prefix and WebkitLineClamp is not in the unitless exceptions list so 'px' is added to the final rule.

**How**:
Replace lineClamp with WebkitLineClamp in the unitless exceptions.

**Checklist**:
- [x] Documentation N/A
- [ ] Tests 
- [x] Code complete N/A

Wasn't sure what kind of test you are looking for, would you mind pointing me in the right direction here? Sorry if I've missed them but I don't currently see any existing tests for this area of the code.

### DIFF
--- a/packages/emotion-utils/src/index.js
+++ b/packages/emotion-utils/src/index.js
@@ -35,7 +35,6 @@ export const unitless: { [string]: 1 } = {
   gridColumnSpan: 1,
   gridColumnStart: 1,
   fontWeight: 1,
-  lineClamp: 1,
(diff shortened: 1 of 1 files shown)


tkh44:
Thanks!

PR #580, opened 2018-02-28
Update theming.md to no longer follow dead link

woahitsraj:
Documentation was linking to a markdown file that was not there. I updated the link to follow the same path as a link earlier in the documentation

The link would send the user to a 404 page which is bad

Updated markdown file

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete


mitchellhamilton:
Thanks!

PR #583, opened 2018-03-02
Output dir support

wardpeet:
**What**: #180

**Why**: Prevents polluting src tree when preparing production / static builds.

**How**: By saving the optional key `outputDir` in the babel plugin state and doing extra stuff if that key is present at the end of the babel transformation

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete

I used the code from pr #396 and just updated it to be complient with the latest version. I changed import style to absolute just to make sure that we can't go wrong here.

### DIFF
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -453,33 +453,32 @@ export default function(babel: Babel) {
         exit(path: BabelPath, state: EmotionBabelPluginPass) {
           if (state.staticRules.length !== 0) {
             const toWrite = state.staticRules.join('\n').trim()
-            let filenameArr = path.hub.file.opts.filename
+            let cssFilename = path.hub.file.opts.sourceFileName
 
             if (state.outputDir) {
-              filenameArr = nodePath.join(
-                state.outputDir,
-                path.hub.file.opts.sourceFileName
-              )
+              cssFilename = nodePath.join(state.outputDir, cssFilename)
             }
 
-            filenameArr = filenameArr.split('.')
+            const cssFilenameArr = cssFilename.split('.')
+            // remove the extension
+            cssFilenameArr.pop()
+            // add emotion.css as an extension
+            cssFilenameArr.push('emotion.css')
 
-            filenameArr.pop()
-            filenameArr.push('emotion', 'css')
-
-            const cssFilename = filenameArr.join('.')
+            // use absolute path so resolving can't go wrong
+            cssFilename = nodePath.resolve(cssFilenameArr.join('.'))
             const baseCssName = nodePath.basename(cssFilename)
-            const dirPath = cssFilename.replace(baseCssName, '')
 
             const exists = fs.existsSync(cssFilename)
-            addSideEffect(path, './' + nodePath.basename(cssFilename))
+            addSideEffect(path, cssFilename)
             if (
               exists ? fs.readFileSync(cssFilename, 'utf8') !== toWrite : true
             ) {
               if (!exists) {
                 if (state.outputDir) {
-                  mkdirp.sync(dirPath)
+                  mkdirp.sync(nodePath.dirname(cssFilename))
                 }
+
                 touchSync(cssFilename)
               }
               fs.writeFileSync(cssFilename, toWrite)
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-emotion/src/index.js
```diff
@@ -441,6 +442,8 @@ export default function(babel: Babel) {
             // path.hub.file.opts.filename !== 'unknown' ||
             state.opts.extractStatic
 
+          state.outputDir = state.opts.outputDir
```
mitchellhamilton: There's no need to do this, just use `state.opts.outputDir`. The only reason it's done for `extractStatic` is because it used to be determined by other things like as you can see where there was a filename passed to babel.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -450,16 +453,31 @@ export default function(babel: Babel) {
...
+              cssFilename = nodePath.join(state.outputDir, cssFilename)
+            }
+
+            const cssFilenameArr = cssFilename.split('.')
+            // remove the extension
+            cssFilenameArr.pop()
+            // add emotion.css as an extension
+            cssFilenameArr.push('emotion.css')
+
+            // use absolute path so resolving can't go wrong
```
mitchellhamilton: We can't rely on absolute paths. There are so many times when it would cause problems, for example, distributing an npm package, the imports wouldn't work since they would point to paths that wouldn't exist on a consumers computer. Please use [`path.relative`](https://nodejs.org/api/path.html#path_path_relative_from_to) to get relative paths instead.

wardpeet:
so require statements are made relative but disk operations are resolved paths. I've fixed the tests and tested it on a local project.

wardpeet:
@tkh44 what's wrong with babel 7? or nothing we should deal with yet?

tkh44:
I hit this problem but got called away and didn’t investigate yet.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -495,19 +496,43 @@ export default function(babel: Babel) {
         exit(path: BabelPath, state: EmotionBabelPluginPass) {
           if (state.staticRules.length !== 0) {
             const toWrite = state.staticRules.join('\n').trim()
-            const filenameArr = path.hub.file.opts.filename.split('.')
-            filenameArr.pop()
-            filenameArr.push('emotion', 'css')
-            const cssFilename = filenameArr.join('.')
-            const exists = fs.existsSync(cssFilename)
-            addSideEffect(path, './' + nodePath.basename(cssFilename))
+            let cssFilename = path.hub.file.opts.sourceFileName
```
mitchellhamilton: I think if you replace `path.hub.file.opts.sourceFileName` with `path.hub.file.opts.generatorOpts ? path.hub.file.opts.generatorOpts.sourceFileName : path.hub.file.opts.sourceFileName` the babel 7 error will be fixed.

tkh44:
Thanks @wardpeet!

ai:
Also I afraid to file name conflict. What if we have `a/one.js` and `b/one.js`. If I understood the code correctly, they both will produce `outputDir + 'one.emotion.css'` and one override another.

ai:
Maybe we can have old behavior on `outputDir`?

wardpeet:
The behaviour should be the following 
outputdir + /a/one.emotion.css 
outputdir + /b/one.emotion.css

eatsjobs:
@wardpeet @tkh44  @mitchellhamilton I got a strange result with extract static. I know it's a deprecated feature but at the moment i can't switch to linaria o css-literal-loader. Please(on my knees) is it possibile to only have this fix in the babel plugin?

Problem: 
When transpiling with babel cli:

```javascript
babel src --out-dir esm --ignore 'src/**/*.spec.js,src/**/*.test.js' --copy-files
```
I got this result:
- *.css are generated but outside my folder project
- the import '*.css' in the transpiled code is somenthing like:
```javascript
import "../../Projects/src/Accordion/AccordionItem.emotion.css";
```

- OS: macOS Mojave
- node version v8.12.0

```javascript
    "@babel/cli": "^7.1.2",
    "@babel/core": "^7.0.0"
"emotion": "^9.2.8",
    "emotion-theming": "^9.2.6",
    "react": "^16.4.1",
    "react-dom": "^16.4.1",
    "react-emotion": "^9.2.8",
"babel-plugin-emotion": "^9.2.11"
```

```javascript
//.babelrc
["emotion",{
                        "autoLabel": true,
                        "hoist": true,
                        "sourceMap": false,
                        "extractStatic": true,
                        "outputDir": "esm"
                    }
                ]
```

PR #584, opened 2018-03-02
Preserve existing options passed to styled calls

### LINKED ISSUE #582: [enhancement - babel plugin] Respect custom options in styled calls
**babel options**
```js
module.exports = {
	plugins: [
		['emotion', { autoLabel: process.env.NODE_ENV === 'development' }],
	]
}
```

**input**
```js
const StyledBar = styled(Row, { test: true })`
	padding: 1em;
`
```

**actual**
```js
const StyledBar = styled(Row, {
	label: 'StyledBar',
	target: 'e1j58gbc0'
})`
	padding: 1em;
`
```

**expected**
```js
const StyledBar = styled(Row, {
	test: true,
	label: 'StyledBar',
	target: 'e1j58gbc0'
})`
	padding: 1em;
`
```

**Use case**
Custom emotion wrappers, I want to put some styling logic into a centralized, single piece of code - my own version of `styled` that I'm going to use across my entire app. It will be only a thin wrapper around standard emotion, so ideally I'd like to use existing babel plugin.

**Implementation**
I'm ofc willing to work on this, would like to get approval from you first though before I start to work on this.

Andarist:
fixes #582

**What**:

Preserving existing options passed to styled calls when using emotion's babel plugin

**Why**:

It would ease creating custom emotion wrappers.

**How**:

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

TODO:
- [ ] fix flow typings - I'd appreciate help with this, dunno why they are not compatible atm
- [x] handle more cases (like `buildStyledObjectCallExpression`)
- [x] decide what about rest of the arguments, they are lost at the moment too, should we preserve them?

### DIFF
--- a/packages/babel-plugin-emotion/src/index.js
+++ b/packages/babel-plugin-emotion/src/index.js
@@ -192,9 +192,10 @@ const buildFinalOptions = (t, options, ...newProps) => {
     existingProperties = options.properties
   }
 
-  return t.objectExpression(
-    [...existingProperties, ...newProps].filter(Boolean)
-  )
+  return t.objectExpression([
+    ...existingProperties,
+    ...newProps.filter(Boolean)
+  ])
 }
 
 export function buildStyledCallExpression(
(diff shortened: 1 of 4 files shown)


mitchellhamilton:
Let's preserve the rest of the arguments and for the flow errors, just add `// $FlowFixMe` comments for now.

Andarist:
So I think this is code-complete, you can make a code review of this. I'd appreciate hints on how to fix those flow types, I always struggle with a little bit more complicated cases :s

Issue #585, opened 2018-03-03
For prop getting swallowed


philholden:
- `emotion` version: 8.0.12, 9.0.2
- `react` version: 16

Relevant code.

https://codesandbox.io/s/rl1n7r0m3o

```javascript
import React from "react";
import { render } from "react-dom";
import styled from "react-emotion";

const Label = styled("label")`
  color: red;
`;

const App = () => (
  <div>
    <h2>styled swallowing "for" prop</h2>
    <div>
      <input type="checkbox" name="aa" value="1" id="one" />
      <Label for="one">Styled Label (click on label does nothing)</Label>
    </div>
    <input type="checkbox" name="bb" value="1" id="two" />
    <label for="two">
      Label element (un-styled click on label toggles check)
    </label>
  </div>
);

render(<App />, document.getElementById("root"));
```
What you did:
clicking on `styled('label')` with `for` prop for a `<input type="checkbox" id="id1" />` 

What happened:

does not update the checkbox because the `for` prop is not forwarded to the label.

Reproduction:

https://codesandbox.io/s/rl1n7r0m3o

Suggested solution:

update whitelist and add 'for'

-----

If you know any temporary way to force a prop not to be swallowed that will work with emotion 8.0.12 that would be much appreciated.

If none exists may be nice to add an escape hatch to both omit & include props to be forwarded (whitelist may be brittle)

PR #586, opened 2018-03-03
Add `for` to whitelist

philholden:
**What**:
Add `for` to the props whitelist

**Why**:
`for` attributes not getting forwarded to `styled('labels')` making checkboxes unresponsive when label clicked.

**How**:
Tried adding `for` to props whitelist radios and checkboxes now respond when clicked

- [N/A] Documentation
- [N/A] Tests
- [N/A] Code complete

https://github.com/emotion-js/emotion/issues/585

### DIFF
--- a/packages/create-emotion-styled/src/props.js
+++ b/packages/create-emotion-styled/src/props.js
@@ -57,6 +57,7 @@ const props = {
   download: true,
   draggable: true,
   encType: true,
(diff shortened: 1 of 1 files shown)


aaronjensen:
AFAICT this is actually a bug when using preact since preact uses `for` instead of `htmlFor`

PR #594, opened 2018-03-08
Add localIdent option to support same scenario as css-loader (WIP)

wardpeet:
**What**:
I'm adding a feature to choose how the babel plugin exports css classes to an external stylesheet. It's based on the code from [css-loader from webpack](https://github.com/webpack-contrib/css-loader/tree/master).

**Why**: I want my classnames to be unique the way I want them to be. I don't want to change my classNames when I change something in the file

**How**:
Added localIdentName to babel options so we can change the generated identifier. By default it fallbacks to the old scenario so this isn't a breaking change.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

Just wanted some feedback if you're willing to add this feature.

This makes it possible to generate classes like `components-header--headerClass`

I would also like to add a `getLocalIdent` function where you can generate it inside babelrc. This will only work with babel 7 where babelrc can be a js file.

### DIFF
--- a/packages/babel-plugin-emotion/src/babel-utils.js
+++ b/packages/babel-plugin-emotion/src/babel-utils.js
@@ -92,14 +92,6 @@ export function addRuntimeImports(state: EmotionMacroPluginPass, t: Types) {
     state.emotionImports = undefined
   }
 }
-export function getName(identifierName?: string, prefix: string) {
-  const parts = []
-  parts.push(prefix)
-  if (identifierName) {
-    parts.push(identifierName)
-  }
-  return parts.join('-')
-}
 
 export function getLabel(
   identifierName?: string,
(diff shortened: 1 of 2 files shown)


wardpeet:
THIS IS JUST TO GET FEEDBACK

wardpeet:
@tkh44 @mitchellhamilton do you guys have an opinion on this? If you guys don't think it's a good idea, just let me know.

mitchellhamilton:
Sorry for not replying to this for ages but we added a [`labelFormat` option](https://emotion.sh/docs/babel-plugin-emotion#labelformat) in #563 which I believe covers this use case and also in general I would rather have good defaults than adding a bunch of options. I'm gonna close this since I think `labelFormat` covers this use case but feel free to discuss more if you don't think it does.

PR #605, opened 2018-03-21
Fix SSR with multiple createEmotion calls with the same context

### LINKED ISSUE #603: SSR doesn't work with multiple emotion instances.
## Versions

- `emotion` version: 9.0.2
- `react` version: N/A

## Code

```javascript
const createEmotion = require('create-emotion')
const createEmotionServer = require('create-emotion-server')

const context = {}

// instances should work together because they share context
const emotion1 = createEmotion(context)
const emotion2 = createEmotion(context)

// emotionServer can be bootstrapped from either instance.
const { renderStylesToString } = createEmotionServer(emotion1)

// Rendering works fine on components created by the first instance
const className1 = emotion1.css`color: magenta;`
console.log(renderStylesToString(`<h1 className="${className1}"></h1>`))
// <style data-emotion-css="madiid">.css-madiid{color:magenta;}</style><h1 className="css-madiid"></h1>

// Style tags are empty when rendering components created by the first instance
const className2 = emotion2.css`color: rebeccapurple;`
console.log(renderStylesToString(`<h1 className="${className2}"></h1>`))
// <style data-emotion-css="l7wj9y"></style><h1 className="css-l7wj9y"></h1>
```

## Steps to reproduce

Tried to use emotion server-side rendering with react-static.  Turns out it triggered an edge case in emotion.

To be clear, react-static is not required to reproduce the bug.  Steps to reproduce:

1. Create two instances of emotion that share a single context object.  I did this following the example in the [docs](https://emotion.sh/docs/create-emotion#context).
2. Create a css class using `css` exported from the second instance.
3. Interpolate the class into some html.
4. Try to inline styles into the html using `renderStylesToString`

## Problem description

* `renderStylesToString` creates style tags with empty contents.
* The style tags are empty because `emotion.caches.inserted[id]` is an empty string.  ([source](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion-server/src/inline.js#L18))
* `emotion.caches.inserted[id]` is an empty string because `current` never get's updated with the css rules generated by stylus via `stylis-rule-sheet` (You can test this yourself by adding `console.log(current)` after [this line](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L259)).
* [`current`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L79) never gets updated because the [`insertRule`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L81) function has captured a closure referencing `current` from the first invocation of `createEmotion`.  (Sorry if I'm not using the right technical terms.  I'm not used to discussing the innards of JS.)
* The closure points to the `current` variable from the first instance because it's used by a stylus plugin that is only registered if [`caches`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L78) has not been created yet. (i.e. the first invocation of createEmotion) ([source](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L90))

## Runable example

Minimal test-case here: https://github.com/whmountains/emotion-bug-repro
Try online at: https://codesandbox.io/s/github/whmountains/emotion-bug-repro/tree/master/?expanddevtools=1&view=editor

## Temporary workaround

I think this will work, though I haven't tested it.

Replace this:

```js
import createEmotion from 'create-emotion'

const context = typeof global !== 'undefined' ? global : {}

export const {
  flush,
  hydrate,
  cx,
  merge,
  getRegisteredStyles,
  injectGlobal,
  keyframes,
  css,
  sheet,
  caches
} = createEmotion(context)
```

With this:

```js
import createEmotion from 'create-emotion'

const world = global !== 'undefined' ? global : {}
const emotion = world.__EMOTION_INSTANCE__
  ? world.__EMOTION_INSTANCE__
  : (world.__EMOTION_INSTANCE__ = createEmotion({}))

export const {
  flush,
  hydrate,
  cx,
  merge,
  getRegisteredStyles,
  injectGlobal,
  keyframes,
  css,
  sheet,
  caches
} = emotion
```

## Suggested Solutions

I have a few ideas.  Haven't tested any of them.

1. Move [`let current`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L79) outside the [`createEmotion`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/index.js#L64) function.

2. Keep `current` inside the context object.

3. Get stylus and the stylesheet manager out of the context `object`.

mitchellhamilton:
**What**:
Fix SSR with multiple createEmotion calls with the same context

**Why**:
Closes #603 

**How**:
Store the emotion instance instead of the caches on the context object.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/create-emotion/src/index.js
+++ b/packages/create-emotion/src/index.js
@@ -15,8 +15,6 @@ type StylisPlugins = Function[] | null | Function
 type EmotionCaches = {|
   registered: { [key: string]: string },
   inserted: { [key: string]: string | true },
-  stylis: (scope: string, styles: string) => string,
-  sheet: StyleSheet,
   nonce?: string,
   key: string
 |}
@@ -62,9 +60,12 @@ type EmotionOptions = {
 }
 
 function createEmotion(
-  context: { __SECRET_EMOTION__?: EmotionCaches },
+  context: { __SECRET_EMOTION__?: Emotion },
   options?: EmotionOptions
 ): Emotion {
+  if (context.__SECRET_EMOTION__ !== undefined) {
+    return context.__SECRET_EMOTION__
+  }
   if (options === undefined) options = {}
   let key = options.key || 'css'
   if (process.env.NODE_ENV !== 'production') {
@@ -74,10 +75,7 @@ function createEmotion(
       )
     }
   }
-  // $FlowFixMe
-  let caches: EmotionCaches = context.__SECRET_EMOTION__
   let current
-
   function insertRule(rule: string) {
     current += rule
     if (isBrowser) {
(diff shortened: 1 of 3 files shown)


whmountains:
This is awesome 🎉   When can we expect a new release?

Issue #611, opened 2018-04-03
Export testOmitPropsOnStringTag for external usage


Andarist:
**Problem description:**

I have custom SVG components and I'd like to filter out DOM props before passing them to a component that renders `svg` (and it doesnt do anything else than that).

**Suggested solution:**

Just export this util so it can be reused in application code. Probably could be a separate package? Can prepare a PR for this if find it useful.

Maybe we could just expose `omitFn` as static property on the returned `Styled` class (or accept it as config option in `createStyled`) - that way one could override it nicely without introducing unnecessary component layer.

Andarist:
@mitchellhamilton cool! thank you ❤️ any plans to release this in 9.x line (I can live with duplication for now, just wondering)

Also could u comment on this one:
> Maybe we could just expose `omitFn` as static property on the returned `Styled` class (or accept it as config option in `createStyled`) - that way one could override it nicely without introducing unnecessary component layer.

mitchellhamilton:
It's already used in 9.x.
https://github.com/emotion-js/emotion/blob/10b404147df2b61fbfa4038299e6d0f6aad6af76/packages/create-emotion-styled/src/utils.js#L3

In regards to the omitFn option, I think it definitely could be useful(especially for styled-system which I really like so 👍 ) so I'd be happy to accept a PR though it should probably be called `shouldForwardProp`. I think it should probably be a option to styled since a static property feels hacky. Since we modify the options to styled in the babel plugin, I'm thinking we should accept options from the second and third arguments where the second overwrites the third and if there's a second argument passed to styled that's not an object literal, we put the options from babel on the third argument. This way, if someone uses a helper or something for the options argument, the options won't be overwritten and labels and stuff with still work.

@tkh44 do you have any thoughts on this?

Andarist:
Actually existing options are preserved (like proposed `shouldForwardProp`) while transforming since https://github.com/emotion-js/emotion/pull/584 . So it should be just matter of supporting this option in the code, no change to arguments order etc should be needed.

mitchellhamilton:
[I meant if this happens](https://github.com/emotion-js/emotion/pull/584/files#diff-3cbe27131024d4d2b0388471e182ea8eR188) but it's probably pretty uncommon so let's just leave it.

Andarist:
Yeah, it's rather uncommon especially that it's not officially supported to pass there anything more than what babel plugin adds and which is already an object.

Gonna prepare a PR soon for this so we can discuss it further under it.

PR #616, opened 2018-04-05
Support shouldForwardProp in styled's options

Andarist:
related to #611 

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/create-emotion-styled/src/utils.js
+++ b/packages/create-emotion-styled/src/utils.js
@@ -33,7 +33,12 @@ export const pickAssign: (
   return target
 }
 
-export type StyledOptions = { e: string, label: string, target: string }
+export type StyledOptions = {
+  e?: string,
+  label?: string,
+  target?: string,
+  shouldForwardProp?: (?string) => boolean
+}
 
 type CreateStyledComponent = (...args: Interpolations) => *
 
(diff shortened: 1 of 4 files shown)


packages/create-emotion-styled/src/utils.js
```diff
@@ -7,14 +7,14 @@ export function setTheme(theme: Object) {
...
+export const testPickPropsOnStringTag: (key: string) => boolean = memoize(
   isPropValid
 )
-export const testOmitPropsOnComponent = (key: string) =>
+export const testPickPropsOnComponent = (key: string) =>
   key !== 'theme' && key !== 'innerRef'
 export const testAlwaysTrue = () => true
 
-export const omitAssign: (
+export const pickAssign: (
```
Andarist: imho `omit` was misleading here, and `pick` is better suited because `pickAssign` actually picks properties if predicate (`testFn`) returns truthy, `omit` would pick them if predicate would return falsy

Andarist:
There are still some flow errors, it seems type inference doesnt handle this accurately. Let me know if u want this refactored or should I just use FlowFixMe? Kinda hating to adjust actual, valid, code just to satisfy inference algorithm

mitchellhamilton:
Use FlowFixMe

Andarist:
Done

PR #617, opened 2018-04-05
Add defaultProps using the options argument in `styled`

tkh44:
**What**:
You can specify default props in the options argument of `styled`

```javascript
const MyComponent = styled('input', {
  props: {
    'type': 'checkbox'
  }
})({
  height: 40,
  width: 40,
  color: 'green'
})
```

**Why**:
I got annoyed.

**How**:
I did a simple merge in `render`. I may be missing something obvious but this should work fine.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/docs/styled.md
+++ b/docs/styled.md
@@ -4,7 +4,6 @@ title: "Styled Components"
 
 `styled` is a way to create React or Preact components that have styles attached to them. It's available from [react-emotion](/packages/react-emotion) and [preact-emotion](/packages/preact-emotion). `styled` was heavily inspired by [styled-components](https://www.styled-components.com/) and [glamorous](https://glamorous.rocks/)
 
-
 ### Styling elements and components
 
 `styled` is very similar to `css` except you call it with an html tag or React/Preact component and then call that with a template literal for string styles or a regular function call for object styles.
@@ -48,26 +47,6 @@ render(
 )
 ```
 
-### Default Props
-
-Default props can be set via the second argument of `styled`. 
-
-`this.props` is merged with and overwrites any overlapping keys in `defaultProps`.  
-
-*Does not work with shorthand syntax*
-
-```jsx live
-const MyComponent = styled('input', {
-  props: {
-    'type': 'checkbox'
-  }
-})({
-  height: 40,
-  width: 40,
-  color: 'green'
-})
-```
-
 ### Styling any component
 
 `styled` can style any component as long as it accepts a `className` prop.
@@ -84,7 +63,6 @@ const Fancy = styled(Basic)`
 
 render(<Fancy />)
 ```
-
 
 ### `withComponent`
 
--- a/docs/styled.md
+++ b/docs/styled.md
@@ -4,6 +4,7 @@ title: "Styled Components"
 
 `styled` is a way to create React or Preact components that have styles attached to them. It's available from [react-emotion](/packages/react-emotion) and [preact-emotion](/packages/preact-emotion). `styled` was heavily inspired by [styled-components](https://www.styled-components.com/) and [glamorous](https://glamorous.rocks/)
 
+
 ### Styling elements and components
 
 `styled` is very similar to `css` except you call it with an html tag or React/Preact component and then call that with a template literal for string styles or a regular function call for object styles.
@@ -47,6 +48,26 @@ render(
 )
 ```
 
+### Default Props
+
+Default props can be set via the second argument of `styled`. 
+
+`this.props` is merged with and overwrites any overlapping keys in `defaultProps`.  
+
+*Does not work with shorthand syntax*
+
+```jsx live
+const MyComponent = styled('input', {
+  props: {
+    'type': 'checkbox'
+  }
+})({
+  height: 40,
+  width: 40,
+  color: 'green'
+})
+```
+
 ### Styling any component
 
 `styled` can style any component as long as it accepts a `className` prop.
@@ -65,9 +86,9 @@ render(<Fancy />)
 ```
 
 
-### Change the rendered tag using `withComponent`
+### `withComponent`
 
-Sometimes you want to create some styles with one component but then use those styles again with another component, the `withComponent` method can be used for this. This was inspired by [styled-components' `withComponent`](https://www.styled-components.com/docs/api#withcomponent).
+Change the underlying tag of the styled component.
 
 ```jsx live
 // Create a section element
@@ -76,12 +97,6 @@ const Section = styled('section')`
 `
 // Create an aside element with the same styles as Section
 const Aside = Section.withComponent('aside')
-render(
-  <div>
-    <Section>This is a section</Section>
-    <Aside>This is an an aside</Aside>
-  </div>
-)
 ```
 
 ### Targeting another emotion component
--- a/packages/create-emotion-styled/src/index.js
+++ b/packages/create-emotion-styled/src/index.js
@@ -22,20 +22,28 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
         )
       }
     }
+
+    const isReal = tag.__emotion_real === tag
+
     let staticClassName
     let identifierName
-    let stableClassName
-    let defaultProps
+    let stableClassName = tag.__emotion_target
     let shouldForwardProp
+    let extraProps = tag.__emotion_props
 
     if (options !== undefined) {
       staticClassName = options.e
       identifierName = options.label
       stableClassName = options.target
-      defaultProps = options.props
       shouldForwardProp = options.shouldForwardProp
+      if (options.withProps !== undefined) {
+        extraProps =
+          isReal && tag.__emotion_props
+            ? tag.__emotion_props.concat(options.withProps)
+            : [options.withProps]
+      }
     }
-    const isReal = tag.__emotion_real === tag
+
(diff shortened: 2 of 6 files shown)


ChristopherBiscardi:
Cool, Seems fine to me (and useful). Test failures seem to be trivial type issues.

docs/styled.md
```diff
@@ -47,6 +48,26 @@ render(
...
+
+Default props can be set via the second argument of `styled`. 
+
+`this.props` is merged with and overwrites any overlapping keys in `defaultProps`.  
+
+*Does not work with shorthand syntax*
+
+```jsx live
+const MyComponent = styled('input', {
+  props: {
```
Andarist: seems like `defaultProps` instead of just `props` would nicely relate to true React APIs

packages/create-emotion-styled/src/index.js
```diff
@@ -85,7 +88,9 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
           }
         }
         render() {
-          const { props, state } = this
+          const state = this.state
+          const props = Object.assign({}, defaultProps, this.props)
```
Andarist: just a thought - it generally should be the same, but maybe worth to offload that work to builtin React's mechanism of `defaultProps`? u could just assign `static defaultProps = options.props`

diegohaz:
Why not this?
```jsx
const MyComponent = styled('input')({
  height: 40,
  width: 40,
  color: 'green'
})

MyComponent.defaultProps = {
  'type': 'checkbox'
}
```

Andarist:
@diegohaz I guess just convenience - personally I use what u've shown and im just fine with it. It's easily doable in userland and doesnt cost anything in the library (because there is nothing special in it for that case) - actual implementation (as we can see in this PR) is rather minimal though, so from my point of view  🤷‍♂️

ChristopherBiscardi:
I saw some conversation on Twitter about possibly making this a function, so here's a (possible?) use case where I do this currently to make styled work with Formik:

```js
const TextInput = styled.input`
  outline: none;
  padding: 11px 10px;
  border: 1px solid #d4d4d2;
  border-radius: 3px;
  box-shadow: none;
  transition: border 0.3s, color 0.3s;
  margin-bottom: ${({ hasError }) => (hasError ? "8px" : bottomMargin)};
`;

const FormikTextInput = ({ field, ...props }) => <TextInput type="text" {...field} {...props} />;
```

but it might be nice to

```js
const FormikTextInput = styled('input', ({ field, ...props }) => {
    type:'text',
    ...field,
    ...props
  })`
  outline: none;
  padding: 11px 10px;
  border: 1px solid #d4d4d2;
  border-radius: 3px;
  box-shadow: none;
  transition: border 0.3s, color 0.3s;
  margin-bottom: ${({ hasError }) => (hasError ? "8px" : bottomMargin)};
`;
```

both used as 

```js
render={({ errors, touched, isSubmitting }) => (
...
    <Field
      component={FormikTextInput}
      name="name"
      placeholder="Give the query a descriptive name"
      hasError={touched.name && errors && errors.name}
      disabled={isSubmitting}
    />
```

(disclaimer, I haven't thought this through that much. Just pulled some code out of a codebase I had).

diegohaz:
We had a [discussion](https://github.com/styled-components/styled-components/pull/365#issuecomment-271981558) about it on styled-components. I always vote for keeping public API as simple as possible, specially when it can already be achieved in another simple way.

Less bugs, less things to maintain, less things to upgrade in future versions etc.

I see some value on it If you guys are planning to accept functions that return default props based on actual props. But I'd wait for more people to come up with use cases. 😝

mitchellhamilton:
I'm good with adding this in general. I think we should probably call it `withProps` and have a method like `withComponent` but `withProps` to align with glamorous and IMO the method is more convenient but let's have both and implement the method with the option.
Let's definitely allow functions.
Let's just make it the same as glamorous.
https://github.com/kentcdodds/glamorous-website/blob/master/pages/api/content/glamorous.md#withprops

also, this doesn't handle this case 
```jsx
const _Button = styled("button", {
  props: {
    style: { color: "hotpink" }
  }
})`
  color: turquoise;
`;

const Button = styled(_Button)();
```

tkh44:
> I'm good with adding this in general. I think we should probably call it withProps and have a method like withComponent but withProps to align with glamorous and IMO the method is more convenient but let's have both and implement the method with the option.

> Let's definitely allow functions.

How many recompose apis do you think we could provide legit use without ballooning the size? (https://github.com/acdlite/recompose/blob/master/docs/API.md#withprops)

> Let's just make it the same as glamorous.
kentcdodds/glamorous-website:pages/api/content/glamorous.md@master#withprops

I agree.

tkh44:
@ChristopherBiscardi thats pretty much what I had in mind. (https://github.com/acdlite/recompose/blob/master/docs/API.md#withprops)

@diegohaz I see what you mean in this case

```
const MyComponent = styled('input')({
  height: 40,
  width: 40,
  color: 'green'
})

MyComponent.defaultProps = {
  'type': 'checkbox'
}
```

In the case of `className` though we need to merge them in a special way. The default prop would never appear because we apply a class name ourselves so the value in defaultProps will never apply
```
const MyComponent = styled('input')({
  height: 40,
  width: 40,
  color: 'green'
})

MyComponent.defaultProps = {
  'className': 'marketing-needs-this-😖-class-name'
}
```

docs/styled.md
tkh44: I agree with `props` not being optimal. I like @mitchellhamilton's idea of using `withProps` to match recompose.

packages/create-emotion-styled/src/index.js
tkh44: https://github.com/emotion-js/emotion/pull/617/files#diff-77b4dbda613d78e500ded003e8362f64R1125

tkh44: Maybe you are onto something though. Maybe there is a need for `defaultProps` and `withProps`

mitchellhamilton: @tkh44 I don't think it would fail, the example below would fail with defaultProps though and probably does with this current implementation as well

```jsx
const SomeComponent = styled('div', { props: { className:'thing' }})(...styles)

render(<SomeComponent className="other-thing" />)
```

docs/styled.md
```diff
@@ -76,12 +97,6 @@ const Section = styled('section')`
 `
 // Create an aside element with the same styles as Section
 const Aside = Section.withComponent('aside')
-render(
```
mitchellhamilton: The render function is needed to render the example.

browniefed:
If the `props` is a function can theme be injected? I think that would make this more useful than a generic `defaultProps` assignment and or functional component

Andarist:
> In the case of className though we need to merge them in a special way. The default prop would never appear because we apply a class name ourselves so the value in defaultProps will never apply

Isnt `className` just handled in special way in render? https://github.com/emotion-js/emotion/blob/d19b947602d186683ba3225237e088487975578f/packages/create-emotion-styled/src/index.js#L100

I don't see how it would never apply with `defaultProps` approach.

mitchellhamilton:
I've been thinking about this quite a bit and I think we shouldn't add it and instead encourage people to create normal components and use the css prop. For example, the example given could instead be
```jsx
const MyComponent = props => (
  <input
    type="checkbox"
    css={{
      height: 40,
      width: 40,
      color: 'green'
    }}
    {...props}
  />
)
```
I think encouraging this is better because it only requires learning one Emotion specific API and the rest of it is plain React. Also, we don't have to think about supporting a bunch of different edge cases like functions.

> To anyone who tries the example with emotion 9 and class names don't get passed through, that's known and will be fixed in emotion 10

mindplay-dk:
Does anyone have an actually viable solution for TypeScript?

I'm attempting to port a large project from `styled-components` and I am struggling to find any substitute for this. 😐

My codebase has 100s of calls to `.attrs` like the following:

```ts
export const SelectLink = styled(Anchor).attrs({
  linkColor: true,
})`
  margin: 0 5px;
  font-size: 10px;
`
```

Here, `linkColor` is a property defined by `Anchor`.

That is largely the use case, just setting some defaults and passing them to a different component - we don't have any of the more advanced usages with callback functions or any of that stuff.

Even for that simple use case, I've been unable to find a viable migration path.

I understand why setting attributes doesn't directly have anything to do with styling, and as such, I agree, it probably doesn't belong in this package - but I need some way to migrate the existing code.

`recompose` is dead, so I tried a different package that offers `withDefaultProps` but it [didn't work](https://github.com/PacoteJS/pacote/issues/1555).

I'd be perfectly happy to refactor, if you can suggest something that works with TypeScript?

Even for this trivial use case:

```ts
function Hello({ world }: { world?: string }) {
  return <div>Hello, {world}</div>;
}

const HelloWithDefaults = withDefaultProps({ world: 123 }, Hello);
```

I can't find any approach that works.

What I don't want to do is wrap every every component in another component, because that means I have to duplicate the property types (or manually figure out how to reference the types) for every single component - these are mostly third-party components getting wrapped, so I need some way to derive a new component/type from an existing component/type.

I'm not in loves with `.attrs` by any means, I just don't know what to do do instead.

Did anyone solve this?

PR #621, opened 2018-04-10
Adds favicon to emotion site via gatsby-plugin-favicon

claudiopro:
**What**: I'm adding a nice favicon to the emotion site 😄 

**Why**: because it makes it easier for users to spot the tab in the browser and visually scan their bookmarks. Also, since the browser is always requesting `/favicon.ico`, this means the site serves the 404 page (currently, 9.9KB) for each page request resulting into unnecessary $$$ being wasted.

**How**: I added the [gatsby-plugin-favicon](https://github.com/Creatiwity/gatsby-plugin-favicon) (MIT licensed) and updated the Gatsby config for the emotion site, specifying to generate the favicons only. The plugin also allows to generate more favicons e.g. for touch devices but I left them disabled.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

I also tested that the favicon appears in the browser's tab and in the bookmarks. Note that Safari will require a specific treatment to generate a [pinned tab icon](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html).

Chrome:
screen shot 2018-04-10 at 4 05 01 pm
Firefox:
screen shot 2018-04-10 at 4 07 53 pm
Safari:
screen shot 2018-04-10 at 4 17 20 pm

### DIFF
--- a/packages/site/gatsby-config.js
+++ b/packages/site/gatsby-config.js
@@ -30,6 +30,24 @@ module.exports = {
           path: `${__dirname}/../../emotion.png`
         }
       },
+      {
+        resolve: `gatsby-plugin-favicon`,
+        options: {
+          logo: `${__dirname}/../../emotion.png`,
+          injectHTML: true,
+          icons: {
+            android: false,
+            appleIcon: false,
+            appleStartup: false,
+            coast: false,
+            favicons: true,
+            firefox: false,
+            twitter: false,
+            yandex: false,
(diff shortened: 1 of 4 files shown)


claudiopro:
Thanks for accepting the PR 😄Glad to help!

Issue #619, opened 2018-04-06
Add support for printing DOM Elements to `jest-emotion`


rszewczyk:
- `emotion` version: `9.1.0`
- `react` version: `16.3.0`

Relevant code.

```javascript
import React from 'react'
import { render } from 'react-testing-library'
import { createSerializer } from 'jest-emotion'
import * as emotion from 'emotion'
import { Something } from './stuff'

expect.addSnapshotSerializer(createSerializer(emotion))

it('snapshots dom nodes', () => {
  const { container } = render(<Something>hello!</Something>)
  expect(container.firstChild).toMatchSnapshot()
})

```

As expected, this doesn't produce a snapshot with styles and replaced selectors because `react-testing-library` returns DOM elements, not React elements. 

Suggested solution:

Add support for printing styles/replaced selectors for DOM Element snapshots. I'd like to submit a PR for this. I have a [proof of concept here](https://gist.github.com/rszewczyk/4375363cddff6636262d76d8a073b860). What I'm not sure about is some of the implementation details. Should, it be implemented as a separate serializer or an enhancement to the current serializer? If the latter, should be enabled via configuration, always enabled, etc?

Thank in advance for your help/feedback!

mitchellhamilton:
@rszewczyk Thanks for starting work on this!!

I think we should add it to the current serializer and have it enabled by default but add a `dom` option that will disable it when the option is false.

rszewczyk:
Thanks @mitchellhamilton! I'll work on this today.

rszewczyk:
@mitchellhamilton [PR opened](https://github.com/emotion-js/emotion/pull/622)

mitchellhamilton:
Closed via #622

PR #622, opened 2018-04-12
Add support to jest-emotion for printing DOM Elements (#619)

rszewczyk:
**What**:

Modifies the snapshot serializer plugin to:
* add styles and replace class names in DOM Elements
* add a configuration option to disable the above behavior

**Why**:

This will allow `jest-emotion` to work with libraries like `react-testing-library`

**How**:

* Modifies test to check for and include DOM Elements (I borrowed the implementation details from the `jest/pretty-format` [dom element plugin](https://github.com/facebook/jest/blob/master/packages/pretty-format/src/plugins/dom_element.js)
* Modify the function that gets all nodes in the tree to work with both React and DOM Elements
* Add code to get selectors from DOM Elements

The test I wrote for this uses `pretty-format` directly in order to isolate them from the serializer used to test the rest of the packages.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


packages/jest-emotion/README.md
```diff
@@ -57,6 +58,21 @@ expect.addSnapshotSerializer(
   })
 )
 ```
+
+### `disableDOMElements`
```
mitchellhamilton: Could you rename this option to `DOMElements` and make it so that it's true by default and making it false disables printing DOM elements' styles.

packages/jest-emotion/test/printer.test.js
```diff
@@ -0,0 +1,91 @@
...
+import prettyFormat from 'pretty-format'
+import * as emotion from 'emotion'
+import { createSerializer } from '../src'
+
+const { ReactElement, ReactTestComponent, DOMElement } = prettyFormat.plugins
+
+describe('jest-emotion with dom elements', () => {
+  const emotionPlugin = createSerializer(emotion)
+
+  const divStyle = emotion.css(`
```
mitchellhamilton: Could you change this to use `css` as a tagged template literal like this?
```jsx
css`
  color: red;
`
```

rszewczyk: fixed

packages/jest-emotion/README.md
rszewczyk: done

tkh44:
Thank you @rszewczyk!

kentcdodds:
This is awesome! Thanks!

PR #633, opened 2018-04-21
Add warning for invalid content values in objects

mitchellhamilton:
**What**:
Add warning for invalid content values

**Why**:

Seeing `px` as the value for `content` when you pass `''` is confusing.

**How**:

Add a dev warning that checks for invalid content values using some logic from [here](https://github.com/threepointone/glamor/blob/master/src/plugins.js#L49-L62)

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete


tkh44:
This is really great DX. Awesome idea!

PR #635, opened 2018-04-23
Add StyledComponent to allowed Interpolation types

### LINKED ISSUE #625: Typing of reference to other component seems broken
- `emotion` version: 9.1.1
- `react` version: 16.3
- `typescript` version: 2.8.1

Relevant code.

```javascript
const Child = styled.div`
  color: red;
`;

const Parent = styled.div`
  ${Child} {
    color: blue;
  }
`;
```

This results in the following error on `${Child}`:
```
Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'CSSProperties | { [key: string]: ObjectStyleAttributes; } | ((props: ThemedProps<{ innerRef?: str...'.
  Type 'TemplateStringsArray' is not assignable to type '(props: ThemedProps<{ innerRef?: string | ((instance: HTMLDivElement | null) => any) | undefined;...'.
    Type 'TemplateStringsArray' provides no match for the signature '(props: ThemedProps<{ innerRef?: string | ((instance: HTMLDivElement | null) => any) | undefined; } & { children?: ReactNode; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement>, any>): ObjectStyleAttributes'.
```

Workaround for now:
```typescript
const Parent = styled.div`
  ${Child as any} {
    color: blue;
  }
`;
```

mvestergaard:
This fixes an error when referencing another StyledComponent from within
a template interpolation string

Fixes #625


mvestergaard:
I think I was a little too quick in assuming this fix works. This seems to cause props to have implicit any type in other scenarios. 😟

mvestergaard:
I took a different approach.

I added a simple "placeholder" interface called `ComponentRef` that `StyledComponent` extends, and use that as a valid Interpolation value.

This seems to work, and does not break existing typings.

pluma:
Is there anything missing from this PR to get it merged?

mvestergaard:
A "problem" with this change, is that this will be allowed by the typing:
```typescript
const NotStyled: React.SFC = () => <div />;

const Parent = styled.div`
  ${NotStyled} {
    color: green;
  }
`;
```

However personally I would rather that the typing allowed intended behavior, without having to use an escape hatch, than catching unintended usage.

tkh44:
LGTM, but I'd rather have someone more confident in TS confirm to merge.

pluma:
Awesome! :tada:

PR #650, opened 2018-05-09
Remove comments from styles in babel-plugin-emotion

### LINKED ISSUE #647: Support for inline comments?
Apologies if this has been answer before. I searched through prior issues but didn't notice any of the responses directly answering my question.

Using: `react-emotion 9.1.2`

I frequently find myself using this pattern:

I will have some styling that I am experimenting with:

```
export const PaginationControl = styled.div`
  color: #fff;
  padding: 8px;
`;
```

And while I'm working, I want to comment out a style temporarily while I try a different property:

```
export const PaginationControl = styled.div`
  color: #ddd;
  padding: 16px;
  // color: #fff;
  // padding: 8px;
`;
```

I haven't been able to find a way to do this yet without breaking the styling. Is there a method to do something like this?

mitchellhamilton:
**What**:

Remove comments from styles in babel-plugin-emotion

**Why**:

Closes #647 + the comments won't be included in prod so bundles will be smaller

**How**:

Use some stuff from babel-plugin-styled-components and put it into `@emotion/babel-utils` and use that here.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete


StJohn3D:
Hi @mitchellhamilton, Is this supposed to work with `@emotion/styled/macro` ?

From my testing I am still seeing inline comments like the one below appearing in my CRA 3.1.1 production bundle.

```ts

export const StyledTest = styled.div`
  position: absolute;
  overflow: hidden;
  top: 0;
  left: 0;
  bottom: 0;
  right: 0 ;
  /* This comment should not be in production bundles */
`
```

Andarist:
Should work the same, could you prepare a repro case (runnable repository) that could be investigated?

StJohn3D:
Hi @Andarist thanks for the quick reply! I'll try to follow tomorrow.

StJohn3D:
Hi @Andarist, so after further investigation I realized that my issue is that since I'm writing a component library - my components are only compiled with typescript and so there's no babel in the process until lib files are consumed by a real CRA. But by then it's too late. Unless there's some similar macro for ts files I think I'll have to invent one.

Andarist:
Well - with CRA, unfortunately, you have limited configuration possibilities. You'd have to compile your component library with both TS and babel (which shouldn't be that hard to setup - it might be worth doing).

StJohn3D:
@Andarist yup, that's exactly what I did. The article I linked above and this https://github.com/Microsoft/TypeScript-Babel-Starter were both very helpful in getting me there.
Thanks again for all your help!

Issue #655, opened 2018-05-11
Write docs for `shouldForwardProp`


Andarist:
Would be great to have this one documented, it might be used like this:
```js
import isPropValid from '@emotion/is-prop-valid'

const StyledTextarea = styled(TextareaAutosize, { shouldForwardProp: isPropValid })({
  fontSize: '16px',
})
```

Another example can be found in this [test case](https://github.com/emotion-js/emotion/blob/d19b947602d186683ba3225237e088487975578f/packages/react-emotion/test/index.test.js#L1136-L1144).

tricoder42:
Hey @Andarist,
is it possible to set a global/default `shouldForwardProp`? 

I'm using `test` prop to add a testing ID which I use in end-to-end tests and which is removed from production code.

I was thinking about using `data-test` prop instead, but that wouldn't work for components passing `data-test` down to child container:

```jsx
const Button = ({ children, /*more props here */ data-test }) => {
  // some logic based on props

  return <StyledButton data-test={data-test}>{children}</StyledButton>
}
```

This doesn't work because `data-test` can't be a variable. `dataTest` isn't converted to `data-test` either and using spread operator just for one prop is a bit overkill :/

I would like to allow `test` prop in all styled components, but adding `{ shouldForwardProp: isTestProps }` is again bit overkill. I think I'll use it as a workaround `styled('div', withTestProps)`, but curious if I can override the default behavior.

Andarist:
U cant quite override the default behaviour, u could create a custom "proxy" `styled` factory that would just forward slightly modified arguments to the real one and use your proxy across your application.

I would encourage you to rethink if you need to pass data-test around as prop, seems like an anti-pattern to me (unless some specific use case needs it). Most commonly it should be just used as a simple marker, used plain in sight, not as a configurable prop.

tricoder42:
I would like to keep using `test` prop to be honest. `data-test` was just a workaround because it's already valid prop (but invalid variable name in js).

I'll try custom proxy when it becomes more painful. Right now I have just one place in codebase where I want to use `test` prop on styled component. In this case I use good old className instead of styled component.

Thank you for clarification 👍

Eugen-Ask:
Why don't make shouldForwardProp === isPropValid by default? I often use props in css and it's really non-convenient to write this construction every time.

Andarist:
It's a default for DOM tags - it's not for compose components though.

lifeiscontent:
[...]

PR #656, opened 2018-05-13
Tweak babel.md

smhutch:
**What**:
Content change in babel.md.

**Why**:
To avoid ambiguous language. When reading `babel.md`, I misunderstood:
> but not required in version 8 and above

as:
> but you don't need to install it to get the additional features in version 8 and above.

**How**:
Explicitly state that installation is optional in version 8 and above.

**Checklist**:

- [X] Documentation
- [ ] Tests **N/A**
- [ ] Code complete **N/A**

I'm looking to get into helping with open-source projects I use, this is the first opportunity I have found to _hopefully_ help. Really enjoying `emotion` so far 😄

### DIFF
--- a/docs/babel.md
+++ b/docs/babel.md
@@ -2,7 +2,10 @@
 title: "Babel Plugin"
 ---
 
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Awesome :) You might be interested in https://github.com/emotion-js/emotion/issues/655 too

PR #658, opened 2018-05-14
Emotion Primitives

nitin42:
**What**: Add `emotion-primitives` to emotion monorepo

**Why**:

**How**: Integration with React primitives

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

**TODO**

- [x] Render primitives on Web
- [x] Add all emotion features like passing props to interpolate functions, `withComponent` and etc for React Native and Sketch
- [x] Add tests
- [x] Improve error warnings
- [x] Build
- [x] Flow types

* Renders on Web ✅
* Renders on Sketch ✅
* Renders on Native ✅

Once this is done, we can extract parts for React Native and create a separate package `emotion-native`.


packages/emotion-primitives/src/emotionPrimitive.js
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  render(<View color='mistyrose' />)
+
+  const Title = Text.withComponent('Text')
+
+  render(<Title>Hello</Title>)
+*/
+export const emotionPrimitive = tag => {
+  if (Platform.OS === 'web') {
```
Andarist: imho this can be easily split at build time - we can create 2 separate bundles and point from package.json with main/module and react-native keys appropriately

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,57 @@
...
+    return viewStyleProps.includes(propName)
+  } else if (element === 'Image') {
+    return imageStyleProps.includes(propName)
+  }
+
+  return false
+}
+
+export function splitProps(rootEl, props) {
+  const rest = { ...props }
```
Andarist: this will make a copy of `props`, seems redundant as u only read from it anyway

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,57 @@
...
+    return imageProps.includes(propName)
+  }
+
+  return false
+}
+
+export function getPrimitiveProps(element, props) {
+  const acc = {}
+
+  Object.keys(props).forEach(prop => {
```
Andarist: iteration style seems inconsistent with `splitProps` where for similar job you use `.reduce`

packages/emotion-primitives/src/emotionPrimitive.js
nitin42: That would be great.

packages/emotion-primitives/src/splitProps.js
nitin42: Haha ya I think I forgot to refactor before pushing the code. We can remove this since 'props' is pass by value.

packages/emotion-primitives/src/splitProps.js
nitin42: This will change in next commit. I was testing some weird behaviour caused due to primitive props like onLongPress.

tkh44:
@nitin42 I believe you can use this as an example to create the styles.

https://github.com/emotion-js/next/blob/master/packages/serialize/src/index.js#L66

https://github.com/emotion-js/next/blob/032eab919602eca87165e92b0546bb76f95a0733/packages/css/src/index.js#L4

nitin42:
Thanks @tkh44 ! Got it working. Wrapping up the work now by preparing docs, examples and demos for all the platforms (native, web and sketch).

packages/emotion-primitives/src/emotion-native.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
```
Andarist: `import * as React from 'react'` works better with typings

packages/emotion-primitives/src/emotion-native.js
```diff
@@ -0,0 +1,58 @@
...
+import transform from 'css-to-react-native'
+import cssToObjects from 'css-to-object'
```
Andarist: probably should be `cssToObject`

nitin42: Not using this anymore! Haha It's better if you don't review this now 😅

packages/emotion-primitives/src/emotion-native.js
```diff
@@ -0,0 +1,58 @@
...
+    css = cssToObjects(styles.join(''))
+
+    Object.keys(css).forEach(prop => {
+      cssToRN.push([prop, css[prop]])
+    })
+
+    // Resolve the style using StyleSheet.create
+    rnStyles = resolveStyles([transform(cssToRN)])
+  } else {
+    rnStyles = [...styles]
```
Andarist: **i think** spread in general has some extra work to do (like dealing with array holes, using iteration protocol etc), I guess this is known to be an array so `styles.concat()` should be slightly more performant

nitin42: Yup! We can definitely making changes and add bits related to performance. Since that is emotion's trademark 😉

nitin42:
Thanks @Andarist for the initial review. Now I know the contributing parts a little better. Stay tuned for some surprise. This requires some more work but it should be stabled by next week.

packages/emotion-primitives/src/emotion-native.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // Store RN styles
+  let rnStyles
+
+  // If styles are not passed using the object styles
+  if (typeof styles === 'object' && Object.keys(styles) == 0) {
+    css = cssToObjects(styles.join(''))
+
+    Object.keys(css).forEach(prop => {
+      cssToRN.push([prop, css[prop]])
```
Andarist: `cssToRN` is only manipulated here and passed as argument to a function just few lines below, as it's size is known (`Object.keys(css).length`) it might be worth to preallocate it based on that

I mean smth like this:
```js
const keys = Object.keys(css)
const cssToRN = new Array(keys.length)
keys.forEach((prop, i) => {
  cssToRN[i] = [prop, css[prop]]
})
```

packages/emotion-primitives/src/index.js
```diff
@@ -0,0 +1,3 @@
+import { emotionPrimitive } from './emotionPrimitive'
```
Andarist: u can shorten it to just
```js
export { emotionPrimitive } from './emotionPrimitive'
```

packages/emotion-primitives/src/emotion-native.js
nitin42: Hm! Makes sense. But what about `Array` instantiation cost ?

packages/emotion-primitives/src/emotion-native.js
```diff
@@ -0,0 +1,58 @@
...
+    rnStyles = [...styles]
+  }
+
+  // TODO: Add all the emotion features here like theming and etc.
+  return class extends React.Component {
+    render() {
+      // Collect valid style props for the primitive
+      const { styleProps } = splitProps(tag, this.props)
+      // Collect valid props for the primitive
+      const primitiveProps = getPrimitiveProps(tag, this.props)
```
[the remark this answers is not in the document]
nitin42: This will be merged to `splitProps` instead. Like this -

```js
const { styledProps, primitiveProps }  = splitProps(tag, this.props)
```

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,57 @@
...
+  imageProps,
+  viewProps,
+  textStyleProps,
+  viewStyleProps,
+  imageStyleProps
+} from './props'
+
+function isValidStyleProp(element, propName) {
+  if (element === 'Text') {
+    return textStyleProps.includes(propName)
```
Andarist: `includes` requires polyfills to be used by library's users if they want to support older browsers, not sure if it's worth using in that case as `indexOf` equivalent is just a little bit more verbose

packages/emotion-primitives/src/index.js
Andarist: also it seems that u should export a default here (instead of named export) with
```js
export { emotionPrimitive as default } from './emotionPrimitive'
```

packages/emotion-primitives/src/splitProps.js
nitin42: Agree! I will replace this with indexOf > -1

nitin42:
Almost done! 😅 It's ready to be reviewed.

**What was changed?**

Added new implementation for styled component. Overall, the core was refactored.

**Remaining**

Exposing platform specific API and bundling accordingly (@Andarist @tkh44 I need some help over here)

Andarist:
> Exposing platform specific API and bundling accordingly (@Andarist @tkh44 I need some help over here)

Please make sure that you have checked "Allow edits from maintainers." checkbox here - I will try to find some time to add needed changes to this PR tomorrow. This doesn't have to block this PR though - it's just a small optimization that can be done.

nitin42:
Cool! I was thinking of removing react-emotion first from this package but realised that may be we can bundle accordingly! Thanks.

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+    styles += handleInterpolation.call(
+      this,
+      interpolation,
+      styles.charCodeAt(styles.length - 1) === 46
+    )
+
+    if (stringMode === true && strings[i + 1] !== undefined) {
+      styles += strings[i + 1]
+    }
+  }, this)
```
Andarist: wouldnt arrow function be good enough here? passing `this` like that is a little bit surprising for the reader at the first moment

nitin42: It was to me too when I first referenced this with @emotion/css. We can change this, no issues.

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+
+  return styles
+}
+
+function convertStyles(str) {
+  if (typeof str === 'string' && str.length === 0) return str
```
Andarist: we could return `''` here, i doubt it is any different than returning `str` here which is an empty string in that case and making it more explicit would help recognizing the intention a little bit sooner (not that it's particularly hard to see it in current form, but still 🤷‍♂️ )

nitin42: Haha, sure!

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+  const styleObj = []
+
+  const parsedString = str.split(';')
+
+  parsedString.forEach(style => {
+    if (typeof style === 'string') {
+      // Get prop name and prop value
+      const ar = style.split(': ')
+
+      if (ar[0] && ar[1]) {
```
Andarist: both need to be coerced to booleans, we could make this more explicit by checking `ar.length === 2`

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+
+function convertStyles(str) {
+  if (typeof str === 'string' && str.length === 0) return str
+
+  const styleObj = []
+
+  const parsedString = str.split(';')
+
+  parsedString.forEach(style => {
+    if (typeof style === 'string') {
```
Andarist: this if statement seems unnecessary, `style` has to be a string because it's outcome of `str.split(';')`

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+      // Get prop name and prop value
+      const ar = style.split(': ')
```
Andarist: shouldn't we split here by `':'`?

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+
+    let css = createStyles.apply(this, arr)
+
+    let parsedCSS = convertStyles(css)
+
+    // Convert css styles to react native
+    let rnStyles = Array.isArray(parsedCSS) ? transform(parsedCSS) : {}
+
+    return [StyleSheet.create({ style: rnStyles }).style]
+  } else if (styles[0] == null || styles[0].raw === undefined) {
```
Andarist: nit picking: this `else` is redundant as previous block is terminated with `return` anyway, this is equivalent to:
```js
if (a) {
  return val
}

if (b) {
  return val2
}
```

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+    return styles
```
Andarist: while I prefer this style emotion is aiming for high perf so it would probably be better to rewrite this to a for loop or forEach to avoid double iteration with `filter` + `map` combination

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,109 @@
...
+      .filter(style => {
+        if (typeof style === 'object') {
```
Andarist: `style[0] == null` in here and unfortunately `typeof null` is an `'object'` so this one might crash here

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
+import * as React from 'react'
+import reactPrimitives from 'react-primitives'
+import PropTypes from 'prop-types'
+
+import { getStyles } from './getStyles'
+import { convertToRNStyles } from './convertToRNStyles'
+
+const isValidPrimitive = primitive =>
+  ['Text', 'View', 'Image'].indexOf(primitive) > -1
```
Andarist: maybe hoist this constant array above the function to avoid unnecessary garbage collection on this, although probably modern engines should recognize it's constant and hoist it on their own, but it's hard to tell - so I would argue it's better to do that manually

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
...
+  return function emotion(primitive, { displayName } = {}) {
+    return createStyledComponent
+
+    /**
+     * Create emotion styled component
+     */
+    function createStyledComponent() {
+      let styles = []
+
+      styles.push.apply(styles, arguments)
```
Andarist: i think this should be an equivalent of `function createStyledComponent(...styles) { ... }`

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
...
+            primitive,
+            this.props
+          )
+
+          const emotionStyles = evalStyles(this, Styled, styles, styleOverrides)
+
+          return React.createElement(
+            getPrimitive(primitive),
+            {
+              ...toForward,
```
Andarist: `toForward` is a fresh object, it would be good to avoid making additional copy here by just mutating this object with `ref` and `style` props

Andarist: if we go that way then `children` should get attached to that object too instead of passed as 3rd argument

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
...
+            this.props.children || null
+          )
+        }
+      }
+
+      Styled.primitive = primitive
+
+      Styled.withComponent = (newPrimitive, options = {}) =>
+        emotion(getPrimitive(newPrimitive), {
+          ...options
```
Andarist: is making a copy here needed? (im not sure, it might be)

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * Creates a function that renders the styles on multiple targets with same code.
+ */
+export function createEmotionPrimitive(splitProps) {
+  /*
+   * Returns styled component
+   */
+  return function emotion(primitive, { displayName } = {}) {
```
Andarist: passing `displayName` like that is not part of react-emotion's API, do we have to do it here?

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const getStyledMetadata = ({ primitive, styles, displayName }) => ({
+  styles: primitive.styles ? primitive.styles.concat(styles) : styles,
+  primitive: primitive.primitive ? primitive.primitive : primitive,
+  displayName: displayName || `emotion(${getDisplayName(primitive)})`
+})
+
+const getDisplayName = primitive =>
+  typeof primitive === 'string' ? primitive : primitive.displayName || 'Styled'
```
Andarist: I think you should also consider reading `primitive.name` here (taking lower priority than `.displayName` ofc)

packages/emotion-primitives/src/index.js
```diff
@@ -0,0 +1,50 @@
...
+      getters[alias].tag = reactPrimitives[alias]
+      getters[alias].displayName = `emotion.${tag}`
+      return getters
+    }, {})
+  )
+
+  return styled
+}
+
+const emotion = createEmotionPrimitive(splitProps)
```
Andarist: will it be possible to `createEmotionPrimitive` with other `splitProps` implementation? are there use cases for this?

packages/emotion-primitives/src/props.js
```diff
@@ -0,0 +1,131 @@
...
+
+// <View /> Style props
+const viewStyleProps = Object.keys(viewStylePropTypes)
+// <Text /> Style props
+const textStyleProps = Object.keys(textStylePropTypes)
+// <Image /> Style props
+const imageStyleProps = Object.keys(imageStylePropTypes)
+
+// <Text /> primitive props
+const textProps = [
```
Andarist: we could create RegExps out of those array at build time - like [here](https://github.com/emotion-js/next/blob/e4f914824774a47618d23427397c56c367c7bbe2/packages/is-prop-valid/src/index.js#L5), plus testing functions could be memoized to return results from the cache

mitchellhamilton: We should probably use the objects themselves and do a !== undefined check since that’s what the memoized version would do

packages/emotion-primitives/src/convertToRNStyles.js
mitchellhamilton: I did this to save a few bytes so that Babel didn’t need to add `var _this = this`. It won’t make a big difference but it helps. I’m fine with changing it though

packages/emotion-primitives/src/props.js
Andarist: Do u remember why RegExp was introduced there in the first place? Purely as size optimization? If that's the case the same technique should be reused here or this build-time RegExp should get replaced in the other place too.

packages/emotion-primitives/src/convertToRNStyles.js
Andarist: if that's the case I'm fine with keeping it here too

packages/emotion-primitives/src/props.js
mitchellhamilton: It was mainly for size but also we need to check for some things that aren’t on a list like `on*`, `aria-*` and `data-*` and regex is the easiest way to do that so if we already have to do a regex check we might as well put everything there.

The size concern isn’t really here since we’re importing the object from another package so it will be included anyway and if we included regex it would be larger since there would be our regex + the prop type objects

Andarist: That's true about `viewStyleProps`, `textStyleProps`, `imageStyleProps`, but not for `textProps`, `viewProps` and `imageProps`, right? those could be optimized with a RegExp approach

Andarist:
@nitin42 please take a look at my last commit, there is no need to differentiate builds per environment as you have removed Platform checks in the meantime. I've just tweaked rollup build a little bit & added more stuff to emotion-primitives' package.json. Please verify my chages

rollup.config.js
```diff
@@ -44,16 +52,12 @@ const baseConfig = {
...
-      flatMap(dir => {
-        const {
-          dependencies = {},
-          peerDependencies = {}
-        } = require(`${dir}/package.json`)
-
-        return [...Object.keys(dependencies), ...Object.keys(peerDependencies)]
-      }, getLernaPackages())
+  external: makeExternalPredicate(
+    uniq(
```
Andarist: @mitchellhamilton I think I've done something wrong here in the past, can't remember why i did it that way, but maybe u could look at it & verify my current reasoning

we do not have to read all lerna `package.json` files because this script bundles a single package at a time, so it is enough to read only its own `package.json` and build `external` array based on that

packages/emotion-primitives/src/createEmotion.js
nitin42: I don't think so. This here makes sure that we receive interpolation array when using string template literal or a valid object when used object style notation. But if you've an alternate approach to this, you can alter this 😄

packages/emotion-primitives/src/createEmotion.js
nitin42: Is there any cost ?

packages/emotion-primitives/src/index.js
nitin42: Use case is -

To split component props, style props and style overrides so that we can forward them without logging any warnings.

nitin42:
Made some changes. Going to review build setup later this week after I am done with some testing on all the platforms to make sure the basic examples work. From there, we can stabilise it by removing unused bits and making it performant. I hope it's ok.

packages/emotion-primitives/src/createEmotion.js
Andarist: to the first one - yeah, to the second one (children) there is not, u can disregard my comment, this one doesn't have to be attached to `toForward`. I've checked react's source code and for a single child react will do it by itself so it doesnt really matter if it's done here or on react's side.

packages/emotion-primitives/src/convertToRNStyles.js
Andarist: how those elements could be undefined? checking against `.length` here is preferable as we should avoid out of bounds accesses

packages/emotion-primitives/src/convertToRNStyles.js
Andarist: isnt `width:100px;` supported?

packages/emotion-primitives/src/createEmotion.js
nitin42: Ok. I'll tweak this in next commits!

packages/emotion-primitives/src/createEmotion.js
Andarist: not sure what do u mean, u basically spread `arguments` into `styles` array with that `push.apply`, it seems to me that `function createStyledComponent(...styles) { ... }` is 100% the same, only might be more performant (i have no data to back this up though)

could u explain how they might not be the same?

rollup.config.js
mitchellhamilton: I don’t totally remember why that was done but I think it might be necessary because I did a similar thing [here](https://github.com/emotion-js/next/commit/bede91841b3be2f343adc8f2da407ab987a0c297)

packages/emotion-primitives/src/props.js
mitchellhamilton: Yep, you’re right, those can be done with regex.

@nitin42 I was looking at styled-components' native and primitives implementation and I noticed they don’t do prop filtering and I ended up trying a random prop on react native and noticed there are no unknown prop warnings. Is prop filtering necessary? I’m guessing it is because they have to be filtered on the web? I just wanted to ask to make sure

nitin42: Hm... I'll try to reference this with my work on glamorous primitives. Doing this checks seems unnecessary to me, now!

rollup.config.js
Andarist: 👍 makes sense for peerDeps, but the removed code gathered also regular deps - maybe it was just a small mistake, gonna bring this back in a minute

packages/emotion-primitives/test/emotion-primitives.test.js
```diff
@@ -21,6 +23,16 @@ describe('Emotion primitives', () => {
     )
   })
 
+  test('compose via .styles prop', () => {
+    const Text = emotionPrimitive.text`color: hotpink;`
+
+    const Name = emotionPrimitive.text`${Text.styles}`
```
nitin42: For composition, I think we have to rely on `.styles` static property of the component. I don't any other way to tackle this. But still, open to discuss about it.

mitchellhamilton:
Could you change `emotionPrimitive` to be called `styled`(so that syntax highlighting and etc. work) and change `.view` to `.View` and etc.?

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,113 @@
...
+        static propTypes = {
+          innerRef: PropTypes.func
+        }
+
+        onRef = innerComponent => {
+          this.innerComponent = innerComponent
+
+          if (this.props.innerRef) {
+            this.props.innerRef(innerComponent)
+          }
```
mitchellhamilton: Could we support the new createRef api like [this](https://github.com/styled-components/styled-components/pull/1718/files#diff-8835cf55a33e0e0d79683647a6707a5f)? Also, is there a reason to get the ref at all since it doesn't seem like it's used?

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,113 @@
...
+     * Create emotion styled component
+     */
+    function createStyledComponent() {
+      let styles = []
+
+      styles.push.apply(styles, arguments)
+
+      class Styled extends React.Component {
+        static propTypes = {
+          innerRef: PropTypes.func
```
mitchellhamilton: This prop type won't work with `createRef`.

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,113 @@
...
+      `Cannot style invalid primitive ${primitive}. Expected primitive to be one of ['Text', 'View', 'Image']`
+    )
+  } else if (typeof primitive === 'function') {
+    return primitive
+  }
+}
+
+function evalStyles(context, Comp, styles, styleOverrides) {
+  // Assign static property so that the styles can be reused (like in withComponent)
+  Comp.styles = convertToRNStyles.call(context, styles)
```
mitchellhamilton: This seems to be called on each render so the styles assigned as statics on the component will change depending on the props. Maybe the styles array created at the start of `createStyledComponent` should be assigned as a static on the component instead?

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,113 @@
...
+
+import { getStyles } from './getStyles'
+import { convertToRNStyles } from './convertToRNStyles'
+
+const primitives = ['Text', 'View', 'Image']
+
+const isValidPrimitive = primitive => primitives.indexOf(primitive) > -1
+
+const getPrimitive = primitive => {
+  if (typeof primitive === 'string' && isValidPrimitive(primitive)) {
```
mitchellhamilton: Maybe instead of doing this, we could change babel-plugin-emotion to not change the dot syntax if the first character is uppercase?

packages/emotion-primitives/src/createEmotion.js
nitin42: Yea.. I'll fix this!

packages/emotion-primitives/src/createEmotion.js
nitin42: We can do that but, using this has an advantage. We can use reuse the converted styles like -

```js
const Text = emotion.text`color: red; padding: 10px;`

// After calling convertToRNStyles, styles get converted to RN
// like this -> [73]

// If we decide to assign the styles array then we won't be able to
// reuse the styles on another component because styles array
// would be like this [['color: red; padding: 10px;', raw: [...]]].

const Author = emotion.text`${Text.styles} position: absolute; right: 0;` 
```

nitin42: This is good when composing the styles like using `css` on Web.

packages/emotion-primitives/src/createEmotion.js
nitin42: May be we can but I am not sure that `babel-plugin-emotion` will work in other environment. Tweaking this here may work on `web` but not on other platforms.

nitin42: But I really like this idea though 🤔

mitchellhamilton:
Could you use context in the actual component like styled for the web instead of using withTheme so that another component doesn't get added to the stack? 
https://github.com/emotion-js/emotion/blob/master/packages/create-emotion-styled/src/index.js#L80-L91

Also, could we have a `css` function that accepts styles via objects or strings and returns a style object?

Also, does interpolating react-native style objects like this work and if not, is it possible?

```jsx
styled.View`
  color: hotpink;
  ${{
    padding: 8,
    transform: { rotateX: "45deg" }
  }};
`;

```

I also just removed some of the ref stuff since the ref wasn't being used so the ref could be passed down directly. I also removed the PropTypes since we can let React warn if an incorrect value for a ref is passed

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,98 @@
+import { StyleSheet } from 'react-primitives'
+import transform from 'css-to-react-native'
+import { convertRuleOptions } from 'tslint/lib/configuration'
```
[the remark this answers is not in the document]
nitin42: yeah.. `vscode` did it's thing. I'll make the changes 😅

packages/emotion-primitives/src/convertToRNStyles.js
```diff
@@ -0,0 +1,98 @@
...
+function convertStyles(str) {
+  if (typeof str === 'string' && str.length === 0) return ''
+
+  const styleObj = []
+
+  const parsedString = str.split(';')
+
+  parsedString.forEach(style => {
+    // Get prop name and prop value
+    const ar = style.split(': ')
```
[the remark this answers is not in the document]
nitin42: Got it!

packages/emotion-primitives/src/createEmotion.js
nitin42: I'll test and see if it works (doesn't break other things) without using `styles` array.

nitin42:
Yes, we can use context for theming. About interpolating React Native styles, I have no idea how we can do that. I'll try searching something and see if it works. Also, I think we can use `css` function to convert the styles. I'll normalise this stuff in the later commits.

nitin42:
I need some help over configuring Rollup. When I run `yarn rollup` or `yarn build`, I get an error that `You must specify options.file`. Haven't used Rollup so I'm completely noob.

mitchellhamilton:
@nitin42 6a40c31839aac3961aad59b21fe8d748ed756381 should fix it.

Andarist:
Just a note of explanation - `rollup.config.js` in your case is importing `package.json` and using `module` field as input of one of the configs instead of hardcoding the path in the config itself as they have to match anyway.

nitin42:
Thank you for being responsive. I'm on it! @Andarist you were right, that `styles` array was totally unnecessary since we were assigning styles as static property to the component so we can directly use `...arg` in `createStyledComponent`. Thank you 😅

nitin42:
Got it working!

nitin42:
**Tasks completed**

- [x] Theming

- [x] Refactor `createStyledComponent` (omit styles array)

- [x] Fix failing tests

- [x] Composing styles via `.styles` prop

- [x] Fix proxy usage in current env. (Sketch / Web / Native)

**Tasks Remaining**

Enhance prop filtering

Remove unuse / unnecessary bits

mitchellhamilton:
I'm really not big on the `.styles` thing with the way it is right now because it's so unpredictable. For example,

```jsx
const SomeComponent = styled.View`
`;
render(<SomeComponent color="green" />);

const IWouldExpectThisToAcceptAColorPropButInsteadItIsJustGreen = styled.View`
  ${SomeComponent.styles};
`;
```

nitin42:
Ya that's one tradeoff. I am trying out something here, if it doesn't work then I'll remove this!

nitin42:
I've removed the test for `.styles` prop. But I haven't unassigned this static property on `Styled` component because we still use it in `withComponent` to reuse the styles. Let me know what you think!

nitin42:
Now that we're almost done with this implementation, I wanted to discuss something about documentation. Should I just add a basic example, installation, demo (platform specific) ? Because most of the API details can be shared via emotion doc. website !

DavideDaniel:
I would cover any install specific notes and a basic example to start. Also, awesome work @nitin42 !

nitin42:
Done with basic documentation, tested the build on all the three platforms. Works seamlessly!  We can optimise and cover perf. issues later in subsequent release cycles.

To test the beta version,

```
npm install emotion-primitives@1.0.0-beta.4
```

tkh44:
This is looking great @nitin42. Thanks for all the hard work.

nitin42:
Thank you for guiding me through the codebase and providing suggestions 😉. Such a nice experience I'd working with `emotion` codebase. Also props to @mitchellhamilton @Andarist 😅

tkh44:
I've lost context on where this PR progress stands. What do we need to do to get this merged? If you need any help let us know what you need so we can get it worked out 👍

mitchellhamilton:
Could you add a test for this case?
```jsx
const Thing = styled.View`
  background-color: ${props => props.color}
`

render(<Thing color="green" />)

const OtherThing = Thing.withComponent(Text)
// should be hotpink
render(<Thing color="hotpink" />)
```

Also if you could add a `css` function, that would be great!

nitin42:
Sure! One thing about `css` function. Do you want me to make the implementation for parsing styles concrete i.e just one `css` function for doing everything with styles similar to `emotion` ?

Andarist:
@nitin42 I've added 3 simple commits, please take a look

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,94 @@
+import { isValidStyleProp } from './props'
+
+const forwardableProps = {
```
Andarist: 1. by converting it to object we can do a simple lookup (`if (forwardableProps[propName])`) instead of parsing this string with regexp each time this was looping
2. also I have found some duplicate properties in this string, so I've just smply removed them

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Split primitive props, style props and styled component props into an object.
+ */
+export function splitProps(primitive, { innerRef, ...rest }) {
+  // Style overrides are the style props that are directly passed to the styled primitive <Text display='flex' />
+  const styleOverrides = {}
+  // Forward props are the one which are valid primitive props, hence these props are forwarded to the component.
+  const split = { toForward: {}, styleOverrides }
+
+  for (const propName of Object.keys(rest)) {
```
Andarist: this gets compiled to a simple for loop (babel@7 loose mode assumes array here)

packages/emotion-primitives/src/splitProps.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Split primitive props, style props and styled component props into an object.
+ */
+export function splitProps(primitive, { innerRef, ...rest }) {
+  // Style overrides are the style props that are directly passed to the styled primitive <Text display='flex' />
+  const styleOverrides = {}
+  // Forward props are the one which are valid primitive props, hence these props are forwarded to the component.
+  const split = { toForward: {}, styleOverrides }
```
Andarist: could u explain the whole `toForward`/`styleOverrides` distinction? why we have 2 object here?

packages/emotion-primitives/src/splitProps.js
Andarist: also, according to developit (Jason Miller) `.reduce` is slow, so this should be a decent perf boost overall

packages/emotion-primitives/src/props.js
```diff
@@ -0,0 +1,18 @@
+import viewStylePropTypes from 'react-primitives/lib/web/View/ViewStylePropTypes'
+import textStylePropTypes from 'react-primitives/lib/web/Text/TextStylePropTypes'
+import imageStylePropTypes from 'react-primitives/lib/web/Image/ImageStylePropTypes'
+
+export const isValidStyleProp = (element, propName) => {
```
Andarist: similar history here, there is no need to cast those to array of strings with `Object.keys` and check later existance with `indexOf`

we can just do a simple lookup here

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,111 @@
...
+import { convertToRNStyles } from './convertToRNStyles'
+
+const primitives = ['Text', 'View', 'Image']
+
+const isValidPrimitive = primitive => primitives.indexOf(primitive) > -1
+
+// Validate and return the primitive
+const getPrimitive = primitive => {
+  if (typeof primitive === 'string') {
+    if (isValidPrimitive(primitive)) {
```
Andarist: both branches were checking `typeof primitive === 'string'`, so I've just merged those 2 (for readability, not perf reasons)

packages/emotion-primitives/src/createEmotion.js
```diff
@@ -0,0 +1,111 @@
...
+      return reactPrimitives[primitive]
+    }
+    throw new Error(
+      `Cannot style invalid primitive ${primitive}. Expected primitive to be one of ['Text', 'View', 'Image']`
+    )
+  }
+
+  return primitive
```
Andarist: I've also removed `typeof primitive === 'function'` check because it would result in undefined being passed to `createElement`, we still might pass invalid value, but at least react might be able to report the invalid one (that it got) instead of reporting that it got `"undefined"`

mitchellhamilton:
@nitin42 yep, just like `emotion`‘s `css`

packages/emotion-primitives/src/createEmotion.js
nitin42: That's better!

packages/emotion-primitives/src/splitProps.js
nitin42: `styleOverrides` are the style props directly pass to the component like this -

```
<StyledText display='flex' justifyContent='center' />
```

and `toForward` props are the one that are valid primitive (`Text`, `View` and `Image`) props.

I think emotion does support `styleOverrides` ? or if not then we can remove this bit and hence reduce the `lib` size.

packages/emotion-primitives/src/splitProps.js
nitin42: Cool!

packages/emotion-primitives/src/splitProps.js
mitchellhamilton: We don't support `styleOverrides`.

nitin42:
I was poking around `emotion-js/next` and [saw the implementation for css](https://github.com/emotion-js/next/blob/master/packages/css/src/index.js#L8) similar to what we have in `emotion-primitives` except the naming (we use `createStyles` here). 

Although we can normalise this but we will end up with same things. Along with `createStyles`, we just use some utility functions to parse the style and convert them using `StyleSheet.create`. What do you think ?

packages/emotion-primitives/src/splitProps.js
Andarist: just to add - things like this can be achieved with i.e. `styled-system`

nitin42: Oh! That's great too (for bundle size 😅)

nitin42: Didn't know about `styled-system`. Looks interesting!

nitin42:
Checkout this sandbox for the demo.

[Edit 74wo1kxoo1](https://codesandbox.io/s/74wo1kxoo1)

One thing! I am getting this error

```
Cannot style invalid primitive __esModule.
```

when using ES module instead of commonJS. Any ideas ?

mitchellhamilton:
I think it’s because of the proxy and since there’s only a single export from emotion-primitives, it’s assigned directly to module.exports and when it’s imported, the __esModule property is checked and the proxy get handler throws the error

nitin42:
Oops! Sorry accidentally run `git push` on a different tab in my terminal. I'll revert this.

nitin42:
`it’s assigned directly to module.exports and when it’s imported, the __esModule property is checked and the proxy get handler throws the error`

I wasn't aware of this. Thanks! I'll try a different approach for managing unknown primitives.

nitin42:
What do you suggest instead ?

nitin42:
I've removed the `Proxy` usage for validating the primitives which was assigned on the `Styled` constructor. Also, updated the test case for it.

Check out the new beta release -

```
npm install emotion-primitives@1.0.0-beta.6
```

Updated sandbox link:

[Edit 8zmly1mo62](https://codesandbox.io/s/8zmly1mo62)

nitin42:
Great! Before this release graduates from beta, I'll be removing the dependency "css-to-react-native" and will be writing a custom css function (which may be added as a package called emotion-rn), also we can augment it later!

nitin42:
[#364](https://github.com/airbnb/react-sketchapp/pull/364)

packages/emotion-primitives/README.md
```diff
@@ -0,0 +1,90 @@
...
+yarn add emotion-primitives
+```
+
+This package also depends on `react` and `react-primitives` so make sure you've them installed.
+
+## Example
+
+```js
+import React from 'react'
+import emotion from 'emotion-primitives'
```
mitchellhamilton: Could you change this to be imported as styled so that syntax highlighting and etc. just work?

packages/emotion-primitives/README.md
```diff
@@ -0,0 +1,90 @@
...
+const Description = emotion.Text({
+  color: 'hotpink'
+})
+
+const Image = emotion.Image`
+  padding: 40px;
+`
+
+const emotionLogo =
+  'https://emotion.sh/static/emotion-a76dfa0d18a0536af9e917cdb8f873b9-a69fb.png'
```
mitchellhamilton: Could you use `https://cdn.rawgit.com/emotion-js/emotion/master/emotion.png` instead because the url there could change?

packages/emotion-primitives/README.md
nitin42: Sure!

nitin42:
Tested the new build on all the three platforms. I think we can now graduate this release from beta version. Let me know how official versions are managed in the `emotion-js` org.

mitchellhamilton:
@nitin42 I made some changes.

I added a `css` function and made styled use it internally. It'll accept basically any styles you throw at it and because styled uses it internally now, styled will also accept basically anything.

```jsx
import { css } from 'emotion-primitives'

const someStyle = css`
  color: hotpink
`

const anotherStyle = css({
  color: 'hotpink'
})

const compositionWorks = css`
  background-color: green;
  ${someStyle};
`
```
I also removed styleOverrides like we discussed https://github.com/emotion-js/emotion/pull/658#discussion_r195661456 and added flow everywhere.

Also, I was planning on putting this on the emotion npm scope(which is what we're probably going to do for packages in the future) so it would be `@emotion/primitives` instead of `emotion-primitives`, does that sound good to you?

The way releases work in the monorepo is that everything is on the same version and when we do a release, every package that has changed since the last release gets published under the same version number.

packages/emotion-primitives/src/index.js
```diff
@@ -1,22 +1,17 @@
...
 const assignPrimitives = styled => {
-  Object.assign(
-    styled,
-    primitives.reduce((getters, alias) => {
-      getters[alias] = styled(reactPrimitives[alias])
-      return getters
-    }, {})
-  )
+  // $FlowFixMe
+  createStyled.Text = createStyled(Text)
```
nitin42: This is much cleaner and intuitive!

packages/emotion-primitives/src/styled.js
```diff
@@ -0,0 +1,102 @@
...
+          theme: (state !== null && state.theme) || props.theme || {}
+        })
+
+        const emotionStyles = evalStyles(this, Styled, styles)
+
+        return React.createElement(
+          component,
+          pickAssign(pick, {}, props, {
+            ref: props.innerRef,
+            style: StyleSheet.flatten(emotionStyles)
```
nitin42: 👍

mitchellhamilton:
I just pushed a fix for the build.

nitin42:
Got it working, thanks!

mitchellhamilton:
Also, you don't need to do releases once we merge it in, we release all the packages together with lerna.

packages/babel-plugin-emotion/src/index.js
```diff
@@ -335,14 +345,18 @@ export function buildStyledObjectCallExpression(
...
 
-  return t.callExpression(
-    t.callExpression(identifier, [
-      tag,
-      buildFinalOptions(t, styledOptions, targetProperty, labelProperty),
-      ...restStyledArgs
-    ]),
-    args
-  )
+  let createStyledCall =
```
Andarist: maybe just `styledCall`? `createStyledCall` implies this is a function

nitin42:
Oh! That's great.

packages/emotion-primitives/src/css.js
```diff
@@ -0,0 +1,101 @@
...
+    vals = args
+  } else {
+    vals = interleave(args)
+  }
+
+  let styles = []
+  let buffer = ''
+  let lastType
+
+  let handleInterpolation = (interpolation: *, i: number, arr: Array<*>) => {
```
Andarist: `css` being possibly a hot path it would be good to avoid closure here, right?

packages/emotion-primitives/src/css.js
```diff
@@ -0,0 +1,101 @@
...
+            'let dynamicStyle = (props) => css`color: ${props.color}`\n' +
+            'It can be called directly with props or interpolated in a styled call like this\n' +
+            'let SomeComponent = styled.View`${dynamicStyle}`'
+        )
+      } else {
+        handleInterpolation(interpolation(this.mergedProps), i, arr)
+      }
+      return
+    }
+    let isFalsy = interpolation == null || type === 'boolean'
```
Andarist: this doesn't catch `false`
```js
false == null // false
```

nitin42:
Again tested the new build and everything seems great. Plus, `css` function is 🔥

[Edit pwkxvvx7k0](https://codesandbox.io/s/pwkxvvx7k0)

packages/emotion-primitives/src/css.js
mitchellhamilton: It has `type === 'boolean'` so it does catch false but I should probably rename it since it doesn't actually mean falsy it's meant to mean something that we don't care about.

packages/emotion-primitives/src/styled.js
```diff
@@ -0,0 +1,104 @@
...
+  interleave
+} from './utils'
+import {
+  testPickPropsOnPrimitiveComponent,
+  testPickPropsOnOtherComponent
+} from './test-props'
+import { css } from './css'
+
+function isPrimitiveComponent(component: React.ElementType) {
+  switch (component) {
```
Andarist: is switch case going to be faster than indexOf or object property lookup?

mitchellhamilton: Yes. switch cases are realllllllllllllly fast.

nitin42: Object property lookup!

packages/emotion-primitives/src/styled.js
```diff
@@ -0,0 +1,104 @@
...
+        })
+        let stylesWithStyleProp = styles
+        if (props.style) {
+          stylesWithStyleProp = styles.concat(props.style)
+        }
+        const emotionStyles = css.apply(this, stylesWithStyleProp)
+
+        return React.createElement(
+          component,
+          pickAssign(pick, {}, props, {
```
Andarist: `pick` -> `pickTest`?

packages/emotion-primitives/src/utils.js
```diff
@@ -1,17 +1,23 @@
...
 
-export function setTheme(theme) {
+export function setTheme(theme: Object) {
   this.setState({ theme })
 }
 
 export const testAlwaysTrue = () => true
 
-export const pickAssign = function(testFn, target) {
+export const pickAssign: (
```
Andarist: would be good to package those common utils at some point

packages/emotion-primitives/src/css.js
Andarist: oh, the name threw me off 😉

packages/emotion-primitives/src/css.js
Andarist: after a thought im not sure if this creates a closure 🤔 kinda blur line for me sometimes between closure and scope lookup, anyway - as it being hot it would be good to have single copy of this function anyway

packages/emotion-primitives/src/styled.js
mitchellhamilton: We can't do a object property lookup since our keys are functions but switch cases are wayyyyyyyyyyyyyy faster anyway. https://esbench.com/bench/5b3f224bf2949800a0f61e14

packages/emotion-primitives/src/css.js
```diff
@@ -3,71 +3,82 @@ import transform from 'css-to-react-native'
...
+    interpolation.forEach(handleInterpolation, this)
+  }
+  lastType = type
+}
 
-    if (type === 'string') {
-      buffer += interpolation
+export function css(...args: any) {
+  let vals
+  styles = []
```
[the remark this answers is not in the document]
mitchellhamilton: I'm not sure if it's really worth it/will make a difference. We create a bunch of arrays since `css-to-react-native` accepts styles like this `[['color', 'hotpink'], ['background-color', 'yellow']]` so I don't thinking creating one less here will matter. Also, I don't think mutating it is a good idea since `StyleSheet.flatten` could have memoization/could have it in the future and using the same array would break that.(that might never happen but I think passing an object to a function that we don't control and mutating it isn't a great idea)

nitin42:
Did [`f8e6958`](https://github.com/emotion-js/emotion/pull/658/commits/f8e69581e6d6b94138dec0e11407e1088eb6a571) break anything locally ?

mitchellhamilton:
Yeah, one sec, I'll fix it.

nitin42:
Sorry I accidentally used `git` alias on my system instead of pulling changes from the remote branch `emotion-primitives` 😅

mitchellhamilton:
All good!

I did that on master a few days ago 😂

mitchellhamilton:
I'm gonna rename this to `@emotion/primitives`, merge and release this tomorrow unless anyone has any objections.

mitchellhamilton:
Just published `@emotion/primitives` 🎉

PR #659, opened 2018-05-15
Forward .defaultProps when reusing __emotion_base

### LINKED ISSUE #654: Component loses its default props when wrapped with styled()
- `emotion` version: 9.1.3
- `react` version: 16.3.2

```js
const Title = styled('h1')`
  text-align: center;
  ${props => ({
    color: props.color
  })};
`;

Title.defaultProps = {
  color: 'red'
};

const Title2 = styled(Title)`
  font-style: italic;
`;
```

What you did:

Created a new styled component by wrapping an existing styled component in `styled()`.

What happened:

The wrapped styled component lost its defaultProps. In the example above `<Title>` is displayed as red, but `<Title2>` is not.

<b>CodeSandbox: https://codesandbox.io/s/317r028k6q</b>

Andarist:
fixes #654 

I feel though that we should figure out more holistic approach as `.defaultProps` is only part of the problem - any other config (i.e. `shouldForwardProp`) will get lost too when reusing `__emotion_base`. This should act like `.withComponent` in this regard.

Thoughts?

### DIFF
--- a/packages/create-emotion-styled/src/index.js
+++ b/packages/create-emotion-styled/src/index.js
@@ -140,6 +140,10 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
                 : baseTag.displayName || baseTag.name || 'Component'
             })`
 
+      if (tag.defaultProps !== undefined) {
+        // $FlowFixMe
+        Styled.defaultProps = tag.defaultProps
+      }
       Styled.contextTypes = contextTypes
       Styled.__emotion_styles = styles
       Styled.__emotion_base = baseTag
--- a/packages/react-emotion/test/__snapshots__/index.test.js.snap
+++ b/packages/react-emotion/test/__snapshots__/index.test.js.snap
@@ -678,6 +678,30 @@ exports[`styled random object expression 1`] = `
 </h1>
 `;
 
+exports[`styled should forward .defaultProps when reusing __emotion_base 1`] = `
+.emotion-0 {
+  text-align: center;
+  color: red;
+}
+
+.emotion-2 {
+  text-align: center;
+  color: red;
+  font-style: italic;
+}
+
+<div>
+  <h1
+    className="emotion-0 emotion-1"
+    color="red"
+  />
+  <h1
+    className="emotion-2 emotion-3"
+    color="red"
+  />
+</div>
+`;
+
 exports[`styled theme prop exists without ThemeProvider 1`] = `
 .emotion-0 {
   color: green;
--- a/packages/react-emotion/test/index.test.js
+++ b/packages/react-emotion/test/index.test.js
@@ -1148,4 +1148,31 @@ describe('styled', () => {
       .toJSON()
     expect(tree).toMatchSnapshot()
   })
+
+  test('should forward .defaultProps when reusing __emotion_base', () => {
+    const Title = styled('h1')`
+      text-align: center;
+      ${props => ({
+        color: props.color
+      })};
+    `
+
+    Title.defaultProps = {
+      color: 'red'
+    }
+
+    const Title2 = styled(Title)`
+      font-style: italic;
+    `
+
+    const tree = renderer
+      .create(
+        <div>
+          <Title />
+          <Title2 />
+        </div>
+      )
+      .toJSON()
+    expect(tree).toMatchSnapshot()
+  })
 })


Andarist:
I've looked into the code and it seems to me that for `isReal` components we could return this:
```js
return tag.__emotion_real.withComponent(tag.__emotion_base, options)
```
to fix `shouldForwardProp` situation, that wouldn't help with `.defaultProps` though. Also not sure if that would be totally safe, not sure how things like staticClassName, identifierName, stableClassName are shared for both of those cases.

Andarist:
> I think for shouldForwardProp we could set it as a static property on Styled

Problem with that approach is that we have to think about it & add those static properties whenever we introduce any new option. Could you explain in short (or link me to some resources) on how exactly staticClassName, identifierName, stableClassName are used? Do they suffer from the same problem? Or they simply shouldn't get forwarded?

> For defaultProps, I think the only way to solve all the edge cases for it would be to opt out of the component flattening if defaultProps is defined since this solution wouldn't account for setting defaultProps on a styled component, passing it to styled again and setting defaultProps on that component as well.

Developer can (although it's a little bit quirky) extend `defaultProps` on his/her own.

mitchellhamilton:
[We don't do component flattening with `extractStatic`](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion-styled/src/index.js#L36) so `staticClassName` doesn't have to be forwarded.

`identifierName` is used for the displayName of components and adding appending that name to the class name via adding the [label property](https://emotion.sh/docs/labels). The label property is already in the styles that are stored on the static property and there wouldn't be a way to use displayName since the component would be flattened so there's no point in forwarding it.

`stableClassName` is the class name used for component selector targeting. This is the only one other than shouldForwardProp that we might want to forward because currently in the example below `AnotherComponent` isn't targeted by targeting `SomeComponent` but we might want it to.

```jsx
  color: hotpink;
`

const AnotherComponent = styled(SomeComponent)`
  color: green;
`

render(
  <div
    css={css`
      ${SomeComponent} {
        color: darkcyan;
      }
    `}
  >
    <AnotherComponent />
  </div>
)
```

Andarist:
Thanks for the explanation. The only thing I'm worrying about is that in the future we might want to "passthrough" more options and static properties approach doesnt scale that well.

Also as to `shouldForwardProp` - it seems to me that actually it should not just override (although it's easier code-wise), but should reuse existing `shouldForwardProp`. So it would work like `.every`. Thoughts? 

As to `defaultProps`, do u want me to adjust the PR to opt out of component flattening?

mitchellhamilton:
I think that using something that doesn't scale well for options is good since it encourages us to not add options because adding options means users have to learn more specifics about the API and IMO that's not something we want.

For reusing `shouldForwardProp`, I can't think of any specific use case where you would want it to be reused. Also, if shouldForwardProp is reused then we have to make assumptions about which function should take precedence. As an example, what should happen here? should it pass down someProp or not? since if we did `firstShouldForwardProp(prop) || secondShouldForwardProp(prop)` `someProp` would get passed down but I can't see how you would want `someProp` to be passed down here.
```jsx
const SomeComponent = styled('div', {
  shouldForwardProp: prop => {
    switch (prop) {
      case 'someProp':
        return false
      default:
        return true
    }
  }
})`
  color: hotpink;
`

const AnotherComponent = styled(SomeComponent, {
  shouldForwardProp: prop => {
    return true
  }
})`
  background-color: green;
`
```

I'm just gonna merge this as is because the edge case I presented is going to be ridiculously rare so if someone opens an issue about it then let's fix it but right now, let's leave it like this.

Andarist:
I've proposed it working like `.every`, so it wouldn't actually work like `firstShouldForwardProp(prop) || secondShouldForwardProp(prop)` but rather as `firstShouldForwardProp(prop) && secondShouldForwardProp(prop)`.

As you have mentioned it doesnt make sense to forward `someProp` in this example, but it feels to me that if we have component that omits `someProp` from being passed down and we wrap it with `styled` again and omit any other prop too with new `shouldForwardProp` then both `someProp` and `anotherProp` should automatically not get forwarded downwards.

If we just replace `shouldForwardProp` with next one this would suddenly allow passing `someProp` which was not the original intention of `SomeComponent`.

PR #663, opened 2018-05-19
Add TS typings to create-emotion

Ailrun:
**What**: Add typings for create-emotion package

**Why**: To make TS user be able to use create-emotion package.

**How**: Using [csstype](https://github.com/frenic/csstype) package, who crawls css data from MDN.

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

I'm willing to add more strict typing (and the typing providing helpful autocompletions) for emotion package, but I think I should add typings for create-emotion, to avoid repeated code and revesed dependencies between typings.

### DIFF
--- a/packages/create-emotion/types/index.d.ts
+++ b/packages/create-emotion/types/index.d.ts
@@ -6,30 +6,25 @@ import * as CSS from 'csstype';
 export interface MultiDimensionalArray<T> extends Array<T | MultiDimensionalArray<T>> {}
 
 export type CSSBaseObject = CSS.PropertiesFallback<number | string>;
-export type CSSPseudoObject<C> = { [K in CSS.Pseudos]?: CSSObject<C> };
-export interface CSSOthersObject<C> {
-  [propertiesName: string]: Interpolation<C>;
+export type CSSPseudoObject = { [K in CSS.Pseudos]?: CSSObject };
+export interface CSSOthersObject {
+  [propertiesName: string]: Interpolation;
 }
-export interface CSSObject<C> extends CSSBaseObject, CSSPseudoObject<C>, CSSOthersObject<C> {}
+export interface CSSObject extends CSSBaseObject, CSSPseudoObject, CSSOthersObject {}
 
-export type FunctionInterpolation<C> = (mergedProps: any, context: C) => Interpolation<C> | void;
+export interface ArrayInterpolation extends Array<Interpolation> {}
 
-export interface ArrayInterpolation<C> extends Array<Interpolation<C>> {}
-
-export type Interpolation<C> =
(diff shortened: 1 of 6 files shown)


packages/create-emotion/types/tests.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+css();
+css(1);
+css('abc');
+css(true);
+
+css([]);
+css([1]);
+css([['abc', 'asdf'], 'efw']);
+
```
mitchellhamilton: Could you add a test with the string syntax, for example:
```jsx
css`
 color: ${'hotpink'};
`
```

packages/create-emotion/types/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface ArrayInterpolation<C> extends Array<Interpolation<C>> {}
+
+export type Interpolation<C> =
+  | undefined | null | boolean | string | number
+  | CSSObject<C>
+  | ArrayInterpolation<C>
+  | FunctionInterpolation<C>
+  ;
+
+export type FunctionClassNameArg = () => ClassNameArg | void;
```
mitchellhamilton: Same as above. We recently deprecated it in #643 so it would be nice to not have it in the typings.

packages/create-emotion/types/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface MultiDimensionalArray<T> extends Array<T | MultiDimensionalArray<T>> {}
+
+export type CSSBaseObject = CSS.PropertiesFallback<number | string>;
+export type CSSPseudoObject<C> = { [K in CSS.Pseudos]?: CSSObject<C> };
+export interface CSSOthersObject<C> {
+  [propertiesName: string]: Interpolation<C>;
+}
+export interface CSSObject<C> extends CSSBaseObject, CSSPseudoObject<C>, CSSOthersObject<C> {}
+
+export type FunctionInterpolation<C> = (mergedProps: any, context: C) => Interpolation<C> | void;
```
mitchellhamilton: Could you remove this? We recently deprecated it in #643 so it would be nice to not have it in the typings.

packages/create-emotion/types/tests.tsx
Ailrun: I added them! Thank you for your feedback!

packages/create-emotion/types/index.d.ts
Ailrun: OK. I removed them.

Ailrun:
@mitchellhamilton Could I ask about release schedule? When will this be released?

PR #668, opened 2018-05-22
Add typing for create-emotion-styled

Ailrun:
**What**: Add typing for create-emotion-styled package

**Why**: To allow TS users to use create-emotion-styled

**How**: By consulting with glamorous typing and old react-emotion typing

**Checklist**:

- [N/A] Documentation
- [ ] Tests
- [ ] Code complete

I'm not sure this is complete typing. Could you give me more idea for testing? What usecases did I miss?

After typing for this package is completed and #667 is merged without complex problems, I want to fix `react-emotion` typing and add `preact-emotion` typing.

### DIFF
--- a/packages/create-emotion-styled/types/index.d.ts
+++ b/packages/create-emotion-styled/types/index.d.ts
@@ -17,7 +17,7 @@ export interface StyledOptions {
   e?: string;
   label?: string;
   target?: string;
-  shouldForwardProp?: (name?: string) => boolean;
+  shouldForwardProp?: (name: string) => boolean;
 }
 
 export type Themed<P extends object, T> = P & { theme: T };
--- a/packages/create-emotion-styled/types/index.d.ts
+++ b/packages/create-emotion-styled/types/index.d.ts
@@ -1,0 +1,101 @@
+// Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
+// TypeScript Version: 2.3
+
+import { Emotion, Interpolation as BaseInterpolation } from 'create-emotion';
+import React, { ComponentClass, ReactHTML, ReactSVG, Ref, SFC } from 'react';
+
+export interface ArrayInterpolation<Props> extends Array<Interpolation<Props>> {}
+export type FunctionInterpolation<Props> = (props: Props, context: any) => Interpolation<Props>;
+
+export type Interpolation<Props> =
+  | BaseInterpolation
+  | ArrayInterpolation<Props>
+  | FunctionInterpolation<Props>
+  ;
+
+export interface StyledOptions {
+  e?: string;
+  label?: string;
+  target?: string;
+  shouldForwardProp?: (name?: string) => boolean;
+}
+
+export type Themed<P extends object, T> = P & { theme: T };
+
+export type StyledStatelessComponentProps<P extends object, T> =
+  & P
+  & { theme?: T }
+  ;
+export type StyledOtherComponentProps<P extends object, T> =
+  & StyledStatelessComponentProps<P, T>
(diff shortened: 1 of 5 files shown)


vadistic:
Wow, what a nice suprise! Thank you, It will come useful 👍 

```ts
// Btw. please please fix StyledOptions in react-emotion typings
export interface Options {
  string?: string;
}
```

Ailrun:
@vadistic Yes. It broken in many ways, wrong typed `Options`, using `Component` instead of `ComponentClass`, ....
Because of https://github.com/paypal/glamorous/issues/419, I want to migrate to emotion, but these typing issue does not allow me to migrate.

packages/create-emotion-styled/types/index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+  | BaseInterpolation
+  | ArrayInterpolation<Props>
+  | FunctionInterpolation<Props>
+  ;
+
+export interface StyledOptions {
+  e?: string;
+  label?: string;
+  target?: string;
+  shouldForwardProp?: (name?: string) => boolean;
```
mitchellhamilton: This should be `(name: string) => boolean`, the name argument will always be there.

packages/create-emotion-styled/types/test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <TestFunComp complex={props.complex} />
+  </div>
+);
+
+const StyledClassComp = createStyled(TestClassComp)({
+  width: '200px',
+}, (props) => ({
+  height: props.theme ? '200px' : '100px',
+}));
+
```
mitchellhamilton: Could you add some tests with the string syntax?

Ailrun: Ah, tests for strings. Thx! I missed them!

Ailrun:
@mitchellhamilton I added/fixed those! Could you suggest more tests I missed? I don't think those tests are enough...

packages/create-emotion-styled/types/test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  <div>
+    {props.world}
+    <TestFunComp0 complex={props.complex} />
+  </div>
+);
+
+const StyledClassComp0 = createStyled(TestClassComp)({
+  width: '200px',
+}, (props) => ({
+  height: props.theme ? '200px' : '100px',
```
mitchellhamilton: `props.theme` will never be falsy, if there isn't a `ThemeProvider`, `props.theme` will be an empty object.

packages/create-emotion-styled/types/test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  content: context.name,
+
+  color: props.color,
+}));
+
+const StyledFunComp1 = createStyled(TestFunComp0)`
+  display: inline;
+  position: fixed;
+  flexGrow: 20;
+  ${(props: StyledFunProps, context) => `
```
mitchellhamilton: Returning a string with css properties isn't something we recommend because it breaks syntax highlighting, formatting and minifying. Could you change it to this?
```jsx
css`
  content: ${context.name};
  color: ${props.color};
`
```

packages/create-emotion-styled/types/test.tsx
Ailrun: Oh, I didn't know that. I should fix typing also.

packages/create-emotion-styled/types/test.tsx
Ailrun: OK. i will change it.

PR #670, opened 2018-05-22
Add failing test for shouldForwardProp composition

brentertz:
**What**:

Add failing test for `shouldForwardProp` composition on composed `styled` components

**Why**:

I am investigating the migration of a [library](https://github.com/mineral-ui/mineral-ui) that is currently built on Glamorous, [which may soon be deprecated](https://github.com/paypal/glamorous/issues/419).

When composing styled components, I have noticed that Emotion and Glamorous seem to work differently.  Glamorous respects the inner component's filtered props, whereas Emotion does not appear to.

In Mineral, [in our emotion branch](https://github.com/mineral-ui/mineral-ui/blob/692-emotion/src/library/styles/createStyledComponent.js#L40), we are attempting to provide a wrapper around `styled`, which adds a custom `shouldForwardProp` that enables filtering of props based on the HTML tag being rendered, a feature that also exists in Glamorous.  This works fine until styled components are composed, at which point the outer `shouldForwardProp` takes precedence and the inner one is completely ignored, thus the resulting component does not receive the expected props.

I see that this issue was also discussed in
https://github.com/emotion-js/emotion/pull/659

Is this functionality desired, planned, achievable, or other?

**How**:

I have not implemented any changes, but rather am trying to understand the current implementation and suggest a change in behavior.

**Checklist**:
- [ ] Documentation - N/A
- [x] Tests
- [ ] Code complete - N/A


packages/create-emotion-styled/src/index.js
```diff
@@ -25,11 +25,18 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
     let identifierName
     let stableClassName
     let shouldForwardProp
+    let customShouldForwardProp
     if (options !== undefined) {
       staticClassName = options.e
       identifierName = options.label
       stableClassName = options.target
-      shouldForwardProp = options.shouldForwardProp
+      shouldForwardProp = customShouldForwardProp =
```
Andarist: do we need to keep 2 variables for this?

Andarist:
LGTM, can't merge it though, so you'll have to w8 for @mitchellhamilton 😉

PR #671, opened 2018-05-24
Fix typing for create-emotion-styled

Ailrun:
**What**:
1. Originally its types are mixed in single file, so it's hard to write preact type using common things.
    ~I do really hate global JSX namespace idea.~
1. It used ReactHTML, but it's not an interface for props, but for factories. Therefore it does not provide HTML attributes like `onClick` as props.
1. ReactHTML problem occurs because of lack of tests, so I add some tests too.

**Why**: Typing introduced in #668 is wrong.

**How**: By using `JSX.IntrinsicElements`

**Checklist**:

- [N/A] Documentation
- [x] Tests
- [x] Code complete


Ailrun:
@mitchellhamilton
Oh, I also forgetted about to ask it. Does babel plugin work with create-emotion-styled too?

mitchellhamilton:
@Ailrun yes, the babel plugin does work with create-emotion-styled.

packages/create-emotion-styled/types/react.d.ts
```diff
@@ -69,3 +69,11 @@ export interface CreateStyled<Theme extends object = any> {
     options?: StyledOptions,
   ): CreateStyledOtherComponent<IP, Theme>;
 }
+
+export type CreateStyledShorthands<Theme extends object> = {
+  [T in keyof JSX.IntrinsicElements]: CreateStyledOtherComponent<JSX.IntrinsicElements[T], Theme>;
+};
+
+export interface CreateStyled<Theme extends object = any>
```
mitchellhamilton: Is there a reason this needs `Theme extends object = any` whereas the shorthand one doesn't need `= any`?

Ailrun: @mitchellhamilton `CreateStyled` is access point of all API, where `CreateStyledShorthands` is just a base interface (well, it's not really interface, but in some way, it's interface) of `CreateStyled`.
Therefore, `CreateStyled` is one used by user, so I put the default type on this interface only.

PR #675, opened 2018-05-25
Fix react-emotion typing

Ailrun:
**What**: Fixing react-emotion typing

**Why**: Current typing is redundant.

**How**: Using `create-emotion-styled` typing

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/react-emotion/types/index.d.ts
+++ b/packages/react-emotion/types/index.d.ts
@@ -1,120 +1,21 @@
 // TypeScript Version: 2.3
-// tslint:disable-next-line:no-implicit-dependencies
-import { StatelessComponent, ComponentClass, CSSProperties } from 'react';
-import { Interpolation as EmotionInterpolation } from 'emotion';
+import {
+  CreateStyled,
+  Interpolation,
+  StyledComponent,
+  StyledOptions,
+  Themed,
+} from 'create-emotion-styled';
 
 export * from 'emotion';
 
-export type InterpolationFn<Props = {}> =
-  (props: Props) =>
-    | EmotionInterpolation
-    | InterpolationFn<Props>;
-
-export type InterpolationTypes<Props = {}> =
-  | InterpolationFn<Props>
-  | EmotionInterpolation;
-
-export type Interpolation<Props = {}> =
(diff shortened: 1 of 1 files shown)


Ailrun:
Why could not yarn fetch packages?

mitchellhamilton:
The yarn registry is down right now. https://status.yarnpkg.com/

packages/react-emotion/types/index.d.ts
```diff
@@ -1,120 +1,21 @@
...
+  Themed,
+} from 'create-emotion-styled';
+
+export {
+  CreateStyled,
+  Interpolation,
+  StyledComponent,
+  StyledOptions,
+  Themed,
+} from 'create-emotion-styled';
```
mitchellhamilton: Was importing everything twice intentional?

Ailrun: @mitchellhamilton Oh... no, it's my mistake. Thank you for pointing this!

mitchellhamilton:
Is there anything that should be updated in the ts docs? https://github.com/emotion-js/emotion/blob/master/docs/typescript.md

Ailrun:
@mitchellhamilton Hm... I will check it.
However, there's other problem I missed. It looks like `styled` could take other StyledComponent as well, am I right?

mitchellhamilton:
Yes, styled can take styled components.

Ailrun:
@mitchellhamilton I didn't know that... Does it works with `emotion` as well?
I mean,
```tsx
import { css } from 'emotion';
const Comp = styled.div({});

const x = css`
  ${Comp} {
    color: blue;
  }
`;
```

is also valid?

mitchellhamilton:
Yes, that also works.

Ailrun:
@mitchellhamilton Then I should update other typings first and revise this PR. I close this now.

LukasWerfel:
[This](https://emotion.sh/docs/styled#targeting-another-emotion-component) still doesnt seem to work with TS.
```
Type error: A computed property name must be of type 'string', 'number', 'symbol', or 'any'. 
```

PR #676, opened 2018-05-26
Add ClassInterpolation to typings

Ailrun:
**What**: Add ClassInterpolation to typings

**Why**: Original typings do not accept class interpolations, i.e., `StyledComponent` as interpolation.

**How**: Add ClassInterpolation with private properties.

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/create-emotion-styled/types/test.tsx
+++ b/packages/create-emotion-styled/types/test.tsx
@@ -141,7 +141,7 @@ const StyledCompWithFunComp2 = StyledFunComp0.withComponent(TestFunComp1);
 
 const StyledCompShorthand0 = createStyled.a({
   textAlign: 'center',
-})
+});
 const StyledCompShorthand1 = createStyled.label`
   display: block;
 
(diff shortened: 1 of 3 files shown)


packages/create-emotion-styled/types/test.tsx
```diff
@@ -168,13 +168,13 @@ const getEditorLogoColor = (editor: 'vscode' | 'emacs' | 'sublime') => {
...
-}
+};
 
 const StyledCompShorthandWithProps0 = createStyled.div<ShorthandProps>(props => ({
   backgroundColor: getEditorLogoColor(props.editor),
 }));
-const StyledCompShorthandWithProps1 = createStyled.section<ShorthandProps>`
-  backgroundColor: ${props => getEditorLogoColor(props.editor)};
+const StyledCompShorthandWithProps1 = createStyled.section`
+  backgroundColor: ${(props: ShorthandProps) => getEditorLogoColor(props.editor)};
```
mitchellhamilton: What's the reason for this change?

Ailrun: @mitchellhamilton It cannot pass check in TS 2.3.
We should add `yarn test:typescript` to TravisCI...

mitchellhamilton: Yeah, I removed the ts tests from travis because it was crashing and I didn't have time to figure out why, if you want to fix it and add it back [here](https://github.com/emotion-js/emotion/blob/master/package.json#L9) that would be great.

Ailrun: @mitchellhamilton I will add it after this one will be merged.

PR #678, opened 2018-05-26
Fix react-emotion typing

Ailrun:
**What**: Fixing react-emotion typing

**Why**: Current typing is redundant.

**How**: Using `create-emotion-styled` typing

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

I will send a PR for fixing typescript document after this PR is merged.

PR #679, opened 2018-05-26
Update typescript document to match with current types

Ailrun:
**What**: Update typescript document

**Why**: Typings are changed, so document is outdated.

**Checklist**:
- [x] Documentation
- [N/A] Tests
- [N/A] Code complete

This document depends on #678 and #680, so please do not merge this before them.


docs/typescript.md
```diff
@@ -2,9 +2,77 @@
...
-### html elements
+const bodyStyle = css({
+  display: 'flex',
+  flowDirection: 'column-reverse',
+});
+```
+
+## create-emotion-styled
+
+Current typing for `create-emotion-styled` is only compatible with React, and will not work with Preact. For detail typing, see following `react-element` section.
```
mitchellhamilton: Could you change "Current typing for `create-emotion-styled` is" to "The current typings for `create-emotion-styled` are only"?

docs/typescript.md
```diff
@@ -2,9 +2,77 @@
 title: "Typescript"
 ---
 
-Emotion includes TypeScript definitions for `styled` components and has type inferences for both html elements and React components.
+Emotion includes TypeScript definitions for `create-emotion`, `emotion`, `create-emotion-styled`, and `react-emotion` packages. These definition also could infer types for css property (with object syntax), and HTML/SVG tag name, and props types.
```
mitchellhamilton: Is there a circumstance where it can't infer types for css properties?
If not, could you replace "These definition also could infer types for css property" with "These definitions also infer types for css properties"?

Also, could you replace "HTML/SVG tag name" with "HTML/SVG tag names"?

Ailrun: Allright. I will change them.

docs/typescript.md
```diff
@@ -2,9 +2,77 @@
...
+});
+
+const subtitleStyle = myEmotion.css`
+  boxSizing: 'border-box';
+  width: 100;
+  height: 60;
+`;
+```
+
+Typescript checks css properties in object style syntax using csstype package, so following code will emit errors.
```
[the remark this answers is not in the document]
Ailrun: But it would be helpful when users meet some type issues.... wouldn't it?

docs/typescript.md
Ailrun: OK. Thank you for your review!

docs/typescript.md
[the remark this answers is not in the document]
Ailrun: Oh... you're right. I will add the link too.

docs/typescript.md
```diff
@@ -184,3 +225,30 @@ const Button = styled('button')`
 
 export default Button
 ```
+
+## preact-emotion
+
+Types are almost same with `react-emotion` package.
```
mitchellhamilton: If they're almost the same and not the exact same, could you explain how they're different?

docs/typescript.md
```diff
@@ -184,3 +225,30 @@ const Button = styled('button')`
...
+
+const bodyStyle = myEmotion.css({
+  display: 'flex',
+  flowDirection: 'column-reverse',
+});
+```
+
+## create-emotion-styled
+
+The current typings for `create-emotion-styled` are only compatible with React, and will not work with Preact. For detail typing, see `react-element` section above.
```
mitchellhamilton: This isn't true anymore about not supporting Preact, right?

docs/typescript.md
```diff
@@ -131,33 +169,36 @@ const Component: SFC = ({ label, className }) => (
...
 
-Unfortunately, when you pass custom props to a styled component, TypeScript will stop inferring your Component props, and you will need to re-type them.
-
 ### Define a Theme
 
 By default, the `props.theme` has `any` type annotation and works without error.\
 However, you can define a theme type by creating a another `styled` instance.
 
-_styled.tsx_
+_styled.jsx_
```
mitchellhamilton: Why the change in extension, isn't `tsx` generally used?

docs/typescript.md
Ailrun: OK. I will add following.
> However, this typing use Preact component types like `ComponentConstructor` and `FunctionalComponent` instead of React component types.

Is this enough?

docs/typescript.md
Ailrun: Oh, my mistake. Sorry. I will change these.

docs/typescript.md
Ailrun: FYI, Preact and React both modify `JSX.Element`, so we cannot import both typings in one file. That's why I did not provide typing for preact version of `create-emotion-styled`. (Its main typing file already import React typing)

Ailrun: The only way I can image, to deal with the problem, is, providing index typing without import any (P)React related typings and making users import them.

However, this may confuse users, because it will not work properly when users do not import (P)React related typings.

docs/typescript.md
```diff
@@ -47,48 +84,42 @@ const App = () => <Link href="#">Click me</Link>
...
 
 type ImageProps = {
   src: string,
+  width: number;
 }
 
-const Image = styled<ImageProps, 'div'>('div')`
-  background: url(${props => props.src}) center center;
+const Image0 = styled('div')`
+  width: ${(props: ImageProps) => props.width};
```
SpaceK33z: Is there a reason you are not using `styled<ImageProps, 'div'>('div')`? I think that looks way clearer, especially because then you don't have to do `(props: ImageProps) => ...` for every prop function.

docs/typescript.md
```diff
@@ -47,48 +84,42 @@ const App = () => <Link href="#">Click me</Link>
...
+// Or with generic type
 
-// Or shorthand
-
-const Image = styled.div<ImageProps>({
-  backgroundSize: contain;
-}, ({ src }) => ({
-  background: `url(${src}) center center`,
-}))
+const Image1 = styled('div')<ImageProps>({
```
SpaceK33z: This doesn't work, since TS expects a second argument, `styled('div')<ImageProps, 'div'>. TypeScript 2.9 has a fix for this, but 2.9 has not yet been released.

docs/typescript.md
SpaceK33z: Why are you using `width: ${(props: ImageProps)` instead of typing the component with `styled<ImageProps>('div')`? This is a bit confusing for me, since I see the latter as a much clearer way to type your component.

Ailrun: Because with updated typing, there're no type parameters Props for `styled`.

Ailrun: In TS 2.9 (which is in RC), you can use template string generic, however, before it, you should give a type to function and let TS infer correct type for component.

SpaceK33z: Why? Typing every prop function with `${(props: ImageProps)` is more work, right? The redundant `div` in `styled<ImageProps, 'div'>` is a bit frustrating, but it will not be necessary anymore as soon as TypeScript 2.9 is released.

SpaceK33z: > what you mentioned occurs only when you use more than two function argument, and as far as I experienced, it's not that common

In my experience I have a lot of these functions, if you are building a component library you can easily have ten of those prop functions. But if this isn't relevant anymore with TS 2.9 I digress.

Ailrun: Maybe my opinion is from my **object syntax** based experience. Even when I make a library, I rarely use more than two function arguments.

Ailrun: And, thank you for your feedback!

felipeleusin:
Hi folks,

I just hit an issue. I have a button component that uses custom props but fallsback to theme props. But if I use 

(props: ButtonProps) => whatever

I lose typing to Themed. I've been trying to get this to work and failing. Do you have any pointers? Should I make something Themed<ButtonProps> for this cases? Maybe it's a good thing to add in the docs

Ailrun:
@felipeleusin If you use TS >= 2.9, you can use something like
```
styled('div')<ButtonProps>`
  ${props => something}
`;
```

felipeleusin:
Oh, ok! Will try this. Thanks a lot!

Ailrun:
I should add that to the document...

felipeleusin:
I can add a PR for that, was just checking it's working first :)

Ailrun:
@felipeleusin Oh, I'm not start yet. I'm discussing about typing in #721.

PR #683, opened 2018-05-27
Fix incorrect code and a type in TypeScript docs

SpaceK33z:
Two of the examples did not work and caused some confusion to me when trying to implement something similar in my project (they gave a compile time error). Also fixed one small typo.


Ailrun:
I believe these will be fixed with #679.
However, that PR is much bigger than this one, it could take some time...

Ailrun:
Close this in favor of #679. Thank you for your concern and effort!

PR #690, opened 2018-05-29
Fix typings for using components as selectors.

### LINKED ISSUE #685: Typescript and component selectors
- `emotion` version: 9.1.3
- `react` version: 16

Consider `Wrapper` defined like this:

```typescript
interface IconProps {
  name: string
  className?: string
}
const ReactIcon : React.StatelessComponent<IconProps> = .... // some react component
const Icon = styled(ReactIcon)``

const Wrapper = styled.div`
  ${Icon} {
      color: black;
  }
`
```

Then this component will not compile:
```
<Wrapper>
  <Icon name="menu" />
</Wrapper>
```

Typescript error:
```
TS2322: Type '{ children: Element[]; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<IconProps & { children?: ReactNode; } & ...'.
  Type '{ children: Element[]; }' is not assignable to type 'Readonly<IconProps & { children?: ReactNode; } & ClassAttributes<HTMLDivElement> & HTMLAttributes...'.
    Property 'name' is missing in type '{ children: Element[]; }'.
```

mgroenhoff:
**What**: Fix the typings to allow react components to be used as component selectors (fixes #685).

**Why**: Because when using Typescript 2.8.3 i get the following error:

> Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'CSSProperties | { [key: string]: ObjectStyleAttributes; } | ((props: ThemedProps<RowProps, Theme>...'.
>   Type 'TemplateStringsArray' is not assignable to type '(props: ThemedProps<RowProps, Theme>) => ObjectStyleAttributes'.
>     Type 'TemplateStringsArray' provides no match for the signature '(props: ThemedProps<RowProps, Theme>): ObjectStyleAttributes'.

**How**: I added a type to the list of allowed interpolation types.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

**_Update_**: I changed it to only allow styled components and not regular (stateless) components.

### DIFF
--- a/packages/react-emotion/types/index.d.ts
+++ b/packages/react-emotion/types/index.d.ts
@@ -1,6 +1,6 @@
 // TypeScript Version: 2.3
 // tslint:disable-next-line:no-implicit-dependencies
-import { StatelessComponent, ComponentClass, CSSProperties } from 'react';
+import { ReactElement, StatelessComponent, ComponentClass, ComponentType, CSSProperties } from 'react';
 import { Interpolation as EmotionInterpolation } from 'emotion';
 
 export * from 'emotion';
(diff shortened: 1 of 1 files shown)


Ailrun:
@tkh44
It looks like this and #678 have some level of redundancy. Could you check it too?

tkh44:
@Ailrun @mgroenhoff can you 2 coordinate on this? It looks like #678 can be applied on top of this but I'm not a TypeScript expert.

Ailrun:
@mgroenhoff #678 is for remove redundancy between `create-emotion-styled` and `react-emotion`. Since we already have typings for `create-emotion-styled`, we don't need to add complex typings in `react-emotion`.

mgroenhoff:
Okay, what do I need to change?

Ailrun:
@mgroenhoff I mean... this PR is about changing the typings for `react-emotion`, and #678 is about removing typings (and replacing by typings from `create-emotion-styled`), so these intents make conflicts... don't they?

mgroenhoff:
I think the idea behind #676 (which I missed) is the same as this PR. #678 indeed invalidates this one because it makes use of those changes. I'll have to test it out tomorrow morning but I think this will do the job too.

mgroenhoff:
I have tested #678 and that solved my issue. Please merge that one 👍

PR #692, opened 2018-05-29
Add Typescript definitions for 'create-emotion-server' and 'emotion-server'

mgroenhoff:
**What**: Typescript definitions added for `create-emotion-server` and `emotion-server`

**Why**: Because Typescript!

**How**:

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [ ] Code complete N/A


mgroenhoff:
Please update `yarn.lock` and commit it yourself because I can't seem to get it to generate the correct registry (https://github.com/yarnpkg/yarn/issues/3330)

packages/emotion-server/types/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dtslint.json",
+    "rules": {
+        "no-relative-import-in-test": false
+    }
+}
```
Ailrun: You miss the final linefeeding for this file

packages/emotion-server/package.json
```diff
@@ -20,8 +23,10 @@
     "emotion": "^9.1.3"
   },
   "devDependencies": {
+    "@types/react-dom": "16.0.5",
```
Ailrun: It should depend on `@types/react-dom` as dependencies, or, at least as peerDependencies, shouldn't it?

mgroenhoff: The only use is inside the tests so I don't think it should.

Ailrun: @mgroenhoff Oh, you're right. My miss.

mgroenhoff:
Thanks!

PR #694, opened 2018-05-30
Add react/preact as peer-dependencies to react-emotion/preact-emotion

### LINKED ISSUE #684: react-emotion is missing react (peer)dependency
There is a direct dependency on react in your code but there is no reference to it in package.json

https://github.com/emotion-js/emotion/blob/d9a9e332dfbc13ea78325082393ccdd54d34afd6/packages/react-emotion/src/index.js#L2

https://github.com/emotion-js/emotion/blob/d9a9e332dfbc13ea78325082393ccdd54d34afd6/packages/react-emotion/package.json#L21-L27

This would normally not be an issue when using a package manager (NPM) that optimizes the resulting dependency tree since react will be placed at the root of the node_modules directory. Otherwise you'll get an error like me:

> Module not found: Error: Can't resolve 'react' in ...

mgroenhoff:
Fixes #684

### DIFF
--- a/packages/create-emotion-styled/package.json
+++ b/packages/create-emotion-styled/package.json
@@ -21,13 +21,13 @@
     "@emotion/is-prop-valid": "^0.6.1"
   },
   "peerDependencies": {
-    "prop-types": ">= 15"
+    "prop-types": "^15.6.1"
   },
   "devDependencies": {
     "@types/react": "16.0.16",
     "dtslint": "^0.3.0",
     "npm-run-all": "^4.0.2",
-    "prop-types": "^15.6.0",
+    "prop-types": "^15.6.1",
     "rimraf": "^2.6.1",
     "rollup": "^0.51.3"
   },
--- a/packages/create-emotion-styled/package.json
+++ b/packages/create-emotion-styled/package.json
@@ -21,7 +21,7 @@
     "@emotion/is-prop-valid": "^0.6.1"
   },
   "peerDependencies": {
-    "prop-types": "^15.6.1"
+    "prop-types": "15.x"
   },
   "devDependencies": {
     "@types/react": "16.0.16",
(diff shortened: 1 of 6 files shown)


packages/react-emotion/package.json
```diff
@@ -23,7 +23,8 @@
     "create-emotion-styled": "^9.1.2"
   },
   "peerDependencies": {
-    "emotion": "^9.1.3"
+    "emotion": "^9.1.3",
+    "react": ">= 15"
```
Andarist: Those `>=` ranges are too wide, we can't guarantee being compatible with future versions of react/preact, so it's "dangerous" to make this assumption here. If we are gonna be compatible with future versions then we should loosen up this range in the future to include next versions. 

Let's just stick to something like `15.x || 16.x` here and `8.x` for `preact`

mgroenhoff: Allright. I think these need to be changed these as well then.
https://github.com/emotion-js/emotion/blob/d9a9e332dfbc13ea78325082393ccdd54d34afd6/packages/create-emotion-styled/package.json#L24
https://github.com/emotion-js/emotion/blob/d9a9e332dfbc13ea78325082393ccdd54d34afd6/packages/emotion-theming/package.json#L50-L51

Andarist: yeah, imho they should be too, good catch 👍

packages/create-emotion-styled/package.json
```diff
@@ -21,13 +21,13 @@
     "@emotion/is-prop-valid": "^0.6.1"
   },
   "peerDependencies": {
-    "prop-types": ">= 15"
+    "prop-types": "^15.6.1"
```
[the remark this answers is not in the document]
mgroenhoff: Right, will do. I did not see where the peer dependency came from by looking at the source so I thought it did not matter but now i see it is actually emitted in the bundle.

PR #695, opened 2018-05-30
Generate the same class names for server and client if source maps are different

### LINKED ISSUE #688: Client generated different class names than server-side when using source-maps
I found that when I enable source maps, React is giving me a warning:

> Warning: Prop `className` did not match. Server: "css-18slqts-Cell e148iowh2" Client: "css-ngphzi-Cell e148iowh2"

I looked into this and i found that when generating the class name, a hash is created from the generated CSS + some identifier name.

https://github.com/emotion-js/emotion/blob/d9a9e332dfbc13ea78325082393ccdd54d34afd6/packages/create-emotion/src/index.js#L262

The styles in my case contains an inline source map URL and this URL is what is different (they're different because one is has a 'sourceRoot' property and contains Windows style paths, and the other does not have a 'sourceRoot' property and contains Unix style paths).

My suggestion is to only generate a hash of the styles up to the start of the source mapping URL comment (when process.env.NODE_ENV !== 'production' ofcourse).

I have tried a quick fix and replaced the previous line with the following and this seems to do the job for me:

```js
      name = hashString(styles.slice(0, styles.indexOf("/*# sourceMappingURL=") || styles.length) + identifierName) + identifierName;
    } else {
      name = hashString(styles + identifierName) + identifierName;
    }
```

mgroenhoff:
**What**: Generate the hash for the class name from the generated CSS without the source maps

**Why**: A different babel setup for server and client can change the way how source maps are generated (like the `sourceRoot` property or different paths to source files) but the resulting styles are the same.

**How**: Do not include the source maps when creating the hash used for the class name

Resolves #688.


Andarist:
Could u describe in more detail why ur babel setup is different for client & server?

mgroenhoff:
For my server-side code I run typescript and then babel. For my client-side code I run typescript and then i run webpack with the babel-loader. I run typescript once for both my client and server code. However i run babel twice, once for my client side code (with the browser target) and once for my server code (with the node target) but they output code in different directories. The emotion babel plugin that generates the source maps generates different paths (because they come from a different place). However the CSS has not changes because it comes from the same source.

Lets say the emotion babel plugin adds a source map to some piece of CSS. The decoded source map for the client will look something like:
```js
{
  version: 3,
  sources: [ 'path/to/client/code.jsx' ], // different
  names: [],
  mappings: [ '...' ], // this is identical
  file: 'path/to/client/code.jsx', // different
  sourceRoot: 'path/to/workspace/root',
  sourcesContent: [ '...' ], // this is identical
}
```
And this will be the equivalent on the server:
```js
{
  version: 3,
  sources: [ 'path/to/server/code.jsx' ], // different
  names: [],
  mappings: [ '...' ], // this is identical
  file: 'path/to/server/code.jsx', // different
  // for some reason it does not include a sourceRoot property
  sourcesContent: [ '...' ], // this is identical
}
```

packages/create-emotion/src/index.js
```diff
@@ -259,7 +259,17 @@ function createEmotion(
       identifierName += `-${p1}`
       return ''
     })
-    name = hashString(styles + identifierName) + identifierName
+    if (process.env.NODE_ENV !== 'production') {
+      name =
```
Andarist: Does this problem affect production env only? I suspect it does not and any solution for this should be applied regardless of the environment

Andarist: Oh, after investigating the source code some more it seems that source maps are handled only in non-production code anyway, so this check here is OK. I would hoist `sourceMapRegEx` up and reuse it at both places.

mgroenhoff: One has the global flag and the other has not. I could however do something like this:
```js
const sourceMapRegEx = /\/\*#\ssourceMappingURL=data:application\/json;\S+\s+\*\//
const sourceMapRegExGlobal = new RegExp(sourceMapRegEx, 'g')
```
Or do you have another suggestion?

Ailrun:
Resolves #688

mitchellhamilton:
Could you explain more about your use case? We don't support source maps in SSR so I don't understand why this would happen.

mgroenhoff:
The issue is not that I want to have source maps server-side, it is that if I want them client-side, I need the server-side to generate the same class names. The class name hash is generated based on the CSS which now includes the base64 encoded inline source map. I need it to create the class name hash only from the actual CSS so that when source maps are different or absent server-side so that React does not error about not being able to hydrate because the class name attribute differs.

mgroenhoff:
Wrong comment button...

mitchellhamilton:
Okay, makes sense. This change means we're checking the string for the source map twice, could you change it so this check also sets the source map variable and remove the other one?

Also, checking process.env is expensiveish in node so it's best to do the checks at initialization time(though don't define a isDev variable or something since that means the code won't be removed in production on the web). So could you change it so it's something like the production checks in `index.js` and `utils.js` in `create-emotion`?

fforres:
Currently we are having this same issue on SSR our CMS, so really excited by this PR. 
Awesome @mgroenhoff !! :D

Andarist:
I've done the merging, once CI gets green gonna merge it.

mgroenhoff:
Thanks guys!

PR #699, opened 2018-06-03
chore: add Node.js 10

DanielRuf:
Also test against Node.js 10.


DanielRuf:
> WDYT?

Makes sense.

PR #703, opened 2018-06-07
Rename a type

mitchellhamilton:
**What**:

Rename the `EmotionOption` type to 

I also updated the `dtslint` versions in react-emotion and emotion-theming so that they're the same as the other packages.

**Why**:

It's optionS so the type should be plural.

**How**:

**Checklist**:

- [ ] Documentation (I'm gonna change it in the other PR)
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/react-emotion/package.json
+++ b/packages/react-emotion/package.json
@@ -29,7 +29,7 @@
   "devDependencies": {
     "@types/react": "16.0.16",
     "cross-env": "^5.0.5",
-    "dtslint": "^0.2.0",
+    "dtslint": "^0.3.0",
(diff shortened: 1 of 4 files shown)


packages/react-emotion/package.json
```diff
@@ -29,7 +29,7 @@
   "devDependencies": {
     "@types/react": "16.0.16",
     "cross-env": "^5.0.5",
-    "dtslint": "^0.2.0",
+    "dtslint": "^0.3.0",
```
Ailrun: Oh, I missed it! Thanks!

PR #708, opened 2018-06-07
Update types to fix error when build with declaration true

guilhermegregio:
**What**:
reexports StyledOtherComponent in react-emotion and import * as React in create-emotion-styled

**Why**:
to fix problem of build with declaration true in tsconfig.json

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete


Ailrun:
Could you also change the tsconfig settings in those typings? They use `allowSyntheticDefaultImports: true` in their settings.

guilhermegregio:
@Ailrun Tks this work fine to me.

PR #709, opened 2018-06-07
Fix serializer replacing non emotion class names

### LINKED ISSUE #706: jest-emotion 9.2.0 replacing non-emotion class names
- `emotion` version: 9.2.1 **(jest-emotion: 9.2.0)**
- `react` version: 16.4.0

Relevant code.

```javascript
import React from "react";
import { css, cx } from "emotion";
import renderer from "react-test-renderer";

const classes = cx(
  "net-42",
  "net",
  css`
    color: darkorchid;
  `
);

it("renders with correct styles", () => {
  expect(renderer.create(<div className={classes} />)).toMatchSnapshot();
});
```

What you did:

Upgraded jest-emotion from 9.1.3 to 9.2.0.

What happened:

Serialized snapshot tests failed due to incorrect classes being applied.

```css
- Snapshot
+ Received

<div
-  className="net-42 net emotion-0"
+  className="emotion-0 net emotion-1"
/>
```

Reproduction:

https://github.com/wKovacs64/jest-emotion-testing

Problem description:

CSS classes that match the [`/e[a-zA-Z0-9-]+[0-9]+/`](https://github.com/emotion-js/emotion/blob/ee71446ac21861f938f51b5696db956d75e8ce87/packages/jest-emotion/src/replace-class-names.js#L8) regex (e.g. `net-42`) are being treated as `emotion` classes.

Suggested solution:

I think this came from https://github.com/emotion-js/emotion/pull/641 and the regex just needs updating?

mitchellhamilton:
**What**:
Fix serializer replacing non emotion class names

**Why**:
Closes #706 

**How**:

Make the regex used to replace the component selector class names more strict.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/jest-emotion/src/replace-class-names.js
+++ b/packages/jest-emotion/src/replace-class-names.js
@@ -5,7 +5,7 @@ function defaultClassNameReplacer(className, index) {
 
 export type ClassNameReplacer = (className: string, index: number) => string
 
-const componentSelectorClassNamePattern = /e[a-zA-Z0-9-]+[0-9]+/
+const componentSelectorClassNamePattern = /^e[a-zA-Z0-9]+[0-9]+$/
 
 export const replaceClassNames = (
   classNames: Array<string>,
--- a/packages/jest-emotion/test/__snapshots__/printer.test.js.snap
+++ b/packages/jest-emotion/test/__snapshots__/printer.test.js.snap
@@ -1,5 +1,15 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`does not replace class names that are not from emotion 1`] = `
+".emotion-0 {
+  color: darkorchid;
+}
+
+<div
+  className=\\"net-42 net emotion-0\\"
+/>"
+`;
+
 exports[`jest-emotion with DOM elements disabled does not replace class names or insert styles into DOM element snapshots 1`] = `
 "<div
   class=\\"css-17f3slo\\"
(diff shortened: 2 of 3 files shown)


packages/jest-emotion/src/replace-class-names.js
```diff
@@ -5,7 +5,7 @@ function defaultClassNameReplacer(className, index) {
 
 export type ClassNameReplacer = (className: string, index: number) => string
 
-const componentSelectorClassNamePattern = /e[a-zA-Z0-9-]+[0-9]+/
+const componentSelectorClassNamePattern = /^e[a-zA-Z0-9]+[0-9]+$/
```
Ailrun: Where is the `'-'`? I mean, is it OK to remove `'-'`?
This pattern does not match with `emotion-0`....

mitchellhamilton: It's not meant to match `emotion-0`.

It's meant to match class names like this `e17ivhiq0`.

Component selector class names used to be like `css-17ivhiq0` so this included `-` and I just forgot to remove it when I changed the shape of the component selector class names

Ailrun: Thank you for your detailed answer! :)

Ailrun: @mitchellhamilton Wa...it, this also does not match with `e17ivhiq0`, does it?
It matches with `e17` or `eifjwe5239`, but does not with `e17ivhiq0`.

Ailrun: Never mind. I'm the fool.

PR #715, opened 2018-06-11
Add third party packages section to document

### LINKED ISSUE #638: Documenting Ember add-on
I've written an Ember add-on to integration Emotion into Ember apps, called [`ember-emotion`](https://www.npmjs.com/package/ember-emotion). I would love to add it to the documentation so that people in the Ember community know that it's an option when looking at Emotion's website.

Would you be interested in me adding it to the *Packages* section of the documentation site even though it isn't an "official" package that's part of your organization? Would you be interested moving the repo under your umbrella?

Ailrun:
**What**: Documents for third party packages

**Why**: To support and encourage third party package authors.

**How**: Rename "Packages" section into "Official Packages" and add "Third Party Packages" section.

**Checklist**:
- [x] Documentation
- [N/A] Tests
- [N/A] Code complete

Resolves #638.

### DIFF
--- a/docs/docs.yaml
+++ b/docs/docs.yaml
@@ -43,3 +43,7 @@
     - create-emotion-styled
     - create-emotion-server
     - preact-emotion
+
+- title: Third Party Packages
+  items:
+    - ember-emotion
--- a/docs/docs.yaml
+++ b/docs/docs.yaml
@@ -30,8 +30,8 @@
     - typescript
     - extract-static
 
-# This loads the READMEs instead of files in docs/
-- title: Packages
+# Official Packages loads the READMEs instead of files in docs/
+- title: Official Packages
   items:
     - emotion
     - react-emotion
--- a/docs/ember-emotion.md
+++ b/docs/ember-emotion.md
@@ -1,0 +1,49 @@
+# ember-emotion
+
+> Use emotion in Ember.js
+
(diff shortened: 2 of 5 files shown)


Ailrun:
What is that weird error? Why couldn't it find a `handlebar`?

Ailrun:
@mitchellhamilton Ah, his document use `handlebars`... I miss that part. I will fix that.

Ailrun:
@mitchellhamilton It still emits a same error. Where do I need to add `handlebars` related import (or settings, ...)?

mitchellhamilton:
I've been thinking about this and I'm not sure if it's a good idea because if we add the documentation for one package then we have to add the documentation for every package that wants it and it's effectively saying we recommend whatever packages that are there.

I think a better solution is to add a top level Community page, add it to the links at the top and have the contents be from [emotion-js/awesome-emotion](https://github.com/emotion-js/awesome-emotion) which I just created, similar to https://www.styled-components.com/ecosystem.

Ailrun:
@mitchellhamilton Wow, awesome opinion! Thank you for making that page. However, I wish that there is a link to your repository from our site. Is it OK for you?

mitchellhamilton:
I was thinking, instead of linking to it, embed the contents of it in the site.

Ailrun:
@mitchellhamilton All contents? Just like styled component did?

Ailrun:
Close this in favor of #720.

PR #720, opened 2018-06-13
Use emotion-next in the site, update to gatsby v2 and add Community page

mitchellhamilton:
**What**:
Use emotion-next in the site, update to gatsby v2 and add Community page

**Why**:
I wanted to see what the migration to emotion@10 will be like

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

This isn't ready yet, I just wanted to open the PR to get people's thoughts on the Community page


packages/emotion-theming/src/theme-provider.js
```diff
@@ -2,15 +2,16 @@
...
 import createBroadcast from './create-broadcast'
 import { channel, contextTypes, type Theme } from './utils'
 
 const isPlainObject = test =>
   Object.prototype.toString.call(test) === '[object Object]'
 
 type Props = {
-  theme: Theme
+  theme: Theme,
+  children?: ReactNode
```
Andarist: should this be optional? i get that it's technically possible to use it without `children` but it also doesn't make sense

PR #722, opened 2018-06-13
Resolve type incompatibilities with legacy type

### LINKED ISSUE #721: Typing of props in restyled components is broken in latest build
Big changes were made to the typings in the last few releases.
I tried to upgrade from 9.1.3 to 9.2.3, where it appears that inference of props types is broken.

- `emotion` version: 9.2.3
- `react` version: 16.4.0
- `typescript` version: 2.9.1

Relevant code.

```typescript
import React from 'react';

interface MyProps {
  className?: string;
  someBehavior: boolean;
}

const MyComponent: React.SFC<MyProps> = ({className}) => <div className={className} />

const MyStyledComponent = styled(MyComponent)`
  border: ${props => props.someBehavior ? 1 : 0}px;
`;
```

Results in this error:
```
Property 'someBehavior' does not exist on type '{ theme: any; }'.
```

Ailrun:
**What**: Enhance type inference for following case.
```tsx
interface Props {
  w: number;
}
declare const Comp: SFC<Props>;
const X = styled(Comp)`
  width: ${props => props.w};
`
```

and add type parameter for compatibilities with old type

```tsx
interface Props {
  w: number;
}
const X = styled<Props, 'div'>('div')`
  width: ${props => props.w};
`
```

**Why**: #721.

**How**: By adding default type parameter

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Code complete

Resolves #721.

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -71,16 +71,30 @@ const App = () => (
 ```tsx
 import styled from 'react-emotion';
 
-const NotALink = styled('div')`
+const NotALink = styled<{}, 'div'>('div')`
   color: red;
 `
 
 const Link = NotALink.withComponent('a')
 
+type TextProps = {
+  className: string;
+  text: string;
+}
+const Text = ({ className, text }: TextProps) => <div className={className}>{text}</div>
+const StyledText = styled<{}, TextProps>(Text)`
+  fontSize: 20px;
+  color: white;
+`;
+
+const OtherLink = StyledText.withComponent('a');
+
 const App = () => <Link href="#">Click me</Link>
 
 // No errors!
 ```
+
+When you use `withComponent`, you should explicitly provide the type of props used to make style. If you don't, TS will think all props for `Text` are used to make style, and request user to give all those props when using the result component of `withComponent`.
 
 ### Passing Props
 
@@ -90,26 +104,26 @@ You can type the props of styled components.
 import styled from 'react-emotion'
 
 type ImageProps = {
-  src: string,
+  src: string;
   width: number;
 }
 
-const Image0 = styled('div')`
-  width: ${(props: ImageProps) => props.width};
+const Image0 = styled<ImageProps, 'div'>('div')`
+  width: ${(props) => props.width};
   background: url(${(props: ImageProps) => props.src}) center center;
   background-size: contain;
 `
 
 // Or with object styles
 
-const Image1 = styled('div')({
+const Image1 = styled<ImageProps, 'div'>('div')({
   backgroundSize: 'contain',
-}, (props: ImageProps) => ({
+}, (props) => ({
   width: props.width;
   background: `url(${props.src}) center center`,
 }));
 
-// Or with a generic type
+// Or with a generic function
 
 const Image1 = styled('div')<ImageProps>({
   backgroundSize: 'contain',
@@ -119,7 +133,8 @@ const Image1 = styled('div')<ImageProps>({
 }));
 ```
 
-* The generic type version only works with object styles due to https://github.com/Microsoft/TypeScript/issues/11947.
+* The generic function version only works with object styles in TS <= 2.8 due to https://github.com/Microsoft/TypeScript/issues/11947.
+* If you use TS > 2.9, generic function will works with string styles too, but it will break VSCode syntax highlighting. See https://github.com/emotion-js/emotion/issues/721#issuecomment-396954993
 
 ### React Components
 
@@ -128,8 +143,8 @@ import React, { SFC } from 'react'
 import styled from 'react-emotion'
 
 type ComponentProps = {
-  className?: string,
-  label: string
+  className?: string;
+  label: string;
 }
 
 const Component: SFC<ComponentProps> = ({ label, className }) => (
@@ -158,12 +173,12 @@ const App = () => (
 import React, { SFC } from 'react'
 import styled from 'react-emotion'
 
-type ComponentProps = {
-  className?: string,
-  label: string
+type ComponentProps0 = {
+  className?: string;
+  label: string;
 }
 
-const Component: SFC<ComponentProps> = ({ label, className }) => (
+const Component0: SFC<ComponentProps0> = ({ label, className }) => (
   <div className={className}>{label}</div>
 )
 
@@ -171,12 +186,33 @@ type StyledComponentProps = {
   bgColor: string
 };
 
-const StyledComponent0 = styled(Component)`
+const StyledComponent00 = styled<StyledComponentProps, ComponentProps0>(Component)`
   color: red;
-  background: ${(props: StyledComponentProps) => props.bgColor};
+  background: ${props => props.bgColor};
 `
 
-const StyledComponent1 = styled(Component)<StyledComponentProps>({
+const StyledComponent01 = styled(Component)<StyledComponentProps>({
+  color: 'red',
+}, props => ({
+  background: props.bgColor,
+}));
+
+type ComponentProps1 = {
+  className?: string;
+  label: string;
+  bgColor: string;
+}
+
+const Component1: SFC<ComponentProps1> = ({ label, className }) => (
+  <div className={className}>{label}</div>
+)
+
+const StyledComponent10 = styled(Component)`
+  color: red;
+  background: ${props => props.bgColor}
+`
+
+const StyledComponent11 = styled(Component)({
   color: 'red',
 }, props => ({
   background: props.bgColor,
@@ -184,8 +220,10 @@ const StyledComponent1 = styled(Component)<StyledComponentProps>({
 
 const App = () => (
   <div>
-    <StyledComponent0 bgColor="red" label="Some cool text" />
-    <StyledComponent1 bgColor="red" label="Some more cool text" />
+    <StyledComponent00 bgColor="red" label="Some cool text" />
+    <StyledComponent01 bgColor="red" label="Some more cool text" />
+    <StyledComponent10 bgColor="red" label="Much more cool text" />
+    <StyledComponent11 bgColor="red" label="The most cool text" />
   </div>
 )
 ```
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -94,7 +94,7 @@ const App = () => <Link href="#">Click me</Link>
 // No errors!
 ```
 
-When you use `withComponent`, you should explicitly provide the type of props used to make style. If you don't, TS will think all props for `Text` are used to make style, and request user to give all those props when using the result component of `withComponent`.
+The prop types used to construct the resulting style definition of `withComponent` should be explicitly provided to avoid TypeScript compiler error. If you don't, TS will think all props for `Text` are used to construct style definition, and request user to give all those props when using the result component of `withComponent`.
 
 ### Passing Props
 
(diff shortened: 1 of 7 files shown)


packages/create-emotion-styled/types/test.tsx
```diff
@@ -131,9 +131,9 @@ const StyledTag1 = createStyled('label')`
 const StyledCompWithButton = StyledClassComp0.withComponent('button');
 const StyledCompWithFunComp2 = StyledFunComp0.withComponent(TestFunComp1);
 
-<StyledCompWithButton />;
```
tkh44: Should we keep this case?

Ailrun: @tkh44 This is the problem what I mentioned in #721.
We cannot distinguish props that is used to make style and props that is passed to internal component without type parameter, so there's two ways, considering them as props for style, and considering them as a internal-only props.

#721, @mvestergaard said complaint about second choice (my original choice), so I changed this into first one.

Ailrun: @tkh44 Could you discuss about this with @mvestergaard ?

mvestergaard: This might be naive, but wouldn't this make that use case work?
```typescript
withComponent<T extends keyof JSX.IntrinsicElements>(
     tag: T,
     options?: StyledOptions,
   ): StyledOtherComponent<JSX.IntrinsicElements[T], JSX.IntrinsicElements[T], Theme>;
```

Basically force Props to be the same as InnerProps.

Ailrun: @mvestergaard The problem occurs in original style. Original style already depends on old `Props`, so we cannot change them freely. If we do so, it will break the type.

mvestergaard: Anyway, I'm not sure what i can discuss with @tkh44 
I've spoken my case on #721 already. The gist is basically, that a breaking change was made. Something that worked just fine before, now requires explicit typing.
In my head this is wrong, but I can't force you guys to change it based on that. So you need to decide internally how you intend the library to be used.
The most common use case should be prioritized.

My personal opinion is that restyling components:
```
const Restyled = styled(OtherComponent)``
```
is a more common use case than switching the internal component
```
const SomeStyledThing = styled.div``;
const SomeOtherStyledThing = SomeStyledThing.withComponent('span');
```
But @Ailrun believes it's the other way around

Ailrun: @mvestergaard Thank you for your summary!
What I mean with 'discussion' is about 'Which one is more common and preferable use case'.
How about your opinion, @tkh44?

Ailrun:
@tkh44 Why did codecov coverage decrease? Does codecov coverage include .ts files?

mitchellhamilton:
@Ailrun It said that because this branch was out of date with master

docs/typescript.md
```diff
@@ -90,26 +104,26 @@ You can type the props of styled components.
...
-  src: string,
+  src: string;
   width: number;
 }
 
-const Image0 = styled('div')`
-  width: ${(props: ImageProps) => props.width};
+const Image0 = styled<ImageProps, 'div'>('div')`
+  width: ${(props) => props.width};
   background: url(${(props: ImageProps) => props.src}) center center;
```
mitchellhamilton: Should `props: ImageProps` be changed to `props` here?

docs/typescript.md
```diff
@@ -90,26 +104,26 @@ You can type the props of styled components.
...
   background-size: contain;
 `
 
 // Or with object styles
 
-const Image1 = styled('div')({
+const Image1 = styled<ImageProps, 'div'>('div')({
```
mitchellhamilton: Could you explain why people would want to do this over the generic way below?

docs/typescript.md
```diff
@@ -119,7 +133,8 @@ const Image1 = styled('div')<ImageProps>({
 }));
 ```
 
-* The generic type version only works with object styles due to https://github.com/Microsoft/TypeScript/issues/11947.
+* The generic function version only works with object styles in TS <= 2.8 due to https://github.com/Microsoft/TypeScript/issues/11947.
+* If you use TS > 2.9, generic function will works with string styles too, but it will break VSCode syntax highlighting. See https://github.com/emotion-js/emotion/issues/721#issuecomment-396954993
```
mitchellhamilton: Could you replace `generic function will works with string styles too` with `generic functions will work with string styles too`?

docs/typescript.md
```diff
@@ -158,34 +173,57 @@ const App = () => (
...
+const Component0: SFC<ComponentProps0> = ({ label, className }) => (
   <div className={className}>{label}</div>
 )
 
 type StyledComponentProps = {
   bgColor: string
 };
 
-const StyledComponent0 = styled(Component)`
+const StyledComponent00 = styled<StyledComponentProps, ComponentProps0>(Component)`
```
mitchellhamilton: Why is a prop type passed here but in other places `div` is passed (shouldn't this be a component type)?

docs/typescript.md
```diff
@@ -71,17 +71,31 @@ const App = () => (
...
+`;
+
+const OtherLink = StyledText.withComponent('a');
+
 const App = () => <Link href="#">Click me</Link>
 
 // No errors!
 ```
 
+When you use `withComponent`, you should explicitly provide the type of props used to make style. If you don't, TS will think all props for `Text` are used to make style, and request user to give all those props when using the result component of `withComponent`.
```
mitchellhamilton: Could you elaborate on `TS will think all props for Text are used to make style`, I don't really understand what it means.

packages/create-emotion-styled/types/test.tsx
```diff
@@ -50,15 +50,15 @@ const TestFunComp1 = (props: TestFunProps1) => (
   </div>
 );
 
-const StyledClassComp0 = createStyled(TestClassComp)({
+const StyledClassComp0 = createStyled(TestClassComp)<{}>({
```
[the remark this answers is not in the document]
Ailrun: @mitchellhamilton Now, all props of component that passed to `createStyled` (or `styled`) considered as props for styling, so if you want to use `withComponent` and don't use any props of passed component (or use only some of them), you should restrict props type by providing type parameter.

Ref: https://github.com/emotion-js/emotion/pull/722#discussion_r195192495

docs/typescript.md
Ailrun: @mitchellhamilton Oh, thanks! I will change it.

docs/typescript.md
Ailrun: @mitchellhamilton Actually, I don't think this is the preferable way.... I add this because of compatibilities. If you think 'Document only for compatibility reason is not needed', I will change this.

docs/typescript.md
Ailrun: @mitchellhamilton You mean `generic function will work with string styles too`, right? I will fix them.

Ailrun: Oh, my mistake! I will fix it

docs/typescript.md
Ailrun: @mitchellhamilton This is for compatibilities, and, sadly, old typings did so...
https://github.com/emotion-js/emotion/blob/24fa6ed67314ae0b8d79276f7ff2815341f40caa/packages/react-emotion/types/index.d.ts#L101-L108

docs/typescript.md
Ailrun: @mitchellhamilton I mean `TS will consider all props for Text as props to make style`. Is this sentence more sensible?

docs/typescript.md
Ailrun: However, in case of string style, as you already know, situation is little different, since we cannot use generic template string in TS < 2.9, and even in TS >= 2.9, `vscode-styled-components` does not support syntax highlight.

docs/typescript.md
mitchellhamilton: No, I still don’t understand it. I don’t understand what you mean by `as props to make style`. how do the props `make` the style?

Ailrun: @mitchellhamilton I mean something like following
```
(props) => ({
  backgroundColor: props.bgColor
})
```
(`bgColor` props is used to make style)

docs/typescript.md
mitchellhamilton: Could you move it below the one that people should use and add a comment about it only being there for compatibility and that people shouldn’t write new code with it?

docs/typescript.md
Ailrun: For example,
1. When you have `Text` component that requires `bgColor`, `content`, you can make following styled component.
    ```
    const StyledText = styled(Text)({
      fontSize: '50px',
    }, props => ({
      backgroundColor: props.bgColor,
    }));
    ```
    In this `StyledText`, `bgColor` props is used to make style, but `content` isn't.
2. Now, suppose that you want to make a new styled component with different inner (base?) component `Link`, by using `withComponent`. Suppose that `Link` requires only `target` as a props. You probably do something like following.
    ```
    const StyledLink = StyledText.withComponent(Link);
    ```
3. However, the problem occurs in TypeScript. TypeScript cannot determine which props are used to make style, i.e., it cannot find `bgColor` is the only props used in style of `StyledText`. Therefore it should do one of following things.
    1. Suppose all props are used in style
    2. Suppose no props are used in style

    However, if you choose first one, TypeScript request to use `bgColor`, `content`, `target` to `StyledLink`, even if `Link` does not require `content`. This is why I choose second one, but legacy typing choose first one and some user complain about 'Why are there no props inference?'.

docs/typescript.md
Ailrun: OK, I will move this and add a comment.

docs/typescript.md
Ailrun: @mitchellhamilton Glamorous do something with second choice (not exactly same, but similar), and `styled-components` have something similar with first one, but with some wrong typed `withComponent`.

Ailrun:
@mitchellhamilton Could you check this again?

docs/typescript.md
Ailrun: @tkh44 I just changed docs. Could you check this again?

docs/typescript.md
```diff
@@ -71,17 +71,31 @@ const App = () => (
...
+}
+const Text = ({ className, text }: TextProps) => <div className={className}>{text}</div>
+const StyledText = styled<{}, TextProps>(Text)`
+  fontSize: 20px;
+  color: white;
+`;
+
+const OtherLink = StyledText.withComponent('a');
+
 const App = () => <Link href="#">Click me</Link>
```
tkh44: Should `Link` be `OtherLink` in this example?

Ailrun: We also have `Link`... Do I need to use all these components (`Link`, `OtherLink`, ...)?

docs/typescript.md
mitchellhamilton: Is this the preferred way to do it now?
If not, could you add a note saying that it's only for compatibility and tell people what they should use now.

Ailrun: @mitchellhamilton OK. I will add the note.

Ailrun: @mitchellhamilton
I notice that this is the preferred way.... There is no walkaround for this case, to be compatible with old one.
I will try better typing at upcoming emotion@10

Ailrun:
@mitchellhamilton Sorry for delay. I just came back from my vacation.

PR #725, opened 2018-06-14
create-emotion-server: refactor inline for performance

apapirovski:
The current version of renderStylesToString is highly inefficient as it allocates new arrays over and over, which isn't necessary. Instead just do a single pass hydrate which improves performance roughly 30x in most scenarios.

We're using emotion at Postmates and this was a significant bottleneck to our SSR. (This change, which we're currently floating, took most of our style hydrates from 25-30ms to sub-1ms.)

I tested this by comparing the output of several pages and both the new and old produce identical markup. It doesn't seem like the tests in the library itself are quite detailed enough.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/create-emotion-server/src/inline.js
+++ b/packages/create-emotion-server/src/inline.js
@@ -1,26 +1,13 @@
 // @flow
 import type { Emotion } from 'create-emotion'
 
-function toTag(
-  emotion: Emotion,
-  ids: Array<string>,
-  thing: { keys: Array<string> },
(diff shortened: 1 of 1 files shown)


apapirovski:
> would u be able to provide a simple benchmark for this? would be super cool to track this over time to prevent potential future regressions

If I get some time, I'll definitely make a PR for those in the future. 👍

Issue #693, opened 2018-05-29
CSS prop incompatible with typescript


cdock1029:
- `emotion` version: `9.1.3`
- `react` version: `16.4.0`
- `react-emotion`: `9.1.3`
- `react-app-rewire-emotion`: `4.0.0`
- `babel-plugin-emotion`: `9.1.2`
- `@types/react`: `16.3.4`
- `@types/react-dom`: `16.0.5`

Relevant code.

```javascript
class App extends Component {
  render() {
    return (
      <div
        className="App"
        css={`
          color: pink;
        `}
      >
        <p>hello world</p>
      </div>
    )
  }
}
```

**What you did:** 

try to compile `App.tsx`

**What happened:**  

Compilation error:
```
TS2339: Property 'css' does not exist on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'.
No lint errors found
Version: typescript 2.8.3, tslint 5.10.0
```

**Problem description:**

Typescript doesn't understand `css` prop.

Ailrun:
@cdock1029 Is this still relavant with 9.2.2?

cdock1029:
No this seems to be resolved with current version, thanks.

Ailrun:
Close this based on https://github.com/emotion-js/emotion/issues/693#issuecomment-396315083

aaronjensen:
@Ailrun afaict this seems to not work w/ preact-emotion (thanks for typing that btw!)

Ailrun:
@aaronjensen Oh, really? I need to check Preact typing again.

emattias:
I had to add `react-emotion` to the `types` array in my `tsconfig.json`: (it works with `emotion` and `create-emotion` aswell)

```
"types": ["node", "react-emotion"]
```

for the addition of the `css` prop/attribute to the `HTMLAttributes` interface to be included: https://github.com/emotion-js/emotion/blob/v9.2.8/packages/create-emotion/types/index.d.ts#L94-L98

karol-majewski:
This definition now lives in the core package — so the entry in `tsconfig.json` should say ` "@emotion/core"` instead of `"react-emotion"`.

https://github.com/emotion-js/emotion/blob/34031439e084d68fa610b8491163618352fa9b97/packages/core/types/index.d.ts#L84-L88

Kovensky:
You're augmenting the wrong thing (rather, the compiler's JSX support is weird in _a lot_ of ways).

[...]

PR #729, opened 2018-06-17
Add tests for css props

aaronjensen:
**What**:
This adds typescript tests for the css prop for both React and Preact. The
Preact tests are currently failing, so they'd need to be fixed before this can
be merged. See https://github.com/emotion-js/emotion/issues/693#issuecomment-397858379

**Why**:

The tests weren't there before and the css prop only worked with React.

**How**:

I'm not sure how to answer this question.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Code complete N/A


Ailrun:
@aaronjensen Thank you for your PR! Could you add tests for `create-emotion-styled` too?

aaronjensen:
@Ailrun done

packages/create-emotion-styled/types/common.d.ts
```diff
@@ -23,7 +23,7 @@ export type Themed<P extends object, T extends object> = P & { theme: T };
 
 export type StyledStatelessProps<P extends object, T extends object> =
   & P
-  & { theme?: T }
+  & { theme?: T, css?: CSSObject | string }
   ;
```
Ailrun: This will only work with StyledComponent. As far as I understand, you can use `css` props in any components. You can see how I added `css` props in https://github.com/emotion-js/emotion/blob/a8f4f8923e89aad7871ebcebd22e9c052197a2cb/packages/emotion/types/index.d.ts#L19-L23

Ailrun: Or does it work differently with preact-emotion/babel-plugin-emotion?

aaronjensen: Ah, ok, thanks. I'll try that w/ preact. Should I leave the above for `StyledComponent` as well?

Ailrun: @aaronjensen Oh, and as an answer to your question, as you can see in the https://emotion.sh/docs/css, it does not require anything but `emotion` and `babel-plugin-emotion`.

aaronjensen: @Ailrun updated

packages/create-emotion-styled/types/common.d.ts
```diff
@@ -23,7 +23,7 @@ export type Themed<P extends object, T extends object> = P & { theme: T };
 
 export type StyledStatelessProps<P extends object, T extends object> =
   & P
-  & { theme?: T }
+  & { theme?: T, css?: BaseInterpolation }
```
mitchellhamilton: This shouldn’t be here since the css prop isn’t specific to styled components

aaronjensen: I can remove it, but then the tests in `create-emotion-styled` fail unless I add `import 'emotion';`. Would you rather that or that I remove the tests?

aaronjensen: Updated w/ the tests in place, let me know if you want me to remove them.

aaronjensen: Done, as a separate commit. Let me know if you want me to squash them.

PR #730, opened 2018-06-17
No duplicated ast nodes

### LINKED ISSUE #726: Emotion Babel Macro duplicating nodes
- `emotion` version: 9.2.3
- `react` version: N/A

Relevant code.

```javascript
// typography.js
import { css } from 'emotion/macro'

const normal = css`
  font-weight: 400;
`

const demibold = css`
  font-weight: 600;
`

export { normal, demibold }

// compile.js
const fs = require('fs')
const src = fs.readFileSync('typography.js', 'utf-8')
const babel = require('babel-core')
const result = babel.transform(src, {
  filename: '/Users/matt/Projects/sandbox/j2/typo.js',
  plugins: [
    require('babel-plugin-macros'),
    require('@babel/plugin-transform-modules-commonjs'),
  ],
  babelrc: false,
  retainLines: true,
  sourceMaps: 'inline',
})

console.log(result.code)

// <result.code>
"use strict";Object.defineProperty(exports, "__esModule", { value: true });exports.demibold = exports.normal = void 0;var _emotion = require("emotion");

const normal = /*#__PURE__*/(0, _emotion.css)("font-weight:400;");exports.normal = normal;



const demibold = /*#__PURE__*/_css("font-weight:600;");exports.demibold = demibold;
```

What you did:

Used emotion/macro and babel-plugin-macros to transform code, together with plugin-transform-modules-commonjs

What happened:

The second reference to `_css` was not replaced with a live reference, causing the output to generate invalid javascript ("_css is not defined" error)

Reproduction:

See above code

Problem description:

Unlike the first reference to the macro import ("css"), the second reference is not replaced with a live reference by the commonjs module transform plugin. This is because when the `emotion/macro` runs, it is inserting duplicate nodes in the tree - and the module transform plugin is configured to skip nodes it has already visited.

Suggested solution:

Create a new node whenever inserting nodes into the AST, and as an extra verification step, use the `checkDuplicateNodes` helper:
https://github.com/babel/babel/blob/d383659ca6adec54b6054f77cdaa16da88e8a171/packages/babel-helper-transform-fixture-test-runner/src/index.js#L128

Andarist:
**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

fixes #726

### DIFF
--- a/packages/babel-plugin-emotion/test/util.js
+++ b/packages/babel-plugin-emotion/test/util.js
@@ -1,10 +1,11 @@
 // @flow
 /* eslint-env jest */
 import plugin from 'babel-plugin-emotion'
-import { transform as babel6Transform } from 'babel-core'
-import { transform as babel7Transform } from '@babel/core'
+import * as babel6 from 'babel-core'
+import * as babel7 from '@babel/core'
 import stage2 from 'babel-plugin-syntax-object-rest-spread'
 import makeCases from 'jest-in-case'
+import checkDuplicatedNodes from 'babel-check-duplicated-nodes'
 import { basename } from 'path'
 import * as fs from 'fs'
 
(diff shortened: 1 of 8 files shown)


packages/babel-plugin-emotion/test/util.js
```diff
@@ -25,38 +26,43 @@ jest.mock('fs')
 fs.existsSync.mockReturnValue(true)
 fs.statSync.mockReturnValue({ isFile: () => false })
 
-const createInlineTester = transform => opts => {
+const isBabel7 = babel => parseInt(babel.version.split('.')[0], 10) === 7
```
Andarist: babel@6 duplicates some nodes, so we just want to validate this when using babel@7 to avoid false positives causes by babel itself

Andarist:
@mitchellhamilton done

Andarist:
@mitchellhamilton as to the registry change - I couldn't install `@babel/preset-env` yesterday (failing on different packages) from yarn registry. It also seems that yarn's registry might be deprecated in the future - https://github.com/yarnpkg/yarn/issues/5891

PR #731, opened 2018-06-18
Fix jest-emotion for preact

aaronjensen:
**What**:

Preact, when rendered by preact-render-to-json does not always have props, so we
need to account for this.

**Why**:

So it works.

**How**:

Assume no classes if there are no props.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/jest-emotion/src/utils.js
+++ b/packages/jest-emotion/src/utils.js
@@ -4,8 +4,7 @@ function getClassNames(selectors, classes) {
   return classes ? selectors.concat(classes.split(' ')) : selectors
 }
 
-function getClassNamesFromTestRenderer(selectors, node) {
-  const props = node.props
+function getClassNamesFromTestRenderer(selectors, { props = {} }) {
   return getClassNames(selectors, props.className || props.class)
 }
 
--- a/packages/jest-emotion/test/preact.test.js
+++ b/packages/jest-emotion/test/preact.test.js
@@ -1,0 +1,44 @@
+/* eslint-disable react/no-unknown-property */
+import { h } from 'preact'
+import render from 'preact-render-to-json'
+import prettyFormat from 'pretty-format'
+import * as emotion from 'emotion'
+import { createSerializer } from '../src'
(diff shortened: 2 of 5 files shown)


Andarist:
Could u add a test for this?

aaronjensen:
Probably not easily. I don’t have time to do it right now. I mentioned this in the PR

aaronjensen:
@Andarist just added a test.

packages/jest-emotion/test/preact.test.js
```diff
@@ -0,0 +1,43 @@
...
+  const divStyle = emotion.css`
+    color: red;
+  `
+  const svgStyle = emotion.css`
+    width: 100%;
+  `
+
+  it('replaces class names and inserts styles into preact test component snapshots', () => {
+    const tree = render(
+      <div class={divStyle}>
```
Andarist: is `preact` accepting `class` instead of `className`?

There is a linting problem found in this file:
> Unknown property 'class' found, use 'className' instead  react/no-unknown-property

If `class` is correct in `preact` you can just disable this rule for this file (with top file comment or smth)

aaronjensen: Thanks. Fixed. And yes, `preact` takes all the html style instead of js style (`class` and `for`, for example)

packages/jest-emotion/src/utils.js
```diff
@@ -6,7 +6,7 @@ function getClassNames(selectors, classes) {
 
 function getClassNamesFromTestRenderer(selectors, node) {
   const props = node.props
-  return getClassNames(selectors, props.className || props.class)
+  return props ? getClassNames(selectors, props.className || props.class) : []
```
mitchellhamilton: I think it should be `: selectors` instead of `: []`

aaronjensen: Nice catch, fixed.

tkh44:
Thanks for the PR @aaronjensen

PR #739, opened 2018-06-20
styled: fix order of static class names

apapirovski:
**What**:
Make the order of static class names match the desired style tag order.

**Why**:
While the extractStatic doesn't suffer from this problem, we're doing something similar in our SSR and since the SSR inlines style tags based on the order of classNames, this makes it impossible to implement this optimization.

**How**:
Flipped the order of staticClassName within the component.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

This is a fairly straightforward change but allows us to implement something that caches component styles for `styled` instead of having to render & hash them on each render. The latter is computationally expensive and accounts for good 20-40ms runtime (in production) for some of our more complex pages.

I would also consider the current behaviour a bug, since I would expect the order of classnames form the following to be reversed.

```
const H1 = styled.h1`color: blue;`;
const RedH1 styled(H1)`color: red;`;
```

What I will get is `className=RedH1 H1` rather than `className=H1 RedH1`.

### DIFF
--- a/packages/create-emotion-styled/src/index.js
+++ b/packages/create-emotion-styled/src/index.js
@@ -121,11 +121,14 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
               styles.concat(classInterpolations)
             )
           } else {
-            className += staticClassName
+            className = `${staticClassName} ${className}`
           }
 
           if (stableClassName !== undefined) {
-            className += ` ${stableClassName}`
+            if (staticClassName === undefined) {
+              className += ' '
+            }
+            className += stableClassName
           }
 
           return view.createElement(
--- a/packages/emotion/test/extract/__snapshots__/extract.test.js.snap
+++ b/packages/emotion/test/extract/__snapshots__/extract.test.js.snap
@@ -10,7 +10,7 @@ exports[`styled basic render nested 1`] = `
 
 exports[`styled className prop on styled 1`] = `
 <h1
-  className="some-class emotion-0 emotion-1"
+  className="emotion-0 some-class emotion-1"
 >
   hello world
 </h1>
--- a/packages/react-emotion/test/__snapshots__/index.test.js.snap
+++ b/packages/react-emotion/test/__snapshots__/index.test.js.snap
@@ -1,5 +1,21 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`should append classNames in the correct order 1`] = `
+.emotion-0 {
+  color: blue;
+}
+
+.emotion-1 {
+  color: red;
+}
+
+<h1
+  className="emotion-0 emotion-1 emotion-2 emotion-3"
+>
+  Test
+</h1>
+`;
+
 exports[`styled basic render 1`] = `
 .emotion-0 {
   color: blue;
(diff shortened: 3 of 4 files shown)


apapirovski:
Also, if there's any chance, it would be really nice to get a release sometime after this lands — we're hoping to stop rolling our own custom renderer for https://github.com/emotion-js/emotion/commit/0847bd2ba1f2c79beaa5340f8deaeb0ea657e6d6.

packages/create-emotion-styled/src/index.js
```diff
@@ -121,11 +121,14 @@ function createEmotionStyled(emotion: Emotion, view: ReactType) {
               styles.concat(classInterpolations)
             )
           } else {
-            className += staticClassName
+            className = `${staticClassName} ${className}`
           }
 
           if (stableClassName !== undefined) {
-            className += ` ${stableClassName}`
+            if (staticClassName === undefined) {
```
Andarist: I'll see why other line got changed, but could u explain this one? is it just to prevent unnecessary space being added? does it cause any harm?

Andarist:
Would be great to cover both of those changes with simple dedicated snapshot tests.

apapirovski:
@Andarist I've updated this with a snapshot test that will fail on master.

mitchellhamilton:
Could you explain more about your use case and how this change fixes it?

I wouldn’t consider the current behaviour a bug since in regular css, the order of class names doesn’t matter.

apapirovski:
@mitchellhamilton The inline.js in server-side rendering (which exports `renderStylesToString`) uses a RegExp to match `css-` classNames and then inserts style tags in the order that those classNames appear. The ordering for `staticClassName` is incorrect, it needs to be inverted as I do in this PR.

mitchellhamilton:
I get that but one of the most important parts of emotion is that we don’t guarantee the order of css across css rules.

```jsx
import { css } from 'emotion'

let firstClass = css`
  color: hotpink;
`

let secondClass = css`
  color: green;
`

render(<div className={`${firstClass} ${secondClass}`} />)
```
In this example, you would expect it to be green and in most cases it will be but we can’t guarantee that otherwise we couldn’t cache anything at all.


```jsx
import { css } from 'emotion'

let classThatIsBeingUsedSomewhereElse = css`
  color: green;
`

let firstClass = css`
  color: hotpink;
`

let secondClass = css`
  color: green;
`

render(<div className={`${firstClass} ${secondClass}`} />)
```
In this example, it won’t be green since the green css is inserted before the hotpink css.

So ultimately, this change won’t fix anything, just make the problem occur less common and encourage people to do something that we don’t recommend.

This is why we strongly encourage [composition](https://emotion.sh/docs/composition).

apapirovski:
Ok, I ended up tracing this issue to `renderStylesToString`. There was a small regression in the code that I provided last time. I'm making a PR that fixes it.

artemirq:
```
import { css } from 'emotion';

const a0 = css`
    background-color: red;
`;

const a1 = css`
    ${a0};
    color: white;
`;
```

**a0** = `css-gxfs0h`
**a1** = `css-1ag5384`

@mitchellhamilton @Andarist There appears question: Is it bad idea to reuse class `css-gxfs0h`? Why emotion create new class with the same styles? (most likely this question arose many times, but I did not find an answer to it, sorry)

PR #740, opened 2018-06-20
ssr: fix insertion order in edge cases

apapirovski:
This fixes a very minor and unlikely to happen regression in `renderStylesToString` where the inserted elements aren't re-inserted in the correct order. This still retains the performance profile of the original changes and keeps the algorithm O(n).

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete

### DIFF
--- a/packages/create-emotion/test/inline.test.js
+++ b/packages/create-emotion/test/inline.test.js
@@ -42,6 +42,15 @@ describe('renderStylesToString', () => {
       )
     ).toMatchSnapshot()
   })
+  test('retains the original style order', () => {
+    const style1 = emotion.css`color: blue;`
+    const style2 = emotion.css`color: red;`
+    expect(
+      emotionServer.renderStylesToString(
+        renderToString(<div className={`${style2} ${style1}`} />)
+      )
+    ).toMatchSnapshot()
+  })
 })
 describe('hydration', () => {
   afterAll(() => {
--- a/packages/create-emotion-server/src/inline.js
+++ b/packages/create-emotion-server/src/inline.js
@@ -19,11 +19,13 @@ const createRenderStylesToString = (emotion: Emotion, nonceString: string) => (
   const regex = new RegExp(`<|${cssKey}-([a-zA-Z0-9-]+)`, 'gm')
 
   const seen = {}
+  const order = {}
 
   let result = ''
   let globalIds = ''
   let globalStyles = ''
 
+  let i = 0
   for (const id in inserted) {
     if (inserted.hasOwnProperty(id)) {
       const style = inserted[id]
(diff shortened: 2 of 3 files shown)


apapirovski:
@mitchellhamilton Been a bit backlogged but will do this weekend.

apapirovski:
@mitchellhamilton I've added a test now. Sorry for the delay.

packages/create-emotion/test/inline.test.js
```diff
@@ -42,6 +42,15 @@ describe('renderStylesToString', () => {
       )
     ).toMatchSnapshot()
   })
+  test('retains the original style order', () => {
+    const style1 = emotion.css`color: blue;`
+    const style2 = emotion.css`color: red;`
+    expect(
+      emotionServer.renderStylesToString(
+        renderToString(<div className={`${style2} ${style1}`} />)
```
mitchellhamilton: I don't think this is something we should guarantee since we can't guarantee it in all cases and we encourage [composition](https://emotion.sh/docs/composition) over relying on insertion order.

For example, if this was changed to something like this then style2 override style1
```jsx
<Fragment>
  <div className={style1} />
  <div className={style2} />
  <div className={`${style2} ${style1}`} /> {/* blue */}
</Fragment>
```
It would also break without SSR if something like this was done.
```jsx
// in some other random place
css`color: red;`

const style1 = css`color: blue;`
const style2 = css`color: red;` 
render(<div className={`${style2} ${style1}`} />) // blue
```
whereas using `cx` would always work.
```jsx
// in some other random place
css`color: red;`

const style1 = css`color: blue;`
const style2 = css`color: red;` 
render(<div className={cx((style1, style2)} />) // red
```

apapirovski: I mean, this just verifies that the new algorithm generates the exact same output as the old one.

apapirovski: Also, this is in part why I didn't create a test case originally. It tests unsupported behaviour but IMO it's important that the two versions match.

mitchellhamilton: I don’t get the static class names bit, are you referring to extractStatic? If so then how is it working with extractCritical?

I wouldn’t call this a breaking change, react@16.4 made bugs surface more often with getDerivedStateFromProps and this is just like that.
More generally, what is and isn’t a breaking change can’t be strictly defined, for example what if someone read toString from a function and relied on the value and the value changed, while doing someone like that is strongly discouraged, so is relying on insertion order in emotion.

apapirovski: Fine by me. We don't rely on this at Postmates so I don't really care. Figured we would want the algorithm to match 1 to 1 to what it used to be but your call.

(FWIW it's possible I'm missing other edge cases where this can happen so my preference would still be to land this.)

mitchellhamilton: I don't think the algorithm really needs to match perfectly, since if you're relying on this behavior, your code can already break so I think I'm gonna close this but thanks for the PR!

Issue #728, opened 2018-06-17
autoLabel isn't useful with grouped components and index.js files


mherodev:
- `emotion` version: ^9.2.3
- `react` version: ^16.4.0

Relevant code.

```javascript
["emotion", { "sourceMap": true, "autoLabel": true, "labelFormat": "[filename]-[local]" }]
```

What you did:

Our pattern for handling styled components is to group them inside an `Elements` variable. In addition, our project structure is largely made up of folders and `index.js` files.

e.g.

Inside `components/Header/index.js` we might have the following styled components:

```js
const Elements = {
  LiHeaderMenuItem: styled.li(props => ({
    display: 'inline-block',
    margin: '0 10px',
    opacity: props.isActive ? '1' : '.5',
  })),
  DivDesktopHeader: styled.div(({ theme }) => ({
    display: 'none',
    [theme.breakpoints.m]: {
      display: 'block',
      background: 'url(/dist/images/logo.png)',
    },
  })),
};
```

What happened:

The result is that `autoLabel` names every single class in the following format:

```html
class="css-1108j53-index-Elements"
```

Instead of naming them uniquely, e.g.

```html
class="css-1108j53-Header-LiHeaderMenuItem"
```

Suggested solution:

It seems clear the `autoLabel` feature needs to be expanded to handle these two cases:

1. In the case of an `index.js` filename, it should use the parent folder name (possibly as an option).
2. It should detect when the local variable is not equal to the component itself, and then instead report the property name the object lives on inside that object.

Thanks!

PR #741, opened 2018-06-21
Improve labels and upgrade @emotion/stylis

### LINKED ISSUE #727: autoLabel doesn't work for styles assigned to component's static properties
- `emotion` version: 9.2.3
- `react` version: 16.4.0
- `TypeScript`, `babel-plugin-emotion`

If I assign class name created by css() function or component created by styled() function to React component's static properties, the auto labels are not generated for them. For example for this MyEditor component:
```
class MyEditor extends React.Component {
render() {
  const E = MyEditor;
  return (
    <E.Container>
     <div className={E.itemCss} />
    </E.Container>)
}

static Container = styled.div({
    display: "flex",
    flexDirection: "column"
    justifyContent: "center",
})

static itemCss = css({
    flex: "1 1 auto",
})
}
```
no label is appended to the itemCss class name and some internal name is appended to the Container component's class name (eg `<div class="css-h5haei ebf56m67"></div>`). It would be great if the labels were set to the respective static properties names.

The autoLabel option is set to true in my .babelrc config and the auto labeling works well in other cases.

mitchellhamilton:
**What**:

**Why**:

Closes #727 and part of #728

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/test/__snapshots__/css-prop.test.js.snap
+++ b/packages/babel-plugin-emotion/test/__snapshots__/css-prop.test.js.snap
@@ -120,7 +120,7 @@ exports[`babel css prop inline babel 6 label in higher order component 1`] = `
 
 const foo = W => class extends Component {
   render() {
-    return <div className={/*#__PURE__*/_css(\\"color:brown;label:foo(Component);\\")}>Hello</div>;
+    return <div className={/*#__PURE__*/_css(\\"color:brown;label:foo;\\")}>Hello</div>;
   }
 };"
 `;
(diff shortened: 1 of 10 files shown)


packages/babel-plugin-emotion/test/__snapshots__/css-prop.test.js.snap
```diff
@@ -288,7 +288,7 @@ const foo = W => class extends Component {
   render() {
     return <div className={
     /*#__PURE__*/
-    _css(\\"color:brown;label:foo(Component);\\")}>Hello</div>;
```
mitchellhamilton: This wouldn't have broken before since class names can't have brackets

PR #747, opened 2018-06-30
Change build system and update Jest

mitchellhamilton:
**What**:
This changes the build system from some npm scripts to some node scripts that run rollup on every package. This is extracted from https://github.com/emotion-js/next

This also moves the site out of the packages folder into the root directory because I wanted to make the packages folder only contain packages that are published to npm. The benchmarks package was also moved to the new scripts folder for the same reason.

**Why**:

- Faster builds since a new process doesn't need to be started for each build
- Everything runs through the same build system

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete


packages/babel-plugin-emotion/lib/macro-styled.js
```diff
@@ -0,0 +1 @@
+module.exports = require('babel-plugin-emotion').macros.styled
```
Andarist: whats the reason behind those 2 "proxy" files?

packages/create-emotion-styled/package.json
```diff
@@ -26,10 +22,7 @@
   "devDependencies": {
     "@types/react": "16.0.16",
     "dtslint": "^0.3.0",
-    "npm-run-all": "^4.0.2",
-    "prop-types": "^15.6.1",
-    "rimraf": "^2.6.1",
-    "rollup": "^0.60.1"
+    "prop-types": "^15.6.1"
```
Andarist: seems weird to have prop-types in devDeps, are they needed here?

mitchellhamilton: It was already there. It's probably not strictly necessary but it was probably added there because it's a peerDependency and it's needed in dev. If you think it should be removed them I'll remove it but IMO, it doesn't really matter.

scripts/build/add-basic-constructor-to-react-component.js
```diff
@@ -0,0 +1,40 @@
+// this is to avoid the spread argument helper for react component constructors
+module.exports = function(babel) {
+  const { types: t } = babel
+
+  return {
+    name: 'ast-transform', // not required
+    visitor: {
+      Class(path) {
+        if (
```
Andarist: could be made more readable with `buildMatchMemberExpression` & I would extract `hasConstructor` too
```js
const isReactComponent = t.buildMatchMemberExpression('React.Component')

// ...

if (isReactComponent(path.node.superClass) && hasConstructor(path.node.body.body)) {
  // ...
}
```

scripts/build/add-basic-constructor-to-react-component.js
```diff
@@ -0,0 +1,40 @@
...
+          path.node.body.body.length &&
+          !path.node.body.body.some(node => {
+            return (
+              t.isClassMethod(node) &&
+              t.isIdentifier(node.key) &&
+              node.key.name === 'constructor'
+            )
+          })
+        ) {
+          path.node.body.body.unshift(
```
Andarist: lately i've become fan of `template`, IMHO it increases readability a lot - although not sure now if it will be possible to create a class method with it - we could always create dummy class with constructor only and just unpack constructor node though

scripts/build/add-basic-constructor-to-react-component.js
```diff
@@ -0,0 +1,40 @@
+// this is to avoid the spread argument helper for react component constructors
```
Andarist: could u explain what exactly is avoided with this and why this is a problem?

packages/babel-plugin-emotion/lib/macro-styled.js
Andarist: gotcha, maybe add a comment that this can be removed with next major?

packages/create-emotion-styled/package.json
Andarist: peer dep thing makes sense 👍

Andarist:
Quite a lengthy PR 😉 have to go to sleep now, but gonna finish reviewing in few hours

scripts/build/add-basic-constructor-to-react-component.js
mitchellhamilton: https://runkit.com/mitchellhamilton/5b370eeaf5b34c0012874d92
Without this, the constructor's arguments will be iterated over and put in an array unnecessarily.

Andarist: nice! this would be useful as part of `preset-react`, gonna propose it later to rest of the babel team

scripts/build/add-basic-constructor-to-react-component.js
```diff
@@ -0,0 +1,37 @@
...
+        t.isIdentifier(node.key) &&
+        node.key.name === 'constructor'
+      )
+    })
+  return {
+    name: 'ast-transform', // not required
+    visitor: {
+      Class(path) {
+        if (
+          isReactComponent(path.node.superClass) &&
```
Andarist: not yet sure how & where this is used, maybe it doesnt matter in current context - but would be good to support `PureComponent` in `isReactComponent` function

scripts/build/add-basic-constructor-to-react-component.js
mitchellhamilton: BTW, this implementation isn't perfect since it should technically pass the second argument to super as well (and maybe the third but I don't know if that's strictly necessary). The reason I didn't do that here is that I know I won't access context in those cases.

https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L21-L29

scripts/build/change-packages.js
```diff
@@ -0,0 +1,45 @@
...
+
+  await Promise.all(
+    packages.map(async ({ pkg, path: pkgPath }) => {
+      // you can transform the package.json contents here
+      let devDeps = pkg.devDependencies
+      if (pkg.devDependencies) {
+        delete devDeps['npm-run-all']
+        delete devDeps['cross-env']
+        delete devDeps['babel-cli']
+        delete devDeps['rollup']
```
Andarist: this should be accessed with dot notation for consistency

scripts/build/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "build",
+  "version": "0.5.7",
+  "private": true,
+  "dependencies": {
+    "chalk": "^2.3.2",
+    "del": "^3.0.0",
+    "rollup": "^0.62.0",
+    "rollup-plugin-alias": "^1.4.0",
+    "rollup-plugin-babel": "^4.0.0-beta.5",
```
Andarist: it's safer to pin beta packages to exact versions, also - this could get upgraded, there is `beta.7` already

scripts/build/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "build",
+  "version": "0.5.7",
```
Andarist: does it make sense to version this?

scripts/build/inline-isBrowser.js
```diff
@@ -0,0 +1,56 @@
...
+              state,
+              babel
+            })
+          }
+          return false
+        })
+
+        if (!path.node.specifiers.length) {
+          path.remove()
+        }
```
Andarist: as this is used as `filter` predicate would be good to return explicit boolean here

scripts/build/rollup.config.js
```diff
@@ -0,0 +1,108 @@
...
+          '@babel/flow'
+        ],
+        plugins: [
+          '@babel/plugin-transform-flow-strip-types',
+          require('./add-basic-constructor-to-react-component'),
+          'codegen',
+          ['@babel/proposal-class-properties', { loose: true }],
+          require('./fix-dce-for-classes-with-statics'),
+          isBrowser && require('./inline-isBrowser'),
+          isBrowser &&
```
Andarist: isnt this duplicating what `inline-isBrowser.js` do (except filtering imported specifiers)? why there are 2 similar plugins?

scripts/build/rollup.config.js
```diff
@@ -0,0 +1,108 @@
...
+                  },
+                  ReferencedIdentifier(path, node) {
+                    if (path.node.name === 'shouldSerializeToReactTree') {
+                      path.replaceWith(t.booleanLiteral(false))
+                    }
+                  }
+                }
+              }
+            }),
+          '@babel/plugin-proposal-object-rest-spread'
```
Andarist: might be a good moment to update babel@7 deps too and use `loose` for this plugin

scripts/build/utils.js
```diff
@@ -0,0 +1,98 @@
+const path = require('path')
+const del = require('del')
```
Andarist: seems like both `rimraf` and `del` are used across the codebase - not a big deal, but could be unified

scripts/build/change-packages.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
Andarist: is this file used? cant find the usage in this PR, I also don't quite get why this is needed - especially when you have removed those devDeps etc in this PR

or is it a helper script which u have used to modify packages in this PR + it can be used to modify them in the future? if so it would be great to add a comment about this

scripts/build/add-basic-constructor-to-react-component.js
Andarist: no idea how `updater` arg might be used 🤔  we could do `super(props, context, updater)` by default + an option to optimize it to just `super(props)`

Not sure if `context` here is only for legacy purposes, or is the new one also passed here - anyway doesn't matter as long as this arg is here

scripts/build/package.json
mitchellhamilton: I think some tools(I _think_ it's lerna and yarn but I'm not totally sure) complain when there isn't a version so since it doesn't really hurt anything, we might as well leave it.

scripts/build/inline-isBrowser.js
mitchellhamilton: That part isn't in the filter predicate.

scripts/build/utils.js
mitchellhamilton: I think the usages of rimraf were only in npm scripts and I removed those ones so I don't think rimraf is being used anymore.

scripts/build/change-packages.js
mitchellhamilton: It's a helper script that can be used to modify the packages in the future, I'll add a comment.

scripts/build/rollup.config.js
mitchellhamilton: `inline-isBrowser.js` is replacing the usages of `@emotion/utils` in packages whereas this plugin is changing the declarations inside of `@emotion/utils`

scripts/build/inline-isBrowser.js
Andarist: oh, right, my bad 👍

scripts/build/rollup.config.js
Andarist: could we maybe add a filename guard to this second plugin so it wont run accidentally on other packages? I think there are some nice new options in babel@7 - like `overrides` etc, maybe they could be leveraged to add this plugin only for certain filepaths/directories, this would make the intent more explicit

Andarist:
The whole idea & setup looks way better than previous one 👍 

I feel though that monorepo setup should be something "on top" of regular git repository - it's job is to ease dependency management & local development. IMHO even if it's kinda artificial problem most of the time I always strive for separating packages, their tooling etc - what I mean by that is that having a monorepo script i.e. to build stuff is a good thing for development but even having such it would be good to embed build tooling + devDeps etc in each package so it can be easily extracted from the monorepo and moved away. Proposed build setup could be packaged and handle both monorepo and single package use cases. I understand it's not really important right now so I propose to screw that for now, but create an issue about this once this PR gets merged.

scripts/build/utils.js
```diff
@@ -0,0 +1,96 @@
...
+    ret.configs = []
+    if (ret.pkg.main && !ret.pkg.main.includes('src')) {
+      ret.configs.push({
+        config: makeRollupConfig(ret),
+        outputConfigs: getOutputConfigs(ret)
+      })
+    }
+    if (ret.pkg['umd:main']) {
+      ret.configs.push({
+        config: makeRollupConfig(ret, true, true),
```
Andarist: would prefer changing signature to:
```
makeRollupConfig(data, { isBrowser = false, isUmd = false } = {})
```

right now looking at this call site it's hard to tell what those booleans do

Andarist: not yet sure about this but - I can't find a package using "browser" field, so not sure how exactly we need to handle it, but a good alternative to transforming packages with babel and replacing stuff with browser values it might be clearer to introduce `BROWSER` env variable, produce code conditionally with `process.env` ternaries in the actual code and prepare 2 builds with appropriate `rollup-plugin-replace` setup

Andarist:
Continuing https://github.com/emotion-js/emotion/pull/747#discussion_r199341939

Should be possible to prepare multiple "source" directories - with only process.env replaced in them and build final bundles from them + ofc redirecting types to appropriate directories. It would be quite unconventional setup and slightly more complicated, but IMHO would be clearer in the actual source what are the possible code shapes, right now the babel plugin idea is hidden and can be broken more easily.

mitchellhamilton:
I was thinking of using plain globals but using `process.env` would probably just work since flow knows about `process.env`.

The other problem though is that the second variable `shouldSerializeToReactTree`, also depends on `process.env.NODE_ENV` when it's not in the browser build so either the check for that would have to be inlined or imported from `@emotion/utils` which is just an extra thing to deal with. 

TBH, I think that even though it's not perfect, I'd rather just leave it right now since it's not getting used on the current packages here yet and it works.

TrySound:
Would be good to change `es` suffix to `esm` which is more descriptive like [esm loader](https://github.com/standard-things/esm). The latest version of rollup also has `esm` alias for format which will replace `es` in the future. [material-ui uses `es`](https://unpkg.com/@material-ui/core@1.3.0/es/AppBar/AppBar.js) to ship es6 (not only modules) syntax.

mitchellhamilton:
Thanks for your review @Andarist!!

Andarist:
Not a problem, it was a cool PR to review! 😃

Issue #746, opened 2018-06-29
Enhance typings for ClassNameArg in cx


tomshane:
- `emotion` version: 9.2.4
- `react` version: 16.4.1

Relevant code.

```javascript
const trueBool = true;
const undefBool: boolean | undefined;
const text = "hello";
const emptyText: string | undefined;
const customType: "value1" | "value2" | "value3" | undefined = "value1";

const classes = cx(
  { "cls1": trueBool },           // works
  { "cls2": undefBool },          // doesn't work, should ignore the class
  { "cls3": text },               // doesn't work, should apply the class
  { "cls4": emptyText }           // doesn't work, should ignore the class
  { "cls5": customType }          // doesn't work, should apply the class
);
``` 

Problem description:

Please note, _"doesn't work_" actually doesn't mean it is not working in run-time, just TS displays error, because of types definition. In run-time everything seems to work as expected.

The problem is in `ClassNameArg` type, where the "object" part is defined as` { [key: string]: boolean }`.

Suggested solution:

The easiest solution probably would be to change typing to ` { [key: string]: any }`, but I'm not sure everyone would be fine with it.

Ailrun:
@tomshane Thank you for reporting this! Introducing `any` always comes with worries, but for this case, IMO, `{ [key: string]: any }` looks like the correct type.
Could you make a PR for this?

tomshane:
I'm sorry, I don't have git environment up and running on my current machine. I could do it when I'm back from holiday, though.

Kovensky:
At least DefinitedlyTyped generally prefers to keep `boolean` for things that semantically are boolean even though other falsy values are accepted for falsehood. Your code can be fixed by just prefixing your non-boolean value with `!!`.

PR #751, opened 2018-07-06
classnames also accept undefined as value

pasix:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


packages/create-emotion/types/index.d.ts
```diff
@@ -32,7 +32,7 @@ export interface ArrayClassNameArg extends Array<ClassNameArg> {}
 
 export type ClassNameArg =
   | undefined | null | boolean | string
-  | { [key: string]: boolean }
+  | { [key: string]: boolean | undefined }
```
tkh44: Would this need to support all of these types as values?

Ailrun: We may need `any`, as described in #746.

Andarist: while `any` is technically correct I doubt we should actually recommend it, typings give us power to restrict usages to sane ones

pasix: yes it's better with `undefined | null | boolean | string`

osi-oswald:
Why was type `number` not added as well? Consider the following usecase:
```
cx('count', { isOdd: count % 2 }) // TS complains, emotion handles it just fine
```
I now have to type one of the following just to make TS happy again (which I find unnecessary):
```
cx('count', { Boolean(isOdd: count % 2) })
cx('count', { !!(isOdd: count % 2) })
```
Actually type `any` would really be the best type for it. Otherwise someone might come and ask "why did you not add type XXX as well"?

Andarist:
Imho relaying on auto bool coercion for numbers there is not the best way to do. While it ofc works in JavaScript, type systems should strive to be more explicit about such things and I would consider the current lack of number support a good thing.

Issue #755, opened 2018-07-07
Adding support for styling and rendering components on VR platform


nitin42:
- `emotion` version:
- `react` version:

Relevant code.

```javascript

```

What you did:

What happened:

Reproduction:

Problem description:

Using `@emotion/primitives` with `react-360` to style VR apps across the platforms. But, `@emotion/primitives` depends on `react-primitives` which when installed in `react-360` project, gives an error that

```
couldn't resolve module react-vr
```

But when we try to install `react-vr`, it conflicts with `react-360`.

Suggested solution:

We can detach the `css` function from `@emotion/primitives` and create a separate package for it, and then use it according to the environment.

Something like this:

```js
// Emotion primitives

import { createCssFunction } from 'emotion-primitive-css'

import { StyleSheet } from 'react-primitives'

// This will use StyleSheet method for styling components in react-native, Web and Sketch
const css = createCssFunction(StyleSheet)

export default css
```

```js
// Emotion VR

import { createCssFunction } from 'emotion-primitive-css'

import { StyleSheet } from 'react-360'

// This will use StyleSheet method for styling components in react-360
const css = createCssFunction(StyleSheet)

export default css
```

mitchellhamilton:
I'm good with adding something like this though I'd rather call it `createCss` and have a `createStyled` function like this and call the package `@emotion/primitives-core`
```jsx
// @flow
import * as React from 'react'

export function createCss(StyleSheet) {
  // ...
}

type Options = {
  getShouldForwardProp: (
    Component: React.ElementType
  ) => (prop: string) => boolean
}

let defaultPickTest = prop => prop !== 'theme' && prop !== 'innerRef'
export function crateStyled(
  StyleSheet,
  { getShouldForwardProp = () => defaultPickTest }: Options = {}
) {
  // ...
}

```
And then it can be used in `@emotion/primitives` like this
```jsx
// @flow
import * as React from 'react'
import { View, Image, Text, StyleSheet } from 'react-primitives'
import {
  testPickPropsOnPrimitiveComponent,
  testPickPropsOnOtherComponent
} from './test-props'
import { createStyled, createCss } from '@emotion/primitives-core'

function getShouldForwardProp(component: React.ElementType) {
  switch (component) {
    case View:
    case Text:
    case Image: {
      return testPickPropsOnPrimitiveComponent
    }
  }
  return testPickPropsOnOtherComponent
}

let styled = createStyled(StyleSheet, { getShouldForwardProp })

[...]

PR #759, opened 2018-07-09
Emotion Native

nitin42:
**What**: Styling Native components using emotion and a new `primitives-core` package for shared utils.

**Why**: 

**How**:  This PR adds support for styling React Native components using emotion, adds a new package `primitives-core` for primitive shared utils and also refactors the `primitives` package.

**Checklist**:

- ✅ Documentation
- ✅ Tests
- ✅ Code complete

 

Related #755


nitin42:
Almost done! Just need some work on flow types.

nitin42:
I don't know why there are some flow types error in other packages (I haven't touch any). Plus, typescript tests seems to be broken.

packages/primitives-core/package.json
```diff
@@ -0,0 +1,26 @@
+{
+	"name": "primitives-core",
```
mitchellhamilton: This should be `@emotion/primitives-core` and it should be imported as `@emotion/primitives-core`

packages/primitives/package.json
```diff
@@ -3,14 +3,10 @@
...
-	"files": [
-		"src",
-		"dist"
-	],
+	"files": ["src", "dist"],
 	"dependencies": {
-		"@emotion/is-prop-valid": "^0.6.3",
 		"babel-plugin-emotion": "^9.2.5",
-		"css-to-react-native": "^2.2.0"
+		"primitives-core": "1.0.0-beta"
```
mitchellhamilton: `@emotion/primitives-core`

packages/primitives/test/css.test.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import { css } from '@emotion/primitives'
+import { css } from '../src'
```
mitchellhamilton: This should be `@emotion/primitives`, all package names are aliased to their src.

packages/primitives/test/emotion-primitives.test.js
```diff
@@ -8,7 +8,7 @@ import reactPrimitives from 'react-primitives'
 import { ThemeProvider } from 'emotion-theming'
 import { render, unmountComponentAtNode } from 'react-dom'
 
-import styled from '@emotion/primitives'
+import styled from '../src'
```
mitchellhamilton: Same as above

packages/primitives/test/no-babel/basic.test.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import '../mock-primitives'
 import * as React from 'react'
-import styled, { css } from '@emotion/primitives'
+import styled, { css } from '../../src'
```
mitchellhamilton: Same as above

packages/native/package.json
```diff
@@ -0,0 +1,23 @@
+{
+	"name": "emotion-native",
```
mitchellhamilton: `@emotion/native`

packages/native/src/index.js
```diff
@@ -0,0 +1,28 @@
+import reactNative from 'react-native'
+import { createCss } from '@emotion/primitives-core'
+
+import { styled } from './styled'
+
+const css = createCss(reactNative.StyleSheet)
+
+const components = `ActivityIndicator ActivityIndicatorIOS ART Button DatePickerIOS DrawerLayoutAndroid
```
mitchellhamilton: Could you make this into an array since it's easier to read and there's no need to split

packages/native/test/native-css.test.js
```diff
@@ -0,0 +1,115 @@
+import { css } from '@emotion/native'
+
+let returnArguments = (...args) => args
+
+describe('Emotion native css', () => {
```
mitchellhamilton: I'm not really sure if we need all these tests duplicated, maybe it could just be a small smoke test to make sure css works

packages/primitives-core/README.md
```diff
@@ -0,0 +1,39 @@
+# @emotion/primitives-core
+
+> Shared utilities for emotion primitives and native
+
+## Introduction
+
+This package contains some shared utilities which are consumed by `@emotion/primitives` and `@emotion/native` for styling and rendering components.
```
mitchellhamilton: Could you add a note saying that most people won't need this package and should use `@emotion/primitives` and `@emotion/native`?

packages/primitives-core/src/styled.js
```diff
@@ -0,0 +1,93 @@
...
+
+type State = {
+  theme: Object
+} | null
+
+let defaultPickTest = prop => prop !== 'theme' && prop !== 'innerRef'
+
+export function createStyled(
+  StyleSheet: Object,
+  getShouldForwardProp: Function = (cmp: React.ElementType): Function =>
```
mitchellhamilton: This should be `getShouldForwardProp: (cmp: React.ElementType) => (prop: string) => boolean = () => defaultPickTest` and could you make this argument into an object so that if there needs to be more options in the future it's easier?

packages/primitives/src/styled.js
```diff
@@ -1,120 +1,37 @@
...
+import { StyleSheet, View, Text, Image } from 'react-primitives'
+import { createStyled } from '@emotion/primitives-core'
 import {
   testPickPropsOnPrimitiveComponent,
   testPickPropsOnOtherComponent
 } from './test-props'
-import { css } from './css'
 
-function isPrimitiveComponent(component: React.ElementType) {
+function shouldForwardProp(component: React.ElementType) {
```
mitchellhamilton: Should be called getShouldForwardProp

packages/primitives/test/emotion-primitives.test.js
```diff
@@ -140,17 +140,17 @@ describe('Emotion primitives', () => {
...
+  //   const Text = styled('p')`
+  //     color: hotpink;
+  //   `
+  //   let ref = React.createRef()
+  //   const rootNode = document.createElement('div')
+
+  //   render(<Text innerRef={ref} id="something" />, rootNode)
+  //   expect(ref.current).toBe(rootNode.querySelector('#something'))
+  //   unmountComponentAtNode(rootNode)
+  // })
```
mitchellhamilton: Any reason this was commented out?

packages/primitives-core/README.md
```diff
@@ -0,0 +1,39 @@
...
+import { createCss } from '@emotion/primitives-core'
+
+const css = createCss(StyleSheet)
+
+<View style={css`background-color: red;`}><Text style={css`font-size: 40px;`}>Hello World</Text></View>
+```
+
+### `createStyled(StyleSheet)`
+
+`createStyled` also accepts a platform specific `StyleSheet` method for creating styles, and returns a styled component. You can assign primitives to it for example - `View`
```
mitchellhamilton: "returns a styled component", createStyled returns a function that returns a function that returns a styled component right?

packages/native/README.md
```diff
@@ -0,0 +1,74 @@
...
+}
+
+AppRegistry.registerComponent('ExampleApp', () => App);
+```
+
+The API surface remains same for `@emotion/native` which means you can use all the API methods similar to `emotion`.
+
+## Theming
+
+Use `emotion-theming` for theming support.
```
mitchellhamilton: An example maybe?

packages/native/README.md
```diff
@@ -0,0 +1,74 @@
...
+        />
+      </Container>
+    )
+  }
+}
+
+AppRegistry.registerComponent('ExampleApp', () => App);
+```
+
+The API surface remains same for `@emotion/native` which means you can use all the API methods similar to `emotion`.
```
mitchellhamilton: Reading this is confusing, the same as what? emotion? or `@emotion/primitives`? and you can't use everything that `emotion` has, e.g. `keyframes` and `injectGlobal`

nitin42: Oops! I missed those parts of emotion API. I'll fix it.

packages/native/README.md
nitin42: Hm... That'd be great!

packages/native/src/index.js
nitin42: Sure! No issues.

packages/native/test/native-css.test.js
nitin42: I was thinking of this too! I'll need some help over here so will ping you later when I comeback to this after fixing basic stuff.

packages/primitives-core/src/styled.js
nitin42: Yup! I'll fix it

packages/primitives/test/emotion-primitives.test.js
nitin42: Hm.. can't recall may be its related to emotion primitives PR work. Anyway, I'll uncomment and update the test.

nitin42:
I also noticed some flow types errors for CreateStyled in @emotion on running yarn flow.

mitchellhamilton:
if you merge from master, those flow errors should be fixed.

nitin42:
I pulled from master and now getting this error when running `yarn flow` 😅

```
.flowconfig:14 Unsupported config section: "declarations"
```

Any idea ?

nitin42:
I pulled latest changes from master and now getting this error on running `yarn flow` 😅

```
.flowconfig:14 Unsupported config section: "declarations"
```

Any ideas ?

mitchellhamilton:
@nitin42 run `yarn install`

nitin42:
Still no luck! Getting that same error on running `yarn install` plus tests seems to be broken when I pulled the changes from master. Can you fix this and then I pull the updates ?

mitchellhamilton:
@nitin42 done.

nitin42:
ITS WORKING !!!

packages/primitives-core/src/styled.js
```diff
@@ -0,0 +1,94 @@
...
+
+type State = {
+  theme: Object
+} | null
+
+let defaultPickTest = prop => prop !== 'theme' && prop !== 'innerRef'
+
+export function createStyled(
+  StyleSheet: Object,
+  getShouldForwardProp: (
```
mitchellhamilton: Could you make this argument into an object in case we need to add more options in the future?

packages/native/package.json
```diff
@@ -0,0 +1,27 @@
...
+		"emotion-theming": "^9.2.5",
+		"react": "^16.4.1",
+		"react-native": "0.46.0"
+	},
+	"dependencies": {
+		"@emotion/primitives-core": "1.0.0"
+	},
+	"peerDependencies": {
+		"prop-types": "15.x",
+		"react-native": "0.46.0"
```
mitchellhamilton: Not sure about this version, maybe it should be looser since iirc the latest version of react-native is 0.56

nitin42: For testing purposes, I fallback to 0.46.0. I'll update it!

nitin42:
Updated!

nitin42:
Thanks @tkh44 for the review!

packages/native/package.json
```diff
@@ -0,0 +1,27 @@
...
+	"files": ["src", "dist"],
+	"devDependencies": {
+		"emotion-theming": "^9.2.5",
+		"react": "^16.4.1",
+		"react-native": "^0.56.0"
+	},
+	"dependencies": {
+		"@emotion/primitives-core": "1.0.0"
+	},
+	"peerDependencies": {
```
Andarist: shouldnt a peer on react be here?

packages/native/README.md
```diff
@@ -0,0 +1,181 @@
...
+import React from 'react';
+import { AppRegistry, StyleSheet, Text, View } from 'react-360';
+import styled from '@emotion/native';
+
+const StyledName = styled.Text`
+	font-size: 40px;
+	color: hotpink;
+`;
+
+export default class App extends React.Component {
```
Andarist: indentation is different here than in other examples (in this file), this could be enforced with prettier + lint-staged on precommit hook (with husky) as it can handle `.md` too, cc @mitchellhamilton

packages/native/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+    configurable: false,
+    get() {
+      return styled(reactNative[comp])
+    }
+  })
+)
+
+export { css }
+
+export default /* #__PURE__ */ styled
```
Andarist: this doesnt make much sense here, `#__PURE__` comments are useful for call expressions, this is just exported identifier

Not sure what actual value it gives in context of react-native (it's not that much bytes-constrained as web), it certainly won't hurt, but would have to be used differently.

I would convert this to:
```js
export default components.reduce((acc, comp) =>
  Object.defineProperty(acc, comp, {
    enumerable: true,
    configurable: false,
    get() {
      return styled(reactNative[comp])
    }
  })
, styled)
```

This would make it self contained expression that can be more safely annotated with PURE.

You could also use my plugin to insert those PURE annotations automatically - https://github.com/Andarist/babel-plugin-annotate-pure-calls , but ofc that's not obligatory 😉

packages/native/src/index.js
```diff
@@ -0,0 +1,71 @@
+import reactNative from 'react-native'
+import { createCss } from '@emotion/primitives-core'
+
+import { styled } from './styled'
+
+const css = createCss(reactNative.StyleSheet)
```
Andarist: this could be annotated with PURE

packages/native/src/styled.js
```diff
@@ -0,0 +1,9 @@
+import { StyleSheet } from 'react-native'
+import { createStyled } from '@emotion/primitives-core'
+
+/**
+ * a function that returns a styled component which render styles in React Native
+ */
+let styled = createStyled(StyleSheet)
```
Andarist: this is also a good candidate for PURE

packages/primitives-core/README.md
```diff
@@ -0,0 +1,42 @@
...
+import { createCss } from '@emotion/primitives-core'
+
+const css = createCss(StyleSheet)
+
+<View style={css`background-color: red;`}><Text style={css`font-size: 40px;`}>Hello World</Text></View>
+```
+
+### `createStyled(StyleSheet)`
+
+`createStyled` also accepts a platform specific `StyleSheet` method for creating styles, and returns a function that returns a function that returns a styled component. You can assign primitives to it for example - `View`
```
Andarist: > and returns a function that returns a function that returns a styled component

while this is technically correct, it sounds like a gibberish 😅 could we rephrase it ?

packages/primitives-core/src/styled.js
```diff
@@ -0,0 +1,96 @@
...
+
+let defaultPickTest = prop => prop !== 'theme' && prop !== 'innerRef'
+
+type options = {
+  getShouldForwardProp: (cmp: React.ElementType) => (prop: string) => boolean
+}
+
+export function createStyled(
+  StyleSheet: Object,
+  { getShouldForwardProp = () => defaultPickTest }: options = {}
```
Andarist: I've noticed that this API differs from web one. Regular emotion accepts `shouldForwardProp` - is there any reason why this is different here?

Andarist: the whole `createStyled` also differs from regular `create-emotion-styled` API which accepts whole emotion (created upfront with `create-emotion`). It seems to me that react-native equivalent would be
```js
const css = createCss(StyleSheet)
const styled = createStyled(css)
```
or 
```js
const css = createCss(StyleSheet)
const styled = createStyled({ css })
```
I'm not sure if native version has some other API than `css`/`styled` (web one has), so I'm not sure which one would be more appropriate here.

packages/primitives-core/src/styled.js
```diff
@@ -0,0 +1,96 @@
...
+) {
+  const css = createCss(StyleSheet)
+
+  return function createEmotion(component: React.ElementType) {
```
Andarist: on web this is called `createEmotionStyled`

packages/primitives-core/src/styled.js
```diff
@@ -0,0 +1,96 @@
...
+            })
+          )
+        }
+      }
+
+      Styled.contextTypes = contextTypes
+
+      Styled.displayName = `emotion(${getDisplayName(component)})`
+
+      return Styled
```
Andarist: would be cool to support component as selectors, not sure if possible though and can be added later if yes

packages/primitives/src/index.js
```diff
@@ -1,7 +1,10 @@
 // @flow
-import { Text, View, Image } from 'react-primitives'
+import { StyleSheet, Text, View, Image } from 'react-primitives'
+import { createCss } from '@emotion/primitives-core'
 
-import { createStyled } from './styled'
+import { styled as createStyled } from './styled'
+
+const css = createCss(StyleSheet)
```
Andarist: this is a good candidate for PURE

packages/primitives-core/src/styled.js
Andarist: regarding `shouldForwardProp` - I think I've found the answer, this is used at 1 layer above than regular `shouldForwardProp`, right? would be good to support normal `shouldForwardProp`, but it can also be added later (maybe create some kind of TODO list of things that could be added to achieve parity with regular `emotion`)

packages/native/package.json
nitin42: Oops! Done ✅

packages/native/README.md
nitin42: That'd be great. This will also resolve all the formatting issues!

packages/native/src/index.js
nitin42: That makes sense. I'll refactor the code to use your solution.

packages/native/src/styled.js
nitin42: are there any side-effects of using PURE here ?

packages/primitives-core/src/styled.js
nitin42: I haven't looked into it yet to check whether its possible or not. Let's save it up for later!

nitin42:
I've made some changes. I am still not sure about annotating call expressions with `PURE` because it may have side-effects ? Still not sure. Let me know! 😅

packages/native/README.md
mitchellhamilton: @Andarist If you could submit a PR for it, that would be great!

Andarist: I'll try to remember to do that later :)

packages/primitives-core/src/styled.js
Andarist: > The API here differs because, we may add some more options along with getShouldForwardProp in the future.

Making it an object is ofc preferrable, but that was not the question. Normal `emotion` also accepts configuration objects - my question was about API differences.

And what about 1st arg to `createStyled`?

nitin42: There are no such API differences except we use `StyleSheet` constructor and a little refactored code taken from `@emotion/primitives`. The goal is to keep the signature of functions similar across the env. whether it's web, native or sketch. I liked your suggestion and will make the changes.

nitin42: Your argument is valid here since having such API reduces the maintenance cost.

nitin42:
I think we can merge this now. The remaining changes don't affect the API design!

nitin42:
@mitchellhamilton have you published the package ? Haha someone jumped in my Twitter DM to ask about our work on emotion native 😅

mitchellhamilton:
@nitin42 Yup, it’s published!

PR #765, opened 2018-07-14
Rewrite @emotion/provider, add weak memoize package and ship flow files for next packages

mitchellhamilton:
**What**:
Rewrite @emotion/provider, add weak memoize package and ship flow files for next packages

**Why**:
`@emotion/provider` wasn't async safe at all

**How**:
memoization and WeakMaps(I ❤️ WeakMaps)

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/next-packages/provider/src/index.js
+++ b/next-packages/provider/src/index.js
@@ -1,82 +1,60 @@
 // @flow
 import * as React from 'react'
-import { CSSContext } from '@emotion/core'
+import { withCSSContext, CSSContext } from '@emotion/core'
 import type { CSSContextType } from '@emotion/utils'
-import createCache from '@emotion/cache'
+import weakMemoize from '@emotion/weak-memoize'
 
 type Props = {
-  theme?: Object | ((Object | void) => Object),
-  children?: React.Node,
-  cache?: CSSContextType
+  theme: Object | ((Object | void) => Object),
+  children: React.Node
 }
 
-function getTheme(
-  theme: Object | ((Object | void) => Object),
-  outerTheme: Object | void
-) {
-  if (typeof theme === 'function') {
-    const mergedTheme = theme(outerTheme)
+let createCreateCacheWithTheme = weakMemoize(cache => {
+  return weakMemoize(theme => {
+    return {
+      ...cache,
+      theme
+    }
+  })
+})
+
+let createGetTheme = weakMemoize((outerTheme: Object | void) => {
+  return weakMemoize((theme: Object | ((Object | void) => Object)) => {
+    if (typeof theme === 'function') {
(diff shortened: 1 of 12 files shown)


next-packages/weak-memoize/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@emotion/weak-memoize",
+  "version": "0.1.0",
+  "description": "A memoization function that uses a WeakMap",
+  "main": "dist/index.cjs.js",
```
Andarist: when flat bundling I've found putting pkg name into file name more readable when debugging (so smth like `dist/weak-memoize.cjs.js` here)

next-packages/provider/src/index.js
```diff
@@ -35,48 +30,24 @@ function getTheme(
     )
   }
 
-  if (outerTheme === undefined) {
```
Andarist: how this is achieved for top-level provider?

next-packages/provider/src/index.js
```diff
@@ -35,48 +30,24 @@ function getTheme(
...
-        {this.props.children}
-      </CSSContext.Provider>
-    )
-  }
-  render() {
-    return <CSSContext.Consumer children={this.consumer} />
+  })
+})
+
+export default withCSSContext((props: Props, context) => {
```
Andarist: im slightly confused about those things:
- this package is called `provider` now but in fact it handles only theme and we now have things like `CSSContext.Provider`, not sure if generic `provider` name is a good one for this one here
- it exposes `withCSSContext`, but `@emotion/core` exposes it too (I realize their implementations differ but naming both the same is confusing)

mitchellhamilton: > this package is called provider now but in fact it handles only theme and we now have things like CSSContext.Provider, not sure if generic provider name is a good one for this one here

You're totally right about the naming. I'm not too focused on package naming right now since some stuff is likely going to move to other packages. For example, this component is likely going to be moved to emotion-theming and be called ThemeProvider.

> it exposes withCSSContext, but @emotion/core exposes it too (I realize their implementations differ but naming both the same is confusing)

I'm not sure if I'm understanding you correctly but this doesn't expose `withCSSContext`, it only uses it to define a component.

Andarist: > You're totally right about the naming. I'm not too focused on package naming right now since some stuff is likely going to move to other packages. For example, this component is likely going to be moved to emotion-theming and be called ThemeProvider.

👍 

> I'm not sure if I'm understanding you correctly but this doesn't expose withCSSContext, it only uses it to define a component.

Oh lol, i've misread this in a hurry 🙈

Issue #689, opened 2018-05-29
outputDir do not work in new Babel 7 beta and could have name conflict


ai:
- `emotion` version: 9.1.3
- `react` version: 16.4
- `babel` version: 7.0.0-beta.49

Relevant code.

```javascript
{
  "plugins": [
    [
      "babel-plugin-emotion",
      {
        "extractStatic": true
      }
    ]
  ]
}
```

What you did:

Try to compile `component/loader.js`

What happened:

1. Emotion creates `loader.emotion.css` in the root (since `outputDir` is `""` by default).
2. Keep old relative `require("./loader.emotion.css")`
3. Raise `Error: Cannot find module './loader.emotion.css'`

Reproduction:

Because we need some discussion from “Problem description” I didn’t create reproduction case. Could do it if it will be needed (sorry, on the train, have no time 😅).

Problem description:

Also, I am worried about changes in https://github.com/emotion-js/emotion/pull/583. I didn’t find a code to prevent file name conflict. What if I have `components/loader.js` and `layout/loader.js`? They both will use the same `loader.emotion.css` in `outputDir`?

Suggested solution:

For `component/loader.js` creates `outputDir + "component/loader.emotion.css"`). By default, it will have old 9.0 behavior. But on `outputDir` CSS files will be created in separated dir (so @wardpeet problem will be solved too).

wardpeet:
the correct behaviour should be
```
outputdir + /a/one.emotion.css
outputdir + /b/one.emotion.css
```

i'll have a look to fix both issues in the next week.

ai:
Any updates?

Andarist:
This is not high on our priority list - especially considering the fact that `extractStatic` is going to be removed in v10 ([reference](https://github.com/emotion-js/emotion/issues/637)). I think we could merge a community PR to v9 for this though.

wardpeet:
@ai PR is up #772 not sure if it fixes the issue with babel 7 which I find a bit weird that it occurs if it's empty as it should have the old behaviour. Could you put up a small repro?

How it works now in babel 6 at least:
outputDir: ""
file: "component/loader.js" => component/loader.emotion.css

OutputDir: ".tmp"
file: "component/loader.js" => .tmp/component/loader.emotion.css

OutputDir: ".tmp"
file: "../component/loader.js" => .tmp/_/component/loader.emotion.css

ai:
Awesome! Hope it will be accepted soon.

PR #772, opened 2018-07-21
Keep all files in outputdir

wardpeet:
**What**: Nameconflicts and files getting out of tmpdir.

**Why**: To make sure we put every css file in our outputDir

**How**:
All files that start with ../ are replaced with a _/ so the file stays inside the outputDir. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


wardpeet:
Fixes part of #689

wardpeet:
It doesn't fix cli but tools that use the node api like babel loader. 

If you like I can put a webpack example online.

It does not fix any babel 7 issue but I'm not sure there is an issue

mitchellhamilton:
An example would be great!

PR #774, opened 2018-07-24
Add no response bot

mitchellhamilton:
**What**:
Add no response bot

**Why**:
So we don't have to deal with issues where the author hasn't provided enough information

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 14
+# Label requiring a response
+responseRequiredLabel: needs more information
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
+  to our request for more information from the original author. With only the
+  information that is currently in the issue, we don't have enough information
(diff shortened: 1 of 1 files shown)


.github/no-response.yml
```diff
@@ -0,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
```
Andarist: i think I've seen bots that post a comment first with a warning that the issue will be closed if it stays without a response for next X days, IMHO would be better to use such behaviour, because sometimes author has provided enough info, but we didnt have time to look into it - so posting a warning before closing gives author opportunity to keep the thread alive

mitchellhamilton: [probot/stale](https://github.com/probot/stale)? I was intending to use this by commenting on issues and asking for more information and then adding the label, this one doesn't do anything unless you add the label. I think we could add that one as well but I think this one is still useful because it's better for closing things that don't have necessary information faster.

Andarist: not sure about all intricacies of those 2, after making a rly quick look - I think probot/stale is better (more user friendly)

not sure how they both can be used at the same time (stale seems to overlap functionalities with no-response), but as mentioned - I have not thought this through really :p if you believe both are useful or one is of them is better then just go with ur gut

PR #777, opened 2018-07-26
add serializer file

kentcdodds:
**What**: This adds a root-level `serializer.js` file

**Why**: It will make it much easier to use this package for the common use-case. People will be able to configure the serializer using the [`snapshotSerializers`](https://jestjs.io/docs/en/configuration.html#snapshotserializers-array-string) config:

```json
{
  "snapshotSerializers": [
    "jest-emotion/serializer"
  ]
}
```

**How**: Created a new file called `serializer.js` in the root of the project that simply creates the serializer as documented.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/jest-emotion/serializer.js
+++ b/packages/jest-emotion/serializer.js
@@ -1,4 +1,4 @@
 const { createSerializer } = require('./dist')
 const emotion = require('emotion')
 
-expect.addSnapshotSerializer(createSerializer(emotion))
+module.exports = createSerializer(emotion)
--- a/packages/jest-emotion/serializer.js
+++ b/packages/jest-emotion/serializer.js
@@ -1,0 +1,4 @@
+const { createSerializer } = require('./dist')
+const emotion = require('emotion')
+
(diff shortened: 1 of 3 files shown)


packages/jest-emotion/serializer.js
```diff
@@ -0,0 +1,4 @@
+const { createSerializer } = require('./dist')
```
mitchellhamilton: Could you change this to requiring `jest-emotion` instead of `./dist`?

kentcdodds: Interesting! Will that work!?

tkh44:
Thanks @kentcdodds

packages/jest-emotion/serializer.js
mitchellhamilton: Yeah, node should look in the node_modules folder, go into the jest-emotion folder and look at the main property in the package.json.

ndelangen:
Awesome! just what I needed!

Issue #776, opened 2018-07-25
[docs] Prerendering issues - advise disabling speedy


Andarist:
@andrewdelprete had an issue with this ([thread](https://twitter.com/AndrewDelPrete/status/1022180246391382016)) and reached out for help, seems like we could include a note about this in some caveats/faq section - if having problems then solution is to disable speedy and ensure that happens before creating any styles (due to import hoisting it might be required to move that to other file)

oleblaesing:
I hope this will solve the problem for all puppeteer users. :)

Andarist:
I think it should be closed as #779 got merged in. Thanks @oleblaesing !

PR #779, opened 2018-07-27
Add docs for using Puppeteer as a prerenderer

oleblaesing:
Added a section of using Puppeteer as a prerenderer to the SSR docs.

I have already tried to implement prerendering with Puppeteer in my react-emotion app, without success. The [recommended way](https://twitter.com/AndrewDelPrete/status/1022180246391382016) in the [relating issue](https://github.com/emotion-js/emotion/issues/776) works very well.


docs/ssr.md
```diff
@@ -69,3 +69,21 @@ module.exports = {
...
+## Puppeteer
+
+If you are using Puppeteer to prerender your application, you have to disable emotion's `speedy` option to fully render all needed CSS inside the DOM.
+
+```jsx
+import { sheet } from 'emotion'
+
+import App from './App'
+
+sheet.speedy(false)
```
mitchellhamilton: Doing this won’t work if you use `css` in App or any of its dependencies, could you show it in another file and that being the first import?

Also, is there a way to detect if the page is being prerendered with puppeteer and if so, could you make this so it only disables speedy when it’s being prerendered?

docs/ssr.md
```diff
@@ -69,3 +69,21 @@ module.exports = {
...
+
+ReactDOM.render(<App />, document.getElementById('root'))
+```
+
+> Note:
+>
+> You have to disable this option before any styles are written. Putting it close to your entrypoint (without any other styling) is a good approach.
```
mitchellhamilton: I wouldn’t really say good approach since there isn’t really any alternative so it should probably be you have to.

docs/ssr.md
```diff
@@ -69,3 +69,44 @@ module.exports = {
...
+
+If you are using Puppeteer to prerender your application, you have to disable emotion's `speedy` option to fully render all needed CSS inside the DOM.
+
+index.js
+
+```js
+import { sheet } from 'emotion'
+
+import mount from './mount'
+
```
mitchellhamilton: This still won't work since `./mount` will run before speedy is disabled. It needs to be something like this. (though replace the comments with the actual code)
```jsx
// index.js
import './disable-speedy'
// ...other imports
// ...mounting and stuff

// disable-speedy.js
import { sheet } from 'emotion'
// ...disable speedy
```

docs/ssr.md
```diff
@@ -69,3 +69,44 @@ module.exports = {
...
+    ReactDOM.hydrate(<App />, root)
+  } else {
+    ReactDOM.render(<App />, root)
+  }
+}
+```
+
+> Note:
+>
+> You have to disable this option before any styles are written or before you use any styled component.
```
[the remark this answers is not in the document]
oleblaesing: Ah, I haven't thinked about module hoisting. Thank you, I will take a look at it in my personal application. :)

PR #782, opened 2018-07-29
Change live code and add precommit hook for prettifying markdown

mitchellhamilton:
**What**:
Change live code and add precommit hook for prettifying markdown

**Why**:

The live code thing previously worked by having a different language in the code block but that stopped prettier from formatting it so I changed it so the live code thing works by adding a `// @live` comment. I'm also working on the docs for emotion 10 and we needed to do this sometime so I wanted to do this now so there isn't a bunch of annoying conflicts to deal with later.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

I just wanna do it on markdown since we run eslint on ci which will fail if code is formatted incorrectly and most people are using eslint editor extensions that do auto fix on save and doing it as a precommit hook would make committing slower which is annoyyyyying.

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,2 +1,6 @@
 semi: false
 singleQuote: true
+overrides:
+- files: "docs/*.md"
+  options:
+    printWidth: 60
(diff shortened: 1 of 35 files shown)


nitin42:
Finally 😅 Thanks!

tkh44:
Thanks @mitchellhamilton

PR #785, opened 2018-08-02
Change keyframes in @emotion/core and a bunch of perf/size stuff

mitchellhamilton:
**What**:
Change keyframes in @emotion/core so that it can be used the same way as in v9. (it's implemented very different though)

**Why**:
Because the way keyframes work in @emotion/core  isn't great right now.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/next-packages/core/__tests__/css.js
+++ b/next-packages/core/__tests__/css.js
@@ -1,9 +1,8 @@
 // @flow
 /** @jsx jsx */
 import 'test-utils/next-env'
-import { jsx, keyframes } from '@emotion/core'
+import { jsx, css } from '@emotion/core'
 import ThemeProvider from '@emotion/provider'
-import css from '@emotion/css'
 import renderer from 'react-test-renderer'
 
 const SomeComponent = (props: { lol: true }) => (props.lol ? 'yes' : 'no')
@@ -24,62 +23,13 @@ test('thing', () => {
   expect(tree.toJSON()).toMatchSnapshot()
 })
 
-test('keyframes', () => {
-  const animation = keyframes(css`
-    from {
-      color: green;
-    }
-    to {
-      color: hotpink;
-    }
-  `)
-  const tree = renderer.create(
-    <div>
-      <div
-        css={css`
-          animation: ${animation.name} ${true};
-          ${animation.styles};
-        `}
-      >
-        {animation.name}
-      </div>
-    </div>
-  )
-
-  expect(tree.toJSON()).toMatchSnapshot()
-})
-
-test('keyframes insert in css call', () => {
-  const animation = keyframes(css`
-    from {
-      color: green;
-    }
-    to {
-      color: hotpink;
-    }
-  `)
-  const tree = renderer.create(
-    <div>
-      <div
-        css={css`
-          animation: ${animation.name};
-          ${animation.styles};
-        `}
-      >
-        {animation.name}
-      </div>
-    </div>
-  )
-
-  expect(tree.toJSON()).toMatchSnapshot()
-})
-
 test('css call composition', () => {
-  const first = css`
+  let first = css`
     color: hotpink;
   `
-  const second = css({ ':hover': first })
-  expect(second).toMatchSnapshot()
+  let tree = renderer.create(<div css={css({ ':hover': first })} />)
+
+  expect(tree.toJSON()).toMatchSnapshot()
 })
 
 test('theming with the css prop', () => {
--- a/next-packages/core/src/global.js
+++ b/next-packages/core/src/global.js
@@ -34,7 +34,7 @@ class InnerGlobal extends React.Component<InnerGlobalProps> {
     this.updateStyles()
   }
   updateStyles() {
-    let serialized = serializeStyles(this.props.context.registered, [
+    let serialized = serializeStyles(undefined, this.props.context.registered, [
       this.props.styles
     ])
     if (serialized.name === this.styleName) {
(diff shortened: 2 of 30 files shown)


next-packages/core/__tests__/css.js
```diff
@@ -24,62 +23,13 @@ test('thing', () => {
...
-      </div>
-    </div>
-  )
-
-  expect(tree.toJSON()).toMatchSnapshot()
-})
-
 test('css call composition', () => {
-  const first = css`
+  let first = css`
```
Andarist: why the const -> let change?

next-packages/core/src/global.js
```diff
@@ -67,19 +67,19 @@ class InnerGlobal extends React.Component<InnerGlobalProps> {
       this.sheet.flush()
     }
 
-    rules.forEach(rule => {
-      this.sheet.insert(rule)
-    })
+    rules.forEach(this.sheet.insert, this.sheet)
```
Andarist: 👌

next-packages/core/src/global.js
```diff
@@ -67,19 +67,19 @@ class InnerGlobal extends React.Component<InnerGlobalProps> {
...
 
   componentWillUnmount() {
     this.sheet.flush()
   }
   render() {
     if (!isBrowser) {
-      const serialized = serializeStyles(this.props.context.registered, [
-        this.props.styles
-      ])
+      const serialized = serializeStyles(
```
Andarist: it seems to me that the most common use case is to pass `undefined` as first argument, maybe it would be better to reorder the arguments for this?

next-packages/serialize/src/index.js
```diff
@@ -72,8 +85,8 @@ export function handleInterpolation(
   }
   if (interpolation.__emotion_styles !== undefined) {
     if (
-      interpolation.toString() === 'NO_COMPONENT_SELECTOR' &&
-      process.env.NODE_ENV !== 'production'
+      process.env.NODE_ENV !== 'production' &&
```
Andarist: is this change purely because of the code style or does it impact something?

next-packages/utils/src/index.js
```diff
@@ -52,7 +52,9 @@ export const insertStyles = (
     )
     context.inserted[insertable.name] = true
 
-    if (!isBrowser) {
+    if (isBrowser) {
```
Andarist: just reassuring myself - this is inlined at build time and 4 bundles are created, right? 2 for module/main and 2 for their browser alternatives

Andarist: also - is reordering this if/else logic important in any way?

next-packages/core/src/keyframes.js
```diff
@@ -1,15 +1,23 @@
...
-|}
+  styles: string,
+  anim: 1,
+  toString: () => string
+|} & string
 
-export const keyframes = (arg: ScopedInsertableStyles): Keyframes => {
-  const name = `animation-${arg.name}`
+export const keyframes = (...args: *): Keyframes => {
+  let insertable = css(...args)
```
Andarist: could u explain how this works now and how did it work before (im curious about the reasons behind this)?

Andarist: im interested in the design reasons why emotion@9 has returned a class name (string) from `css` call, but emotion@10 is switching to outputting objects, isnt it bad-ish for GC? I realize that GC could be pressured previously in other places

next-packages/core/src/keyframes.js
```diff
@@ -1,15 +1,23 @@
...
+import css from '@emotion/css'
 
 type Keyframes = {|
   name: string,
-  styles: string
-|}
+  styles: string,
+  anim: 1,
+  toString: () => string
+|} & string
```
Andarist: how does it work 😨 ? I mean intersection with string

mitchellhamilton: It’s not really good at all and there has to be a FlowFixMe comment but it’s like this so people can interpolate it in a string without flow errors

next-packages/utils/src/index.js
mitchellhamilton: Yes, it's inlined. The reordering isn't important, I just did it to avoid the negation.

next-packages/serialize/src/index.js
mitchellhamilton: Without this change, uglify can't remove the `interpolation.toString()` call since it doesn't know if it has side effects. We could probably add a pure comment but this is an easy change and will work regardless of the minifier.

next-packages/core/__tests__/css.js
mitchellhamilton: No real reason, I just felt like it. https://jamie.build/const.html

next-packages/core/src/keyframes.js
mitchellhamilton: By this, i'm assuming you mean keyframes?

Previously, they were used like this
```jsx
/** @jsx jsx */
import { keyframes, css, jsx } from "@emotion/core";

let animation = keyframes(css(...someKeyframes));

render(
  <div
    css={css`
      animtion: ${animation.name};
      ${animation.styles};
    `}
  />
);
```
but this is inconvenient since you have to interpolate the name and the styles and you can't define keyframes inline in a css call.

So this is how it works in this PR.
```jsx
/** @jsx jsx */
import { keyframes, css, jsx } from "@emotion/core";

let animation = keyframes(...someKeyframes);

render(
  <div
    css={css`
      animtion: ${animation};
    `}
  />
);
```

emotion@10 is returning objects from css because it only inserts styles inside the react tree.

While it will likely be marginally worse for the GC, it'll be better for memory usage because of the optimizations in #767 and part of those optimizations relies on information that's only available inside the React tree and the fact that we have objects to get the hash and styles from rather than storing large strings on a cache which isn't great in terms of memory usage.

PR #790, opened 2018-08-07
Fix config being silently ignored in babel-plugin

danielberndt:
Hi there, I just created a failing test case for you to inspect. 

The babel plugin seems to ignore the second parameter of `styled` if it's not an object but a reference to an object.

The code below works as expected when not using the babel-plugin, but when the babel-plugin is applied, the `cfg` seems to get ignored.

```jsx
const Button = ({ isRed, ...rest }) => (
  <button {...rest}>{isRed ? 'passed' : 'not passed'}</button>
)
const StyledButton1 = styled(Button, {shouldForwardProp: p => p !== 'isRed'})({})
<StyledButton2 isRed /> // -> isRed is not forwarded, as expected

const cfg = { shouldForwardProp: p => p !== 'isRed' }
const StyledButton2 = styled(Button, cfg)({})
<StyledButton2 isRed /> // -> isRed gets forwarded!
```

### DIFF
--- a/packages/babel-plugin-emotion/__tests__/__snapshots__/styled.js.snap
+++ b/packages/babel-plugin-emotion/__tests__/__snapshots__/styled.js.snap
@@ -175,34 +175,55 @@ styled(
   window.whatever
 )\`\`
 
+const cfg = { shouldForwardProp: window.whatever }
+styled('button', cfg)({})
+
 
       ↓ ↓ ↓ ↓ ↓ ↓
 
 import _styled from \\"@emotion/styled-base\\";
 
+function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }
+
 /*#__PURE__*/
 _styled('div', {
   shouldForwardProp: window.whatever,
   target: \\"ek6pnb20\\"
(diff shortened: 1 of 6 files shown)


Andarist:
I guess the problem is that the babel plugin needs to extend the used config but it's not obvious how to do that - maybe we could reuse babel's `extend` helper? 🤔

danielberndt:
Another option I'd also be fine with would be to just bail and let the user know that this construct is not supported, or simply not apply any babel transforms to such `styled` calls.
I'm not really too experienced with babel transforms to know what's possible.

danielberndt:
Ah I just found this piece:
https://github.com/emotion-js/emotion/blob/bb99600fcfecbd1d381cdc558d885732142c3112/packages/babel-plugin-emotion/src/index.js#L198-L201

So you are warning about this behaviour, but this warning is nowhere to be seen in my dev environment. (I'm using `create-react-app` with `react-app-rewire-emotion`)

danielberndt:
Alright. I think I just found a way to make it work. Instead of ignoring non-object options, I spread them onto the finalOptions.
So based on the code sample above, the transformed code will look something like this:

```
const cfg = { shouldForwardProp: p => p !== 'isRed' }
const StyledButton = styled(Button, {
  ...cfg,
  target: "e8nbyc50"
})()
```

Is this an approach you're happy with? If so, what would I need to do in order to make it shippable?

tkh44:
What do you think of @danielberndt's solution @Andarist?

Andarist:
This expects user having set up spread operator transpilation, which would be practical (let's face it - everyone does it) but a bad practice (we shouldn't assume such things).

I think it's possible to reuse babel's runtime helpers & that we should probably use them anyway (instead of inline ones) to reduce (slightly) emotion libs footprint - let me look into this one in following week or so and I'll report back with the findings.

Andarist:
I think I know how to reuse the existing babel helper in here, need to prepare a PR to babel though to use cjs/esm fatigue. Gonna try to do it in following week.

Andarist:
Took me way longer than I've expected but I got ball rolling and created this PR - https://github.com/babel/babel/pull/8885 which (if merged) will allow us to reuse babel `extends` helper consistently across CJS and ESM formats.

Andarist:
The babel PR doesnt move along right now, so I've rebased & tweaked this PR to just use babel's `extends` helper directly and rely on user's helperGenerator (implementation detail of babel - basically it's a generator that can be overriden by plugins, such as @babel/plugin-transform-runtime to insert helpers i.e. coming from @babel/runtime package, by default it just inserts a helper into a transformed file).

cc @mitchellhamilton could u take a look at this? CI is failing on flow task, but it doesnt seem to be related to this PR

Andarist:
Sure, gonna work on that later. Thanks for the preliminary review.

Andarist:
Ok, I've moved the test and added one more - but the fixture didnt cause a snapshot update, I'm not sure why 🤔

PR #795, opened 2018-08-08
Emotion 10 typing

Ailrun:
**What**: Typings for emotion@10

**Why**: For TS users

**How**: Using TS

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete


Ailrun:
@mitchellhamilton It's finished for `@emotion` scoped packages if we will drop preact support.

packages/serialize/types/index.d.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export interface ComponentSelector {
+  __emotion_styles: any;
+}
+
+/**
+ * @desc Following type includes `& string` because of `animation` and `animationName`.
+ * It's hard to add exceptions only for those CSS property to make them accept `Keyframes`.
+ */
```
mitchellhamilton: What would be the point of adding exceptions when using this on anything other than animation and animationName is useless?

packages/styled-base/types/tests.tsx
```diff
@@ -0,0 +1,278 @@
...
+  <Canvas0 />
+);
+(
+  <Canvas1 />
+);
+
+interface PrimaryProps {
+  readonly primary: string;
+}
+const Button2 = styled<'button', PrimaryProps>('button')`
```
mitchellhamilton: Why does `'button'` have to be duplicated?

packages/utils/types/index.d.ts
```diff
@@ -7,31 +7,32 @@ export interface RegisteredCache {
...
+  key: string;
   insert(rule: string): void;
   flush(): void;
+  tags: Array<HTMLStyleElement>;
 }
 
-export interface CSSContextType {
-  stylis: (key: string, value: string) => Array<string>;
+export interface EmotionCache {
+  stylis(key: string, value: string): Array<string>;
```
mitchellhamilton: Not that it affects the actual types but `key` and `value` would be better named `selector` and `styles` respectively.

Also, what's the significant of changing it from an arrow function to the method thing?

packages/core/types/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ClassAttributes, ComponentClass, Context, Provider, SFC, ReactElement, ReactNode, Ref, createElement } from 'react';
+
+export {
+  Interpolation,
+  css,
+};
+
+export const ThemeContext: Context<object>;
+export const CacheProvider: Provider<EmotionCache | null>;
+export function withEmotionCache<Props, RefType = any>(func: (props: Props, context: EmotionCache | null, ref: Ref<RefType>) => ReactNode): SFC<Props & ClassAttributes<RefType>>;
```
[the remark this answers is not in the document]
Ailrun: @mitchellhamilton It is `null` when user is not in a browser.
https://github.com/emotion-js/emotion/blob/7a4ff8ebae5ddce2b3dd81afa258bc15398512b0/packages/core/src/context.js#L6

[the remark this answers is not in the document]
mitchellhamilton: Yes, that is the value of the actual context but when using `withEmotionCache`, the value that get's passed here is never null.

packages/core/types/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Keyframes } from '@emotion/serialize';
+import { ClassAttributes, ComponentClass, Context, Provider, SFC, ReactElement, ReactNode, Ref, createElement } from 'react';
+
+export {
+  Interpolation,
+  css,
+};
+
+export const ThemeContext: Context<object>;
+export const CacheProvider: Provider<EmotionCache | null>;
```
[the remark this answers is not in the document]
Ailrun: I think asymmetric type for `withEmotionCache` and `CacheProvider` make user confused... If you think that's OK, I will change this one only.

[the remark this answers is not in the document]
mitchellhamilton: As I explained above, someone using `withEmotionCache` will never receive null.

packages/core/types/index.d.ts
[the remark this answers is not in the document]
Ailrun: @mitchellhamilton How? SSR never run this HOC?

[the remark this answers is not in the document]
mitchellhamilton: https://github.com/emotion-js/emotion/blob/master/packages/core/src/context.js#L45-L83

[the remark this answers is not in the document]
Ailrun: @mitchellhamilton Alright. I will update both of them.

packages/serialize/types/index.d.ts
Ailrun: @mitchellhamilton So what you said is we should add `& string` type even if it is not actual string?

packages/styled-base/types/tests.tsx
mitchellhamilton: Yes, that's what I mean, could you add a comment explaining that for future reference?

packages/core/types/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+export function withEmotionCache<Props, RefType = any>(func: (props: Props, context: EmotionCache | null, ref: Ref<RefType>) => ReactNode): SFC<Props & ClassAttributes<RefType>>;
+
+export const jsx: typeof createElement;
```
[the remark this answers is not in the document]
Ailrun: @mitchellhamilton In TS, `createElement` does nothing with props (I really hate this).
Props only depends on interfaces in `JSX` namespace, so if we want to modify props type, we should access them.

packages/serialize/types/index.d.ts
mitchellhamilton: Yes, users should see it as a string.

Also, slightly unrelated but does TS have [some kind of opaque type like flow](https://flow.org/en/docs/types/opaque-types/)?

packages/styled-base/types/tests.tsx
Ailrun: @mitchellhamilton Of course. How about adding a comment in the typing itself? (where the type parameter is defined)

packages/cache/types/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Plugin as StylisPlugin, Prefix } from '@emotion/stylis';
+import { EmotionCache } from '@emotion/utils';
+
+export {
+  StylisPlugin,
+  Prefix,
+  EmotionCache,
+};
+
+export interface CreateCacheOptions {
```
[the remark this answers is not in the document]
Ailrun: @mitchellhamilton OK. It looks like `Options` is more consistent with typing of `@emotion/sheet` too.

packages/styled-base/types/tests.tsx
mitchellhamilton: Both places would be great because people might look at the tests and type their stuff based on it.

packages/serialize/types/index.d.ts
Ailrun: @mitchellhamilton TypeScript (almost) always follows structural typing, so there is no equivalent. In terms of PL theory, flow is much more beautiful... The only reason I don't use it is they don't support dependencies between flow-typed typing packages... (I don't know whether now they support or not, but at least this May, they didn't support it)

Ailrun:
@mitchellhamilton All fixed.

Ailrun:
@mitchellhamilton However, the problem of `prettier` is that it requires many rule changes for tslint too...

mitchellhamilton:
Use https://github.com/alexjoverm/tslint-config-prettier maybe?

packages/core/types/tests.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface TestTheme0 {
+  resetStyle: any
+}
+
+;<Global styles={(theme: TestTheme0) => [theme.resetStyle]} />
+
+declare const getRandomColor: () => string
+
+const ComponentWithCache = withEmotionCache((_props: {}, context) => {
+  if (context === null) {
```
mitchellhamilton: This can be removed

packages/create-emotion/types/index.d.ts
```diff
@@ -1,107 +1,122 @@
...
+    | ((key: string, value: string, context: 1 | 2 | 3) => boolean)
+  key?: string
+  container?: HTMLElement
 }
 
-export default function createEmotion(context: EmotionContext, options?: EmotionOptions): Emotion;
+export default function createEmotion(
+  context: EmotionContext,
+  options?: EmotionOptions
+): Emotion
```
mitchellhamilton: The `context` argument here was removed btw.

Ailrun: Actually, there are more broken types for those pre-typed packages, but I think fixing those could be handled separately.

Ailrun:
@mitchellhamilton `tslint-config-prettier` does not support dtslint. More correctly, dtslint does not support additional `"extends"`.

Ailrun:
@mitchellhamilton I manually modify them.

Issue #792, opened 2018-08-07
Allow function-based style mapping using withProps


Gejsi:
- `emotion` version: ^9.2.5
- `react` version: ^16.2.0

Relevant code.

```javascript
const Title = withProps({
  style: ({ big }) => ({
    fontSize: big ? '32px' : '18px'
  })
})(styled("h1") `
  color: green;
`);
```

What you did:
I created a simple component to test if props can be added inside the `withProps` HOC.

What happened:
React threw this error:
_The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX._
I changed my code to fix the error:
```javascript
const Title = withProps({
  style: { fontSize: ({ big }) => (big ? "32px" : "18px") }
})(styled("h1")`
  color: green;
`);
```
No error is thrown however the attribute `style` is not rendered.

Reproduction:
Simply write a function inside the `withProps` HOC.
https://codesandbox.io/s/k5vrzm19v5

Problem description:
The reproduction I provided is clearly useless in a normal environment because it is just a bunch of static styles. But you can not write dynamic styles in template literals otherwise you will end up generating too many classes.
Imagine moving a component depending on the mouse position, there are two ways of doing it: 

1. Set the attribute to the component: 
```javascript
<MyComponent style={{ left: mouseX + 'px' }} />
```
2. Set the attribute using the `withProps` HOC (best method)
 ```javascript
const MyComponent = withProps({
   style: { left: ({ mouseX }) => mouseX + 'px' }
})(styled('div')`
/** styles here */
`)
```

Suggested solution:
I do not really know how to fix this problem: the HOC itself works properly.

**Comment by Andarist** [comment]
Functions as style object values are not supported atm (should they? 🤔 cc @mitchellhamilton )

If they were u could do this:
```js
const getProp = prop => props => props[prop]

const Title = withProps({
  css: { fontSize: ({ big }) => (big ? "32px" : "18px") }
})(styled("h1")`
  color: green;
  ${getProp('css')}
`);
```

mitchellhamilton:
tbh, I think we should stop recommending `withProps` all-together since it's just another thing to learn without any real benefit over [creating a regular react component with the css prop](https://github.com/emotion-js/emotion/pull/617#issuecomment-388284739). The example above could be expressed like so.
```jsx
let Title = ({ big, ...props }) => (
  <h1
    style={{
      fontSize: big ? 32 : 18
    }}
    css={css`
[...]

PR #797, opened 2018-08-08
Allow for functions as interpolated object values

Andarist:
Ref https://github.com/emotion-js/emotion/issues/792#issuecomment-411337648


Andarist:
Sure, I just thought I'd propose it so we can decide if it's worth it. I'm attached to it, would never write a code that would go through that path myself 😉

PR #799, opened 2018-08-11
Emotion 10 docs

mitchellhamilton:
This is very WIP and I'm aware a bunch of stuff is broken and I haven't totally figured out how we're going to fit the docs for the old api with the new docs but I wanted to open this anyway

https://next-docs.emotion.sh


Andarist:
Worth reading through - https://github.com/facebook/react/issues/13456 . I suspect those problems don't apply as much to us as in react-redux, but still would be good to benchmark this for emotion@10

docs/with-props.md
```diff
@@ -1,26 +1,8 @@
...
-  type: 'password'
-})(styled('input')`
-  background-color: red;
-`)
-
-render(<RedPasswordInput />)
-```
-
-Alternatively, you can use the css prop and create a regular component like this.
+Some css-in-js libraries offer APIs to attach props to components, instead of having our own API to do that, we recommend creating a regular react component, using the css prop and attaching props like you would with any other React component.
```
Andarist: while this obviously should not require an example, maybe it's worth to add one, just so we can refer people to it when they start asking

mitchellhamilton: The example for it is below.

fforres:
🎉🎉🎉❤️❤️❤️

PR #803, opened 2018-08-16
add spectrum badge to README

satyarohith:
Noticed emotion has spectrum community so added spectrum badge to README.


mxstbr:
:raised_hands:

PR #807, opened 2018-08-17
feat(babel-plugin-emotion): Add name for anonymous exports

billyvg:
**What**: This will attempt to add a name for an anonymously exported styled component, if the component has an available name.

**Why**: This will improve `autoLabel` for anonymous/default exports.

**How**: If there is no identifier name for the styled component use the component name that we are styling. The label becomes `styled(Component)`.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Code complete

Please let me know if there are better utility functions to use to get the name of the component of the argument to `styled`.

### DIFF
--- a/packages/babel-plugin-emotion/test/__snapshots__/styled.test.js.snap
+++ b/packages/babel-plugin-emotion/test/__snapshots__/styled.test.js.snap
@@ -1,5 +1,81 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`styled extract babel 6 autoLabel anonymous styled anonymous class component 1`] = `
+"import \\"./emotion.emotion.css\\";
+
+export default styled(class extends React.Component {}, {
+  e: \\"css-1e93lt2\\",
+  target: \\"e8nbyc50\\"
+})();
+
+
+emotion.emotion.css
+.css-1e93lt2{color:blue;}"
+`;
+
+exports[`styled extract babel 6 autoLabel anonymous styled class component 1`] = `
+"import \\"./emotion.emotion.css\\";
+
+export default styled(class Profile extends React.Component {}, {
+  e: \\"css-1aj1g6z\\",
+  target: \\"e8nbyc50\\"
+})();
+
+
+emotion.emotion.css
+.css-1aj1g6z{color:blue;}"
+`;
+
+exports[`styled extract babel 6 autoLabel anonymous styled dom component with object styles 1`] = `
+"
+export default /*#__PURE__*/styled('h1', {
+  target: 'e8nbyc50',
+  label: 'styled(h1)'
+})({
+  color: 'blue'
+});"
+`;
+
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-emotion/test/__snapshots__/styled.test.js.snap
```diff
@@ -1,5 +1,81 @@
...
+
+
+emotion.emotion.css
+.css-1aj1g6z{color:blue;}"
+`;
+
+exports[`styled extract babel 6 autoLabel anonymous styled function component 1`] = `
+"import \\"./emotion.emotion.css\\";
+
+export default styled(function Profile() {}, {
```
Andarist: shouldnt this get a label?

packages/babel-plugin-emotion/test/__snapshots__/styled.test.js.snap
```diff
@@ -1,5 +1,81 @@
...
+
+emotion.emotion.css
+.css-1aj1g6z{color:blue;}"
+`;
+
+exports[`styled extract babel 6 autoLabel anonymous styled function component via variable identifier 1`] = `
+"import \\"./emotion.emotion.css\\";
+
+const Profile = () => <div />;
+export default styled(Profile, {
```
Andarist: this could get a label too

packages/babel-plugin-emotion/test/__snapshots__/styled.test.js.snap
```diff
@@ -349,6 +425,80 @@ emotion.emotion.css
...
+  e: \\"css-1e93lt2\\",
+  target: \\"e8nbyc50\\"
+})();
+
+
+emotion.emotion.css
+.css-1e93lt2{color:blue;}"
+`;
+
+exports[`styled extract babel 7 autoLabel anonymous styled class component 1`] = `
```
Andarist: and this

billyvg:
Ah, yeah I got this all wrong, my intention is to have a display name (and I guess `label` is used for the generated class names) for a component. Where would be the correct package to handle this in @mitchellhamilton ?

billyvg:
Hmm looks like what I'm trying to do already exists 🤷‍♂️: https://github.com/emotion-js/emotion/blame/master/packages/create-emotion-styled/src/index.js#L144

Andarist:
@billyvg this is not used in the same way as label though, so it isn't attached to generated class names and this behaviour is what u were looking for, right?

billyvg:
@Andarist I actually wasn't referring to the generated CSS class names (sorry for the confusion). I was more interested in the meta `displayName` that's used when using something like jest (snapshots) and React dev tools.

So the problem I'm seeing is that something, (I think `babel`?) is generating an `identifierName` of `_default` for these anonymous default exports - and because of that we don't get the nice `Styled(ComponentName)` as `displayName`.

Right now I'm trying to confirm where the `_default` name is coming from.

PR #809, opened 2018-08-19
Prettify also .js files on precommit hook

Andarist:


TrySound:
Committing with prettier is quite fast. Running eslint on precommit is long, yes.

Andarist:
> I'm not big on this since we run eslint on CI which runs prettier and fails if any js isn't formatted with prettier.

That's kinda my reason why I like this change - it just eliminates such failed CI builds. Even if many people have IDE integrations they dont necessarily do format-on-save & it's somewhat easy to miss some IDE warning highlight.

tkh44:
I like it. I think we should merge this. We can always remove it if real negative crops up.

mitchellhamilton:
Okay 👍 I don’t mind that much so if people really want it the then I’m good with it.

Issue #637, opened 2018-04-27
Emotion 10


mitchellhamilton:
Emotion 10 is going to introduce some new APIs for React users.

The main aim with these new APIs is to move all insertion to inside the React tree, this enables a couple of things:

* SSR just works because we can render style elements directly into the React tree.
* Setting things like nonces, container elements and stylis plugins is much easier since it only requires a Provider at the top of the tree.
* Global styles can change over time and be removed
* It forces people to think in terms of styling elements instead of creating class names
* The css prop works without the babel plugin because it's implemented as a custom `createElement` (this isn't necessarily a result of moving things inside the React tree but a side-effect of the new implementation)

The main new APIs are a custom `createElement` for the css prop and a `Global` component for global styles.

### Example

> The exact APIs are likely to change before Emotion 10 is released.

```jsx
/** @jsx jsx */
import { jsx, Global } from '@emotion/core'
import css from '@emotion/css'
import { render } from 'react-dom'

render(
  <div>
    <h1
      css={css`
        font-size: 500px;
      `}
    >
      This is really big
    </h1>
    <Global
      css={{
        body: {
          backgroundColor: 'hotpink'
        }
      }}
    />
  </div>,
  document.getElementById('root')
)
```

If you want to try this, look at https://github.com/emotion-js/next but note that there **will** be breaking changes very often for now.

Another change that emotion@10 will bring is that `styled.div` and etc. will work without a babel plugin. The tag list will be included in `@emotion/styled` but if you use `@emotion/styled.macro` with [babel-plugin-macros](https://github.com/kentcdodds/babel-plugin-macros) or the babel plugin, `styled.div` will be replaced with `styled('div')` and `@emotion/styled-base` will be imported instead which doesn't include the tag list.

## Compatibility with Preact and non-React-like libraries

While the majority of Emotion users use React, there are some users who don't use React and we have to figure out how to cater for them.

### Non-React-like libraries

Since all the packages on the `@emotion` scope are very modular, we can build the current emotion APIs with the packages on the `@emotion` scope. This should be pretty simple.

### Preact

There are only a few APIs that we use that are new.

##### Fragment

Fragments are only used for SSR so we could have an alternative SSR API for that similar to `extractCritical` now.

##### createContext

We could tell people to use `create-react-context` and essentially polyfill `createContext`.

##### forwardRef

[...]

PR #812, opened 2018-08-21
Deprecate extract static

mitchellhamilton:
**What**:

Add warnings when using extractStatic and mention in the docs that it's deprecated.

**Why**:

#637 

**How**:

**Checklist**:

- [x] Documentation
- [ ] ~~Tests~~ it's not worth it
- [x] Code complete


docs/extract-static.md
```diff
@@ -4,7 +4,15 @@ title: "Extract Static"
...
+## DEPRECATED
+
+extractStatic is deprecated and will be removed in emotion@10. We recommend disabling extractStatic or using other libraries like [linaria](https://github.com/callstack-io/linaria) or [css-literal-loader](https://github.com/4Catalyzer/css-literal-loader).
+
+### Why is extractStatic being deprecated?
+
+The emotion team isn't very interested in static extraction in its current state as it negates lots of the benefits of emotion, most notably, [composition](/docs/composition.md). There is also very little interest from the community in maintaining it and maintaining `extractStatic` takes time away from maintaining more widely used parts of emotion. We don't think static extraction is a bad idea but the implementation that we have doesn't provide a good enough experience and we believe there are more important problems to be solved first. We're very interested in the work being done with [prepack](https://github.com/facebook/prepack) though and if we add some form of static extraction back, it will likely be done with prepack.
```
tkh44: I wrote this as an alternative. Let me know what you think.

> Static extraction was designed and built for earlier versions of emotion which functioned in subtly different ways than current versions. As the team added more features which expanded the power of [composition](/docs/composition.md) and made it more performant, static extraction became less important. Because most of the community is leveraging the power of composition, the time cost in maintaining `extractStatic` has become too high. Libraries such as [linaria](https://github.com/callstack-io/linaria) and [css-literal-loader](https://github.com/4Catalyzer/css-literal-loader) do static extraction well and the developers working on them are focused on this specific problem. As a team, we believe that projects like [prepack](https://github.com/facebook/prepack), when they are ready, will provide a better and more forward thinking solution to the problem or pre-compiling styles.

mitchellhamilton: I like it, I’ll probably use that with a few small changes

PR #814, opened 2018-08-21
Updated isPropValid props list

### LINKED ISSUE #734: `for` as a prop should be whitelisted for preact
- `emotion` version: 9.2.3
- `preact` version: 8.2.9

Relevant code.

```javascript
<label for="foo">click</label>
```

What you did:

Wrote the above code.

What happened:

`for` was not passed through.

Reproduction:

Problem description:

Preact uses `for`, `class` and `autofocus` instead of `htmlFor`, `className` and `autoFocus`.  There may be others, but those are the main ones I've used.

Suggested solution:

Those should all be whitelisted either for all or via some opt-in that is preact specific.

### LINKED ISSUE #813: Some missing props in the is-prop-valid package
@Andarist recommended we check out `is-prop-valid` https://github.com/styled-components/styled-components/issues/1694#issuecomment-414642019 and it looks to be pretty solid. However, there are a couple missing props based on our existing tests:

- "challenge" (attribute on the keygen element)
- "x-*" <- mozilla tags, similar to data-* or aria-*
- "valueLink" (react 15 era deprecated prop)

Andarist:
fixes #813 fixes #734 cc @probablyup

### DIFF
--- a/next-packages/is-prop-valid/src/props.js
+++ b/next-packages/is-prop-valid/src/props.js
@@ -10,6 +10,8 @@ const props = {
   defaultChecked: true,
   innerHTML: true,
   suppressContentEditableWarning: true,
+  // deprecated react prop
+  valueLink: true,
 
   // https://github.com/facebook/react/blob/d7157651f7b72d9888ctrue123e191f9b88cd8f41e9/src/renderers/dom/shared/HTMLDOMPropertyConfig.js
   /**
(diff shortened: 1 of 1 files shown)


next-packages/is-prop-valid/src/props.js
```diff
@@ -454,10 +457,12 @@ const props = {
   yChannelSelector: true,
   z: true,
   zoomAndPan: true,
+  // preact
   for: true,
-  class: true
+  class: true,
+  autofocus: true
```
mitchellhamilton: `autofocus` is unnecessary as I made the regex case insensitive and `autoFocus` is already there

mitchellhamilton:
Thanks!!

PR #818, opened 2018-08-22
babel plugin: undefined in className

slorber:
**What**: try to fix the babel plugin so that it filters out undefined from final className

**Why**: it's bad to have class="xxx undefined" in production

**How**:  by modifying the code

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A

-------------------------------

I tried to add inline cases/snapshots the problematic cases:

```
  'undefined className': {
    code: `(<div className={undefined} css={\`color: $\{color};\`}></div>)`
  },
  'undefined const className': {
    code: `const cls = undefined; (<div className={cls} css={\`color: $\{color};\`}></div>)`
  },
  '"undefined" className': {
    code: `(<div className="undefined" css={\`color: $\{color};\`}></div>)`
  },
```

As of now, it's not perfectly working, I don't know how to avoid the "x" + "undefined" in the 3nd case.

@mitchellhamilton any idea? I gave it a try but I don't really know what I'm doing here

### DIFF
--- a/packages/babel-plugin-emotion/src/css-prop.js
+++ b/packages/babel-plugin-emotion/src/css-prop.js
@@ -101,7 +101,12 @@ export default function(
     }
   }
 
+  function isUndefinedIdentifier(node) {
+    return node.type === 'Identifier' && node.name === 'undefined'
+  }
   function add(a, b) {
+    if (isUndefinedIdentifier(a)) return b
+    if (isUndefinedIdentifier(b)) return a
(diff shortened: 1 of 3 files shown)


mitchellhamilton:
You would have to add a ternary and check at runtime for undefined.

Just FYI, this will be fixed in v10.

mitchellhamilton:
This is fixed in v10

PR #824, opened 2018-08-25
Reimplement vanilla emotion, babel-plugin-emotion with new utils and change react-emotion tests to new styled

mitchellhamilton:
**What**:
Reimplement vanilla emotion and babel-plugin-emotion with new utils and change react-emotion tests to new styled

**Why**:
For emotion  10

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


BREAKING_CHANGES_IN_V10.md
```diff
@@ -0,0 +1,8 @@
+# Breaking Changes
+
+* Only supports react@16.3 and above(+ preact)
+* functions in interpolations are stringified in css and cx calls(probably won't affect you very much, there's a warning about it in v9)
+* `create-emotion-styled` is gone, use the new styled api and a provider
+* The css prop doesn't work via the babel plugin. `jsx` can be manually imported from `@emotion/core` or [babel-plugin-jsx-pragmatic](https://github.com/jmm/babel-plugin-jsx-pragmatic) can be used. (should we make a babel preset to that?)
```
Andarist: an eslint rule with a fixer would be cool for this

next-packages/cache/src/rule-sheet.js
```diff
@@ -22,30 +22,56 @@ let ruleSheet: StylisPlugin = (
...
+        case 64: {
+          // @import
+          if (depth === 0) {
+            current.push(content + ';')
+            return ''
+          }
+          break
+        }
+        // charcode for l
+        case 108: {
```
Andarist: this https://github.com/xtuc/charcodes/tree/master/packages could be used to better distinguish char codes and plugin phases/contexts here (both being numbers)

BREAKING_CHANGES_IN_V10.md
mitchellhamilton: 👍 love the idea!

package.json
```diff
@@ -57,6 +57,7 @@
     "codecov": "^2.3.1",
     "css": "^2.2.1",
     "cssjanus": "^1.2.0",
+    "cssom": "^0.3.4",
```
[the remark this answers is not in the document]
mitchellhamilton: I added it because jsdom depends on it and an older version didn’t support `@supports` so adding it forced it to update.

next-packages/utils/src/index.js
```diff
@@ -41,17 +53,20 @@ export const insertStyles = (
...
   }
   if (context.inserted[insertable.name] === undefined) {
-    let rules = context.stylis(
-      `.${context.key}-${insertable.name}`,
-      insertable.styles
-    )
+    let rules = context.stylis(`.${className}`, insertable.styles)
     context.inserted[insertable.name] = true
 
+    if (insertable.map !== undefined && process.env.NODE_ENV !== 'production') {
```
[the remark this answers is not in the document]
mitchellhamilton: It’s better the way it is since the check is pure and cheaper than checking process.env in node

packages/emotion/test/warnings.test.js
```diff
@@ -51,13 +51,6 @@ it('does warn when invalid values are passed for the content property', () => {
...
   expect(console.error).toBeCalledWith(
-    "Interpolating functions in css calls is deprecated and will be removed in the next major version of Emotion.\nIf you want to have a css call based on props, create a function that returns a css call like this\nlet dynamicStyle = (props) => css`color: ${props.color}`\nIt can be called directly with props or interpolated in a styled call like this\nlet SomeComponent = styled('div')`${dynamicStyle}`"
-  )
-})
-
-it('does warn when functions are passed to cx calls ', () => {
-  cx(() => 'cls')
-  expect(console.error).toBeCalledWith(
-    'Passing functions to cx is deprecated and will be removed in the next major version of Emotion.\nPlease call the function before passing it to cx.'
+    "Functions that are interpolated in css calls will be stringified.\nIf you want to have a css call based on props, create a function that returns a css call like this\nlet dynamicStyle = (props) => css`color: ${props.color}`\nIt can be called directly with props or interpolated in a styled call like this\nlet SomeComponent = styled('div')`${dynamicStyle}`"
```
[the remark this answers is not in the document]
mitchellhamilton: Yeah, I’d love to use them but we haven’t updated to a version that supports it iirc, the last time I tried updating jest some stuff broke though.

next-packages/cache/src/rule-sheet.js
mitchellhamilton: Not sure if it’s worth it, especially because we would have to add it as a regular dependency so that people who use flow wouldn’t get flow errors

next-packages/utils/src/index.js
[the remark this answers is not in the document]
Andarist: hm, it think (not sure) that this check is pure only if one uses `pure_getters` or similar option, if you are concerned with process.env implications maybe it would be sufficient to cache it upfront and use the cached value here?

next-packages/styled-base/src/index.js
```diff
@@ -103,15 +103,31 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
 
-      const ele = React.createElement(
-        baseTag,
-        // $FlowFixMe
-        pickAssign(shouldForwardProp, {}, props, {
-          className,
-          ref: props.innerRef
-        })
-      )
+      for (let key in props) {
```
Andarist: was `pickAssign` much slower?

mitchellhamilton: It was a little slower, probably because it used `arguments`.

next-packages/styled-base/src/index.js
```diff
@@ -158,6 +158,7 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
               : baseTag.displayName || baseTag.name || 'Component'
           })`
 
+    Styled.defaultProps = tag.defaultProps
```
Andarist: 👌

Andarist: extending defaultProps might still be an issue though, because it's quite confusing for newcomers 🤔

mitchellhamilton: This isn’t something new, it was there in create-emotion-styled, iirc it’s for when styled components get flattened. Now that I’m thinking about it, it should probably be more like this, thoughts?

```jsx
if (tag.somePropThatOnlyStyledComponentsHave !== undefined) {
  Styled.defaultProps = tag.defaultProps
}
```

Andarist: Oh, kinda thought it didnt work before. Im not sure if doing it only for emotion components is a good thing, it seems like this behaviour is useful in any case - I can't imagine when I wouldn't want those defaultProps to transition to the wrapping styled (they might be useful in function interpolations). Seems like `styled(Tag)` should ideally be completely transparent and the output component should behave pretty much the same as `<Tag/>` itself (with styling adding on top of that ofc)

scripts/build/utils.js
```diff
@@ -108,7 +108,8 @@ function getOutputConfigs(pkg, isBrowser = false) {
   let configs = [
     {
       format: 'cjs',
-      file: cjsPath
+      file: cjsPath,
+      exports: 'named'
```
Andarist: ❤️ auto mode generates horrible interop problems 😢

next-packages/serialize/src/index.js
```diff
@@ -75,6 +79,8 @@ if (process.env.NODE_ENV !== 'production') {
   }
 }
 
+let cursor
```
Andarist: a comment explanation why this is kept at module scope would be good for future code readers

next-packages/serialize/src/index.js
```diff
@@ -173,15 +173,15 @@ function createStringFromObject(
       string += handleInterpolation(mergedProps, registered, obj[i], false)
     }
   } else {
-    for (let key in obj) {
-      if (typeof obj[key] !== 'object') {
-        if (registered[obj[key]] !== undefined) {
-          string += `${key}{${registered[obj[key]]}}`
+    // using var instead of let here because using let here makes babel use a closure
```
Andarist: not entirely sure, but seems like a deopt in babel that could be fixed, gonna try to debug this later, this causes it:
```js
          value.forEach(value => {
            string += `${processStyleName(key)}:${processStyleValue(
              key,
              value
            )};`
          })
```
but original `value` is not referenced inside the callback so I don't see why would it deopt because of this

next-packages/serialize/src/index.js
```diff
@@ -193,11 +193,10 @@ function createStringFromObject(
...
         }
         if (
-          Array.isArray(obj[key]) &&
-          (typeof obj[key][0] === 'string' &&
-            registered[obj[key][0]] === undefined)
+          Array.isArray(value) &&
+          (typeof value[0] === 'string' && registered[value[0]] === undefined)
         ) {
-          obj[key].forEach(value => {
+          value.forEach(value => {
```
Andarist: wouldnt this be faster as regular for loop? or is perf diff negligible nowadays?

mitchellhamilton: Probably negligible but I'm gonna change it to a for loop anyway.

Andarist: if u do so u can use `let` for a `key` again :p

next-packages/serialize/src/index.js
```diff
@@ -173,8 +173,7 @@ function createStringFromObject(
       string += handleInterpolation(mergedProps, registered, obj[i], false)
     }
   } else {
-    // using var instead of let here because using let here makes babel use a closure
-    var key
+    let key
     for (key in obj) {
```
Andarist: u could move that let declaration to the parenthesis

lerna.json
```diff
@@ -11,6 +11,12 @@
     }
   },
   "version": "9.2.8",
-  "npmClient": "yarn",
-  "useWorkspaces": true
+  "npmClient": "npm",
```
Andarist: this surprises me - the commit message say about using yarn, but here u are changing the npm client to, well, npm

mitchellhamilton: Yeah, lerna is only used for publishing and I want to use npm to publish because yarn doesn't respect publishConfig.tag and npm does.

Andarist: ok, makes sense - maybe add a note to some kind of Contributing / Developing section about this?

Andarist:
would it be possible to merge this into master or at least to some next branch? its becoming really huge and its imho hard to follow along right now

mitchellhamilton:
@Andarist yep, I'm gonna merge this into master very soon

PR #826, opened 2018-08-29
styled-elements proof of concept

tkh44:
**What**:
This is adding a new feature to the emotion 10 custom `createElement`. It allows the developer to style elements in a context provider that are applied to the elements as they are created.

```js
/** @jsx jsx */

<StyledElementsProvider
  components={{
    div: { display: 'flex', color: 'hotpink' },
    h1: { fontSize: 18 },
    button: { outline: 'none', border: 'none', backgroundColor: 'purple' }
  }}
>
  <div>
    <h1>Hello</h1>
    <button>Click Me</button>
  </div>
</StyledElementsProvider>
```

Will output:
```html
.emotion-2 {
  display: -webkit-box;
  display: -webkit-flex;
  display: -ms-flexbox;
  display: flex;
  color: hotpink;
}

.emotion-0 {
  font-size: 18px;
}

.emotion-1 {
  outline: none;
  border: none;
  background-color: purple;
}

<div
  className="emotion-2"
>
  <h1
    className="emotion-0"
  >
    Hello
  </h1>
  <button
    className="emotion-1"
  >
    Click Me
  </button>
</div>
```

A few more supported cases as of right now:
```js
<StyledElementsProvider
  components={{ div: ({ fontSize }) => ({ fontSize: fontSize || 8 }) }}
>
  <div fontSize={24} />
</StyledElementsProvider>
```

```js
<StyledElementsProvider
  components={{ div: ({ theme }) => ({ color: theme.color }) }}
>
  <ThemeProvider theme={{ primary: 'green' }}>
    <div />
  </ThemeProvider>
</StyledElementsProvider>
```

**Why**:

Talking to @ChristopherBiscardi about how MDX custom component providers work I wanted to see if we could build something like that into emotion. This concept allows for a more focused approach to styling default elements. This concept should be expanded to include custom elements and add proper support for nested `StyledElementProvider`'s.

**How**:

Added some simple hooks into the `jsx` but I fear I may added some perf regressions with this initial attempt. Any help here would be appreciated.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/next-packages/core/__tests__/__snapshots__/styled-elements.js.snap
+++ b/next-packages/core/__tests__/__snapshots__/styled-elements.js.snap
@@ -1,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Styled Elements - basic element 1`] = `
+.emotion-0 {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  color: hotpink;
+}
+
+<div
+  className="emotion-0"
+/>
+`;
+
+exports[`Styled Elements - css 1`] = `
+.emotion-0 {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  color: hotpink;
+  color: green;
+}
+
+<div
+  className="emotion-0"
+/>
+`;
+
+exports[`Styled Elements - nested elements 1`] = `
+.emotion-2 {
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  color: hotpink;
+}
+
+.emotion-0 {
+  font-size: 18px;
+}
+
+.emotion-1 {
+  outline: none;
+  border: none;
+  background-color: purple;
+}
+
+<div
+  className="emotion-2"
+>
+  <h1
+    className="emotion-0"
+  >
+    Hello
+  </h1>
+  <button
+    className="emotion-1"
+  >
+    Click Me
+  </button>
+</div>
+`;
+
+exports[`Styled Elements - prop function 1`] = `
+.emotion-0 {
+  font-size: 24px;
+}
+
+<div
+  className="emotion-0"
+  fontSize={24}
+/>
+`;
+
+exports[`Styled Elements - prop function with theming 1`] = `
+<div
+  className="emotion-0"
+/>
+`;
+
+exports[`Styled Elements - theming 1`] = `
+.emotion-0 {
+  color: hotpink;
+  color: green;
+}
+
+<div
+  className="emotion-0"
+/>
+`;
--- a/next-packages/core/__tests__/styled-elements.js
+++ b/next-packages/core/__tests__/styled-elements.js
@@ -1,0 +1,88 @@
+// @flow
+/** @jsx jsx */
+import 'test-utils/next-env'
+import { jsx } from '@emotion/core'
+import { StyledElementsProvider } from '@emotion/core/src/styled-element-context'
+import ThemeProvider from '@emotion/provider'
+import renderer from 'react-test-renderer'
+
+test('Styled Elements - basic element', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider
+      components={{ div: { display: 'flex', color: 'hotpink' } }}
+    >
+      <div />
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
+
+test('Styled Elements - nested elements', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider
+      components={{
+        div: { display: 'flex', color: 'hotpink' },
+        h1: { fontSize: 18 },
+        button: { outline: 'none', border: 'none', backgroundColor: 'purple' }
+      }}
+    >
+      <div>
+        <h1>Hello</h1>
+        <button>Click Me</button>
+      </div>
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
+
+test('Styled Elements - css', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider
+      components={{ div: { display: 'flex', color: 'hotpink' } }}
+    >
+      <div css={{ color: 'green' }} />
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
+
+test('Styled Elements - prop function', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider
+      components={{ div: ({ fontSize }) => ({ fontSize: fontSize || 8 }) }}
+    >
+      <div fontSize={24} />
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
+
+test('Styled Elements - theming', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider components={{ div: { color: 'hotpink' } }}>
+      <ThemeProvider theme={{ primary: 'green' }}>
+        <div css={theme => ({ color: theme.primary })} />
+      </ThemeProvider>
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
+
+test('Styled Elements - prop function with theming', () => {
+  const tree = renderer.create(
+    <StyledElementsProvider
+      components={{ div: ({ theme }) => ({ color: theme.color }) }}
+    >
+      <ThemeProvider theme={{ primary: 'green' }}>
+        <div />
+      </ThemeProvider>
+    </StyledElementsProvider>
+  )
+
+  expect(tree.toJSON()).toMatchSnapshot()
+})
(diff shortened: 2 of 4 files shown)


mitchellhamilton:
Seems interesting but what benefit does this have over targeting those elements with child selectors?

Ailrun:
Sound interesting too, but I agree @mitchellhamilton. The feature looks fancy, but do we really need this?

tkh44:
> Seems interesting but what benefit does this have over targeting those elements with child selectors?

That's a valid point in some situations but I think where this will shine is styling basic elements directly and these base styles will be mixed in with css props and className using emotions composition.

In the most simple, devs will set these styles at the top level of their app and just use the semantic html markup while working on their app instead of importing and using styled components. The idea would be to get the basic styles how you want them at a high level and then use the `css` prop to do final touch-ups at the point of use.

tkh44:
Just to clarify, this feature should probably be in its own package, but this is a proof of concept and I wanted to get it working.

Ailrun:
We can already do following, what's the negative effect of that?
```typescript
const App = styled('div')`
  & div {
    display: flex;
    color: hotpink;
  }

  & h1 {
    font-size: 18px;
  }
`;

(
  <App>
    <div>
      hi
    </div>
    <h1>Oh My!</h1>
  </App>
)
```

mitchellhamilton:
What @Ailrun said.

My bigger concern though with this is that imagine there is a new developer on a team using this and they're looking at some styles, how do they know where the styles come from? they look at the element and it doesn't show them at all so they have to ask someone else how it works whereas with global styles/child selectors it can see seen in devtools where it comes from.

tkh44:
The element tag selector has the highest specificity so it would a challenge to customize any of the elements at the point of use.

```js
const App = styled('div')`
  & div {
    display: flex;
    color: hotpink;
  }

  & h1 {
    font-size: 18px;
  }
`;

(
  <App>
    <div css={{ display: 'block' }}> // <-- no effect
      hi
    </div>
    <h1>Oh My!</h1>
  </App>
)
```

I made an example showing this here https://codesandbox.io/s/jpp3mwz4yv

mitchellhamilton:
That's not true with the css prop. https://codesandbox.io/s/3vv2vxx201

mitchellhamilton:
There might be some ridiculously rare cases(as in you would practically have to purposely intend to do it) where the styles with the child selector would be override the ones directly on the element. You could also completely avoid those cases by using the Global component though. (the Global component inserts its style elements before other styles)

mitchellhamilton:
Wait no sorry, my example was wrong, you're right.

mitchellhamilton:
I think using the Global component is still probably better though. Is a there a case where you would only want to apply this to part of an app?

Andarist:
Additional downside of this - might be really confusing for people that not all elements within a subtree are targeted with this, because naturally only those managed with emotion can be.

tkh44:
> I think using the Global component is still probably better though. Is a there a case where you would only want to apply this to part of an app?

I could see an application having the top-level definitions defining the base styles and resets and then making adjustments deeper in the application where the literal context of the styling changes. Ex. The shell of a content page will have different styles than the actual content or article portion of the page.

I see your point about this being confusing if it were used as a primary method of styling. I wonder if having an option that controls whether Provider styles cascade using composition or if they set a hard boundary and scoped only to the closest parent. My intuition here is to set it as scoped by default.

Component authors could also use this as a way for consumers to style the internals of the component. By exposing the `components` as a prop to the end users, they are able to provide custom styling functionality without much work. This could mean both open source component libraries and your average app with a common components folder. I think both are valid use cases.

I would like to figure out a way to match on custom components but I'm unsure the best way to do this.

mitchellhamilton:
>if they set a hard boundary and scoped only to the closest parent

I don't really understand what you mean by this.

WRT matching custom components, you could change the components prop to something like what a Map's constructor accepts

```jsx
<StyledElementsProvider
  components={[
    ['div', { display: 'flex', color: 'hotpink' }],
    ['h1', { fontSize: 18 }],
    [
      'button',
      {
        outline: 'none',
        border: 'none',
        backgroundColor: 'purple'
      }
    ]
  ]}
>
  <div>
    <h1>Hello</h1>
    <button>Click Me</button>
  </div>
</StyledElementsProvider>
```

tkh44:
This is neat

```js
/** @jsx jsx */
import system from 'system-components/emotion'

<StyledElementsProvider
  components={{
    div: system('space', 'width', 'color')
  }}
>
  <div>Click Me</div>
</StyledElementsProvider>
```

mitchellhamilton:
Here's my idea(not related to emotion at all), it kind of like `react-select@2`'s new components prop and the MDX component provider you talked about before.

You use it like this, it could easily be statically typed and it uses forwardRef so it's completely transparent.
```jsx
let {
  ComponentProvider,
  components: { SomeComponent }
} = createComponents({
  SomeComponent: props => (
    <h1
      {...props}
      css={{
        color: "green"
      }}
    />
  )
});

```

https://codesandbox.io/s/01klkyl7vv

tkh44:
I like that quite a bit. If you were to give each component the emotion theme on the props it would be fairly useful.

mitchellhamilton:
You can already use the theme from the css prop in emotion 10.
https://github.com/emotion-js/next#with-the-css-prop

Ailrun:
@mitchellhamilton Is there any reason for not using an array of `[Component, style]`? I mean, with that object syntax, it's hard to see which one correspond to the style. (For example, function.name, component.displayName, ... can be the key of that object)

mitchellhamilton:
@Ailrun could you give an example of how that would look?

Ailrun:
@mitchellhamilton I mean, something like
```typescript
<React.Fragment>
  <SomeComponent>Hello CodeSandbox</SomeComponent>
  <ComponentProvider
    // the components prop can also be a function that
    // accepts the parent components, like ThemeProvider
    // the value of the components props should be declared
    // outside of render(or cached) so that it doesn't change the context on every render
    components={[
      [SomeComponent, props => (
        <h1
          {...props}
          css={{
            color: "hotpink"
          }}
        />
      )]
    ]}
  >
    <SomeComponent>Hello Again!</SomeComponent>
  </ComponentProvider>
</React.Fragment>;
```

mitchellhamilton:
@Ailrun I don't really understand why that would be better? I think it would make it more confusing and hard to use since if you wanted to let someone customize the components, you would have to export the provider _and_ the components rather than only the provider.

The name and displayName properties are never used at all. It's only the key of the object that's used.

Ailrun:
@mitchellhamilton Oh,
```jsx
props => (
        <h1
          {...props}
          css={{
            color: "hotpink"
          }}
        />
      )
```
*is the definition* of component? I thought it somehow reference original component.

mitchellhamilton:
Also, that would be much harder to statically type since currently it could be typed like this in flow whereas I have no idea how it would be typed with your API.
```jsx

type createComponents = <Components: { [key: string]: React.ComponentType<*> }>(
  defaultComponents: Components
) => {
  components: Components,
  ComponentProvider: React.ComponentType<{ components: Components }>
}
```

mitchellhamilton:
@Ailrun Yes, that's correct.

Ailrun:
That's my total misunderstanding. Can I hide related comments to focus on discussion?

mitchellhamilton:
Yeah, if you want to.

Ailrun:
However, what you mentioned is still hard to give types. Specifically, in TS, to give children of `SomeComponent` to `ComponentProvider`, you need a way to give a type to value `SomeComponent`.

mitchellhamilton:
@Ailrun not really sure what you mean but I just tried typing it in TS and it seems to work fine. https://codesandbox.io/s/j3p000wo95

Ailrun:
@mitchellhamilton I mean, your line 42 use the type defined in line 7, not a type defined in line 22. You cannot use type of line 22 without explicitly accessing it.

mitchellhamilton:
I still don't understand, what's on line 42 is just `</ComponentProvider>`.

Ailrun:
@mitchellhamilton Oh, I mean line 41. My miss.

mitchellhamilton:
Oh, I get what you mean. The point is that they should have the same type definition though, you should just be switching out the implementation with the provider.

mitchellhamilton:
And them having the same type definition is enforced by the type system.

mitchellhamilton:
That restriction isn’t meant to be specific to TS, it’s how it should always be used.

I don’t understand why the second thing you mentioned is bad.

Ailrun:
@mitchellhamilton Second one is the problem you mentioned in https://github.com/emotion-js/emotion/pull/826#issuecomment-417142619, isn't it?
And for restriction, even if you said it should satisfy that restriction, but many JS users will change the props type, won't they?

mitchellhamilton:
The problem I was referring to in https://github.com/emotion-js/emotion/pull/826#issuecomment-417142619 was that you would have to have the component to change the implementation, the current API doesn't require you to have the component to change the implementation, you of course need the component to use it though.

> And for restriction, even if you said it should satisfy that restriction, many JS users will change the type of props, won't they?

That's true, that's just the cost of not using a type system. I doubt that would happen often though because there shouldn't really be a very big incentive for doing that since the intention for it was that say if I create a component or a bunch of components that use a set of components internally, I can allow someone who uses the root component(s) to change the implementation of the internal components. As in the consumer wouldn't render the internal components themselves.

PR #828, opened 2018-08-29
Add ClassNames component

mitchellhamilton:
**What**:
Add ~~Style~~ ClassNames component.

`css` and `cx` work just like the exports from `emotion` except they work with emotion 10's SSR and the options that can be provided via context.

It also accepts the `theme`.

```jsx
<ClassNames>
  {({ css, cx, theme }) => {
    let secondClassButItsInsertedFirst = css`
      color: green;
    `
    let firstClassButItsInsertedSecond = css`
      color: hotpink;
    `

    return (
      <div
        className={cx(
          firstClassButItsInsertedSecond,
          'some-other-class',
          secondClassButItsInsertedFirst
        )}
      />
    )
  }}
</ClassNames>
```

**Why**:

To account for cases when components accept props like `wrapperClassName` and etc.

**How**:

**Checklist**:

* [ ] Documentation (gonna add to #799 after #824 is merged)
* [x] Tests
* [x] Code complete

I'm not totally sure about the API, the name of the component and whether it should be in the core package. I'd love to hear people's thoughts on it!!

> Note: This is based on #824 not master.


next-packages/core/src/style.js
```diff
@@ -0,0 +1,127 @@
...
+    let arg = args[i]
+    if (arg == null) continue
+
+    let toAdd
+    switch (typeof arg) {
+      case 'boolean':
+        break
+      case 'function':
+        if (process.env.NODE_ENV !== 'production') {
+          console.error(
```
Andarist: is this needed? it's already "next major version of Emotion" :D

next-packages/core/src/style.js
```diff
@@ -0,0 +1,127 @@
...
+import { withCSSContext } from '@emotion/core'
+
+type ClassNameArg =
+  | string
+  | boolean
+  | (() => ClassNameArg)
+  | { [key: string]: boolean }
+  | Array<ClassNameArg>
+
+let classnames = (args: Array<ClassNameArg>): string => {
```
Andarist: can u explain more why this is duplicated? I suspect it's like a global caching registry issue, but couldnt this share the implementation with original cx, css & merge?

next-packages/core/src/style.js
mitchellhamilton: Nope, I made this a while ago and copied `classnames` from v9.

next-packages/core/src/style.js
mitchellhamilton: I’m just lazy, I think I’ll move it and merge into `@emotion/utils`. There’s no point in sharing css and cx since they need slightly different things in here and in create-emotion.

mitchellhamilton:
Just FYI, I’m mainly looking for a review of the API over code review.

mitchellhamilton:
Because there needs to be somewhere to render the style tags and the exact styles that need to be rendered there must be known when the style tag is being rendered, that's not possible if using something like `css` from `emotion`.(by SSR, I'm referring to emotion 10's automatic SSR)

Andarist:
> that's not possible if using something like css from emotion

Why? Isnt the css call synchronous anyway?

mitchellhamilton:
Let's say you have this component, how is it possible to do automatic SSR?
```jsx
let SomeComponent = () => (
  <div
    className={css`
      color: hotpink;
    `}
  />
)
```

mitchellhamilton:
Also, I just want to repeat this, this is only intended for cases when components accept props like `wrapperClassName` and etc.

Andarist:
Ok, I get now how it's impossible for auto-SSR.

The API looks good, although it slightly shows that even though current approach brings easier SSR to the table it also has its drawbacks such as having to know that sometimes `<Style/>` has to be used.

Also - not sure what's the best name for this component, maybe `<ServerStyle/>`?

mitchellhamilton:
This isn't just for auto-SSR though, it's also for providing the cache via context so that things like working in iframes, getting the theme and etc. work.

The `ServerStyle` name doesn't really make sense since it's not just for SSR.

Andarist:
> This isn't just for auto-SSR though, it's also for providing the cache via context so that things like working in iframes, getting the theme and etc. work.

Is this meant to be a replacement for previously presented APIs? You've already demoed i.e. rendering to iframes etc.

Sorry for asking so many questions, I'm not too familiar with v10 APIs - lacking a "big picture" and how existing pieces fit together.

mitchellhamilton:
No, this isn't meant to be a replacement for the css prop or styled, it's meant for for the small number of cases that they can't solve.

Sorry, when I said providing the cache via context I meant consuming it via context.

mitchellhamilton:
Yeah, that's mostly correct though this isn't very different from the other consumers, they all work in the same way, they just provide different APIs to insert styles. This API is just less convenient but more flexible.

I just thought of another name, `ClassNames`, any thoughts on that?

tkh44:
I like this name better.

I just thought of this, but what if it was just called `Emotion`

mitchellhamilton:
@Andarist suggested that too, I feel like it doesn’t convey much meaning about what it does though.

Ailrun:
For me, it's also hard to understand. How about `SerialStyles` or something like that?

mitchellhamilton:
I like `ClassNames` because it conveys that it’s about class names since what it does is let you get a class name from some styles and merge together class names. 

I don’t really get what `SerialStyles` means and how it would convey what this component does

Ailrun:
Oh, I thought about `css` props part only. With `cx` props, what you said make sense.

Ailrun:
`css` props is different original `css`, so the component name should convey the differences, shouldn't it?

mitchellhamilton:
There isn’t really a difference between the css function in this component and the css function in v9 except for the implementation.

mitchellhamilton:
Are you talking about the css prop in emotion 10 or the css function in this component?

Ailrun:
It's different in SSR, isn't it?

mitchellhamilton:
Kind of but the usage of it is the same which is the part that people care about

Ailrun:
I mean the difference between `css` function from `@emotion/core` and from props of this component.

Ailrun:
(I believe) I know interfaces of those functions are same, but if it work differently it should convey those differences as a name, shouldn't it?

mitchellhamilton:
If it works the same from the user’s perspective, why should they have to learn a new name for something that does the same thing?

Ailrun:
Actually, that's where my `SerialStyle` comes from, since I thought those difference should be reflected in the name of component, not a name of actual props.
My reasoning is like following.
1. If we have different name for `css` function from `@emotion/core` and the `css` function for the children of this component, refactoring codes those use `css` function becomes harder.
2. But IMO, since there are differences, those difference should be reflected somewhere.
3. Where would be appropriate for this? I cannot think other than component name.

But what you said in https://github.com/emotion-js/emotion/pull/828#issuecomment-417815963 makes sense, so I need to think more about this.

mitchellhamilton:
> If we have different name for css function from @emotion/core and the css function for the children of this component, refactoring codes those use css function becomes harder.

`css` from `@emotion/core` and the `css` function for the children of this component return totally different thing things though, the former returns a description of the styles while the latter returns a class name.

Ailrun:
Oh, yes, then we don't need to worry about refactoring. Then instead of concerning about new name for component, giving appropriate name for `css` in this component would be better. Thank you for enlightening.

TrySound:
Guys, there is a better solution than wrapping every component with another one render prop.

We can have the same api but with some requirements about where functions may be called.

```js
import * as React from 'react';
import { Provider, css } from '@emotion/contextual';

export const Component = () => (
  <Provider>
    <div className={css({ width: 100, height: 100, background: '#000' })}>
      Content
    </div>
  </Provider>
)
```

The point here is that since 16.5 we are able to consume context with `Context.unstable_read`, which works in render phase. This api actually doesn't work with ssr yet (that's why it's unstable), but we already can start experiments on client side.

Ref https://github.com/facebook/react/pull/13139

mitchellhamilton:
This component doesn’t only need context though, it also needs to render elements.

pksunkara:
This is needed for emotion to work with https://github.com/reactjs/react-modal. Check [this](https://github.com/storybooks/storybook/commit/775fe4e9c7de3d45b442b8092dc3ffa0bf746f34) commit in @storybooks.

mitchellhamilton:
I'm gonna merge this into #824 because I want to try some stuff out. If anyone has any comments on it, you can still leave them here though.

Issue #644, opened 2018-05-07
Stop using componentWillMount


luisherranz:
- `emotion` version: 9.1.3
- `react` version: 16.3

Relevant code.

```javascript
        componentWillMount() {
          if (this.context[channel] !== undefined) {
            this.unsubscribe = this.context[channel].subscribe(
              setTheme.bind(this)
            )
          }
        }
```

What you did:
Update to React 16.3 and use `strict mode`.

What happened:
This warning appeared: `componentWillMount: Please update the following components to use componentDidMount instead: Styled(div), Styled(li), Styled(span), Styled(ul)`

Reproduction:
Just use `styled` within React `strict-mode` or [React `Async`](https://github.com/koba04/react-fiber-resources):
https://codesandbox.io/s/732mw5w53j

Problem description:
`componentWillMount` is going to be deprecated and should be avoided from now on.

Suggested solution:
This is the `componentWillMount` which is causing the warning:
https://github.com/emotion-js/emotion/blob/82fd71a8f9fda7df31b4810b17a255a75500686b/packages/create-emotion-styled/src/index.js

That logic could be moved to the `componentDidMount` or the `constructor`. Whichever is the case, let me know and I'll do a PR.

ChrisCallanJr:
made PR [#830](https://github.com/emotion-js/emotion/pull/830) to fix this

mitchellhamilton:
As I mentioned in #830, this is going to be fixed in emotion 10.

FezVrasta:
Closing as fixed by Emotion 10

PR #830, opened 2018-08-30
updated componentWillMount to componentDidMount

ChrisCallanJr:
**What**:
This is addressing issue [ #644](https://github.com/emotion-js/emotion/issues/644)

**Why**:
`componentWillMount` is deprecated as of React 16.3.

**How**:
By replacing it with the `componentDidMount` method

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ x] Code complete

This is my first ever open source PR, so sorry if I'm doing anything wrong!


mitchellhamilton:
Hey @ChrisCallanJr, thank you so much for the PR!!!!

We're working on emotion 10 right now which will remove the usage of `componentWillMount` so this isn't necessary and doing it like this would break some things(theming in SSR) so unfortunately we can't accept this.

Thanks again for the PR though!!

PR #836, opened 2018-09-02
inferno-emotion: Initial package

LucaT1:
**What**:
Added a new package to implement almost native support for inferno in emotion

**Why**:
The only current supported way to use inferno with emotion is via `inferno-compat` wich provides unacceptable performance.

**How**:
Created a new package called `inferno-emotion` specular to the React one, implemented all possible tests, tough need some help on the babel-plugin configuration since it looks like it's not picking it up and there are some other questions tough the code.

**Checklist**:
- [x] Documentation
- [x] Tests
- [ ] Code complete ( Is this relevant? )

There are lots of comments through the code, so a careful revision is highly advised. Furthermore, I need some information about why the babel plugin is not working in my tests with jest. Anyway, it seems to work pretty nicely with inferno real world apps. 

Although this implementation is not the fastest is the quickest to implement without adding complex voodoo magic to the babel plugin. Here is a more complete explanation from the `index.js` file of the package:
```
Since the main difference between Inferno and React
is the fact that they use different methods to create VNodes,
which boosts speed in the first package, simplicity in the second,
we simply create a mockup for the missing createElement function

Please note that this lowers performance improvements
when using styled, considering that lots of the extra
the speed that inferno has comes from precompiling the type of the vnode
via its babel plugin. This could be ported to emotion as well but would
definitely take some time a modification of the current createEmotionStyled utility

So mind that the current solution is better than using inferno-compat
alias, but also significantly lowers performance compared to a complete
solution that would be harder to achieve.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "postinstall": "opencollective postinstall"
   },
   "resolutions": {
+    "**/inferno": "^5.5.0",
     "**/react": "^16.3.2",
     "**/react-dom": "^16.3.2",
     "**/browserslist": "^3.2.8"
@@ -92,6 +93,7 @@
     "preact-render-to-json": "^3.6.6",
     "prettier": "1.10.2",
     "raf": "^3.4.0",
+    "inferno": "^5.5.0",
     "react": "^16.3.2",
     "react-dom": "^16.3.2",
     "react-router-dom": "^4.2.2",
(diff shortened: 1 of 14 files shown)


mitchellhamilton:
Thanks so much for this PR but unfortunately I don't think this is something we can accept. Supporting Preact has already been a large burden and we'd like to be able to use new React APIs without worrying about whether the new APIs are in React alternatives. We're also making some API changes in the next major version which will involve the removal of create-emotion-styled.

PR #841, opened 2018-09-06
log error message if no units in shorthand properties

### LINKED ISSUE #752: [primitives] Failed to parse declaration margin or padding
Congrats on launching Emotion Primitives 🎉

It worked great on my very basic tests, but not with some styles like `margin` or `padding`.
Specific margin like `margin-bottom` works.

- `@emotion/primitives` version: ^9.2.5
- `react` version: 16.4.1
- `react-native` version: 0.56.0

<div>

</div>

```javascript
import React, { PureComponent } from 'react'

import styled from '@emotion/primitives'

  flex: 1;
  justify-content: center;
  align-items: center;
  background-color: #f5fcff;
`

const Title = styled.Text`
  font-size: 20;
  text-align: center;
  margin: 10; // <- Issue here. margin-bottom works.
  color: hotpink;
`

const Description = styled.Text`
  text-align: center;
  color: #333333;
  margin-bottom: 5;
`

export default class App extends PureComponent {
  render() {
    return (
      <Container>
        <Title>Welcome to Emotion Primitives</Title>
        <Description>Style and render primitives across targets.</Description>
      </Container>
    )
  }
}
```

nitin42:
**What**: Fixes #752 

**Why**:

**How**: Improve error handling when transforming style pairs in `primitives-core`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

```js
const Text = styled.Text`
 margin: 20px;
 padding: 20;
`
```

this will log

screenshot 2018-09-06 at 5 36 57 pm


packages/native/test/native-styled.test.js
```diff
@@ -136,4 +136,15 @@ describe('Emotion native styled', () => {
...
+  it('Log error message if units are not specified when using shorthand properties', () => {
+    const Text = styled.Text`
+      margin: 20px;
+      padding: 50;
+    `
+
+    const tree = renderer.create(<Text>Hello World</Text>).toJSON()
+
+    expect(tree).toMatchSnapshot()
+  })
```
mitchellhamilton: Could you make it so this checks the error message works? something like this.
https://github.com/emotion-js/emotion/blob/76828f1f8394af31f046085cc07e259b92f56341/packages/emotion/test/warnings.test.js

packages/primitives-core/src/css.js
```diff
@@ -121,5 +121,25 @@ function convertStyles(str: string) {
 
   parsedString.forEach(convertPropertyValue, stylePairs)
 
-  return transform(stylePairs)
+  try {
```
Andarist: aint sure entirely about this - but `primitives` might be used on web too, maybe this try/catch should be wrapped with development check? wdyt @mitchellhamilton ?

mitchellhamilton: Yeah, that might be a good idea.

Andarist: Gonna merge this regardless of this. I suspect that we might want to drop primitives & native packages in emotion@11 because it seems we don't have capacity to handle their issues.

The code as-is improves the story around those packages right now, so it IMHO deserves being merged.

PR #843, opened 2018-09-07
Add support for `as` prop

### LINKED ISSUE #837: `as` prop
Styled-components v4 has introduced this, which seems like a useful thing to me.

It is neat to specify the underlaying component dynamically instead of creating all variants upfront. The feature was also introduced previously in completely separate project - https://github.com/reakit/reakit (built in on top of styled-components v3). Not sure if they have consulted/discussed it, but seems like it has proven to be useful as it got implemented in the core API.

Andarist:
closes #837

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


next-packages/styled-base/src/index.js
```diff
@@ -29,30 +28,26 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
   }
   let identifierName
   let shouldForwardProp
+  let optionsShouldForwardProp
```
Andarist: I dont like this, but got the expected behaviour in the least amount of code. Would appreciate refactor tips if you have any

tadeuszwojcik:
Is there a change to merge that change soon ? :) or some other way to play around with it?

Andarist:
not sure, because it seems that this change might be breaking interoperability with some libraries - u can read more about it in this thread - https://github.com/emotion-js/emotion/issues/837

Andarist:
Sure, it can and next major version is the best moment to do it. Some interop solution with other libraries should be considered though, because that's existing ecosystem that we should support. If you have any ideas/opinions on how to do that - join the discussion :)

JoshRosenstein:
@Andarist  looking at the other discussion about breaking some of the semantic-ui / react-bootstrap components, and looking at your current draft, how about having the option to disable the as-prop feature.
For example having a option api 'enableAs' (prob a better name for it) that is default to true.
And pass that to your getShouldForwardProp, and also use it to determine finalTag.

```javascript
const testOmitPropsOnComponent = (asEnabled:boolean)=>(key: string) =>
  key !== "theme" && key !== "innerRef" && asEnabled? key !== "as":true ;

export const testAlwaysTrue = () => true;

export const getShouldForwardProp = (tag, asEnabled) =>
  typeof tag === "string" &&
  // 96 is one less than the char code
  // for "a" so this is checking that
  // it's a lowercase character
  tag.charCodeAt(0) > 96
    ? testOmitPropsOnStringTag
    : testOmitPropsOnComponent(asEnabled);
````

then in finalTag..
```javascript
const finalTag = options.enableAs? props.as || baseTag : baseTag ;
````

so jquense snippet would work by just disabling.. ?
```javascript
import { Button } from 'semantic-ui-react'

const FooterButton = styled(Button,{enableAs:false})`
  margin-top: 30px
`

<FooterButton as="a" href="/link/styled/as/Button"/>
````

For me, im a big fan for the as prop, but i do agree with you in trying to support existing ecosystems. I think adding this additional api, is worth it considering the amount of people that want this feature and the amount of people that also rely on systems like semantic-ui and react-bootstrap.

**Comment by Andarist** [comment]
@mitchellhamilton could u check my logic? I've decided to do some hoops around as & shouldForwardProp compatibility, the code is more convoluted but I think it might reflect better what people would expect (and respecting `shouldForwardProp` of the `finalTag` in most cases aligns with styled-components behaviour)

When we iron out the logic I'll write more tests to cover all edge cases

File: packages/styled/__tests__/as-prop.js [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+      <Title data-testid={h1Title}>My Title</Title>
+      <Title data-testid={h3Title} as="h3">
+        My Subtitle
+      </Title>
+    </article>
+  )
+
+  expect(getByTestId(h1Title).tagName).toBe('H1')
+  expect(getByTestId(h3Title).tagName).toBe('H3')
+})
```
mitchellhamilton: I disagree with this behavior. I think if you pass a composite base, we should forward `as` unless you explicitly pass a `shouldForwardProp`.

mitchellhamilton:
Sorry I didn't respond to this for so long! I think I missed the notification 😢

packages/styled/__tests__/as-prop.js
Andarist: I was not sure how to handle this case. The reasoning behind this solution is that most people don't use `shouldForwardProp` feature and this logic makes more common scenarios easier.

With string `as` here we kinda implicitly know which props should be forwarded regardless of the base type and it seems to me that this would be the expected outcome for most people. It also aligns the default behaviour to `styled-components` so this would make migrating from one lib to another easier.

I'm not overly attached to the logic here, so feel free to decide how this should work and I'll adjust the code.

Andarist:
> Sorry I didn't respond to this for so long! I think I missed the notification 😢

No worries!

packages/styled/__tests__/as-prop.js
mitchellhamilton: I feel like the default should handle the most common use case and I feel like the most common use case for passing a non string base is that it handles `as`, while there are certainly some use cases where styled should handle `as` for composite bases, I think it won't be as common.

mitchellhamilton: Yeah, choosing between compatibility with `styled-components` or (IMO) better behavior is hard. I feel like for now, let's make it so by default composite bases forwards as(unless a shouldForwardProp that doesn't forward as is passed) and see what people think and then go from there. I feel pretty confident with it since I feel like a more real world usecase would be more like this.
```jsx
const Link = styled.a``

<Link as={DifferentLink} />
```

packages/styled-base/src/utils.js
```diff
@@ -17,6 +17,12 @@ export const getShouldForwardProp = tag =>
     ? testOmitPropsOnStringTag
     : testOmitPropsOnComponent
 
+export const hasDefaultShouldForwardProp = (tag, shouldForwardProp) =>
+  shouldForwardProp ===
+  (typeof tag === 'string'
+    ? testOmitPropsOnComponent
```
JoshRosenstein: testOmitPropsOnStringTag ?

packages/styled/__tests__/as-prop.js
tkh44: My take on this.

```
1. if emotion component forward `as`
2. if composite component forward `as`
3. if emotion is writing the final string tag use the `as` prop to overwrite the tag and *do not forward* `as`
4. if custom shouldForwardProp returns `false` for `'as'` *do not forward* `as`
```

mitchellhamilton:
I've updated the implementation here to be what @tkh44's and my thinking is. If it turns out we're wrong, we can always change it but for now, we think this is the best approach so I'm going to merge it and release it in the v10 beta

Andarist:
Thanks @mitchellhamilton for taking care of this! Was planning to do it, but couldn't find time lately to finish the work here.

PR #852, opened 2018-09-14
Upgrade Babel 7 beta to stable version

LinusU:
**What**: This patch bumps all dependencies on the Babel 7 beta to point to the stable version instead.

**Why**: I'm having a react-native app and there is a deep dependency on `babel-plugin-emotion` which pulls in the beta version of some babel packages. I think that this is causing a conflict somewhere, so I'm trying to make sure that all dependencies use the stable version.

**How**:  Bumped the version specifiers in the `package.json` files. Also bumped `react-native` to 0.57 which uses Babel 7 stable as well, and then React to 16.5.0 since that is what react-native uses. Then used yarn to update the lockfiles as well.

**Checklist**:

- [ ] Documentation **N/A**
- [ ] Tests **N/A**
- [ ] Code complete **N/A**


Andarist:
Thanks for the PR! You'd have to fix the test setup though, because CI is failing now.

LinusU:
- Changed the TypeScript code slightly to work in current versions and `next`.
- Removed Babel plugins that aren't used

next-packages/stylis/types/test.ts
```diff
@@ -74,11 +74,7 @@ stylis.set({
   },
 });
 // $ExpectError
-stylis.set({
-  prefix() {
-    return 'hi';
-  },
-});
+stylis.set({ prefix: () => 'hi' });
```
Andarist: cant both be supported? cc @Ailrun

LinusU: The problem was that TypeScript next reports the error on a different line, one further down. It's still the same error as before.

I can add these lines (👇) as well if you want to? I just changed to a arrow function to make it a bit more condense.

```typescript
// $ExpectError
stylis.set({ prefix () { return 'hi'; } });
```

Ailrun: Both supported, but how about just preserving previous test as you mentioned?
I mean, how about just changing it into oneliner?

mitchellhamilton: tbh, I think it's better as an arrow function since i think the reason this broke was that I formatted it with prettier and having it as an arrow function means it will stay on one line.

Ailrun: Oh, it's changed because of `prettier`... Then alright. Let's leave this as an arrow function.

mitchellhamilton:
@LinusU just cut a release.

LinusU:
Thanks!

PR #853, opened 2018-09-14
jest-emotion: sometimes method `getAttribute` is not avail

Schubidu:
**What**: bug in jest-emotion method `getAttribute` is not avail, while jest creating snapshot

**Why**: in combination with `ReactDom.createPortal` it might can be `getAttribute` is not avail

**How**: just check if the method is avail

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


mitchellhamilton:
Could you create a reproduction of this happening? I don’t see how it could and especially how portals would affect it.

Schubidu:
It’s deep in a customer project. My result while debug my issue on the project. Well may I can give the next days the error message I get on the console.

PR #854, opened 2018-09-14
Add typescript types for channel and createBroadcast in emotion-theming

CarlosBonetti:
**What**:
Added typescript types for channel and createBroadcast.

**Why**:
To match exported types with actual objects being exported by emotion-theming.

**How**:
By manually modifying `emotion-theming` typescript type definitions file.

**Checklist**:

- [ ] Documentation - N/A
- [X] Tests
- [X] Code complete

### DIFF
--- a/packages/emotion-theming/types/index.d.ts
+++ b/packages/emotion-theming/types/index.d.ts
@@ -20,3 +20,7 @@ export interface EmotionThemingModule<Theme> {
     ThemeProvider: ThemeProviderComponent<Theme>;
     withTheme<Props>(component: ComponentClass<Props> | SFC<Props>): ComponentClass<OptionalThemeProps<Props, Theme>>;
 }
+
+export const channel: string;
+
+export function createBroadcast(initialState: any): any;
--- a/packages/emotion-theming/types/tests.tsx
+++ b/packages/emotion-theming/types/tests.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import * as emotionTheming from '../';
 // tslint:disable-next-line:no-duplicate-imports
-import { ThemeProvider, withTheme, EmotionThemingModule } from '../';
+import { ThemeProvider, withTheme, EmotionThemingModule, channel, createBroadcast } from '../';
 
 const theme = { primary: "green", secondary: "white" };
 const CompSFC = (props: { prop: boolean }) => <div />;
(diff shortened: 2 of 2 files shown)


packages/emotion-theming/types/index.d.ts
```diff
@@ -20,3 +20,7 @@ export interface EmotionThemingModule<Theme> {
     ThemeProvider: ThemeProviderComponent<Theme>;
     withTheme<Props>(component: ComponentClass<Props> | SFC<Props>): ComponentClass<OptionalThemeProps<Props, Theme>>;
 }
+
+export const channel: string;
+
+export function createBroadcast(initialState: any): any;
```
Ailrun: It does not return `any`. It clearly has `publish`, `subscribe`, `unsubscribe` as its properties.
And you could use generic type to remove `any` for `initialState` also.

packages/emotion-theming/types/index.d.ts
```diff
@@ -20,3 +20,7 @@ export interface EmotionThemingModule<Theme> {
     ThemeProvider: ThemeProviderComponent<Theme>;
     withTheme<Props>(component: ComponentClass<Props> | SFC<Props>): ComponentClass<OptionalThemeProps<Props, Theme>>;
 }
+
+export const channel: string;
```
Ailrun: How about explicitly using `'__EMOTION_THEMING__'` type?

packages/emotion-theming/types/index.d.ts
```diff
@@ -21,6 +21,14 @@ export interface EmotionThemingModule<Theme> {
...
-export const channel: string;
+export const channel: "__EMOTION_THEMING__";
 
-export function createBroadcast(initialState: any): any;
+export type BroadcastListener<Theme = {}> = (state: Theme) => void;
+
+export interface Broadcast<Theme = {}> {
+    publish: (nextState: Theme) => void;
+    subscribe: (listener: BroadcastListener<Theme>) => number;
+    unsubscribe: (unsubID: number) => void;
```
Ailrun: Is there any reason for using arrow function type instead of method types?
(In TS, they are different in terms of assignability)

CarlosBonetti: No special reason. I guess the method type is better in that case?

CarlosBonetti:
I think the `ci/circleci: lint_and_typescript` check is failing for an unrelated reason...

Ailrun:
@CarlosBonetti Yes, it's broken from `@emotion/stylis`. I'll fix it.

PR #856, opened 2018-09-15
Add Typescript typings for `primitives` and `primitives-core`

Iamsamwhoami:
**What**:
Add Typescript types for `@emotion/primitives`.

**Why**:
To be able to use `@emotion/primitives` type-safely in Typescript.

**How**:
Adding `types` folders following the ish same setup as `react-emotion` with some changes to reflect how `@emotion/primitives` works.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

The type setup was quite complicated so I am not 100% sure I've typed everything correctly, but it should work ™️ 🙈

### DIFF
--- a/packages/primitives-core/package.json
+++ b/packages/primitives-core/package.json
@@ -7,7 +7,8 @@
   "module": "dist/index.esm.js",
   "files": [
     "src",
-    "dist"
+    "dist",
+    "types"
   ],
   "scripts": {
     "test:typescript": "dtslint types"
--- a/packages/primitives-core/package.json
+++ b/packages/primitives-core/package.json
@@ -3,14 +3,19 @@
   "version": "9.2.7",
   "description": "Shared utilities for emotion primitives and native",
   "main": "dist/index.cjs.js",
+  "types": "types/index.d.ts",
   "module": "dist/index.esm.js",
   "files": [
     "src",
     "dist"
   ],
+  "scripts": {
+    "test:typescript": "dtslint types"
+  },
   "license": "MIT",
   "dependencies": {
     "@emotion/is-prop-valid": "^0.6.3",
+    "@types/react-native": "^0.56.19",
     "css-to-react-native": "^2.2.1"
   },
   "peerDependencies": {
--- a/packages/primitives-core/types/common.d.ts
+++ b/packages/primitives-core/types/common.d.ts
@@ -11,6 +11,10 @@ import {
   ClassInterpolation
 } from 'create-emotion';
 
+export interface StyledOptions {
+  shouldForwardProp?: (name: string) => boolean;
+}
+
 export type BasePrimitivesStyle = ViewStyle | TextStyle | ImageStyle;
 
 export type BaseInterpolation =
--- a/packages/primitives-core/types/common.d.ts
+++ b/packages/primitives-core/types/common.d.ts
@@ -5,17 +5,21 @@
 
 import { ViewStyle, TextStyle, ImageStyle } from 'react-native';
 
-import {
-  Emotion,
-  ArrayInterpolation as BaseArrayInterpolation,
-  ClassInterpolation
-} from 'create-emotion';
-
 export interface StyledOptions {
   shouldForwardProp?: (name: string) => boolean;
 }
 
 export type BasePrimitivesStyle = ViewStyle | TextStyle | ImageStyle;
+
+export interface BaseArrayInterpolation extends Array<BaseInterpolation> {}
+
+export interface ClassInterpolation extends Function {
+  __emotion_real: any;
+  __emotion_styles: BaseInterpolation[];
+  __emotion_base: ClassInterpolation;
+  __emotion_target: string;
+  __emotion_forwardProp: undefined | null | ((arg: string) => boolean);
+}
 
 export type BaseInterpolation =
   | undefined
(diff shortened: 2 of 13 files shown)


packages/primitives/types/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ImageProps, ViewProps, TextProps } from 'react-native';
+
+export * from 'emotion';
+
+export type ThemedReactEmotionInterface<Theme extends object> = CreateStyled<
+  Theme
+>;
+
+declare module 'react-native' {
+  interface ViewProps {
```
Iamsamwhoami: Typings for `react-primitives` doesn't seem to expose these, so taking them from `react-native` typings instead.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-primitives/index.d.ts

packages/primitives-core/types/index.d.ts
```diff
@@ -0,0 +1,56 @@
+// Definitions by:
+// Junyoung Clare Jang <https://github.com/Ailrun>
+// Sam Pettersson <https://github.com/iamsamwhoami>
+// TypeScript Version: 2.3
+
+import { StyleSheet } from 'react-primitives';
+import { Emotion } from 'create-emotion';
+import * as React from 'react';
```
Ailrun: Why do we need this `react` import?

packages/primitives-core/types/common.d.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type BaseInterpolation =
+  | undefined
+  | null
+  | boolean
+  | string
+  | number
+  | TemplateStringsArray
+  | BasePrimitivesStyle
+  | BaseArrayInterpolation
```
Ailrun: This will allow some weird things, like
```
[someCSSObjectWhichIsNotPrimitiveStyle]
```
Making `BaseArrayInterpolation` in this file would work better.

packages/primitives-core/types/common.d.ts
```diff
@@ -0,0 +1,41 @@
...
+  | ClassInterpolation;
```
Ailrun: This one also make similar problem, since it refers `Interpolation` of `create-emotion` as the type for `__emotion_styles`.

packages/primitives-core/types/react.d.ts
```diff
@@ -0,0 +1,86 @@
+// Definitions by:
+// Junyoung Clare Jang <https://github.com/Ailrun>
+// Sam Pettersson <https://github.com/iamsamwhoami>
+// TypeScript Version: 2.3
+
+import { ComponentType, ComponentClass, Ref, SFC } from 'react';
+import { ClassInterpolation } from 'create-emotion';
```
Ailrun: One more `ClassInterpolation` related thing.

packages/primitives-core/types/common.d.ts
Iamsamwhoami: @Ailrun oh, totally missed that, I've pushed another commit, looking good now? :)

packages/primitives-core/types/common.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface BaseArrayInterpolation extends Array<BaseInterpolation> {}
+
+export interface ClassInterpolation extends Function {
+  __emotion_real: any;
+  __emotion_styles: BaseInterpolation[];
+  __emotion_base: ClassInterpolation;
+  __emotion_target: string;
+  __emotion_forwardProp: undefined | null | ((arg: string) => boolean);
+}
```
Ailrun: Is there any reason for not using `Base` prefix for this one?

packages/primitives-core/package.json
```diff
@@ -3,14 +3,20 @@
...
+    "types"
   ],
+  "scripts": {
+    "test:typescript": "dtslint types"
+  },
   "license": "MIT",
   "dependencies": {
     "@emotion/is-prop-valid": "^0.6.3",
+    "@types/react-native": "^0.56.19",
     "css-to-react-native": "^2.2.1"
```
Ailrun: Missing deps for `create-emotion` and `create-emotion-styled` where this one uses their type.

packages/primitives/package.json
```diff
@@ -3,11 +3,18 @@
...
-    "dist"
+    "dist",
+    "types"
   ],
+  "scripts": {
+    "test:typescript": "dtslint types"
+  },
   "dependencies": {
+    "@types/react-native": "^0.56.19",
+    "@types/react-primitives": "^0.6.2",
```
Ailrun: Likewise, deps on `emotion`.

packages/primitives-core/package.json
Ailrun: @mitchellhamilton @nitin42 Adding such deps is acceptable?

packages/primitives/package.json
Ailrun: Type for this depends on `emotion`. If this package should not depends on it, then type also should.

mitchellhamilton: I'd prefer if this doesn't depend on emotion, create-emotion or create-emotion-styled for the types since it doesn't depend on any of them in the code.

Iamsamwhoami: I'll decouple them fully and update the PR :)

mekwall:
@Iamsamwhoami Are you committed to get this PR approved or are you looking for someone else to pick it up?

Iamsamwhoami:
@mekwall ah, time flies fast, I've not had time to get back to this, so I'll close this, if you feel like you wanna pick it up, go ahead 😄

PR #865, opened 2018-09-19
 Remove Non Valid Characters in Classname 

azizhk:
**What**:
Strip out non-word characters from variables names if any while generating label.

**Why**:
Because they are not valid CSS Selectors
https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors#449000

Some Background:
We have extracted some common components to a different npm module.
In that at build time we use rollup to bundle everything to one file targetting ES Next.
If two files use the same variable name, rollup adds `$1` suffix to the variable in the bundled file.
Then in the repo where we use this module we are using the babel-plugin-emotion during development with `autoLabel: true`. And then those styles don't turn up because `$` is not valid CSS Selector Syntax.

I know that we should run babel-plugin-emotion in our module at build time so that we get correct variable names as in source code, but we then would have to build two bundles one for development and production. Nevertheless, this will help when someone has a `$` in their variable names and is harmless otherwise.

**How**:
Simple Regex

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


packages/babel-plugin-emotion/src/babel-utils.js
```diff
@@ -90,12 +90,19 @@ export function getLabel(
   filename: string
 ) {
   if (!identifierName || !autoLabel) return null
-  if (!labelFormat) return identifierName.trim()
+
+  // Valid Characters in CSS Class Names Selecter
+  // https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors#449000
+  const transformedidentifierName = identifierName.replace(/[^\w-]/g, '')
```
Andarist: this should be properly camelCased, also maybe `normalizedName` (or similar) would be a better and more readable name?

azizhk:
Thanks @Andarist for the review. I've done the requested changes.

mitchellhamilton:
Thanks!!

PR #867, opened 2018-09-20
Upgrade lerna to v3, fixes #864

### LINKED ISSUE #864: Use lerna@3
Emotions packages doesn't contain the LICENSE file when they are published to npm. This is a bug in lerna@2, and lerna@3 fixed it. Can we upgrade to it and publish them again please? Some corporate companies are forbidden to use the packages until this is fixed.

- `emotion` version: next

pksunkara:


pksunkara:
@mitchellhamilton @Andarist Here's the PR which should fix the LICENSE issues.

lerna.json
```diff
@@ -10,7 +9,8 @@
       "Tag: Internal": ":house: Internal"
     }
   },
-  "version": "9.2.10",
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "registry": "https://registry.npmjs.org",
+  "version": "9.2.10",
```
Ailrun: Wrong trailing `,`.

lerna.json
```diff
@@ -10,7 +9,8 @@
       "Tag: Internal": ":house: Internal"
     }
   },
-  "version": "9.2.10",
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "registry": "https://registry.npmjs.org",
```
Ailrun: Do we need this registry?

pksunkara: Yes, because yarn defaults to `yarnpkg` when publishing.

pksunkara: Even though we are using `npm` when publishing, it's better safe to have this.

pksunkara:
Addressed the review and tests are passing.

pksunkara:
Please publish packages (I only care about next-packages) once this is merged.

tkh44:
I'd like @mitchellhamilton to check this over before merging, but lgtm.

mitchellhamilton:
I’m fine with this but I don’t see how it would solve the LICENSE issue?

pksunkara:
Lerna@2 when publishing packages in a monorepo doesn't do anything, thus the packages won't contain the LICENSE because it's generally in the root of the repo. Lerna@3 copies the LICENSE before preparing for publishing.

mitchellhamilton:
Wouldn’t we have to add the LICENSE file to each package and add it to the package.json files array?

pksunkara:
Not sure about the package.json files, but lerna will take of copying the file to each package.

mitchellhamilton:
[It looks like LICENSE files are always published](https://docs.npmjs.com/files/package.json#files) so lgtm

Thanks!!

tkh44:
Thanks @pksunkara and @mitchellhamilton

PR #868, opened 2018-09-20
Remove the generic type / object styles note in the TypeScript doc

elliottsj:
**What**:

Remove the generic type / object styles note in the TypeScript doc.

Also add an example using the tagged template style.

**Why**:

The generic type now works with the tagged template style, so the
note is no longer necessary.

https://github.com/Microsoft/TypeScript/issues/11947 is fixed by
https://github.com/Microsoft/TypeScript/pull/23430.

**How**:

Manually edited the `docs/typescript.md` document.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -111,15 +111,19 @@ const Image1 = styled('div')({
 
 // Or with a generic type
 
-const Image1 = styled('div')<ImageProps>({
+const Image2 = styled('div')<ImageProps>`
+  width: ${props => props.width};
+  background: url(${props => props.src}) center center;
+  background-size: contain;
+`
+
+const Image3 = styled('div')<ImageProps>({
   backgroundSize: 'contain',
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -111,16 +111,20 @@ const Image1 = styled('div')({
...
+const Image3 = styled('div')<ImageProps>({
   backgroundSize: 'contain',
 }, props => ({
   width: props.width;
   background: `url(${props.src}) center center`,
 }));
 ```
 
-* The generic type version only works with object styles due to https://github.com/Microsoft/TypeScript/issues/11947.
-
```
Ailrun: How about changing it to warn for TS<2.9 users?

elliottsj:
Closing as duplicate of #909

Issue #869, opened 2018-09-21
How to use `enzyme` to select css classname generated by `cx` function?


huchenme:
Hi Emotion team,

`css` and `cx` is awesome, but I am facing difficulties to use `enzyme` to select element whose className is generated by `cx` function.

Since enzyme does not support regex css selector, there seems no way to select `cx` generated className?

Andarist:
We don't believe that you should query your elements by their class names in tests - not with emotion, not without it. Please take a look at this - https://testing-library.com/docs/guide-which-query#priority

Sorry that we couldn't get back to you sooner - hope you understand.

PR #871, opened 2018-09-22
feat: Do not add hash in test environment

### LINKED ISSUE #870: Way to disable hash in test environment?
Would it be nice if we could disable hash className and only use the `label` in test environment?

huchenme:
This PR Closes #870 #869

**What**: 

in TEST environment:

```
const cls1 = css`
 font-size: 20px;
 background: green;
 label: label1;
`

const cls2 = css`
 font-size: 20px;
 background: red;
 label: label2;
`

const cn = cx(cls1, cls2, 'other')
```

Before:

```
console.log(cls1) // css-12345-label1
console.log(cls2) // css-12345-label2
console.log(cn) // other css-12345-label1-label2
```

After:

```
console.log(cls1) // css-label1
console.log(cls2) // css-label2
console.log(cn) // label1 label2 other css-12345-label1-label2
```

**Why**:

With this feature, it is easier to use enzyme CSS selectors

**How**:

* If `process.env.NODE_ENV === 'test'` and `label` is not empty, disable hash.
* Do not disable hash for `injectGlobal` as it will break the `emotion-server` tests
* Add className without key to `rawClassName` list (it will be problematic with keys)

**Checklist**:

- [ ] Documentation (any place I should add this?)
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion-server/test/auto-label/__snapshots__/auto-label.test.js.snap
+++ b/packages/emotion-server/test/auto-label/__snapshots__/auto-label.test.js.snap
@@ -10,7 +10,7 @@ Object {
   unicode-range: U+0100-024f,U+1-1eff,U+20a0-20ab,U+20ad-20cf,U+2c60-2c7f,U+A720-A7FF;
 }
 
-@-webkit-keyframes animation-1fps0gg-bounce {
+@-webkit-keyframes animation-bounce {
   from, 20%, 53%, 80%, to {
     -webkit-animation-timing-function: cubic-bezier(0.215,0.610,0.355,1.000);
     animation-timing-function: cubic-bezier(0.215,0.610,0.355,1.000);
@@ -42,7 +42,7 @@ Object {
   }
 }
 
-@keyframes animation-1fps0gg-bounce {
+@keyframes animation-bounce {
   from, 20%, 53%, 80%, to {
     -webkit-animation-timing-function: cubic-bezier(0.215,0.610,0.355,1.000);
     animation-timing-function: cubic-bezier(0.215,0.610,0.355,1.000);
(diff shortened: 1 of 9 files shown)


packages/emotion-server/test/auto-label/__snapshots__/auto-label.test.js.snap
```diff
@@ -2,15 +2,7 @@
 
 exports[`extractCritical returns static css 1`] = `
 Object {
-  "css": "@font-face {
```
huchenme: I might need some help with this, tried for a few hours still no luck to get `injectGlobal` work

huchenme:
Could someone help review, please? @tkh44 @mitchellhamilton

mitchellhamilton:
Thanks for the PR but this would break if two labels are the same so I don't think this is something we can accept. For your original issues, I would recommend adding something like a `data-testid` attribute to the elements that you want to select.

mitchellhamilton:
@huchenme How would that work?

The whole point is that using a hash makes each class name unique and deterministic, how would you maintain the uniqueness and determinism without using hashes.

huchenme:
It is important that in Jest tests, people are able to select by classNames. That is why Jest have mentioned on [their website](https://jestjs.io/docs/en/webpack#mocking-css-modules) and recommended to use [identity-obj-proxy](https://github.com/keyanzhang/identity-obj-proxy).

I want to achieve the same in the PR, why do we need to keep className unique in TEST environment?

Andarist:
I agree with @mitchellhamilton that what u actually might want to use are `data-testid`s, they could also be useful in your E2E tests (where u wont be able to target elements by classnames anyway)

mitchellhamilton:
> why do we need to keep className unique in TEST environment?

Because there could still be clashes.

PR #876, opened 2018-09-26
Upgrade `babel-plugin-emotion` to Babel 7

Spy-Seth:
**What**: Upgrade `babel-core` to use `@babel/core` v7

**Why**:

To avoid multiple version of installer when using the plugin. This kind of disagrement:

```
Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel.
```

This lead to a lot of change in the snapshot: mainly some formatting that change. There is also some generated that change (in the `styled.test.js.snap` file for exemple). 
It seems to be a change of hash algorithm: each hashs are changed everywhere 

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/package.json
+++ b/packages/babel-plugin-emotion/package.json
@@ -14,7 +14,7 @@
     "@emotion/hash": "^0.6.2",
     "@emotion/memoize": "^0.6.1",
     "@emotion/stylis": "^0.7.0",
-    "babel-core": "^6.26.3",
+    "@babel/core": "^7.0.0",
     "babel-plugin-macros": "^2.0.0",
     "babel-plugin-syntax-jsx": "^6.18.0",
     "convert-source-map": "^1.5.0",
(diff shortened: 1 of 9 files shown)


Spy-Seth:
I'm not sure I understand the error on CI right now. Could someone point me in the good direction?

packages/babel-plugin-emotion/package.json
```diff
@@ -14,7 +14,7 @@
     "@emotion/hash": "^0.6.2",
     "@emotion/memoize": "^0.6.1",
     "@emotion/stylis": "^0.7.0",
-    "babel-core": "^6.26.3",
```
mitchellhamilton: Could you move this to `devDependencies` (with the babel 6 version) and remove all the other changes? We use babel 6 so we can test against it.

mitchellhamilton:
Thanks for the PR, I wanted to get this fixed and released before we have v10 in master so I just fixed this by moving babel-core to devDeps in master.

Spy-Seth:
Thanks for taking care of it! I was occupied on other things these day ^^'

PR #883, opened 2018-10-01
Allow `innerRef` prop to be passed down to composite components

Andarist:
Ref forwarding make the innerRef just a regular prop and this should be forwarded to the wrapped component

### DIFF
--- a/packages/styled-base/src/index.js
+++ b/packages/styled-base/src/index.js
@@ -142,11 +142,6 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
               }
             } else {
               newProps.ref = ref
-              if (process.env.NODE_ENV !== 'production' && props.innerRef) {
-                console.error(
-                  '`innerRef` is no longer allowed, please use the `ref` prop instead'
-                )
-              }
             }
 
             const ele = React.createElement(baseTag, newProps)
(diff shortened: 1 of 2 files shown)


packages/styled-base/src/index.js
```diff
@@ -142,11 +142,6 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
               }
             } else {
               newProps.ref = ref
```
Andarist: `innerRef` is popular name in the react ecosystem and numerous libraries use it, IMHO it's undesirable to filter this out now as people might want to use it to get access to wrapped composite components

Andarist:
We can keep a warning (although IMHO that is odd - even though I understand the concern that people might be confused). Anyhow I think we should forward this prop.

mitchellhamilton:
I've been thinking about this a bit and I think we should maybe go so far as to still support innerRef for v10 and remove it in v11 since I think the migration experience is really important and there's no good way to safely codemod innerRef to ref.

Andarist:
Personally I would be against supporting it - refs should be used with caution, so I wouldn't expect them being used in too many places per application.

I understand migration concerns though - it's always a hard choice between one and other, so my personal preference here (which boils down to not wanting to have redundant code & avoiding maintenance costs related to it) shouldn't matter as much.

I can adjust the PR in whatever way you see fit. 

Additional note - styled-components is dropping support for `innerRef` in v4, so at least it wouldn't be as unusual for us to drop it too.

Andarist:
Closing in favour of https://github.com/emotion-js/emotion/pull/923

PR #885, opened 2018-10-01
Replace live playground code examples with codesandbox embeds.

tkh44:
This is a proof of concept for using codesandbox embeds for the code examples. They are generated on the fly and I've got the basics working.

~We will have to fork or patch [react-codesandboxer](https://github.com/Noviny/codesandboxer/tree/master/packages/react-codesandboxer) to enabled autoloading.~(@CompuIves helped me avoid using this component) I believe we could also use a custom gatsby plugin as an alternative method to do this. Create the sandbox during ssr. I didn't have time to explore this though.

With this, I believe we can also remove all the Playground and Live code. I wanted to wait what @mitchellhamilton thinks about this before continuing on. 

Preview of the changes.


site/src/templates/doc.js
```diff
@@ -54,37 +54,110 @@ const ClassName = (props: any) => {
...
+      css={mq({
+        maxWidth: constants.breakpoints[1] + 'em',
+        paddingLeft: [32, 30],
+        paddingRight: [32, 30]
+      })}
+    >
+      {internalCodeStylesClassName => (
+        <div className={internalCodeStylesClassName}>
+          <iframe
+            src={url}
```
tkh44: We could probably use the `onLoad` event on the iframe to display placeholder and reveal iframe when loaded.

tkh44:
I agree on the load time. It is pretty bad.

> Something that I think would be a really good idea though would be adding an open with CodeSandbox link to the examples since that wouldn't affect load times and would mean all the examples would be sharable which I'm guessing was one of the main reasons you wanted to do this?

I like this. Sharing examples is part of it. I liked the idea of devs being able to start a project from any code sample they choose.

One thing we should add is a full next.js example now that codesandbox supports it. Developers can get a boilerplate project going really quickly from an example in the documentation.

One thing I would like to work out is how to give the sandboxes a meaningful title. What do you think would be the best way to give the codeblock a title or name?

mitchellhamilton:
> One thing we should add is a full next.js example now that codesandbox supports it. Developers can get a boilerplate project going really quickly from an example in the documentation.

Great idea! + a gatsby example too

(I made this to show off some stuff in emotion 10 https://codesandbox.io/s/github/mitchellhamilton/emotion-10-next-demo)

> One thing I would like to work out is how to give the sandboxes a meaningful title. What do you think would be the best way to give the codeblock a title or name?

Maybe make it so text after the `@live` bit is used as a name, you'll want to look at https://github.com/emotion-js/emotion/blob/master/site/plugins/gatsby-remark-live-code/index.js

Andarist:
Going to close this as this got really stale by now and chances that somebody would pick this up are rather slim. The idea itself is very good though, it would be great to have links to codesandbox for each example.

PR #886, opened 2018-10-02
Setup AppVeyor CI

Andarist:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,5 @@
 install:
   - ps: Install-Product node 10 x64
-  - npm install -g yarn@latest
   - yarn
 
 shallow_clone: true
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,12 @@
+install:
+  - ps: Install-Product node 10 x64
+  - npm install -g yarn@latest
+  - yarn
+
+shallow_clone: true
+build: off
+
+test_script:
+  - node -v
+  - yarn -v
+  - yarn test
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,6 +6,9 @@ install:
 shallow_clone: true
 build: off
 
+cache:
+  - "%LOCALAPPDATA%\\Yarn"
+
 test_script:
   - node -v
   - yarn -v
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,3 +1,5 @@
+const path = require('path')
+
 let pkgsThatNeedBabelPluginEmotion = [
   'emotion',
   'create-emotion',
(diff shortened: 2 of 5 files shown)


Andarist:
Sure, that could be done later - but unfortunately it's not working yet and I don't have much time to debug this at the moment. For some reasons my console.logs are not shown in AppVeyor output 🙈 and I don't have windows machine at hand (nor experience with running scripts on it) to test this myself (without triggering CI rebuild).

Andarist:
Done, let's see if this changes anything for AppVeyor

.appveyor.yml
```diff
@@ -0,0 +1,12 @@
...
+  - npm install -g yarn@latest
+  - yarn
+
+shallow_clone: true
+build: off
+
+test_script:
+  - node -v
+  - yarn -v
+  - yarn test
```
Ailrun: We probably need `cache` setting.
https://www.appveyor.com/docs/build-cache/

Andarist:
@Ailrun I've added caching as recommended by yarn - https://yarnpkg.com/lang/en/docs/install-ci/

not sure if this is enough though and I'm not sure how to truly verify it

.appveyor.yml
```diff
@@ -6,6 +6,9 @@ install:
 shallow_clone: true
 build: off
 
+cache:
+  - "%LOCALAPPDATA%\\Yarn"
+
```
Ailrun: It looks like our CircleCI setting also caches `node_modules` directory. Is there any reason you don't cache them?

Ailrun: Oh, since that's `yarn`s' recommendation...

mitchellhamilton:
Could you update your local version of yarn and revert the changes to the yarn.lock?

.appveyor.yml
Ailrun: We can add `node_modules` and make the cache depends on `yarn.lock` file, but I think this is good enough for yarn, since yarn is fast enough with (its own) cache...

mitchellhamilton:
It seems like this is taking over ten minutes per build. Having a slow CI is something that I would really dislike and I would honestly rather have a fast CI and not support windows than have a slow CI and support windows.

Maybe we could just run the jest tests on windows?

Ailrun:
Test part takes ~10min and Installation takes ~5min. After cache is installed, it would become faster, but it still takes more than 10min.

Ailrun:
How many thread does it use? Document said appveyor have 2 CPU by default. It looks like there are more things run in parallel...

Andarist:
I'll try to play with it to see if we can make it faster. I'm OK with not testing this on windows - the source changes should be merged though (but I guess you agree on this anyway).

mitchellhamilton:
Yep, I'm fine with the source changes.

mitchellhamilton:
@Andarist I just uninstalled AppVeyor because it was breaking codecov's statuses, if you can get this much faster then I'll reinstall it.

Andarist:
No problem, I actually have wanted to propose same thing. I'm not sure if I really want to work on this though given lack of time, so going to close this and we can revisit in case of problems in the future.

PR #888, opened 2018-10-04
Make is-prop-valid check case sensitive

mxstbr:
See https://github.com/styled-components/styled-components/issues/1969#issuecomment-426943214 for the reasoning. Basically this is currently letting props such as `onboarding` or `onlineSize` through even though those are obviously not event handlers.

By making this check case sensitive it should work as expected.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete


Andarist:
You should verify which one is covered by the regexp now - `autoFocus` or `autofocus` - and add a second one that is not covered at the moment. This is needed for preact compatibility.

PR #890, opened 2018-10-04
Add nohoist option not to hoist dtslint

### LINKED ISSUE #889: test:typescript script cannot find typescript executable after we upgrade to Lerna@3
- `emotion` version: master branch
- `react` version: N/A

**What you did**

1. Remove all `node_modules` directory (`rm -rf **/node_modules`)
1. Reinstall dependencies (`yarn`)
1. Run typescript test (`yarn test:typescript`)

**What happened**

Tests are failed with
```
Error: Cannot find module '<path to repository root>/emotion/node_modules/dtslint/typescript-installs/next/node_modules/typescript'
```

**Suggested solution**
I think this problem occurs because of hoisting of `dtslint`. It can be resolved with `nohoist` option of `yarn`.

Ailrun:
**What**: Adding nohoist option

**Why**: To resolve #889.

**How**: Modifying `package.json`

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,12 +174,17 @@
       }
     ]
   },
-  "workspaces": [
-    "packages/*",
-    "site",
-    "scripts/*",
-    "playgrounds/*"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/*",
+      "site",
+      "scripts/*",
+      "playgrounds/*"
+    ],
+    "nohoist": [
+      "**/dtslint"
+    ]
+  },
   "bugs": {
     "url": "https://github.com/tkh44/emotion/issues"
   },


mitchellhamilton:
Is there a way that dtslint could be run for all the packages that have types from the root rather than have the scripts in every package?

Ailrun:
@mitchellhamilton We can run `dtslint packages/<package1>/types`, `dtslint packages/<package2>/types`, ... parallelly on the root package, though this one is more unreliable IMO, since,
1. We still need to add each packages manually when we add a new typing
2. Since we place dtslint script farther from package itself, when one adds new typing, he/she could have hard time to find where the type testing script is.

Ailrun:
@mitchellhamilton If you think that's really the way we should take, then writing a script for that could be a solution.

mitchellhamilton:
Nah, if dtslint doesn't have an easy way to do this then I'll just merge this.

PR #896, opened 2018-10-05
Drop support for preact

mitchellhamilton:
**What**:
Remove all the preact packages and stuff in the build setup for it.

**Why**:
Supporting preact is getting impractical, it was easy in the past when there were very few differences in the react and preact APIs but as react has added new APIs like `createContext`, `forwardRef` and `Fragment`, it's getting significantly harder. While there are implementations of the new context API for preact, they aren't as performant as React's `createContext` and they can't have some of the upcoming features like `unstable_read` and `static contextType = MyContext`.(before anyone says preact is implementing x and y, this isn't just about the current APIs that are missing in preact, it's about future APIs that we couldn't fully utilize because we would still have to support preact) Also, `preact-emotion` has significantly fewer downloads (~1k vs ~150k downloads a week) than `react-emotion` so the low demand for it doesn't justify the cost of maintaining it.

Note that the `emotion` package will still be usable with preact and if someone wants to maintain a separate implementation of styled for preact, I'd be happy to give up the `preact-emotion` name on npm.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete


Andarist:
@Ailrun right, but at the moment is-prop-valid package is also used by other libraries (i.e. styled-components) and they expect preact support, considering that those are just 3 props, I would be in favour of keeping them here

mitchellhamilton:
I agree with @Andarist.

Issue #875, opened 2018-09-25
BUG: styled method className prop is not passed down


huchenme:
- `emotion` version: 9.2.10
- `react` version: 16.4.2

### Relevant code.

https://codesandbox.io/s/k2p1knxnov

### Problem description:

**Expected:**
The inner class should have style `text-transform: uppercase;`

**Actual:**
The inner component does not have the className passed in so it does not have expected style

In our library, we allow people to pass in className as props and set to the right component, it will be breaking if the passed in className is generated by `emotion`

### Suggested solution:

the passed in className should be one of the className in the list

mitchellhamilton:
This is expected behavior. If emotion just appended the class name there would be issues with specificity. Using emotion class names to target elements is strongly discouraged for this exact reason.

huchenme:
@mitchellhamilton I do not quite agree with this. If both `styled` and `css` are provided by `react-emotion` and `styled` accepts `className`, the example should work.

It is ok that this scenario have never been thought about, but in our project, we have mix of `styled`, `css` and code shown in the example. The only way I can think about to get rid of this issue is to rewrite the components using `styled` to `css`.

But since `react-emotion` provide both, they should work together, that is my point.

Andarist:
@mitchellhamilton maybe we could warn in dev mode about this (or with eslint plugin)?

mitchellhamilton:
@huchenme Could you show a real world of where you would want this behavior? 

> @mitchellhamilton maybe we could warn in dev mode about this (or with eslint plugin)?

Yeah, that's a good idea, it would be possible at runtime so no need for eslint.

huchenme:
We have a large code base of CSS modules and want to convert to CSS-in-JS. Initially, we would want to convert everything to `styled` way and then we realized it is not very practical as we have lots of code which set className to the most outer div to represent `state`:

```js
<div
  className={classNames(
    styles.container,
    open ? styles.open : styles.closed,
    disabled ? styles.disabled : false
  )}
>
  <div className={styles.inner}>Hello</div>
</div>
```

and in CSS:

```css
.container {

}

.inner {
  .disabled & {
	cursor: not-allowed;
  }

  .open & {
	display: block;
  }

  .closed & {
	display: none;
  }
}
```

This works fine if both outer and inner components are styled using `css` method from `emotion`. except we need some way to generate hash for `disabled`, `open` and `closed`.

We used:

```js
const disabled = css``;
const open = css``;
const closed = css``;

[...]

PR #901, opened 2018-10-08
Add warning for interpolating a class name from `css`

mitchellhamilton:
**What**:
Add warning for interpolating a class name from css

**Why**:

To make it clear how unsupported this is. #875 

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -154,6 +154,16 @@ function handleInterpolation(
     // eslint-disable-next-line no-fallthrough
     default: {
       const cached = registered[interpolation]
+      if (
+        process.env.NODE_ENV !== 'production' &&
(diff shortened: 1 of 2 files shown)


packages/serialize/src/index.js
```diff
@@ -154,6 +154,16 @@ function handleInterpolation(
     // eslint-disable-next-line no-fallthrough
     default: {
       const cached = registered[interpolation]
+      if (
+        process.env.NODE_ENV !== 'production' &&
+        couldBeSelectorInterpolation &&
+        cached !== undefined
+      ) {
+        console.error(
```
Andarist: this should be warned only once - to avoid spamming the console too much

PR #904, opened 2018-10-09
Fix repository field in package.json

bertho-zero:


bertho-zero:
@mitchellhamilton It's done !

PR #906, opened 2018-10-10
Fix typings for @emotion/core with missing cases

Ailrun:
**What**: @emotion/core typings

**Why**: There are few missing cases

**How**: Add tests for missing cases, and fix them.

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete


packages/core/types/tests.tsx
```diff
@@ -47,6 +47,13 @@ const ComponentWithCache = withEmotionCache((_props: {}, context) => {
     }}
   />
 </div>
+;<Global
```
Ailrun: GitHub (and my editor emacs) cannot syntax highlight all these things.... Oh my prettier...

Issue #900, opened 2018-10-08
shouldForwardProp in styled component


timomeh:
- `emotion` version: 9.2.9
- `react` version: 16.5.0

**Relevant code.**

```javascript
import styled from '@emotion/native'
import isPropValid from '@emotion/is-prop-valid'

const Text = styled.Text(props => ({
  fontFamily: "Lato"
}))

const Headline = styled(Text, { shouldForwardProp: isPropValid })(props => ({
  color: props.isColored ? props.theme.baseColor : props.theme.textColor,
  fontWeight: '700'
}))

<Headline isColored>Hello, World!</Headline>
```

**What you did:**

I'm using react-native and [react-testing-library](https://github.com/kentcdodds/react-testing-library) together with react-native-web to test my components.

I added `{ shouldForwardProp: isPropValid }` to my styled components in react native to prevent the prop `isColored` from being forwarded.

**What happened:**

Using the setup described above, I noticed that the prop `isColored` is being forwarded. The following error message will be displayed:

`Warning: React does not recognize the isColored prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase iscolored instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

**Problem description:**

It seems like `shouldForwardProp` isn't being called in @emotion/native. I tried passing a `console.warn` to see if `shouldForwardProp` will be called, but nothing will be logged.

```javascript
const shouldForwardProp = () => {
  console.warn('shouldForwardProp called')
}

const Headline = styled(Text, { shouldForwardProp  })(...))
```

**Suggested solution:**

@emotion/native uses `createStyled` from @emotion/primitives-core:

https://github.com/emotion-js/emotion/blob/7a4ff8ebae5ddce2b3dd81afa258bc15398512b0/packages/native/src/styled.js#L1-L7

primitives-core has a default implementation for `shouldForwardProp`:

https://github.com/emotion-js/emotion/blob/7a4ff8ebae5ddce2b3dd81afa258bc15398512b0/packages/primitives-core/src/styled.js#L7-L16

It seems like @emotion/native doesn't override this default implementation. It should provide an API to pass `shouldForwardProp` as option.

Or maybe there is an API for it which isn't documented, or I can't find the documentation, or I can't find a way to do it in the emotion package code.

There is a discussion regarding this in the PR for @emotion/native: https://github.com/emotion-js/emotion/pull/759/files#r204662931
[...]

PR #907, opened 2018-10-10
Add typings of @emotion/is-prop-valid package

Ailrun:
**What**: Adding typings of `@emotion/is-prop-valid` package

**Why**: Since there are users who use `@emotion/is-prop-valid` in their app like https://github.com/emotion-js/emotion/issues/900

**How**:

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/is-prop-valid/types/index.d.ts
+++ b/packages/is-prop-valid/types/index.d.ts
@@ -1,0 +1,5 @@
+// Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
+// TypeScript Version: 2.1
+
+declare function isPropValid(string: string): boolean
+export default isPropValid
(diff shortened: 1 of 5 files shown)


packages/is-prop-valid/types/index.d.ts
```diff
@@ -0,0 +1,5 @@
+// Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
+// TypeScript Version: 2.1
+
+declare const isPropValid: RegExp['test']
```
mitchellhamilton: I’d rather this be defined as `string => boolean` since it’s clearer and if for some reason it changes in the future to not use regex, it’d be more accurate. (I understand it won’t change the actual type)

Ailrun: @mitchellhamilton OK, I will update one.

Ailrun:
@mitchellhamilton It's updated.

PR #908, opened 2018-10-10
Fix typings of @emotion/styled-base to work well with ref

Ailrun:
**What**: Fixing typings of `@emotion/styled-base`

**Why**: It does not work well with `ref` props.

**How**:

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -15,9 +15,11 @@
 import { ComponentSelector, Interpolation } from '@emotion/serialize'
 import * as React from 'react'
 
-import { Overwrapped, PropsOf } from './helper'
+import { Omit, Overwrapped, PropsOf } from './helper'
 
 export { Interpolation }
+
+type JSXInEl = JSX.IntrinsicElements
 
 export type WithTheme<P, T> = P extends { theme: infer Theme }
   ? P & { theme: Exclude<Theme, undefined> }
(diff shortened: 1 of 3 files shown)


mitchellhamilton:
I don't get how the changes here relate to `ref`, could you explain this more?

packages/styled-base/types/index.d.ts
```diff
@@ -33,50 +35,45 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
...
   >(
     ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
   ): StyledComponent<InnerProps, StyleProps, Theme>
   <
-    StyleProps extends Overwrapped<InnerProps, StyleProps> = InnerProps &
-      ExtraProps
+    StyleProps extends Omit<
+      Overwrapped<InnerProps, StyleProps>,
+      ReactClassPropKeys
+    > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>
```
Ailrun: @mitchellhamilton These part is about hiding `ref` and `key` from default `StyleProps`, so that
```jsx
const Comp = styled('div')({}).withComponent('label');
<Comp ref={(el: HTMLLabelElement) => {}} />
```
can make sense.
(Originally, `ref` of `Comp` was `((el: HTMLLabelElement) => void) & ((el: HTMLDivElement) => void)`.)

packages/styled-base/types/index.d.ts
```diff
@@ -33,50 +35,45 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
...
+   * @desc this method is type-unsafe
    */
-  withComponent<NewTag extends keyof JSX.IntrinsicElements>(
+  withComponent<NewTag extends keyof JSXInEl>(
     tag: NewTag
-  ): StyledComponent<JSX.IntrinsicElements[NewTag], StyleProps, Theme>
+  ): StyledComponent<JSXInEl[NewTag], StyleProps, Theme>
+  withComponent<Tag extends React.ComponentType<any>>(
+    tag: Tag
+  ): StyledComponent<PropsOf<Tag>, StyleProps, Theme>
```
Ailrun: With this, we can use type of tag instead of `React.ComponentType` or something like that.
This is required for inferring valid type of ref.
(Originally, `ref` is `(el: React.ComponentType<Something>) => void`.)

packages/styled-base/types/index.d.ts
```diff
@@ -94,20 +91,12 @@ export interface CreateStyledComponentExtrinsic<
...
-    ExtraProps,
-    Theme
-  >
-  <InnerProps, ExtraProps = {}>(
-    tag: React.SFC<InnerProps>,
+  <Tag extends React.ComponentType<any>, ExtraProps = {}>(
+    tag: Tag,
     options?: StyledOptions
-  ): CreateStyledComponentExtrinsic<React.SFC<InnerProps>, ExtraProps, Theme>
+  ): CreateStyledComponentExtrinsic<Tag, ExtraProps, Theme>
```
Ailrun: Likewise, to use valid `ref` with tag type.

packages/styled-base/types/tests.tsx
```diff
@@ -116,13 +116,15 @@ const Button3 = styled<'button', PrimaryProps>('button')(props => ({
 // $ExpectError
 ;<Button3 type="button" />
 
-const Button4 = styled<ReactClassProps0, PrimaryProps>(ReactClassComponent0)`
+const Button4 = styled<typeof ReactClassComponent0, PrimaryProps>(
+  ReactClassComponent0
+)`
```
mitchellhamilton: Why does there have to be `typeof ReactClassComponent0` if the user is passing `ReactClassComponent0` to styled, can't TS infer it?

Ailrun: @mitchellhamilton Since we need `PrimaryProps`. It's same with
https://github.com/emotion-js/emotion/blob/952a2e442c6528a615f5719c5896a14c11deebfa/packages/styled-base/types/tests.tsx#L85-L98

Ailrun: In TS3.2, partial type argument will be supported (though it could be deferred...) and it will become much easier to use these.

PR #916, opened 2018-10-11
Allow extract critical without globals

Brondahl:
**What**:
Add ability to fetch only the specific emotion css rules attached to a block of HTML, excluding any global rules.

Whilst making the change, I did some boy-scouting. Trivial variable renaming for greater clarity, and fix let => const and typof foo = 'undefined' JS best practices.
(Separate commits for each step.)

**Why**:
I have a use case where I needed that subset of the rules.
I imagine other may need it as well, and adding it has no impact on anyone who doesn't care.

**How**:
Current code very explicitly returns css if either:
a) the id is in the HTML.
b) the css is a global rule.
I added an optional boolean parameter to allow exclusion of the second case (defaulting to false, i.e. include Globals, i.e. the current behaviour).

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete

I'm afraid I don't know how to: `Please add a `Tag:` prefixed label from the labels so that this PR shows up in the changelog`. Please advise.

### DIFF
--- a/docs/ssr.md
+++ b/docs/ssr.md
@@ -1,5 +1,5 @@
 ---
-title: "Server Side Rendering"
+title: 'Server Side Rendering'
 ---
 
 Server side rendering works out of the box in emotion 10 and above
@@ -63,7 +63,7 @@ const stream = renderToNodeStream(<App />).pipe(
 
 ### extractCritical
 
-This returns an object with the properties `html`, `ids` and `css`. It removes unused rules that were created with emotion(it still includes rules that were inserted with `injectGlobal`).
+This returns an object with the properties `html`, `ids` and `css`. It removes unused rules that were created with emotion. You may choose whether or not to include rules that were inserted with `injectGlobal`, by setting the second `includeGobals` parameter; by default globals **are** included.
 
 ```jsx
 import { renderToString } from 'react-dom/server'
@@ -73,6 +73,12 @@ import App from './App'
 const { html, ids, css } = extractCritical(
   renderToString(<App />)
 )
+
+const {
+  html: htmlNotInclGlobals,
+  ids: idsNotInclGlobals,
+  css: cssNotInclGlobals
+} = extractCritical(renderToString(<App />), false)
 ```
 
 #### hydrate
--- a/packages/create-emotion-server/src/extract-critical.js
+++ b/packages/create-emotion-server/src/extract-critical.js
@@ -8,23 +8,27 @@ const createExtractCritical = (cache: EmotionCache) => (html: string) => {
 
   let o = { html, ids: [], css: '' }
   let match
-  let ids = {}
+  let idsInHtml = {}
   while ((match = RGX.exec(html)) !== null) {
     // $FlowFixMe
-    if (ids[match[1]] === undefined) {
+    if (idsInHtml[match[1]] === undefined) {
       // $FlowFixMe
-      ids[match[1]] = true
+      idsInHtml[match[1]] = true
     }
   }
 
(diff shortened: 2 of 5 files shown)


Andarist:
> I'm afraid I don't know how to: Please add aTag:prefixed label from the labels so that this PR shows up in the changelog. Please advise.

This has to be done with somebody from emotion team, so don't worry about it.

Could you tell us more about your use case? Why do you need this?

Brondahl:
Do I need to do anything about the Appveyor error? If so, how do I do so - everything seems to work on my machine?

Brondahl:
> Could you tell us more about your use case? Why do you need this?

I can, but alas, I can't do so immediately - I was just getting this PR submitted before doing some other work. Will provide more details later today, hopefully.

Punchline is "NetlifyCMS Preview"

Andarist:
I'm not familiar with that solution so I might be talking bollocks here or ask stupid questions, but I would still like to try explore alternatives (to see if they can handle your use case, there is high change that they might not):
- is the iframe src-less? can you access it's content window safely?
- have you tried using emotion instances? maybe not sharing the instance with netlify CMS would automatically solve your issue?

Brondahl:
> is the iframe src-less?
Yes, there's no src attribute. I belive it's being managed and populated by react-iframe and portals?

> can you access it's content window safely?
Urrr ... I don't know what this question means. (I've never done that much with iframes in the past?)

> have you tried using instances.
No, I didn't know that was a thing. Does sound plausible that it would fix it, but I'm not sure how to implement that? The docs don't immediately make it clear to me how this works. Is the idea that rather than importing `css` directly from the emotion dependency, that instead I have a file which imports create-emotion, invokes it and then exports `css` from the created object, and then I import *THAT into my files?

The [docs](https://github.com/emotion-js/emotion/tree/master/packages/create-emotion) say `If you aren't using SSR, context can be an empty object`. I *am* using heavy SSR, so what should I provide as the context? In all the doc examples, the context object is setup as an empty object?

Andarist:
> Is the idea that rather than importing css directly from the emotion dependency, that instead I have a file which imports create-emotion, invokes it and then exports css from the created object, and then I import *THAT into my files?

Yeah, that's the idea in emotion@9. 

>  I wasn't expecting an extended discussion about why I wanted the functionality?

Well, working on any kind of software is trying to find the best approach for the problem. OSS is often trying to find a balance between supporting every possible use case (feature requests) and keeping the code base clean and focused.

If there are means to achieve what you are trying to do already in the library (im not saying there are for sure) then it's worthwhile to discuss alternative approaches. 

There is always also a possibility of forking the library if your PR doesn't get accepted (speaking in general sense, not about this particular PR).

> Could you explain (again, only so that I understand how the OS community works!!!) why it is that you're interested in these other possible approaches? Is it that you suspect this won't solve my problem? Do you think there's some risk associated with my change, that I haven't noticed? Something else?

Im just concerned about keeping the library small and well organized - less code code === less problems 😉 

Could you maybe join our slack and reach out to me on prv to describe your problem, current setup etc? That could speed up a decision on this.

Brondahl:
FYI, @Andarist Instances did indeed fix this issue.
So I don't personally need this PR any more.

I infer from the previous conversation that your preferred choice is to not take the new functionality?

Given that, what should I do with the PR?
My assumption would be "Close it, and delete my fork"?

Andarist:
Im really glad you were able to work it out! Yeah - if not needed we'd prefer not to merge it. Going to close this now and thanks for being patient and bearing with me 😉

PR #917, opened 2018-10-11
Make styled accept the css prop

mitchellhamilton:
**What**:
Make styled accept the css prop (as in without using `jsx` from `@emotion/core`)
```jsx

let Comp = styled.div(...styles)

<Comp css={...moreStyles} />

```

**Why**:
This is mainly about a performance optimization rather than styled accepting the css prop without using the custom createElement. I think a very common case is going to be adding the css prop to a styled component and doing this means that only one context consumer will be used instead of two(and only one style element in SSR which is the more important thing), while I don't think this is gonna make a huge difference on it's own, I think it might make a difference in the real world since lots of styled components will be rendered. I'm not totally sure about doing this since it would mean there's a way(it's hacky but if there's a way, people will use it) for people to get the css prop passed down to a component.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


jgoux:
Would it be possible to have the same interpolation power that we have when defining a styled component when using the `css` prop ?

The reasoning here is to be able to use the same level of abstraction when integrating third party librairies as when we define a styled component (without having to name them explicitly with a `styled(MyComponent)` )

I find it useful to abstract the theme access behind utilities functions : 

```jsx
import { AppBar } from 'material-ui'

// utility function for interacting with the theme
const scale = (n = 1) => ({ theme }) => `${n * theme.spacing.unit}px`

// it just works, but I don't want to have to name that
const MyAppBar = styled(AppBar)`
  padding: ${scale(2)};
`

// this would be awesome
const App = () => (
  <AppBar
    css={css`
      /* Yay, same syntax than styled(AppBar) without the naming! */
      padding: ${scale(2)};
    `}
     {...materialUiAppBarSpecificProps}
    />
)

// current API
const App = () => (
  <AppBar
    css={theme => css`
      /* Different syntax than styled(AppBar) :(  and more noisy */
      padding: ${scale(2)({ theme })};
    `}
     {...materialUiAppBarSpecificProps}
    />
)
```

mitchellhamilton:
That's something which I'd **REALLY** not like to do, doing that would mean that for **every** css prop, we would have to subscribe to the theme which disables some potential future performance optimisations. Also, with React Hooks, you could call `useContext` with the ThemeContext to get the theme so it wouldn't really be a big gain at all. We might even remove being able to pass a function to the css prop to get the theme in favor of telling people to `useContext`.

PR #924, opened 2018-10-12
Fix typings of jest-emotion package

Ailrun:
**What**: Typings of jest-emotion

**Why**: I missed it last time

**How**: With TS

**Checklist**:
- [N/A] Documentation
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/jest-emotion/types/index.d.ts
+++ b/packages/jest-emotion/types/index.d.ts
@@ -2,21 +2,31 @@
 // TypeScript Version: 2.4
 
 /// <reference types="jest" />
-import { Emotion } from 'create-emotion';
+export interface EmotionMatchers extends jest.ExpectExtendMap {
+  toHaveStyleRule(
+    received: any,
+    property: string,
+    value: any
+  ): { message(): string; pass: boolean }
+}
+export const matchers: EmotionMatchers
 
 export interface CreateSerializerOptions {
-  classNameReplacer?: (className: string, index: number) => string;
-  DOMElements?: boolean;
+  classNameReplacer?: (className: string, index: number) => string
+  DOMElements?: boolean
 }
-
-export function getStyles(emotion: Emotion): string;
-export function createSerializer(emotion: Emotion, options?: CreateSerializerOptions): jest.SnapshotSerializerPlugin;
-export function createMatchers(emotion: Emotion): jest.ExpectExtendMap;
+export function createSerializer(
+  options?: CreateSerializerOptions
+): jest.SnapshotSerializerPlugin
+export const print: jest.SnapshotSerializerPlugin['print']
+export const test: jest.SnapshotSerializerPlugin['test']
+declare const serializer: jest.SnapshotSerializerPlugin
+export default serializer
 
 declare global {
   namespace jest {
     interface Matchers<R> {
-      toHaveStyleRule(property: string, value: any): R;
+      toHaveStyleRule(property: string, value: any): R
     }
   }
 }
--- a/packages/jest-emotion/types/test.ts
+++ b/packages/jest-emotion/types/test.ts
@@ -1,52 +1,45 @@
-/// <reference types="jest" />
-import * as emotion from 'emotion';
-import {
+import serializer, {
+  matchers,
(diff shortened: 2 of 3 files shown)


packages/jest-emotion/types/index.d.ts
```diff
@@ -2,21 +2,33 @@
...
+  classNameReplacer?: (className: string, index: number) => string
+  DOMElements?: boolean
 }
-
-export function getStyles(emotion: Emotion): string;
-export function createSerializer(emotion: Emotion, options?: CreateSerializerOptions): jest.SnapshotSerializerPlugin;
-export function createMatchers(emotion: Emotion): jest.ExpectExtendMap;
+export interface EmotionSerializer {
+  print(val: any, printer: (val: any) => string): string
+  test<T>(val: T): T | boolean
```
mitchellhamilton: Why is the type `T | boolean`? why not just `boolean`?

Ailrun: @mitchellhamilton When `val` is falsy, it will return `val` itself, won't it?

mitchellhamilton: Technically, yes but from an API perspective `test` should be seen as returning a boolean.

mitchellhamilton: If it was meant to be perfectly technically correct, it would be `T | true` but technical correctness isn't the most important thing.(please don't make it `T | true`, make it `boolean`) People should see it as returning a boolean because that's the API that we want to be showing

And tbh, why does `EmotionSerializer` even exist, it's the same type as any jest serializer, do the jest types not have a type for snapshot serializers?

Ailrun: @mitchellhamilton Could you give me more detailed use case of `print` and `test`? If user can put falsy value to `test` by oneself, using `boolean` only could confuse user IMO. If it's not the case, `boolean` may be OK.

Ailrun: @mitchellhamilton TS definition for Jest serializer is too general. It looks like
```
    interface SnapshotSerializerPlugin {
        print(val: any, serialize: ((val: any) => string), indent: ((str: string) => string), opts: SnapshotSerializerOptions, colors: SnapshotSerializerColors): string;
        test(val: any): boolean;
    }
```

mitchellhamilton: No one will ever directly call `test`, they'll generally call `expect.addSnapshotSerializer(serializer)`

> And tbh, why does EmotionSerializer even exist, it's the same type as any jest serializer, do the jest types not have a type for snapshot serializers?

I'd rather not have the `EmotionSerializer` all together though and just have `serializer` be the type of [`SnapshotSerializerPlugin`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L383-L386)

Ailrun: @mitchellhamilton I didn't use `SnapshotSerializerPlugin` because of `T | boolean` part. However, if use case cannot confuse user, I will change that part.

Ailrun: @mitchellhamilton Then why do we export `test`?

mitchellhamilton: So people can just add jest-emotion to their jest config https://jestjs.io/docs/en/configuration#snapshotserializers-array-string

Ailrun: If it's only for test, I think type definition should not export them.

Ailrun: @mitchellhamilton OK, I will change this part.

mitchellhamilton: What's the harm in having it in the types?

Ailrun: @mitchellhamilton 
- Pros
  - Can use it with TS config (As far as I know, jest doesn't support config file in TS, but who knows :))
- Cons
  - User can load it and try to use it (since it is typed, when user uses autocompletion, it will be listed)

PR #939, opened 2018-10-16
Fix cloneElement with elements created with `jsx` from `@emotion/core` that use the css prop

mitchellhamilton:
**What**:
Fix cloneElement with element created with `jsx` from `@emotion/core` that uses the css prop

**Why**:

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/core/src/jsx.js
+++ b/packages/core/src/jsx.js
@@ -6,6 +6,8 @@ import { serializeStyles } from '@emotion/serialize'
 
 let typePropName = '__EMOTION_TYPE_PLEASE_DO_NOT_USE__'
 
+let hasOwnProperty = Object.prototype.hasOwnProperty
+
 let render = (cache, props, theme: null | Object, ref) => {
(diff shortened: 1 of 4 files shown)


mitchellhamilton:
I wanna get this in because I want to test it on a project but if anyone has any comments, comment away.

packages/core/src/jsx.js
```diff
@@ -87,15 +98,26 @@ export const jsx: typeof React.createElement = function(
...
-    if (typeof props.css === 'function') {
-      return (
-        <ThemeContext.Consumer>
-          {theme => render(context, props, type, args, theme)}
-        </ThemeContext.Consumer>
-      )
+
+  let argsLength = args.length
+
+  let createElementArgArray = new Array(argsLength)
```
Andarist: that would be an interesting optimization to make in babel, to postpone argument spreading until they are used 🤔

PR #942, opened 2018-10-17
Remove unsafe styled optimization

mitchellhamilton:
**What**:
Remove unsafe styled optimization

**Why**:
Because the way labels and keyframes worked changed a while ago which means in really rare circumstances, this could break and the optimization that this was trying to do doesn't actually work now.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/react-emotion/test/__snapshots__/index.test.js.snap
+++ b/packages/react-emotion/test/__snapshots__/index.test.js.snap
@@ -373,6 +373,10 @@ exports[`styled withComponent will replace tags but keep styling classes 1`] = `
   color: green;
 }
 
+.emotion-2 {
+  color: green;
+}
+
 <article>
   <h1
     className="emotion-0 emotion-1"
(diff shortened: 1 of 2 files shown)


packages/react-emotion/test/__snapshots__/index.test.js.snap
```diff
@@ -373,14 +373,18 @@ exports[`styled withComponent will replace tags but keep styling classes 1`] = `
   color: green;
 }
 
+.emotion-2 {
+  color: green;
+}
```
mitchellhamilton: If anyone's curious why this snapshot changed and two class names with the same styles exist is that they each have different labels and the optimization that was here broke the label but now that it's not here, the label works so there are two different classes.

Andarist: is it possible to add a test confirming the correct behaviour in previously broken case? except having a different class name here now this test doesnt actually change the observable output

PR #946, opened 2018-10-18
Fix labels of size badges in README

Ailrun:
**What**: Labels of size badges

**Why**: It confuses readers

**How**: Fix the label texts

**Checklist**:
- [x] Documentation
- [N/A] Tests
- [N/A] Code complete


tkh44:
Thanks

PR #951, opened 2018-10-21
SSR benchmarks

mitchellhamilton:
**What**:
Add some SSR benchmarks

**Why**:
The client side benchmarks take longer to run and generally don't have super consistent results and since there is a lot of performance sensitive stuff shared across the client and server, these are useful

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

# THESE BENCHMARKS HAVE PROBLEMS
All benchmarks do.
# DO NOT CHOOSE A LIBRARY BASED ON BENCHMARKS
These are only for measuring performance improvements and will only test specific things.


mitchellhamilton:
I'm gonna merge this because I have things I'm working on that I want to test with these.

PR #956, opened 2018-10-23
Always enable source maps

### LINKED ISSUE #954: Always enable sourceMap in babel plugin
What if we always enabled sourceMap? Instead of just adding adding it to the string though, we'd do an inline `process.env.NODE_ENV !== 'production'` check. This would also have the added advantage that component libraries could have the source map option enabled and consumers could see the source maps but they wouldn't hurt production. For example,
```jsx
let style = css`
  display: flex;
  background-color: hotpink;
  color: ${color};
`

// ↓ ↓ ↓ ↓ ↓ ↓

let style = css(
  'display:flex;background-color:hotpink;color:',
  color,
  ';',
  process.env.NODE_ENV === 'production' ? '' : 'big source map stuff'
)
```
(for that specific case, we'd probably also want to use `+` on the string before it rather than add it as an argument so the `,''` could be removed)

For anyone who's unaware, autoLabel is also automatically enabled when `process.env.NODE_ENV !== 'production'` in the v10 babel plugin but that check is at build time. It's not necessarily safe to assume that the `process.env.NODE_ENV` value at build time will be the same as when the code is running(mainly because of babel-cli) but for labels since they're relatively small, I think it's an okay trade off and they can be helpful in production too, for debugging and etc. whereas source maps cannot be used at all in production because they can't be used with insertRule.

I'm mainly opening this issue because there's some stuff I want to do before this and I don't want to forget about.

mitchellhamilton:
**What**:
Always enable source maps

**Why**:
Closes #954 

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I've been thinking about this more and there's one problem I can see. checking process.env.NODE_ENV is expensive in node and since this would mean it would happen really often, it could be bad. I'm kind of not super worried about it though since most(i think?) SSR apps tend to use webpack to compile server side stuff anyways(e.g. next, razzle both do it)

### DIFF
--- a/packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
+++ b/packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
@@ -7,6 +7,21 @@ import { getSourceMap } from './source-maps'
 import { simplifyObject } from './object-to-string'
 import { appendStringToExpressions, joinStringLiterals } from './strings'
 
+function createSourceMapConditional(t, production, development) {
+  return t.conditionalExpression(
+    t.binaryExpression(
+      '===',
+      t.memberExpression(
+        t.memberExpression(t.identifier('process'), t.identifier('env')),
+        t.identifier('NODE_ENV')
+      ),
+      t.stringLiteral('production')
+    ),
+    production,
+    development
+  )
+}
+
 export let transformExpressionWithStyles = ({
   babel,
   state,
@@ -33,7 +48,6 @@ export let transformExpressionWithStyles = ({
     if (label && shouldLabel) {
       appendStringToExpressions(path.node.arguments, `label:${label};`, t)
     }
-    let node
 
     let isPure = true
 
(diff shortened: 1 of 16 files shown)


packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
```diff
@@ -7,6 +7,21 @@ import { getSourceMap } from './source-maps'
 import { simplifyObject } from './object-to-string'
 import { appendStringToExpressions, joinStringLiterals } from './strings'
 
+function createSourceMapConditional(t, production, development) {
+  return t.conditionalExpression(
```
Andarist: maybe use a babel template for readability?
```js
  const tmpl = template(
    'process.env.NODE_ENV !== "production" ? DEV_VALUE : PROD_VALUE',
    {placeholderPattern: /(DEV_VALUE)|(PROD_VALUE)/},
  )
```

mitchellhamilton: tbh, i think this is readable enough and i'm not a fan of the relying on casing by default of the babel template api so I'd rather have it like this than have to look up the api of babel template to understand what everything does. I doubt this will change much anyway so, I think it's fine how it is.

PR #967, opened 2018-10-26
[WIP] Use Hooks

mitchellhamilton:
**What**:
Use hooks for stuff

**Why**:
There's a slight performance and size improvement(~150 bytes). I also have some ideas on how hooks could help some things so I want to do this so we can use hooks.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

Not gonna merge this until hooks are stable but just wanted to open this to show some people


jgoux:
I made a codesandbox to demonstrate the usage of hooks with emotion 10 : https://codesandbox.io/s/3ykkxov026

Thanks for the advices @mitchellhamilton ! 🥇

packages/emotion-theming/src/with-theme.js
```diff
@@ -3,20 +3,20 @@ import * as React from 'react'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import { ThemeContext } from '@emotion/core'
 
+let useContext: <Value>(context: React$Context<Value>) => Value = (React: any)
+  .useContext
+
 type Props = { theme: Object }
 
 // should we change this to be forwardRef/withCSSContext style so it doesn't merge with props?
+// should we remove this altogether and tell people to useContext
```
Andarist: we should! :) although we could simplify it for the users and provide
```js
const useTheme = () => useContext(ThemeContext)
```

TrySound:
Hey, guys!
I'm personally not ready to add jsx pragma into my project and would be cool to have a way around it. How about this kind of api?

```js
import { useEmotion } from '@emotion/hook';

export const Component = () => {
  const { css, cx, theme, injectKeyframes, injectGlobal } = useEmotion()
  const container = css({ width: '100%', padding: theme.spaces[1] })
  const fade = injectKeyframes({
    from: { opacity: 0 },
    to: { opacity: 1 }
  })
  return (
    <div className={container}></div>
  )
}
```

- it should work well with ssr
- it's easy to add out of the box theming if necessary
- per component caching is also possible
- locally injected keyframes and globals can be destroyed along with component

Andarist:
Why the jsx pragma is a problem for you? Im curious - would be great to know people problems to potentially solve them.

As to the hooks - what would be the advantage of this over just importing "bare" emotion and using it in render?

TrySound:
Some tools (like fastpack) do not support jsx pragma. Flow had also problems with it, not sure about now.

The advantage and the ability to connect with react context to achieve caching and SSR. As far as I understand bare import is not just work in v10.

Also it would be really good to make injectKeyframes and injectGlobal destroyable.

TrySound:
jsx pragma looks to me like a workaround and it still adds a noise in dev tools like styled.

mitchellhamilton:
> Some tools (like fastpack) do not support jsx pragma. Flow had also problems with it, not sure about now.

Then fastpack should support it, why should api decisions be based on a small technical restriction in a bundler that could be fixed? It also works completely fine with flow.

> The advantage and the ability to connect with react context to achieve caching and SSR.

An api like that wouldn’t enable zero config ssr like the other v10 apis allow because we need to render elements into the tree.

> As far as I understand bare import is not just work in v10.

The `emotion` package will still work in v10, it just won’t support zero config ssr.

istarkov:
Guys beautiful library, thank you work your work!!!

IMO ability to access `css` without pragma having zero SSR config is must have.

Few questions:

- What if I already have other `jsx pragma`?

And how to be with types?
Most APIs support typed `className` as a prop (via PropTypes, flow, typescript) but not `css` prop.  To avoid errors I can use such API only like:
```<SomeThirdPartyComponent className={css`blabla:blala;`}```
and in this case all zero config ssr is lost because no `css` property is allowed on `SomeThirdPartyComponent`.

In case above `pragma` is useless for me, as I prefer typed code without errors, I like to see no PropTypes warnings in console.

Hook variant looks good for me, at least we must be able to create such hooks in userspace.
Is it possible with current API?

**Comment by istarkov** [comment]
BTW now hooks version with zero config is impossible to create, Suspense with server rendering support should help but its not even near now

**Comment by FezVrasta** [comment]
I think a `const [css, cx ] = useClassNames()` would be a great way to use hooks in Emotion.

**Comment by TrySound** [comment]
@FezVrasta It won't work for zero config ssr. css/cx calls are tracked when render props is executed and then generated styles are put in `<><style></style>{childrenResult}</>`

Because of this I still use emotion v9 and not gonna upgrade.

**Comment by klarstrup** [comment]
@TrySound v10 with CacheProvider gives a similar dev experience to v9

That said I do imagine a way to make a `useClassNames` or whatever work by using a side effect that makes the closest parent `styled.x`/`CacheProvider`/`EmotionBoundary` render the actual style tags 🤔

**Comment by corysimmons** [comment]
lol I left styled-components several months ago because of this (it seemed like Emotion's Babel plugin fixed this at some point).

Now styled-components has fixed this: https://github.com/styled-components/styled-components/pull/2390

**Comment by FezVrasta** [comment]
Is this PR going to be resurrected? I can take care of it if you need help.

File: packages/emotion-theming/src/theme-provider.js [hunk 2]
```diff
@@ -42,19 +45,15 @@ type Props = {
 }
 
 let ThemeProvider = (props: Props) => {
+  let theme = useContext(ThemeContext)
+
+  if (props.theme !== theme) {
+    theme = createCacheWithTheme(theme)(props.theme)
```
Andarist: seems like this could be memoized to improve perf for nested ThemeProviders etc

packages/styled-base/src/index.js
```diff
@@ -72,99 +74,87 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
+        if (finalShouldForwardProp(key)) {
+          newProps[key] = props[key]
+        }
+      }
+
+      newProps.className = className
+
+      newProps.ref = ref || props.innerRef
+
+      if (process.env.NODE_ENV !== 'production' && props.innerRef) {
```
Andarist: as this can't land without bumping major version I think we can remove support of innerRef here altogether

packages/emotion-theming/src/theme-provider.js
mitchellhamilton: It's already memoized with weak maps

Andarist: Oh, you are right, cool!

packages/core/src/global.js
```diff
@@ -38,111 +37,96 @@ export let Global: React.StatelessFunctionalComponent<
...
+      let sheet: StyleSheet = ((sheetRef.current: any): StyleSheet)
+      if (sheet.tags.length) {
+        // if this doesn't exist then it will be null so the style element will be appended
+        let element = sheet.tags[sheet.tags.length - 1].nextElementSibling
+        sheet.before = ((element: any): Element | null)
+        sheet.flush()
+      }
+      cache.insert(``, serialized, sheet, false)
+    },
+    [cache, serialized]
```
Andarist: previous version was only updating styles on `serialized.name` change, I'm not sure if the previous version was safe - because I don't quite understand the cursor/keyframes thing and what guarantees it gives in regards to `serialized.name`. Would love to learn about it though, could you explain this a little bit? 

If we can't put `serialized.name` in the inputs array then we can remove this array altogether, as `serialized` changes with each render - so technically this input array doesn't give us anything right now

mitchellhamilton: If `serialized.name` is the same, the style is the same since the hash of the contents of the keyframes is in the style it's part of the whole styles' hash.

PR #973, opened 2018-10-28
RFC: Theming in v10 and beyond

mitchellhamilton:
I wrote my thoughts on what I think theming should look like with Hooks as documentation so read it on GitHub or on the deploy preview.

I think this is what we should primarily recommend because I’m not sure about the practicality of #887  in the real world and I’m not sure if the trade off in terms of dynamicism will really translate to an important performance improvement in the real world. I'm thinking of publishing it outside of the emotion packages and maybe trying it in a real project to see how well it works and then think about making it an official emotion package.

We should also absolutely keep on supporting the current theming API for at the absolute minimum, v10 but not provide the theme to the css prop, Global or ClassNames.

### DIFF
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -2,37 +2,92 @@
 title: 'Theming'
 ---
 
-Theming is provided by the library [`emotion-theming`](https://emotion.sh/docs/emotion-theming).
+Emotion 10 and above doesn’t have any explicit theming API, instead, we recommend using React’s Context and Hooks which work well together with Emotion. This has many benefits over having a built in theming API which are explained below.
 
-```bash
-npm install -S emotion-theming
-```
-
-Add `ThemeProvider` to the top level of your app and access the theme with `props.theme` in a styled component or provide a function that accepts the theme as the css prop. The api is laid out in detail [in the documentation](https://emotion.sh/docs/emotion-theming).
+We recommend creating a file which exports a ThemeProvider component and a useTheme hook.
 
 ```jsx
-// @live
-/** @jsx jsx */
-import { jsx } from '@emotion/core'
-import styled from '@emotion/styled'
-import { ThemeProvider } from 'emotion-theming'
+import { createContext, useContext } from 'react'
 
-const theme = {
+let defaultTheme = {
   colors: {
-    primary: 'hotpink'
+    primary: 'hotpink',
+    secondary: 'green'
   }
 }
 
-const SomeText = styled.div`
-  color: ${props => props.theme.colors.primary};
-`
+let ThemeContext = createContext(defaultTheme)
 
-render(
-  <ThemeProvider theme={theme}>
-    <SomeText>some text</SomeText>
-    <div css={theme => ({ color: theme.colors.primary })}>
-      some other text
-    </div>
-  </ThemeProvider>
-)
+export let ThemeProvider = ThemeContext.Provider
+
+export let useTheme = () => useContext(ThemeContext)
 ```
+
+The theme can be used with css prop like this.
+
+```jsx
+/** @jsx jsx */
+import { jsx } from '@emotion/core'
+import { useTheme } from './theme'
+
+let Button = props => {
+  let theme = useTheme()
+  return (
+    <button
+      css={{
+        color: theme.colors.primary
+      }}
+      {...props}
+    />
+  )
+}
+
+let SecondaryButton = props => {
+  let theme = useTheme()
+  return (
+    <button
+      css={{
+        color: theme.colors.secondary
+      }}
+      {...props}
+    />
+  )
+}
+```
+
+Or with the styled api like this
+
+```jsx
+import styled from '@emotion/styled'
+import { useTheme } from './theme'
+
+let Button = styled.button(() => {
+  let theme = useTheme()
+
+  return {
+    color: theme.colors.primary
+  }
+})
+
+let SecondaryButton = styled.button`
+  color: ${() => useTheme().colors.secondary};
+`
+```
+
+## Why don’t we have a built in theming API?
+
+Theming in emotion v9 and earlier was inspired by theming in styled-components which was created before React’s new Context and Hooks existed. This meant having access to a value through context was novel but with React's new features, problems with that API can be solved and having a theming API built in is unnecessary
+
+### Type Safety
+
+The previous theming API was extremely hard to statically type. By using context and hooks directly, a default value can be provided so type systems can guarantee that there's always a theme that can be accessed.
+
+### Global Namespace
+
+Since there’s was a single object which contained the theme, if a component from npm or somewhere wanted to use theming, it had to choose a unique name store it’s theme in which meant name conflicts could occur. By directly using the Context API, this problem doesn't exist.
+
+### Testing
+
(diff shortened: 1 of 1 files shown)


docs/theming.md
```diff
@@ -2,37 +2,92 @@
...
+
+Theming in emotion v9 and earlier was inspired by theming in styled-components which was created before React’s new Context and Hooks existed. This meant having access to a value through context was novel but with React's new features, problems with that API can be solved and having a theming API built in is unnecessary
+
+### Type Safety
+
+The previous theming API was extremely hard to statically type. By using context and hooks directly, a default value can be provided so type systems can guarantee that there's always a theme that can be accessed.
+
+### Global Namespace
+
+Since there’s was a single object which contained the theme, if a component from npm or somewhere wanted to use theming, it had to choose a unique name store it’s theme in which meant name conflicts could occur. By directly using the Context API, this problem doesn't exist.
```
Andarist: This shouldnt be relevant point here because it refers to legacy context and it's not a problem with new context APi.

docs/theming.md
```diff
@@ -2,37 +2,92 @@
...
+
+### Testing
+
+To test components that used the previous theming API, either a default theme has to be provided with defaultProps or a ThemeProvider has to be wrapped around every component which could get really inconvenient. By providing a default value when the theme is created, components can just be rendered.
```
Andarist: I don't feel strongly about this one - you might improve DX for testing here, but you deteriorate (to some extent) DX for writing components

docs/theming.md
```diff
@@ -2,37 +2,92 @@
...
+
+### Over-subscription
```
Andarist: Preferring css variables for theming this also is not quite a problem, right?

docs/theming.md
```diff
@@ -2,37 +2,92 @@
...
-Add `ThemeProvider` to the top level of your app and access the theme with `props.theme` in a styled component or provide a function that accepts the theme as the css prop. The api is laid out in detail [in the documentation](https://emotion.sh/docs/emotion-theming).
+The theme can be used with css prop like this.
 
 ```jsx
-// @live
 /** @jsx jsx */
 import { jsx } from '@emotion/core'
+import { useTheme } from './theme'
+
+let Button = props => {
```
Andarist: this recommendation suffers from `shouldForwardProp` problem, which really gets annoying long term from DX perspective

Andarist: Maybe this example shouldnt mention `button` but rather a composite component

docs/theming.md
Andarist: oh, i have kinda skipped the part where you mention libraries built on top of emotion 😅

docs/theming.md
Andarist: > How does it deteriorate DX for writing components?

Current way of using theme is way more ergonomic (in my opinion) than having to manually use theme context

docs/theming.md
Andarist: right, ive misjudged the example by not thinking about it too much and having styled API with interpolations in mind, css prop doesnt suffer from this

tkh44: We should probably dedicate a page on the docs dedicated to mapping solutions to "known" apis like `withComponent`, `shouldForwardProp`, etc. We should also show how easy it is to add further functionality to components this way.

ianstormtaylor:
I totally agree with this, this is a great direction. `useTheme` will be the easiest way to mentally think about these things.

My only request would be that emotion ships with the following as default:

```js
import { createContext, useContext } from 'react'
const ThemeContext = createContext({})
export const ThemeProvider = ThemeContext.Provider
export const useTheme = () => useContext(ThemeContext)
```

That way for the simple cases we can just `import { ThemeProvider, useTheme }` from emotion directly without having to define the context ourselves. And then people are free to define their own theme context if they really need to, with the default values.

Right?

docs/theming.mdx
```diff
@@ -38,21 +53,71 @@ render(
...
+```jsx
 import styled from '@emotion/styled'
-import { ThemeProvider } from 'emotion-theming'
+import { useTheme } from './theme'
 
-const theme = {
-  colors: {
-    primary: 'hotpink'
+let Button = styled.button(() => {
+  let theme = useTheme()
```
stramel: This usage and the below usage trigger errors in the latest React. 

```console
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```

Andarist: @stramel have you tried to use this with emotion 10 or 11? It can only work with the latter.

stramel: Oh let me try! Thank you for the quick reply!

stramel: @Andarist I'm running it inside Docz v2 using `@emotion/styled` and `@emotion/core` on version `11.0.0-next.9`. 

When starting the project it complains because it can't find `@emotion/styled-base`. 

```console
ERROR #98123  WEBPACK

Generating SSR bundle failed

Can't resolve '@emotion/styled-base' in 'Projects/libs/atoms/src/lib/Button'
```

so I installed `@emotion/styled-base` version `v11.0.0-next.3` and it is telling me i need to use `@emotion/styled`? 

When I tried the `v10` version of `@emotion/styled-base` it obviously failed... 😞

Andarist: Seems like maybe you have 2 versioms of emotion in your node modules for some reason or that you are using v10 of our babel plugin. `@emotion/styled-base` has been removed in v11 in favor of `@emotion/styled/base`.

If you prepare a runnable repro case then I could take a look at it.

stramel: Alright, let me do some debugging on what you mentioned might be the problem before I make to you chase a ghost.

stramel: @Andarist Looks like you were correct. Just tried it in my Next.js app with v11 and no errors. Doing some debugging...

`yarn why @emotion/core` and `yarn why @emotion/styled` both result in v10 and v11 because Docz and the gatsby docz theme both use v10

weyert: Any suggestions how to fix it?

stramel: I filed an issue on docz asking for a bumping to v11 but I'm not sure what the workaround would be. https://github.com/doczjs/docz/issues/1335

weyert: Thank you @stramel! Anyone know the expected release date of v11?

Andarist: I can't promise anything, but hopefully within a month. We have most of the things done, some final touches left, but this is a work we are doing in our free time so there is also a possibility that this will take some more time.

JakeGinnivan:
@mitchellhamilton I think the work on theming in v11 is better than this approach of stripping it out?

Thoughts on closing this?

Andarist:
We dont intend to remove builtin theming but some of the problems with it were nicely outlined here. We should „compile” those parts into docs and recommend using builtin theming only when building apps while using custom theme approach where developing libraries on top of emotion

PR #974, opened 2018-10-29
Remove theme from new APIs

mitchellhamilton:
**What**:
Remove passing the theme to the css prop, Global and ClassNames

**Why**:
Because of #973, even if we release v10 before hooks are available, I don't think we should introduce an API and then remove it soon after, especially since the usage of this would be spread throughout a codebase and practically impossible to codemod.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


Andarist:
Not long time ago you've thought that exposing `theme` in those places were a good idea. What has changed? Just ease of how theme can be accessed? Was it really harder before? 

I realize that this question might sound a little bit harsh (not being native english speaker I'm not always sure how to express my intentions clearly), but that's just an honest question - I'm literally curious.

mitchellhamilton:
> Not long time ago you've thought that exposing theme in those places were a good idea. What has changed? Just ease of how theme can be accessed? Was it really harder before?

Mainly how it can be accessed. Having it accessible without having to add a render prop or HOC is so much more convenient. I've thought the current theming API had many problems(personally, my biggest problem with it is type safety) for a while but I wasn't sure how to adequately solve it and hooks along with context solve it perfectly so I don't want to add an API and then soon after deprecate. I thought adding the theme to these places was a good idea because the tools to build a better theming API didn't exist, but now they do so I want to prepare for it.

> I realize that this question might sound a little bit harsh (not being native english speaker I'm not always sure how to express my intentions clearly), but that's just an honest question - I'm literally curious.

No worries!!! Please be critical!! I want to make emotion as good as it can possibly be and having criticism is important for that!

mitchellhamilton:
I've been thinking about this more and talked to @tkh44 and we've decided to keep this at the absolute minimum for v10(likely for a couple major versions) since it's kind of strange to remove an api when a replacement isn't available yet and the amount of code it requires is small and will get even smaller when hooks are ready

PR #975, opened 2018-10-29
Enable labels by default in vanilla macro

mitchellhamilton:
This is already how it is for the other babel things, so just add it to this one as well.

There's also a small warning message change in this.

### DIFF
--- a/packages/babel-plugin-emotion/src/macro.js
+++ b/packages/babel-plugin-emotion/src/macro.js
@@ -30,7 +30,7 @@ export let createEmotionMacro = (instancePath: string) =>
               babel,
               state,
               path,
-              shouldLabel: state.opts.autoLabel
+              shouldLabel: true
             })
             if (node) {
               path.node.arguments[0] = node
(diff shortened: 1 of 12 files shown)


packages/babel-plugin-emotion/src/macro.js
```diff
@@ -30,7 +30,7 @@ export let createEmotionMacro = (instancePath: string) =>
               babel,
               state,
               path,
-              shouldLabel: state.opts.autoLabel
+              shouldLabel: true
```
Andarist: shouldnt this be dependant on `process.env.NODE_ENV` ("runtime" one) or at least configurable with the default being true?

mitchellhamilton: It's dependent on the build time `process.env.NODE_ENV`, the check is [here](https://github.com/emotion-js/emotion/blob/6cb8d15438b01b8623af42f304d5ea3032332187/packages/babel-plugin-emotion/src/utils/label.js#L28-L30). Since the main use case for the macros will be Create React App and `process.env.NODE_ENV` is set to production for production builds in Create React App, I think doing this is safe.

mitchellhamilton: Yeah, the reason it's like that is that the styled transform doesn't insert the label into the styles, it inserts it into the the styled call so shouldLabel should be disabled there.

PR #976, opened 2018-10-30
Auto label for css prop without babel plugin

emmatown:
**What**:
Automatically add labels in some cases for the css prop without the babel plugin

**Why**:
So that using emotion without the babel plugin is nicer in more cases!

**How**:
Throwing errors and reading stack traces 😛

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/__tests__/__snapshots__/vanilla-emotion.js.snap
+++ b/packages/babel-plugin-emotion/__tests__/__snapshots__/vanilla-emotion.js.snap
@@ -69,7 +69,7 @@ let someCls =
 /*#__PURE__*/
 _css({
   color: window.whatever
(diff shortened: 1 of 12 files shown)


packages/core/src/jsx.js
```diff
@@ -112,6 +128,23 @@ export const jsx: typeof React.createElement = function(
     }
   }
   newProps[typePropName] = type
+  if (process.env.NODE_ENV !== 'production') {
+    try {
+      throw new Error()
+    } catch (e) {
+      if (e.stack) {
+        // chrome
```
Andarist: 🙈 they really should standardize this 😭 although there are horrible compat issues like this https://bugzilla.mozilla.org/show_bug.cgi?id=1498257#c11

packages/core/src/jsx.js
```diff
@@ -112,6 +128,23 @@ export const jsx: typeof React.createElement = function(
     }
   }
   newProps[typePropName] = type
+  if (process.env.NODE_ENV !== 'production') {
+    try {
+      throw new Error()
```
tkh44: This is really clever 😄 Are there any performance implications?

mitchellhamilton: It's only for dev so it doesn't really matter much but I checked the benchmarks and it seems fine.

PR #988, opened 2018-11-03
Simplify browser dead code elimination

mitchellhamilton:
**What**:
Simplify browser dead code elimination

**Why**:
Because the current solution was convoluted, it would have been relatively easy to break the dead code elimination and it was emotion-specific. (I'm thinking about extracting the build stuff for use in other things)

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete

Yes. This does add some duplication. I think it's worth it since it doesn't affect the size of browser builds and it makes this much simpler.


Andarist:
Extracting a reusable helper would be cool. Was thinking about it myself too. I've been using a custom macro for this - https://github.com/andreypopp/react-textarea-autosize/blob/f4f49e84c4beccf29188c72c2a3b21899bc0e85a/src/isBrowser.macro.js and I have already copied it multiple times. 

Was thinking about a macro that would inline `process.env.BROWSER || typeof document !== 'undefined'` by default with a configurable option (macros support config file) with which someebody could point to a `isBrowser.js` file in their project.

mitchellhamilton:
I think writing `typeof document !== 'undefined'`(and the setup required in a build process is also minimal since it's just a replace and it still works without a replace) requires such little effort that I don't really see the value of making a helper for it.

Not a fan of `process.env.BROWSER` since it's completely unstandardised and I don't see why it would or should become standardised when replacing `typeof document` with `'object'` is much easier and doesn't require any additional effort from tooling or users for the same benefit.

PR #992, opened 2018-11-04
Primitive components with the CSS Prop doc

mitchellhamilton:
**What**:

**Why**:
https://github.com/emotion-js/emotion/pull/973#discussion_r228997497

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I'm not totally sure about the name "primitive components" because people could think it's related to react-primitives.


zhDmitry:
But how to use component as selectors then ?

PR #999, opened 2018-11-06
update react-primitives & react-native-web dependencies

pcooney10:
**What**: Updated the `react-primitives` and `react-native-web` dependencies for `emotion-primitives`.

**Why**: React DOM 16.5 changed an unstable API that `react-native-web` depends upon. Namely,  `EventPluginHub` was removed (see [#1096](https://github.com/necolas/react-native-web/issues/1096) & (#506dba9)[https://github.com/necolas/react-native-web/commit/506dba933ce69830feaace08aee85c38ce3e59fb]). This was addressed in 16.5.1, and react-native-web 0.9.0.
I see that this library is already using ^0.9.x of react-native-web, so I figured it'd make sense to update the `react-primitives` dependency that is now also using ^0.9.x.

**How**: I cloned the repo, updated the dependencies, and ran the tests.

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,7 @@
   "resolutions": {
     "**/react": "16.5.2",
     "**/react-dom": "16.5.2",
-    "**/browserslist": "^3.2.8",
-    "**/react-primitives/react-native-web": "0.9.6"
+    "**/browserslist": "^3.2.8"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "**/react": "16.5.2",
     "**/react-dom": "16.5.2",
     "**/browserslist": "^3.2.8",
-    "**/react-primitives/react-native-web": "0.9.3"
+    "**/react-primitives/react-native-web": "0.9.6"
```
mitchellhamilton: If react-primitives depends on the latest version of react-native-web, I’d rather remove this resolution altogether.

pcooney10: Good call. Please see the updated PR.

PR #1007, opened 2018-11-12
Change version ranges

### LINKED ISSUE #1003: Using emotion 10 in a project that also has react-select with a dependency on emotion 9
I get below error. Maybe because the cache is only initialized using the provider, but this calls before react has started rendering:
image

mitchellhamilton:
**What**:
Remove the the caret for dependencies on `10.0.0-beta.x` packages.

**Why**:
Closes #1003 

The reason this is necessary is because `^10.0.0-beta.8` for example will include versions on `10.0.0-really-unsafe-please-do-not-use.x` which is bad because those releases are old.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -19,8 +19,8 @@
     "test:typescript": "dtslint types"
   },
   "dependencies": {
-    "@emotion/cache": "^10.0.0-beta.8",
-    "@emotion/css": "^10.0.0-beta.8",
+    "@emotion/cache": "10.0.0-beta.8",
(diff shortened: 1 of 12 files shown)


Andarist:
Im not exactly sure how prereleases work in regards to semver ranges - but from what i know the prerelease tags are compatible in alphabetic order. So -really is higher than -beta. RxJS had recently similar issues (after releasing some funny name on april fools day) and they had to resolve to using some weirdly named tag starting on “stronger” later.

PR #1017, opened 2018-11-14
Update css-prop.md

tkh44:
Just some doc work.

**What**:

**Why**:

This is going to be an important part of emotion 10.

I would like to add a few more items and gotchas later on for eslint configurations and babel setup for the pragma.

**Checklist**:

- [x] Documentation
- [n/a] Tests
- [n/a] Code complete


docs/css-prop.md
```diff
@@ -32,27 +78,37 @@ render(
 )
 ```
 
-You can also pass objects directly to the css prop, for more usage with objects, look at [the object styles page](/docs/object-styles.md).
+## Style Precedence
 
-```jsx
-// @live
+The `className` prop value overrides the styles provided to the `css` prop which is an important distinction when passing class names generated by emotion. Emotion removes generated class names from the `className` prop value, their underlying styles composed on top of the styles supplied to the `css` prop generating a new, single class name. The precedence order may seem counter-intuitive, but it allows developers predictably compose components and their styles.
```
tkh44: "Emotion removes generated class names from the `className` prop value"

should be

Emotion removes class names generated by itself from the `className` prop value

**I think**

docs/css-prop.md
```diff
@@ -32,27 +82,37 @@ render(
...
+
+The following example demonstrates a potential problem if the `css` prop overwrites styles from `className` prop.
```
jfrolich: This is not very clear. Maybe just say, if the same style is defined in the css prop *and* className, the style of the classname will be used (has a higher precedence).

docs/css-prop.md
```diff
@@ -32,27 +82,37 @@ render(
...
+    {...rest}
   >
-    This has a hotpink background.
-  </div>
+    This text is darkgray
+  </P>
 )
 ```
 
+The author of the `TextBlock` component is expecting the text color of the rendered `P` component to be "darkgray" and further, users of the `TextBlock` component might want to customize the `color` value. The pattern in the example allows developers to effortlessly compose components and overwrite styles in a predictable manner.
```
jfrolich: In this case we can customize the P tag and make sure the the new styles overwrite the default styles.

jfrolich: It would be great to document what to do if you want reverse precedence with an example (maybe same example where we want to make sure the color is *always* hotpink. 
I actually don't know how to do it. `cx` from `ClassNames` cannot deal with css styles (maybe a bug?)

Speaking about it, probably better to expose ClassNames as a hook, it would be less ceremony than a HOC.

jfrolich:
I added some comments

docs/css-prop.md
mitchellhamilton: > It would be great to document what to do if you want reverse precedence with an example (maybe same example where we want to make sure the color is always hotpink.
I actually don't know how to do it. cx from ClassNames cannot deal with css styles (maybe a bug?)

When would you want to that?

> Speaking about it, probably better to expose ClassNames as a hook, it would be less ceremony than a HOC.

`ClassNames` renders elements in the tree so it's not possible. (or at least it would be a sub-optimal api).

docs/css-prop.md
tkh44: Yeah, this is rough. Thanks.

docs/css-prop.md
jfrolich: > `ClassNames` renders elements in the tree so it's not possible. (or at least it would be a sub-optimal api).
But we can expose cx and css using hooks right (as they just use context) as I read in the source code. Maybe I miss the server rendered part but hooks can have side effects.

jfrolich: > When would you want to do that?
Sometimes you might want to add styles after the user provided styles. This is not the most common use case, but still. Maybe override styles based on other props.

mitchellhamilton: > as they just use context

They don't just use context. [They render elements into the tree.](https://github.com/emotion-js/emotion/blob/master/packages/core/src/class-names.js#L115-L131)


> Sometimes you might want to add styles after the user provided styles. This is not the most common use case, but still. Maybe override styles based on other props.

Could you provide a code example of when you would want to do it in the real world? It's hard to talk about stuff like this without seeing actual use cases.

jfrolich: I saw that, but the DOM mutation can be done using `useEffect` I belief? Another option is to make the hook only work client-side, and provide `ClassNames` as something that also server renders.

mitchellhamilton: > I saw that, but the DOM mutation can be done using useEffect I belief?

On the client, it doesn't matter. The DOM stuff can and does happen in render since it's cached and doesn't have any visible side effects.

> Another option is to make the hook only work client-side, and provide ClassNames as something that also server renders.

That would be a really confusing API and terrible for users who start with a client side app and decide to do SSR later, they would have to rewrite a bunch of code later.

jfrolich: > > I saw that, but the DOM mutation can be done using useEffect I belief?
> 
> On the client, it doesn't matter. The DOM stuff can and does happen in render since it's cached and doesn't have any visible side effects.

So it *can* be done using `useEffect` if it's just appending to the DOM? This means that implementing useClassNames as a hook is possible.

it will be this: 

```es6
const Comp = () => {
  const { cx } = useClassNames();
  return <OtherComp className={cx('bla', 'bla')} />
}
```

Instead of this:

```es6
const Comp = () => {
  const { cx } = useClassNames();
  return (
    <ClassNames>
      {
        ({ cx }) => 
          <OtherComp className={cx('bla', 'bla')} />
      }
    </ClassNames>
  )
```

Which in my opinion would be quite a nice improvement.

mitchellhamilton: No, for SSR, the elements need to be rendered into the tree and having one api for the client and one for ssr is bad DX

jfrolich: BTW how does emotion update the DOM (add the style tags) without a higher level component in server render mode without using HoCs. Can we use something similar in this scenario?

mitchellhamilton: > BTW how does css + createElement update the DOM (add the style tags) without a higher level component in server render mode without using HoCs. Can we use something similar in this scenario?

It renders elements into the tree just like other component, it wouldn’t help this.

jfrolich: > What would it do if it doesn’t involve rendering?

So I mean exposing `css` and `cx`, they are now inline functions. We can make it functions that are exported with the themeContext and cacheContext. Then it allows it to be part of a hook because they don't render anything.

mitchellhamilton: But it still needs to render elements so that wouldn't help anything.

jfrolich: Only if it's not server rendered.

jfrolich: Another idea is to get the element as part of the output of the hook, so you can include it in the render tree when it is a server rendered app.

mitchellhamilton: Yes and having an API that only works on the client is something we don't want to do because a large part of the point of emotion 10 is that SSR just works without having to do something special.

mitchellhamilton: > Another idea is to get the element as part of the output of the hook, so you can include it in the render tree when it is a server rendered app.

That would still break the idea of SSR just working because if it's optional, many people won't do it and then they'll have to refactor it later if they want SSR. It also create two APIs that people have to learn, a SSR version and a non-SSR version.

jfrolich: Alright anyway. Let's stop bikeshedding. It's not often used anyway.

jfrolich: Coming back to this:
> Could you provide a code example of when you would want to do it in the real world? It's hard to talk about stuff like this without seeing actual use cases.

Ehh you have darkMode prop that overrides some of the styled applied to the element.

Am I right that reverse precedence is not possible currently?

mitchellhamilton: > Am I right that reverse precedence is not possible currently?

Yes.

> Ehh you have darkMode prop that overrides some of the styled applied to the element.

What would that look like? IMO if a custom style is applied to an component, the consumer of the component likely always wants the style to be applied because it's likely specific to where it's being used and having it be less specific than the components own styles would mean a consumer can't override styles.

Issue #1004, opened 2018-11-09
Unterminated comment literal causes infinite loop


philip-peterson:
Hello! I was recently developing an app and the browser got stuck in an infinite loop. After some investigating, it turned out to be because I had an unterminated comment token.

Relevant code:

```javascript
const { css } = require('emotion');
const x = css`/*`;
```

What happened:

Infinite loop

Suggested solution:

Throw exception for unterminated comment token.

Andarist:
This seems like a stylis bug - gonna report it there right now.

Andarist:
When you have reported this I've added dev-only error in https://github.com/emotion-js/emotion/pull/1020 so it's easy to catch and correct the mistake. Seems like there is nothing more to do here. Thanks for the report!

Slapbox:
@Andarist I just hit this, or something closely related. The issue only occurs after bundling our project. I'd guess I've hit this sort of bug 5 or 6 times over the past year; always only triggered in our bundled production build. Again it does appear to be a Stylis bug though.

The string that was causing an infinite loop was `"line-height:1.45;width:50%;/* &:active,&:focus{border:1px solid #969296;;"`

If we can determine what exactly triggers this issue, it would be amazing to have a dev warning about this.

The input code looked like below. I'm guessing that perhaps nesting a single line comment inside of a multi-line comment caused it?

```js

return css`
  line-height: 1.45;
  width: 50%;

  /*
    &:active,
    &:focus {
      border: 1px solid ${editMode.button.active.border}; //  @TODO 3.14 -- Remove from palette if unused
    } */
`
```


![image](https://user-images.githubusercontent.com/6835891/94314992-ae4cf600-ff4f-11ea-8f94-4d3ba39cc133.png)

Andarist:
> The input code looked like below. I'm guessing that perhaps nesting a single line comment inside of a multi-line comment caused it?

Hard to tell. You would have to prepare a runnable repro case so I could take a look at this. 

> @Andarist I just hit this, or something closely related. The issue only occurs after bundling our project. I'd guess I've hit this sort of bug 5 or 6 times over the past year; always only triggered in our bundled production build. Again it does appear to be a Stylis bug though.

I strongly recommend checking out v11 - it's almost ready and includes Stylis v4 that has resolved all known issues that Stylis v3 had.

Slapbox:
[...]

PR #1020, opened 2018-11-15
Throw in dev mode when attempting to insert styles with unterminated …

Andarist:
Related to #1004

### DIFF
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -56,6 +56,31 @@ let createCache = (options?: Options): EmotionCache => {
   let stylis = new Stylis(stylisOptions)
 
   if (process.env.NODE_ENV !== 'production') {
+    const commentStart = /\/\*/g
+    const commentEnd = /\*\//g
+
+    const oldStylis = stylis
+    stylis = (selector, styles) => {
+      while (commentStart.test(styles)) {
+        commentEnd.lastIndex = commentStart.lastIndex
+
+        if (commentEnd.test(styles)) {
+          commentStart.lastIndex = commentEnd.lastIndex
+          continue
+        }
+
+        throw new Error(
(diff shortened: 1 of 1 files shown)


Andarist:
Sure, gonna make it a dev stylis plugin some time later - that's a good idea. As to the benchmarks - I can do that, but I'd appreciate you sharing a little know-how on how to do it properly, I know the basic idea (it's just running a thing multiple times after all), but maybe you could write up tips on how to make benchmarks reliable and meaningful.

Andarist:
I've written a simple benchmark https://esbench.com/bench/5bf7371a4cd7e6009ef61d0a . It's obviously slower, but the change is not that significant.

packages/cache/src/index.js
```diff
@@ -172,6 +173,32 @@ let createCache = (options?: Options): EmotionCache => {
...
+          while (commentStart.test(content)) {
+            commentEnd.lastIndex = commentStart.lastIndex
+
+            if (commentEnd.test(content)) {
+              commentStart.lastIndex = commentEnd.lastIndex
+              continue
+            }
+
+            throw new Error(
+              'Your styles have unterminated comment ("/*" without corresponding "*/"). Please fix it.'
```
mitchellhamilton:
```suggestion
              `Your styles have an unterminated comment("/*" without corresponding "*/") \`${content}\``
```

Andarist:
Sure, gonna add test for it - maybe would be good to move both dev plugins to separate files and test them separately? Or do you prefer testing them through emotion tests?

mitchellhamilton:
I would prefer testing them through emotion tests(that's how the other one is tested) since then we don't have to care about implementation details. I'm not sure I trust the dead code elimination enough to move them into a separate file anyway.

Andarist:
I've added few basic tests, let me know if this is enough.

PR #1021, opened 2018-11-15
renderStyleToString: Skip undefined styles

### LINKED ISSUE #1014: SSR break when element has css- in attribute value
- `emotion` version: 9.2.12
- `react` version: 16.6.3

Relevant code.

```javascript
import React from 'react';
import {css} from 'emotion';

const red = css`color: red;`;

const App = () => <a href="/css-guide" className={red}>Hello world</a>;

export default App;
```

What you did:

Run the above code in [Razzle](https://github.com/jaredpalmer/razzle), using `renderStylesToString(renderToString(<App />))`.

What happened:

The link appears in blue instead of red as expected.

screen shot 2561-11-13 at 13 21 59

The rendered markup is

```html
<style data-emotion-css="guide 1y2ijxn">undefined.css-1y2ijxn{color:red;}</style><a href="/css-guide" class="css-1y2ijxn" data-reactroot="">Hello world</a>
```

Problem description:

As renderStylesToString scan for `css-xxxx`, it matches the `css-guide` in `href` attribute.

Suggested solution:

createRenderStylesToString should parse the HTML using a parser instead of regular expression. I'm not sure how other CSS-in-JS solution do this but it would worth looking into.

A possible workaround is to rename cssKey into a random string.

whs:
**What**: Fixing SSR returning undefined on finding attribute with `css-` prefix

Closes #1014

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/emotion-server/test/inline.test.js
+++ b/packages/emotion-server/test/inline.test.js
@@ -33,6 +33,17 @@ describe('renderStylesToString', () => {
       emotionServer.renderStylesToString(renderToString(<Page2 />))
     ).toMatchSnapshot()
   })
+  test('skip undefined styles', () => {
+    const { css } = emotion
+    const style = css`
+      color: red;
+    `
+    const component = <a href={`${emotion.cache.key}-fail`} className={style} />
+
+    expect(
+      emotionServer.renderStylesToString(renderToString(component))
+    ).toEqual(expect.not.stringContaining('undefined'))
+  })
   test('renders large recursive component', () => {
     const BigComponent = createBigComponent(emotion)
     expect(
--- a/packages/create-emotion-server/src/inline.js
+++ b/packages/create-emotion-server/src/inline.js
@@ -62,7 +62,7 @@ const createRenderStylesToString = (
     // $FlowFixMe
     const id = match[1]
     const style = inserted[id]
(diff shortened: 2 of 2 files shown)


packages/emotion-server/test/inline.test.js
```diff
@@ -33,6 +33,17 @@ describe('renderStylesToString', () => {
...
+  test('skip undefined styles', () => {
+    const { css } = emotion
+    const style = css`
+      color: red;
+    `
+    const component = <a href={`${emotion.cache.key}-fail`} className={style} />
+
+    expect(
+      emotionServer.renderStylesToString(renderToString(component))
+    ).toEqual(expect.not.stringContaining('undefined'))
```
mitchellhamilton: Could you add a snapshot of the html?

whs:
Changed to snapshot test.

At first I was considering snapshot test, but since you should have failing test first in TDD I choose asserting output directly.

mitchellhamilton:
Could you add the other test back as well so there is a snapshot test and the test asserting that undefined isn't in the string. It's nice to have a snapshot test to see what the output is but also have the test checking for the specific behavior we're looking for.

whs:
Added

PR #1023, opened 2018-11-17
Use preconstruct

mitchellhamilton:
**What**:
Use [preconstruct](https://github.com/preconstruct/preconstruct) which is a build tool for libraries I've been working on.

**Why**:
The old build system _worked_ but it had no tests at all, didn't enforce certain things like package.json fields so I built preconstruct based on the old emotion build system but with much nicer DX and tests.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete


Andarist:
This is cool idea. WDYT though about combining powers and making microbundle more robust and handling all of those things that you want by default?

mitchellhamilton:
I'm not sure, there's lots of things that I want to be really opinionated about and I think that people will likely disagree with some of the opinions. There's also some stuff that microbundle does that I don't want preconstruct to do.

Andarist:
You might be right, although I would still like to drag this discussion a little bit further 😉 Could u be more specific about which of your opinions might be controversial to people and which parts of the microbundle you dont like? I have done some work around in microbundle and have planned to do more, that's why im super interested in this and on the possibility to consolidate both as the overlap seems to be huge

mitchellhamilton:
My main problem with microbundle is that it infers things when it's potentially unsafe, e.g. inferring the umd name from the package.json name field, building bundles for things that aren't configured in the package.json. I'd prefer a tool to explicitly ask what the value of something should be if it can't infer it rather than doing something potentially incorrect so something might require more work to setup but it's guaranteed to be correct.

One potentially controversial thing that preconstruct does is it enforces the names of the dist files so e.g. for cjs builds with a package named `some-package`, the file has to be `dist/some-package.cjs.js` exactly otherwise the build will fail. There's also commands to set all of the fields to the correct values.

It also has some opinionated optimizations like using `object-assign` instead of inserting babel's helper. With that, it also requires that object-assign be a dependency of the package and if it's not, it offers to install it.

I also don't want to support typescript, at least for the foreseeable future.(that may change but right now, I want to make it good for a particular case rather than trying to solve everything)

preconstruct also builds browser builds whenever there is a `typeof window` or `typeof document` and it requires that you accept that browser builds will be generated(and the appropriate package.json fields will be written to) otherwise the build will fail.

The core difference between the two right now is that preconstruct favors failing early whereas microbundle tries to support as many use cases as possible and build as much as it can even if the output is invalid/not useful. (this is just my opinion for my use cases)

Andarist:
> My main problem with microbundle is that it infers things when it's potentially unsafe, e.g. inferring the umd name from the package.json name field

Not sure if this is truly unsafe - it only may yield unexpected result. It shouldn't break anything (other than maybe when migrating from hand-written solution with customized name). I admit though that it's better to force users to fill this in, preferrably with value automatically filled in by the tool (using fix feature).

> One potentially controversial thing that preconstruct does is it enforces the names of the dist files so e.g. for cjs builds with a package named some-package, the file has to be dist/some-package.cjs.js exactly otherwise the build will fail. There's also commands to set all of the fields to the correct values.

Personally I don't see it as that controversial - it's a good default when flat bundling and we can enforce this practice. Not sure how other microbundle contributors would feel about it but I strongly suspect that they wouldnt be againt it.

> It also has some opinionated optimizations like using object-assign instead of inserting babel's helper. With that, it also requires that object-assign be a dependency of the package and if it's not, it offers to install it.

This one could be offered enforced when having react as dep/peerDep and optionally enfored with some option like `preferObjectAssignPackage`. However I'm not sure if it's that good of a solution - personally I'd prefer enforcing runtime helpers and adding `@babel/runtime` as dep of the package. That would make all babel helpers shareable across packages, I realize that `object-assign` might be already in the bundle when using react, but that package is not even tree-shakeable now (doesnt have module field) and it handles symbols (for most cases it doesnt have to). Babel helper weighs 172 bytes (vs over 530 of object-assign) and doesnt suffer from mentioned problems.

> I also don't want to support typescript, at least for the foreseeable future.(that may change but right now, I want to make it good for a particular case rather than trying to solve everything)

Sure - supporting it would be our effort (microbundle contributors).

> preconstruct also builds browser builds whenever there is a typeof window or typeof document and it requires that you accept that browser builds will be generated(and the appropriate package.json fields will be written to) otherwise the build will fail.

That's a good enhancement - so not an argument against merging projects 😉 

> The core difference between the two right now is that preconstruct favors failing early whereas microbundle tries to support as many use cases as possible and build as much as it can even if the output is invalid/not useful. (this is just my opinion for my use cases)

I believe your approach is more sound - we could propose it as a new behaviour of microbundle.

mitchellhamilton:
> Not sure if this is truly unsafe - it only may yield unexpected result. It shouldn't break anything (other than maybe when migrating from hand-written solution with customized name). I admit though that it's better to force users to fill this in, preferrably with value automatically filled in by the tool (using fix feature).

The und name stuff was just an example of wider behaviour, such as generating bundles when the package.json fields aren’t set.

> This one could be offered enforced when having react as dep/peerDep and optionally enfored with some option like preferObjectAssignPackage.

This could be a good idea

> However I'm not sure if it's that good of a solution - personally I'd prefer enforcing runtime helpers and adding @babel/runtime as dep of the package. 

Not really about this but I’ve seen @Babel/runtime and @babel/runtime-corejs2 or something and I don’t really get it, what’s the difference and if there’s two, wouldn’t there be a chance for duplication?

> I realize that object-assign might be already in the bundle when using react, but that package is not even tree-shakeable now (doesnt have module field) and it handles symbols (for most cases it doesnt have to). Babel helper weighs 172 bytes (vs over 530 of object-assign) and doesnt suffer from mentioned problems.

The weight and tree shakability doesn’t really matter if it’s already in the bundle. Rather than it being 172 vs over 530, it’s more 172 vs practically 0.

> Sure - supporting it would be our effort (microbundle contributors).

It’s not really about the direct maintenance cost of a feature, it’s about how a feature like typescript affects other features. For example, preconstruct writes a file that redirects exports the flow types from the entrypoint. Would it also do that for ts or generate declarations or something? These are all questions that would need to be answered and there are even more questions when you have more features.

Andarist:
> The und name stuff was just an example of wider behaviour, such as generating bundles when the package.json fields aren’t set.

Gotcha - I believe though that we could get rid of those behaviours. Most likely they are there because lack of the fixer feature. With it IMHO it doesnt make sense to support them.

> Not really about this discussion specifically but I’ve seen @babel/runtime and @babel/runtime-corejs2 or something and I don’t really get it, what’s the difference and if there’s two, wouldn’t there be a chance for duplication?

There is a chance for duplication, however the corejs2 variant shouldnt be used for libraries - it's more meant to be used with apps (if somebody wants it).

@babel/runtime is smaller and has no dependencies, its helpers are "standalone". corejs2 variant is basically the same thing, but transpiled with @babel/transform-runtime with "usage" option (or smth like that) - so whenever a helper reference some non-es5 global like Promise, Symbol etc then the reference gets replaced to reference to corejs2 module. 

I admit this is kinda confusing.

> The weight and tree shakability doesn’t really matter if it’s already in the bundle. Rather than it being 172 vs over 530, it’s more 172 vs practically 0.

That's true - my thoughts were more about the idea of defaulting to `object-assign` for general purpose libraries, rather than about this specific case of react or when we want to enforce this package.

> It’s not really about the direct maintenance cost of a feature, it’s about how a feature like typescript affects other features. For example, preconstruct writes a file that exports the flow types from the entrypoint. Would it also do that for ts or generate declarations or something? These are all questions that would need to be answered and there are even more questions when you have more features.

Yeah, that's true - but this one should be rather easy to answer. You either have pure JS, flow or TS and you can choose your desired "mode" out of them.

Andarist:
What I propose is that Im going to work (slowly) on improving microbundle and migrating preconstruct features to it so we can see if we can consolidate both at the end, wdyt?

PR #1027, opened 2018-11-21
as prop docs

mitchellhamilton:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/docs/styled.md
+++ b/docs/styled.md
@@ -221,3 +221,5 @@ render(
 ```
 
 This API was inspired by [styled-components](https://www.styled-components.com). ❤️
+
+By default, the as prop only works on html tags so that custom components to use use the `as` prop. It can be overridden by passing a custom [`shouldForwardProp`](#customizing-prop-forwarding) that returns `false` to `'as'`.
--- a/docs/styled.md
+++ b/docs/styled.md
@@ -197,3 +197,27 @@ render(
   </Container>
 )
 ```
+
+### `as` prop
+
+To use styles from a styled component but change the element that's rendered, you can use the `as` prop.
+
+```jsx
+// @live
+import styled from '@emotion/styled'
+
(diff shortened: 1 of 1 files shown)


docs/styled.md
```diff
@@ -197,3 +197,27 @@ render(
   </Container>
 )
 ```
+
+### `as` prop
+
+To use styles from a styled component but change the element that's rendered, you can use the `as` prop.
```
Andarist: It would be great to explain under this section how prop filtering works and how you might change it with custom `shouldForwardProp`

docs/styled.md
```diff
@@ -197,3 +197,29 @@ render(
...
+    href="https://github.com/emotion-js/emotion"
+  >
+    Emotion on GitHub
+  </Button>
+)
+```
+
+This API was inspired by [styled-components](https://www.styled-components.com). ❤️
+
+By default, the as prop only works on html tags so that custom components to use use the `as` prop. It can be overridden by passing a custom [`shouldForwardProp`](#customizing-prop-forwarding) that returns `false` to `'as'`.
```
Andarist: "so that custom components to use use the `as` prop." -> "so custom components have to customize this for themselves." (or smth like that)

You can also mention in the brackets in the end that this customization works also for the html tags

mitchellhamilton: Why mention that it works for html tags? why would someone want to do it and how does telling the user help them?

Andarist: One might write custom `shouldForwardProp` for a html base and make `as` forwarded by accident.

IMHO its a good to mention all possible footguns below describing the most common cases. Otherwise it will be hard to find an explanation of a weird behaviour without looking into the source code.

PR #1033, opened 2018-11-22
Migrating to emotion 10 minor correction

penx:
emotion should be @emotion/core


mitchellhamilton:
Thanks for the PR but this is actually intentional, that import should be from the emotion package.

PR #1035, opened 2018-11-24
Remove react-emotion package

mitchellhamilton:
**What**:
Remove react-emotion package

**Why**:
Having this package is confusing since it's essentially an alias to `@emotion/styled` but people expect it to be like react-emotion from v9. The original reason for keeping it was to make migration easier for people using packages that haven't updated to v10 but there are easier ways to do do with bundlers and package managers that won't cause confusion.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


abisummers:
Is there a migration guide or something? I was confused while looking at the page on Yarn which now has no more readme: https://yarnpkg.com/en/package/react-emotion

Issue #1044, opened 2018-11-28
Bug with uppercase DATA- attributes in the latest is-prop-valid


mxstbr:
Ref: https://github.com/styled-components/styled-components/pull/2251

Andarist:
actually, `dAtA` is valid too 🙈and probably the same thing applies to aria

PR #1045, opened 2018-11-28
Handle uppercased DATA- & ARIA-

Andarist:
fixes https://github.com/emotion-js/emotion/issues/1044

I've prepared this quickly, not an expert of regexp specificity, but i think this correct - please check it though. 

Also - it seems to me that `-.*` should actually be replaced by `-.+`, WDYT?

cc @mxstbr @probablyup and ofc @mitchellhamilton


mitchellhamilton:
I think what we should do is make the regex case insensitive again and handle `on[A-Z].*` separately since it'll make people get more warnings from react if they incorrectly type props like `tabindex` instead of `tabIndex` and we can remove other things like two autofocuses.

Could you benchmark these two implementations and use whatever is fastest?
```jsx
let regex // the current regex but with the extra autofocus removed and case insensitive

function one(prop) {
  return (
    regex.test(prop) ||
    (prop.charCodeAt(0) === 111 &&
      prop.charCodeAt(1) === 110 &&
      prop.charCodeAt(2) < 91)
  )
}

let onRegex = /^on[A-Z]/

function two(prop) {
  return regex.test(prop) || onRegex.test(prop)
}

// in the actual implementation memoize it.
```

Andarist:
Sure, gonna prepare a benchmark later. At this point I think I will also have to add some basic tests. 

Thanks for code suggestions!

Andarist:
According to my benchmark - https://esbench.com/bench/5bfee68a4cd7e6009ef61d23 - it seems that actually sensitive regexp mixed with char codes test for handlers is the best (from all tested cases!). 

Gonna tweak the PR in a moment. How do you feel about keeping the base regexp case-sensitive?

Andarist:
friendly 🏓 @mitchellhamilton

PR #1048, opened 2018-11-28
Use known global (window) in browser build of emotion package

Andarist:


mitchellhamilton:
V10 actually removed the context option but I guess I forgot to stop passing it in the emotion package.

Andarist:
Closing in favour of https://github.com/emotion-js/emotion/pull/1053

PR #1049, opened 2018-11-29
RFC: Add babel preset for css prop

mitchellhamilton:
I have no idea whether this is a good idea but people seem to want it and are already doing it without a preset so maybe we should add a preset to do it.

Please don't review what the exact implementation right now, I just want to know whether people think it's a good idea. It's also important to keep in mind, how should we teach this? Having two "official" ways to do the same thing could get really confusing for consumers because they have to decide which thing they want to use. A babel preset also has the disadvantage that it doesn't work with CRA and other things that don't allow babel configuration so we would have to explain what people should use for their use case. Importing and using the jsx pragma has the advantage though that it works everywhere.


Andarist:
Imho its a good change - as mentioned people will do it anyway, so we can just ease it for them

packages/babel-preset-jsx/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@emotion/babel-preset-jsx",
```
tkh44: We should name this `@emotion/babel-preset-css-prop`. `jsx` is an implementation detail imo and could confuse people.

mitchellhamilton: Totally! That sounds so much better!

ChristopherBiscardi:
@Ailrun I do this already so maybe I can speak to it. Basically, writing a pragma at the top of every file in an application setting gets cumbersome (I still write the pragma for libs). Since it's the same thing over and over, it'd be nice to automate it. It's also "one extra thing" for junior engineers to contend with when they don't have to. Automating the writing of the pragma and the insertion of the import saves headspace to focus on real problems instead of trying to remember the syntax for the pragma, etc. The vast majority of people aren't going to be running multiple css-in-js libs that conflict here, so setting the pragma/import via a babel plugin is a low-cost way to remove a set of repetitive manual labor through automation.

mitchellhamilton:
If people want to review the implementation here, go ahead now.

.circleci/config.yml
```diff
@@ -8,9 +8,9 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - v2-dependencies-{{ checksum "yarn.lock" }}
+            - v3-dependencies-{{ checksum "yarn.lock" }}
```
Andarist: out of curiosity - why the change?

mitchellhamilton: The cache was breaking the build.

packages/babel-preset-css-prop/README.md
```diff
@@ -0,0 +1,37 @@
...
+yarn add @emotion/babel-preset-css-prop
+```
+
+## Usage
+
+**.babelrc**
+
+```json
+{
+  "plugins": ["emotion"]
```
wKovacs64:
```suggestion
  "presets": ["@emotion/babel-preset-css-prop"]
```

packages/babel-preset-css-prop/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+babel --presets @emotion/babel-preset-css-prop script.js
+```
+
+### Via Node API
+
+```javascript
+require('@babel/core').transform(code, {
+  plugins: ['@emotion/babel-preset-css-prop']
```
wKovacs64:
```suggestion
  presets: ['@emotion/babel-preset-css-prop']
```

wKovacs64:
Ran into an issue with fragments when messing around with this in Gatsby. Apparently the fragment shorthand syntax `<> ... </>` will throw an error if you don't set the [`pragmaFrag` option in `@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx.html):
```
Error: transform-react-jsx: pragma has been set but pragmafrag has not been set
```
I guess we just need to add this?
```js
pragmaFrag: 'React.Fragment'
```
⚠️  Disclaimer: I have no idea what I'm talking about. ⚠️

PR #1058, opened 2018-11-30
fix link in README

Fabious:
**What**: fix a broken link in the [emotion documentation](https://emotion.sh/docs/emotion)
CSS Link is broken

**Why**:
We got this message : `You just hit a route that doesn't exist... the sadness.😢`, and i don't want anybody to be sad!

**How**: just fix the link :)


Fabious:
Voilà

PR #1073, opened 2018-12-05
Add dirname value to babel plugin labelFormat option

knpwrs:
**What**:

A new value to use in the `labelFormat` option for the babel plugin.

**Why**:

In our project most styles end up in a file called `styles.js` or `styles.ts`. It would be nice to have the flexibility to name our classes after the directory when in development.

**How**:

Vim.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete


TrevorSayre:
This would be a huge help!

PR #1080, opened 2018-12-06
Fix Sketch support

### LINKED ISSUE #1062: [primitives] [emotion 10] Sketch example: "Can't find variable: document"
- `emotion` 10.0.0:

**Relevant code.**

Sketch Example at https://github.com/emotion-js/emotion/tree/master/packages/primitives

**What you did:**

I tried out the example.

**What happened:**

The following error occured: `ReferenceError: Can't find variable: document`. Traced back to packages/cache/src/index.js  -> line 70 : at the  createCache method.

**Problem description:**

This problem is not specifically caused by the primitives package but rather by the @emotion/cache package. The skpm builder internally uses webpack and is configured to use look for "browser" field in all the packages. The resulting browser build strips out the `if (isBrowser)` statements.

We could argue that the skpm builder uses an invalid webpack configuration and should exclude "browser" from the mainFields, but unfortunately this introduces a lot of errors.

Suggested solution:

I got the example working by removing the `browser` field from packages/cache/package.json.

Andarist:
fixes #1062

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -11,6 +11,10 @@
     "./dist/core.cjs.js": "./dist/core.cjs.js",
     "./dist/core.esm.js": "./dist/core.esm.js"
   },
+  "sketch": {
+    "./dist/core.cjs.js": "./dist/core.cjs.js",
+    "./dist/core.esm.js": "./dist/core.esm.js"
+  },
   "types": "types/index.d.ts",
   "files": [
     "src",


Andarist:
So... it doesn't work 😅My findings:

- all `mainFields` work the same
- but there also `aliasFields` in webpack (always thought `browser` is handled through `mainFields`, but actually it's handled by both) and this is a config which allows "object" aliases
- we can't just use `aliasFields` to redirect to main/module because they will get once more redirected by `"browser"` to browser equivalents

So I think we have 2 options:
- create distinct filenames for sketch files (even if they will be just copies of main/module) and use `aliasFields` for that (so I'd have to prepare another PR to skpm). **the same thing** should actually be done for `react-native` because it also poses a risk now when bundling with webpack
- as `@skpm/builder` is using webpack under the hood we can assume it supports ESM and we can point to a copy of the "module" entry (with distinct name!) from `"sketch"` field

mitchellhamilton:
I don't use the native/primitives stuff so @Andarist if you could choose whichever solution you think is best, I'd be happy with that.

PR #1089, opened 2018-12-08
Use `class` prop instead of `className` for custom components

### LINKED ISSUE #1071: css Prop on Web Components
- `emotion` version: @emotion/core version 10.0.0.2
- `react` version: 16.6.3

Relevant code.

```javascript
/** @jsx jsx */
import {jsx} from @emotion/core;

const myStyles = {
    color: 'blue',
};

function MyElement() {
    return <div css={myStyles}>
        {'I get styles applied.'}
    </div>
}

function MyWebComponent() {
    return <my-element css={myStyles}>
        {'I don't get styles applied.'}
    </my-element>;
}
```

What you did:

I tried to use the `css` prop on a custom, non-react HTML element.

What happened:
The styles detailed in the `css` prop aren't applied. I'm pretty sure `emotion` doesn't even look at the `css` prop.

No error messages, as nothing is technically wrong. The styles are just not applied.

Reproduction:

https://codesandbox.io/s/522r51qrr4

Problem description:

Using the `css` prop on web components, as above, doesn't seem to apply the styles, whereas the same prop on, for example, `div`s and `span`s, does apply the styles.

I was thinking maybe this is a similar situation to how web components need `class` whereas `div`s need `className`, but I don't think so, as the reason for `className` being needed was because "class" is a restricted word. I'm still not entirely clear on why `class` is used on web components, or how `react` gets around the restricted syntax issue. Regardless, "css" is not a restricted word in JavaScript, at least not yet.

I think I can use `style`, but I'm not sure what sorts of transformations are done specifically for the `css` attribute.

Suggested solution:

Treat the `css` prop the same way on web components as on "regular" elements like `div`, `span`, `svg`, etc.

Andarist:
Fixes #1071

**Checklist**:
- [ ] Documentation
- [ ] Tests
- [x] Code complete


mitchellhamilton:
I don't think this is worth doing since the demand for it is so small and it only solves one case.

I think we should wait for React Fire to do anything like this.

overbalance:
Can this addition please be reconsidered? Custom elements make source code much easier to read and to test against.

PR #1098, opened 2018-12-11
Add CSSObject type

### LINKED ISSUE #1067: CSSObject equivalent in Emotion v10?
- `emotion` version: 10.0.1
- `react` version: 16.7.alpha-2
- `typescript` version: 3.2.1

In Emotion v9 there was a useful TypeScript interface `CSSObject` that served as the type of object styles. I can't find a working equivalent in Emotion v10. `ObjectInterpolation` and `CSSProperties` seemed to be the most obvious candidates, but neither fit.

We have lots of code in our codebase that's effectively like this:

```typescript

interface IBasicComponentProps {
  className?: string;
  name: string;
}

const BasicComponent = ({ className, name }: IBasicComponentProps) => (
  <div className={className}>Hello {name}</div>
);

const commonStyling: CSSObject = {
  backgroundColor: 'pink',
  ':hover': {
    backgroundColor: 'cyan'
  }
};

const StyledComponent = styled(BasicComponent)(commonStyling);
```

The question is: what should I replace `CSSObject` with in Emotion v10?

What you did:

I've tried using
- Emotion's `CSSProperties` interface; that's a good start but doesn't allow for the `':hover'` case
- `ObjectInterpolation<any>`

What happened:

Compilation failures with a variety of unhelpful errors.
Btw, I tried reproducing this on codesandbox.io, but it doesn't show the compilation errors reliably.

### LINKED ISSUE #1088: Emotion 10 intrinsics typescript types do not have event handler props
- `emotion` version: 10.0.4
- `react` version: 16.7.0-alpha.2

**Relevant code:**

```javascript
const Form = styled.form({})

const Foo = () => <Form onSubmit={e => { console.log(e) }} />
```

**What you did:**

Upgraded to Emotion 10

**What happened:**

Intrinsics created by `styled` do not appear to have event handler props like `onSubmit`/`onChange` so typescript complains about the implicit any of `e`.

```
error TS7006: Parameter 'event' implicitly has an 'any' type.
```

**Reproduction:**

Codesandbox doesn't import emotion types, so it's not helpful to repro. If there's a way to make that work, or if there's a starter project w/ Typescript + emotion 10 I'd gladly whip up a repro.

aaronjensen:
**What**:

Fixes #1088.
Fixes #1067.

See those for additional context.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -19,6 +19,7 @@ import {
 export {
   ArrayInterpolation,
   ComponentSelector,
+  CSSObject,
   FunctionInterpolation,
   ObjectInterpolation
 } from '@emotion/css'
--- a/packages/css/types/index.d.ts
+++ b/packages/css/types/index.d.ts
@@ -5,6 +5,7 @@ import { Interpolation, SerializedStyles } from '@emotion/serialize'
 export {
   ArrayInterpolation,
   ComponentSelector,
+  CSSObject,
   FunctionInterpolation,
   ObjectInterpolation
 } from '@emotion/serialize'
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -22,6 +22,31 @@ export interface CSSOthersObject<MP> {
   [propertiesName: string]: Interpolation<MP>
 }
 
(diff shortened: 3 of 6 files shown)


aaronjensen:
Done, and from the css package. Also added a basic type test to serialize.

PR #1106, opened 2018-12-13
Update Type Definition 👉 Support css prop

inlet:
**What**:

Update the type def for `@emotion/core` to support the `css` prop as html elements.

**Why**:

The previous approach doesn't work anymore, use `HTMLAttributes` instead.

**How**:

Use `HTMLAttributes` instead of `DOMAttributes`.

**Checklist**:

- [ ] Documentation **N/A**
- [x] Tests
- [x] Code complete

Before:

image

After:

image

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -82,7 +82,7 @@ export function ClassNames<Theme = any>(
 ): ReactElement<any>
 
 declare module 'react' {
-  interface DOMAttributes<T> {
+  interface HTMLAttributes<T> {
     css?: InterpolationWithTheme<any>
   }
 }


packages/core/types/index.d.ts
```diff
@@ -82,7 +82,7 @@ export function ClassNames<Theme = any>(
 ): ReactElement<any>
 
 declare module 'react' {
-  interface DOMAttributes<T> {
+  interface HTMLAttributes<T> {
```
Ailrun: Even with previous code, tests work well. Do you know why?
`HTMLAttributes` extends `DOMAttributes`, so `DOMAttributes` should work.

inlet: demo

Ailrun: Does DOMAttr work with cli TS compiler? If so, it's something related with PHPStorm.
If not, could you tell me which TS version you use?

inlet: Thanks for your prompt reply. The TS Compiler succeeds with `DOMAttributes`, it seems this is related to PHPStorm. I'll close this PR

PR #1108, opened 2018-12-14
Link to @emotion/babel-preset-css-prop from the css prop docs.

tkh44:
Additionally, add table styles for the docs.

**Checklist**:

- [X] Documentation
- [n/a] Tests
- [n/a] Code complete

### DIFF
--- a/docs/css-prop.md
+++ b/docs/css-prop.md
@@ -8,7 +8,13 @@ The primary way to style elements with emotion is the `css` prop. It provides a 
 
 There are 2 ways to get started with the `css` prop.
 
-#### Add `@emotion/babel-preset-css-prop` to babel configuration to enable for entire project
+- [Babel Preset](#babel-preset)
+- [JSX Pragma](#jsx-pragma)
+
+#### Babel Preset
+
+This method will **not** work with [Create React App](https://github.com/facebook/create-react-app) or other projects that do not allow custom babel configurations.
+Use the [JSX Pragma](#jsx-pragma) method instead.
 
 **.babelrc**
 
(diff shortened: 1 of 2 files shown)


docs/css-prop.md
```diff
@@ -6,8 +6,24 @@ The primary way to style elements with emotion is the `css` prop. It provides a
 
 ## Get Started
 
+There are 2 ways to get started with the `css` prop.
+
+#### Add `@emotion/babel-preset-css-prop` to babel configuration to enable for entire project
+
```
mitchellhamilton: Maybe this should say it doesn't work in CRA because some people might not know that CRA doesn't allow babel configuration and might not read the bit below

tkh44: Good call.

docs/css-prop.md
```diff
@@ -6,7 +6,30 @@ The primary way to style elements with emotion is the `css` prop. It provides a
...
+{
+  "presets": ["@emotion/babel-preset-css-prop"]
+}
+```
+
+> [Full `@emotion/babel-preset-css-prop` documentation](https://emotion.sh/docs/@emotion/babel-preset-css-prop)
+
+#### JSX Pragma
+
+Set the jsx pragma at the top of your source file that uses the `css` prop.
```
mitchellhamilton: Maybe this should also say you also have to import it since you don't have to in the babel preset?

tkh44: It says this in the next line. Should I move it?

mitchellhamilton: nah, didn't see it in it's whole, lgtm.

docs/css-prop.md
```diff
@@ -6,7 +6,38 @@ The primary way to style elements with emotion is the `css` prop. It provides a
 
 ## Get Started
 
-#### Set the jsx pragma at the top of your source file.
+There are 2 ways to get started with the `css` prop.
+
+- [Babel Preset](#babel-preset)
+- [JSX Pragma](#jsx-pragma)
+
+Both methods result in the same compiled code. Choose the one that works for you situation.
```
mitchellhamilton: > Choose the one that works for you situation.

This feels kind of pointless, maybe say in a really short way use the babel preset in x cases and use the jsx pragma in y cases(I know it's a lot of repeating but people will skim things a lot)

Issue #1109, opened 2018-12-14
toHaveStyleRule use wrong property for comparison


Werter12:
When you run toHaveStyleRule on the element with css rule which has hover, focus or nested html element css rules. toHaveStyleRule will compare provided property and value with the last rule in css.

- `emotion` version: "10.0.4"
- `jest-emotion` version: "10.0.4"
- `react` version: "16.5.2"

**Relevant code:**

```javascript
    const divStyle = emotion.css`
      color: white;
        &:hover {
          color: yellow;
        }
         &:focus {
          color: black;
        }
    `
    const svgStyle = emotion.css`
       width: 100%;
     `

    const tree = renderer
      .create(
        <div className={divStyle}>
          <svg className={svgStyle} />
        </div>
      )
      .toJSON()

    expect(tree).toHaveStyleRule('color', 'white')
    expect(tree).not.toHaveStyleRule('color', 'yellow')
    expect(tree).not.toHaveStyleRule('color', 'black')

```

**What you did:**
Run toHaveStyleRule on the element with css which has hover, focus or nested html element rule in my project.
Run test with mentioned values in the /jest-emotion/test/matchers.test.js.

**What happened:**
    Error: Expected color to match:
      white
    Received:
     black

**Reproduction:**
Run test with mentioned values in the /jest-emotion/test/matchers.test.js.

**Problem description:**
toHaveStyleRule gives wrong result when you try to check whether the element has css property with specific value. The hover, focus or other css statement will overwrite compassion on the root element.

**Suggested solution:**
Replace this logic  https://github.com/Werter12/emotion/blob/master/packages/jest-emotion/src/matchers.js#L44 . I don't think its a good practice to use Object.assign in this way. Actually its working wrong. Every time its overwrites the previous rule and in the end we have the last rule established in the css. Considering mentioned above example we will receive for comparasion the focus color black rule.
Fixed in https://github.com/emotion-js/emotion/pull/1110

PR #1110, opened 2018-12-14
Extend functionality of the toHaveStyleRule matcher

Werter12:
**What** and **Why**:
Fixed bug https://github.com/emotion-js/emotion/issues/1109 . Added additional param `options` for toHaveStyleRule matcher with properties `target` and `media` which will help developers to test their components with better accessibility (the style rule will be more accessible for testing). `target` - will help to access style rule inside hover, focus or to nested html element. `media` - will help to access css rule in the media queries.

**Checklist**:

- [X] Code complete
- [X] Tests
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Look here 👉 _[emotion babel plugin feature table and documentation](https://g
 - [healthline.com](https://www.healthline.com)
 - [nytimes.com](https://www.nytimes.com)
 - [vault.crucible.gg](http://vault.crucible.gg/)
-- [saldotuc.com](https://saldotuc.com)
+- [render.com](https://render.com)
 - [gatsbythemes.com](https://gatsbythemes.com/)
 - [blazity.com](https://blazity.com/)
 - [postmates.com](https://postmates.com/)
--- a/packages/babel-plugin-emotion/__tests__/__snapshots__/css.js.snap
+++ b/packages/babel-plugin-emotion/__tests__/__snapshots__/css.js.snap
@@ -390,3 +390,37 @@ process.env.NODE_ENV === \\"production\\" ? {
   map: \\"/*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInJlbW92ZS1saW5lLWNvbW1lbnRzLmpzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUVHIiwiZmlsZSI6InJlbW92ZS1saW5lLWNvbW1lbnRzLmpzIiwic291cmNlc0NvbnRlbnQiOlsiaW1wb3J0IGNzcyBmcm9tICdAZW1vdGlvbi9jc3MnXG5cbmNzc2BcbiAgLy8gY29sb3I6IGdyZWVuO1xuICBjb2xvcjogaG90cGluaztcbmBcbiJdfQ== */\\"
 };"
 `;
+
+exports[`@emotion/babel-plugin-core css tagged-template-args-forwarded 1`] = `
+"import css from '@emotion/css'
+
+function media(...args) {
+  return css\`
+    @media (min-width: 100px) {
+      \${css(...args)};
+    }
+  \`
+}
+
+const test = css\`
+  \${media\`color: red;\`};
+\`
+
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+import _css from \\"@emotion/css\\";
+
+function media(...args) {
+  return (
+    /*#__PURE__*/
+    _css(\\"@media (min-width:100px){\\",
+    /*#__PURE__*/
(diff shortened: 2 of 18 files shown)


packages/jest-emotion/test/matchers.test.js
```diff
@@ -120,4 +120,132 @@ describe('toHaveStyleRule', () => {
...
+      .toJSON()
+
+    expect(tree).toHaveStyleRule('color', 'yellow', { target: 'hover' })
+    expect(tree).toHaveStyleRule('color', 'black', { target: 'focus' })
+    expect(tree).toHaveStyleRule('color', 'white')
+  })
+
+  it('matches styles on the nested component or html element', () => {
+    const Svg = styled('svg')`
+      width: 100%;
```
tkh44: If I were to set a color here, `fill: blue;`, which is overridden later to `fill: green;` and the test checks for 'blue' will the tests still pass even though the end visual result is a `green` fill?

packages/jest-emotion/src/matchers.js
```diff
@@ -35,13 +49,35 @@ function valueMatches(declaration, value) {
...
+  const { target = '', media = '' } = options
   const classNames = getClassNamesFromNodes([received])
   const cssString = getStylesFromClassNames(classNames, getStyleElements())
   const styles = css.parse(cssString)
 
-  const declaration = styles.stylesheet.rules
-    .reduce((decs, rule) => Object.assign([], decs, rule.declarations), [])
+  let preparedRules = styles.stylesheet.rules
+  if (media.length > 1) {
+    if (shortMediaRegExp.test(media) || longMediaRegExp.test(media)) {
```
tkh44: Will `longMediaRegExp` break when more than one `and` clause is preset?

`@media screen and (min-width: 30em) and (orientation: landscape)`

packages/jest-emotion/src/matchers.js
```diff
@@ -35,13 +49,35 @@ function valueMatches(declaration, value) {
...
 }
 
-function toHaveStyleRule(received: *, property: *, value: *) {
+function toHaveStyleRule(
+  received: *,
+  property: *,
+  value: *,
+  options?: { target?: string, media?: string } = {}
+) {
+  const { target = '', media = '' } = options
```
mitchellhamilton: There's no need for the default values here, leave them as undefined and handle undefined instead of an empty string where necessary.

packages/jest-emotion/src/matchers.js
Werter12: Yea, it will break I will fix this.

packages/jest-emotion/test/matchers.test.js
```diff
@@ -120,4 +120,132 @@ describe('toHaveStyleRule', () => {
...
+    const tree = renderer
+      .create(
+        <div className={localDivStyle}>
+          <svg className={svgStyle} />
+        </div>
+      )
+      .toJSON()
+
+    expect(tree).toHaveStyleRule('color', 'yellow', { target: 'hover' })
+    expect(tree).toHaveStyleRule('color', 'black', { target: 'focus' })
```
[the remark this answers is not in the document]
Werter12: I agree I will replace it on ':hover' and ':focus'

packages/jest-emotion/test/matchers.test.js
Werter12: The test check for `fill: blue` will fail.

Werter12: The test check for `fill: blue`. The `fill` would be `green`.

packages/jest-emotion/src/matchers.js
Werter12: Just checked this scenario and it is passed. But I think we can get rid of `longMediaRegExp` because it is don't extend `shortMediaRegExp` properly and a invalid `media` can pass this check by adding `() and ()` pattern

packages/jest-emotion/src/matchers.js
```diff
@@ -23,6 +26,8 @@ function isAsymmetric(obj) {
   return obj && isA('Function', obj.asymmetricMatch)
 }
 
+const mediaRegExp = /\([a-z-]+:\s[a-z0-9.]+\)/
```
mitchellhamilton: Would this work with `@media only screen` or something else like that?

packages/jest-emotion/test/matchers.test.js
mitchellhamilton: Could you add a test for that?

Werter12: Yes, I've added it. https://github.com/emotion-js/emotion/pull/1110/files#diff-f37132e84fc7f1a6f14f4ae5bd1d96cdR150

packages/jest-emotion/src/matchers.js
Werter12: @mitchellhamilton I've change regexp to make media with screen property pass

packages/jest-emotion/src/utils.js
```diff
@@ -154,3 +163,27 @@ export function getKeys(elements: Array<HTMLStyleElement>) {
...
+export function hasClassNames(
+  classNames: Array<string>,
+  selectors: Array<string>,
+  target: string
+) {
+  return selectors.some(selector => {
+    if (target === '') {
+      return classNames.includes(selector.slice(1))
+    }
+    return selector.includes(target)
```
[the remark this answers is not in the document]
Werter12: I've added comments in the code what is going on in this function. Should I add comments here in discussion?

packages/jest-emotion/src/matchers.js
mitchellhamilton: I didn't just mean screen, it was an example, I meant all the media types(https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Media_types) and I'd prefer that the media types aren't hardcoded.

packages/jest-emotion/test/matchers.test.js
mitchellhamilton: I don't think that tests it, that looks like it's declaring the fill property once.

Werter12: @mitchellhamilton lets have a chat in slack, I have sent you a message there

mitchellhamilton:
Also could you add a test like this

```jsx
let tree = renderer
  .create(
    <div
      className={css`
        color: green;
        color: hotpink;
      `}
    />
  )
  .toJSON();
expect(tree).not.toHaveStyleRule("color", "green");
expect(tree).toHaveStyleRule("color", "hotpink");
```

packages/jest-emotion/src/matchers.js
```diff
@@ -35,13 +39,35 @@ function valueMatches(declaration, value) {
...
+  const { target, media } = options
   const classNames = getClassNamesFromNodes([received])
   const cssString = getStylesFromClassNames(classNames, getStyleElements())
   const styles = css.parse(cssString)
 
-  const declaration = styles.stylesheet.rules
-    .reduce((decs, rule) => Object.assign([], decs, rule.declarations), [])
+  let preparedRules = styles.stylesheet.rules
+  if (media) {
+    if (mediaRegExp.test(media)) {
```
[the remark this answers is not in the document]
Werter12: Okey

PR #1121, opened 2018-12-17
Fix out of bounds access in @emotion/serialize

### LINKED ISSUE #1119: 'undefined' as a css property
- `emotion` version: 10.0.5
- `react` version: 16.5.2

**Relevant code:** 
```javascript
function tablet(...args) {
  return css`
    @media (min-width: 100px) {
      ${css(...args)};
    }
  `;
}

const H2 = styled('h2')`
  ${tablet`color: red;`}
`;
```

**What you did:** Upgraded from Emotion 9 to Emotion 10

**What happened:** I get `undefined`as a (vendor prefixed!) css property in every one of my media queries.

screenshot 2018-12-17 at 10 35 49

**Reproduction:** https://codesandbox.io/s/llm7j6rnp7

**Problem description:** I've upgraded from Emotion 9 to Emotion 10, and it's mostly been a smooth ride. Since I am doing SSR I especially appreciate how much simpler it is now. The only thing that I haven't been able to work out is some quirkiness with the way I do media queries (which works fine in version 9). In version 10 I get the `undefined` CSS property. I am not sure if this is due to an API change or if it is actually a bug.

**Suggested solution:** N/A

Andarist:
This is an attempt to fix #1119 

I havent yet time to test if this even works, just sending PR to trigger CI as I dont have time right now to check it on my machine. If this won't break anything I'll add a test case for the (hopefully) fixed issue later.

There was similar guard in v9 - https://github.com/emotion-js/emotion/blob/7c582301317cd9c0038a13a118456e798ffcadf8/packages/create-emotion/src/index.js#L270 . I've chosen the length check as supposedly out of bounds access should be avoided for perf reasons.

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -314,7 +314,7 @@ export const serializeStyles = function(
       args[i],
       styles.charCodeAt(styles.length - 1) === 46
     )
-    if (stringMode) {
+    if (stringMode && strings.length > i) {
       styles += strings[i]
     }
   }


Andarist:
The fix is wrong (it would actually fix the "issue", but that's not the point), not sure if the issue needs to be fixed (if not we could at least warn in dev when somebody uses this pattern wrong). Need to investigate further.

PR #1128, opened 2018-12-20
Get options from options parameter :)

emattias:
**What**:
Makes options provided to `babel-preset-css-prop` be used.

**How**:
Takes the options from the second function argument where they are sent to the function :)


packages/babel-preset-css-prop/src/index.js
```diff
@@ -8,14 +8,10 @@ let pragmaName = '___EmotionJSX'
...
-  pragma,
-  sourceMap,
-  autoLabel,
-  labelFormat,
-  instances,
-  ...options
-}) => {
+export default (
+  api,
+  { pragma, sourceMap, autoLabel, labelFormat, instances, ...options }
```
Andarist: is this guaranteed to always be an object?

emattias: `babel-preset-react` seems to assume that: https://github.com/babel/babel/blob/master/packages/babel-preset-react/src/index.js#L10 :)

emattias: Or actually they use this `declare` helper. Can we use that here also? 🤔 https://github.com/babel/babel/blob/master/packages/babel-helper-plugin-utils/src/index.js#L1

emattias: Or just default the argument to `= {}`?

mitchellhamilton:
Could you add a test?

mitchellhamilton:
Something like this:

```jsx
import { transform } from "@babel/core";
import preset from "@emotion/babel-preset-css-prop";
test("works with options", () => {
  transform(
    `
...some code
`,
    {
      presets: [[preset, { ...someOptions }]],
      babelrc: false,
      configFile: false
    }
  );
});
```

Issue #1025, opened 2018-11-19
Typing of white-space


e11en:
- `emotion` version: 9.2.4
- `react` version: 16.4.0
- `react-emotion` version: 9.2.4

When using a mixin to fill my styled component the variable needs to be `whitespace` (lower case S) while in the normal use of white-space I need to use `whiteSpace` (upper case S). I was banging my head for a while trying to fix the error it produced. I think the typing somewhere needs to be fixed.

```javascript
export const nowrap = {
  whitespace: 'nowrap'
};

// With use of mixin
const Container = styled('div')(({theme}: {theme: Theme}) => ({
     ...nowrap
}));

// Normal use without mixin
const Container = styled('div')(({theme}: {theme: Theme}) => ({
     whiteSpace: 'nowrap'
}));

export const StyledDiv = compose<any, any>(
  withTheme()
)(Container);
````

e11en:
Typescript

Ailrun:
Then it's because you didn't put exact type.
`whiteSpace` accepts some specific strings only, where TS infer `{ whiteSpace string }` for your use case. Giving appropriate type will fix this.

Ailrun:
@e11en `whitespace` works since typing should accept **any** key since there is something like `& > &`, `& > span`, ... and so on.
If my previous comment is not understandable to you, just check following example TS code.
[...]

PR #1129, opened 2018-12-20
fix(TS): allow CSSObject on @emotion/styled

kentcdodds:
**What**: This allows: `styled.div(/* this can be an object /*)`

**Why**: Because that API is supported so the typings should allow for it.

**How**: Tried it locally and it typed fine (even helped me fix an issue I had), so I just made this change in the github editor.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [x] Code complete

### DIFF
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -12,19 +12,18 @@
  * a style of that component.
  */
 
-import { ComponentSelector, Interpolation } from '@emotion/serialize'
+import { ComponentSelector, Interpolation, CSSObject } from '@emotion/serialize'
 import * as React from 'react'
 
 import { Omit, Overwrapped, PropsOf } from './helper'
 
 export {
   ArrayInterpolation,
-  CSSObject,
   FunctionInterpolation,
   ObjectInterpolation
 } from '@emotion/serialize'
 
-export { ComponentSelector, Interpolation }
+export { ComponentSelector, Interpolation, CSSObject }
 
 type JSXInEl = JSX.IntrinsicElements
 
@@ -72,7 +71,7 @@ export interface CreateStyledComponentBase<
       ReactClassPropKeys
     > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>
   >(
-    template: TemplateStringsArray,
+    template: TemplateStringsArray | CSSObject,
     ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
   ): StyledComponent<InnerProps, StyleProps, Theme>
 }
--- a/packages/styled-base/types/tests.tsx
+++ b/packages/styled-base/types/tests.tsx
@@ -39,7 +39,7 @@ const Button0 = styled('button')`
 const Button1 = styled('button')({
   color: 'blue'
 })
-const Button2 = styled.button({
+const Button1Dot = styled.button({
   color: 'blue'
 })
 ;<div>
@@ -51,8 +51,8 @@ const Button2 = styled.button({
   <Button1 type="button" />
 </div>
 ;<div>
-  <Button2 />
-  <Button2 type="button" />
+  <Button1Dot />
+  <Button1Dot type="button" />
 </div>
 
 const Input0 = styled('input', {
--- a/packages/styled-base/types/tests.tsx
+++ b/packages/styled-base/types/tests.tsx
@@ -39,9 +39,6 @@ const Button0 = styled('button')`
 const Button1 = styled('button')({
   color: 'blue'
 })
-const Button1Dot = styled.button({
-  color: 'blue'
-})
 ;<div>
   <Button0 />
   <Button0 type="button" />
@@ -49,10 +46,6 @@ const Button1Dot = styled.button({
 ;<div>
   <Button1 />
   <Button1 type="button" />
-</div>
-;<div>
-  <Button1Dot />
-  <Button1Dot type="button" />
 </div>
 
 const Input0 = styled('input', {
--- a/packages/styled-base/types/tests.tsx
+++ b/packages/styled-base/types/tests.tsx
@@ -48,6 +48,11 @@ const Button1 = styled('button')({
   <Button1 type="button" />
 </div>
 
+// $ExpectError
+const BadButton0 = styled('button')({ bad: 'value' })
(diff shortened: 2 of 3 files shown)


kentcdodds:
Hmmm... Just noticed this test:

https://github.com/emotion-js/emotion/blob/6f3ee3ff29ceca0b76ea1e0c6650944ed9219760/packages/styled-base/types/tests.tsx#L39-L41

I'm not sure how that's working because using the object syntax isn't working for me. I'm using `styled.button` rather than `styled('button')`, so maybe that's why...

kentcdodds:
The failure is:

> Property 'button' does not exist on type 'CreateStyled<any>'.

I'm not sure how to fix that 😬

mitchellhamilton:
styled.button and stuff should be in the @emotion/styled package

kentcdodds:
Ok, so I fixed the test, but then I added a few more tests for things that I _think_ should be errors but they're not and I'm not sure why...

Ailrun:
@kentcdodds We cannot emit an error on "wrong properties", since we should handle something like `& > &`, `& > li`, ..., and TS does not support RegExp property type (which they said they will support).

packages/styled-base/types/index.d.ts
```diff
@@ -72,7 +71,7 @@ export interface CreateStyledComponentBase<
       ReactClassPropKeys
     > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>
   >(
-    template: TemplateStringsArray,
+    template: TemplateStringsArray | CSSObject,
```
Ailrun: This case is covered by previous overloading, so you don't need to add one.
`Interpolation<WithTheme<StyleProps, Theme>>` is a (structural) supertype of `CSSObject`.

kentcdodds:
Ah, I think [this](https://github.com/kentcdodds/ts-emotion-border-top-style-issue) is the issue I was having that lead me to open this PR. Would you like me to make a new issue?

---

This is a weird issue and I don't know what's going on. Help appreciated!

Here's the code:

```typescript

const WorkingDivider = styled.hr({borderTopStyle: 'solid'})

const baseStyles = {borderTopStyle: 'solid'}
const NotWorkingDivider = styled.hr(baseStyles)

const otherBaseStyles = {margin: 20}
const AnotherWorkingDivider = styled.hr(otherBaseStyles)

export {WorkingDivider, NotWorkingDivider, AnotherWorkingDivider}
```


Here's the error I'm getting:

```
$ yarn typecheck
yarn run v1.12.3
$ tsc
index.ts:6:37 - error TS2345: Argument of type '{ borderTopStyle: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Type '{ borderTopStyle: string; }' is missing the following properties from type 'TemplateStringsArray': raw, length, concat, join, and 10 more.

6 const NotWorkingDivider = styled.hr(baseStyles)
                                      ~~~~~~~~~~


Found 1 error.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

So from my observations, when I specify `borderTopStyle` in an object and pass it to styled, it'll fail typechecking with the above error. But if I inline the object it works. I have not observed this behavior with any other css property.

Ailrun:
I think the issue you met is #1025.
It's about how TS works...

lychyi:
@kentcdodds @Ailrun is correct, this is due to the inner-workings of TS (and exacerbated by an error messaging issue). Although, to be fair, I think  #1025 makes is hard for some TypeScript users to decipher.

Let me try to explain this phenomenon because when I tripped up on this last time it was kind of hard to find a solution online. Hopefully this helps anyone else out:

### Explanation

The difference in the working/non-working examples above is that `WorkingDivider` is an object literal passed as a function argument, thus **its type is inferred correctly**:
```tsx
// parameter type is effectively { borderTopStyle: 'solid' }
// important here: value is a string literal type
const WorkingDivider = styled.hr({borderTopStyle: 'solid'})
```
vs.
```tsx
// Notice the difference!
// type is { borderTopStyle: string }
// important here: value is a regular string type
const baseStyles = {borderTopStyle: 'solid'}
const NotWorkingDivider = styled.hr(baseStyles)
```
while in this case, we have an object literal assigned to a variable and then passed as a function argument. `const baseStyles` infers its type **before it is passed into the styled function** and the type is different because by default **TS does type-widening on object literals**. And because `borderTopStyle` only takes a specific values of strings (a string union type), it won't just accept `string` as the type. This makes sense, `'frizzy'` or `'asdf'` is not a valid `borderTypeStyle` so you want that type to be strict.

Finally, the last part of your example works because `margin` is designed to take *any* number or string (say, `'20px'`), so type-widening doesn't really affect it. 

```tsx
// type here is { margin: number } which is totally fine
const otherBaseStyles = {margin: 20}
const AnotherWorkingDivider = styled.hr(otherBaseStyles)
```

### Solutions
This is my favorite way to fix it. Do the type inference via annotation:
```tsx
import styled, { Interpolation } from '@emotion/styled';

// You could annotate the type to prevent widening
const baseStyles: Interpolation = { borderTopStyle: 'solid' }
```
I've seen this next solution get thrown around but it's difficult to maintain. i.e. What if you need to change from `'solid'` to `'dotted'`?
```tsx
// Prevent type-widening with an assertion
// ...but if you had multiple properties that needed this, it could get tedious
const baseStyles = { borderTopStyle: 'solid' as 'solid' }
```

### Other reading
* Feature request to disable type-widening on TS https://github.com/Microsoft/TypeScript/issues/20195

Andarist:
@lychyi thanks for the explanation! I've learned something today because of your comment ❤️

Ailrun:
@lychyi Thank you for such detailed explanation!
As one additional side note, it's not just about type widening, but also about how TS use context to infer types (of variables, values, ...). Namely, it (almost) ignore it, and it's a critical difference with Flowtype.

Ailrun:
@lychyi Oh, and in your solution, `Interpolation` can make a problem when you use `props` in your styles.
`CSSObject` is originally meant to be used in that exact case.

kentcdodds:
This has been super helpful. Thank you friends!

lychyi:
@Ailrun Good to know! Thanks for the heads up.

Issue #997, opened 2018-11-06
Unable to compose themed css in emotion10


piotrblasiak:
- `emotion` version: 10.0.0-beta.8
- `react` version: 16.6.0

Relevant code.

```javascript
const css1 = theme => ({background: "black"});
const css2 = {color: "white"};

return <div css={[css1, css2]}/>
```

What you did:

Tried to compose with one of the parts being a function/themed css.

What happened:

"Functions that are interpolated in css calls will be stringified..."

Is this by design or a bug? The workaround I found is to use the spread operator and call on the css function:

````
<div css={theme => ({...css1(theme), ...css2})/>
```

But this forces me to know if css1 is a function or not.

piotrblasiak:
It doesn´t say in the docs that it needs to be a string - and IMO it would make a lot of sense to be able to put a function or an object in there as well.

defusioner:
Would be nice to have this feature. Actually the workaround is to make `css={theme => [css.class1(theme), css.class2]}`

gordonmleigh:
@defusioner Composition order doesn't seem to work with this approach.

Andarist:
@gordonmleigh what do you mean? Could you share repro case and actual/expected outcomes?

gordonmleigh:
Turns out my problem was elsewhere, I was getting confused.

Andarist:
This will be possible in upcoming v11, we’ve just merged in https://github.com/emotion-js/emotion/pull/1130 which implement this

PR #1130, opened 2018-12-22
Added support for passing theme to functions in array css prop

jtmthf:
**What**: When passing an array to the css-prop, functions accepting the theme from `emotion-theming` are now acceptable as array items.

**Why**: This addresses #997. I find that there are many use cases where this is helpful. One example would be when composing a number of style utility functions that may or may not read from the theme. For my use case in particular, I'm working on a reimplementation of [Tailwind CSS](https://tailwindcss.com) using emotion. This is what one example of usage looks like right now.

```jsx
<div
  css={theme => [
    display('flex')
    bg('white')(theme),
    mx("auto")(theme),
    maxW('sm')(theme),
    shadow('lg')(theme),
    rounded('lg')(theme),
    { overflow: "hidden" }
  ]}
/>
```

With this change, the example can be rewritten as:

```jsx
<div
  css={[
    display('flex')
    bg('white'),
    mx("auto"),
    maxW('sm'),
    shadow('lg')
    rounded('lg'),
    { overflow: "hidden" }
  ]}
/>
```

In addition to being shorter, the advantage here is that the developer doesn't need to know if a value in the css-prop is a function that needs to be passed the theme.

**How**: The change was implemented by now wrapping all uses of the array css prop in `ThemeContext.Consumer`. It could potentially be more optimized to check if the array contains a function first before wrapping in the consumer. In the render method, if the css prop is an array, it is now mapped and for each item that is an array it is passed the theme.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] TypeScript typings


luishdz1010:
This would be truly helpful, right now object styles break encapsulation since you need to know whether they are functions or objects

jtmthf:
@simonflk I actually have the feature you described implemented locally, but not yet committed to this branch. I still need to add documentation and TypeScript typings, but am holding off until I get confirmation that this is something the project owner's want.

chibicode:
👍 This is exactly what I need! Would love to see this merged.

a-type:
I'd love to see this as well. It's critical to a project I'm working on where I want to be able to compose in user-provided styles and allow the ability to utilize the theme.

jckw:
Would be amazing to get this merged! Any timeline?

Andarist:
I got this working, made additional changes to babel plugin and Global's styles prop (so its behavior match css prop). I think it's done and would consider it a good change overall - still need @mitchellhamilton approval though.

Issue #1135, opened 2018-12-27
[@emotion/core] cache `createElement`, and make `jsx` to use as polyfill


morlay:
- `emotion` version: 10.x
- `react` version: 16.7

**Relevant code:**

```javascript
import React from "react"
import { jsx } from "@emotion/core"

React.createElement = jsx
```

**What you did:**

If we use emotion for full project, we could hack createElement when initializing application, use code above. 

**What happened:**
But now, `React.createElement` is used directly in the `jsx` function, it will cause a loop call issue.

**Suggested solution:**

we could change a little here

https://github.com/emotion-js/emotion/blob/master/packages/core/src/jsx.js#L107

```javascript
const createElement = React.createElement
export const jsx: typeof React.createElement = function(
  type: React.ElementType,
  props: Object
) {
  let args = arguments

  if (props == null || props.css == null) {
    // $FlowFixMe
    return createElement.apply(undefined, args)
  }
  // ...
}
```

tkh44:
While this is technically working, setting the `jsx` pragma is effectively doing the same thing and in a much safer way imo.

morlay:
@tkh44 
Yes. using the `jsx` is safe but cause other costs: 

* need to import `@emotion/core` in each file. 
* additional configuration in `babel` or `typescript`
  * For TypeScript user, when `jsxFactory` changed, `Fragment` could not be used by shout version  `<></>`

I just hope there is someway way to use `emotion` directly and easier 😄

tkh44:
> I just hope there is someway way to use emotion directly and easier 😄

You don't have to use the css prop.

```javascript
import { css } from 'emotion'

const blueClassName = css({ color: 'blue' })

const Button = (props) => <button className={} />
```

PR #1136, opened 2018-12-27
#1135 cache `createElement`, and make `jsx` to use as polyfill

morlay:


mitchellhamilton:
I'm not a fan of doing this since it could cause unexpected side effects if a package that you're using internally uses a prop named css and I believe overwriting createElement might not work with `import * as React from 'react'`(I'm not totally sure on that though) and our intended solution for applying the css prop in a whole app is https://www.npmjs.com/package/@emotion/babel-preset-css-prop

morlay:
@mitchellhamilton.
it works well for `import * as React from 'react'`, because now `react` exports `React` as namespace which is just an object.

i tried `@emotion/babel-preset-css-prop`, but as a typescript user, i have to use babel too :-(

morlay:
@Andarist 
this PR not change the way of emotion, just make it could be hackable (we will hack `createElement` when we need). 
event in the future, react start to use ESM, these changes still working

Andarist:
The changes from this PR will work, but what you have proposed in your issue:
```js
import React from "react"
import { jsx } from "@emotion/core"

React.createElement = jsx
``` 
won't work in the future (because of namespaces being frozen)

morlay:
@Andarist 
I understand this point. I will remove it in future too, but now

in my case, I'm migrating my projects from `glamor` to `emotion` by replacing `{...css(` to `css={` directly.
I just wan't to simple way to make emotion work in my projects.

Maybe someone have similar issue. just hope this PR could be merged for those developer like me.

Btw I already locate why my projects is broken when using the `babel-preset-css-prop`.
https://github.com/emotion-js/emotion/issues/1147

So, it's fine to close this PR.

Thanks ~

Issue #1139, opened 2018-12-28
Dynamic Global styles with more than one rule crashing on updated


Qrysto:
<!--
- `emotion` version: 10.0.5
- `react` version: 16.7.0

**Relevant code:**

```javascript
// Example usage
<Global
  styles={{
    body: { background: this.state.background },
    html: { color: this.state.color },
  }}
/>
```

**Reproduction:**

https://codesandbox.io/s/yw8lwvmr2v

**Problem description:**

When a Global component with dynamic styles (that may change depending on some variables) having more than one css rule gets updated, it crashes with the error message:

```
Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node. (sheet.browser.esm.js:84)
```

**Suggested solution:**

It seems like [this line](https://github.com/emotion-js/emotion/blob/master/packages/core/src/global.js#L101) caused the problem.

 ```
let element = this.sheet.tags[0].nextElementSibling
```

Shouldn't this be `this.sheet.tags[this.sheet.tags.length - 1].nextElementSibling` instead?

**Comment by Qrysto** [comment]
- `emotion` version: 10.0.5
- `react` version: 16.7.0

**Relevant code:**

```javascript
<Global
  styles={{
    body: { background: this.state.background },
    div: { color: this.state.color }
  }}
/>
```

**Reproduction:**

https://codesandbox.io/s/yw8lwvmr2v

**Problem description:**

When a Global component which
- has more than one css rule
- has dynamic styles (depending on outside variables)
is updated (outside variables change) then it crashes with this error message:

```
Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node. (sheet.browser.esm.js:84)
```

**Suggested solution:**

It seems that [this line](https://github.com/emotion-js/emotion/blob/master/packages/core/src/global.js#L101) is the cause of the issue.

```
let element = this.sheet.tags[0].nextElementSibling
```

Shouldn't it be `this.sheet.tags[this.sheet.tags.length - 1].nextElementSibling` instead?

Qrysto:
If the solution I suggested on the comment above was correct then I can make a PR. I don't have much time to test it throughly or read all the code though, so not sure if it would negatively affect other cases.

Qrysto:
Ok I have created a PR here https://github.com/emotion-js/emotion/pull/1140, please double check

PR #1140, opened 2018-12-28
Fix dynamic Global styles with more than one rule crashing when updated

Qrysto:
More details are in #1139


Andarist:
I've created a followup PR (with your commit included) with test covering this, gonna close this PR. Thank you for the report and your input!

PR #1142, opened 2018-12-28
Fix/1139

### LINKED ISSUE #1139: Dynamic Global styles with more than one rule crashing on updated

Andarist:
fixes #1139

### DIFF
--- a/packages/core/__tests__/global.js
+++ b/packages/core/__tests__/global.js
@@ -2,7 +2,8 @@
 import 'test-utils/dev-mode'
 import * as React from 'react'
 import { render, unmountComponentAtNode } from 'react-dom'
-import { Global, keyframes, css } from '@emotion/core'
+import { Global, keyframes, css, CacheProvider } from '@emotion/core'
+import createCache from '@emotion/cache'
 
 beforeEach(() => {
   // $FlowFixMe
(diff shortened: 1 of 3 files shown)


packages/core/__tests__/global.js
```diff
@@ -2,15 +2,24 @@
...
-import { Global, keyframes, css } from '@emotion/core'
+import { Global, keyframes, css, withEmotionCache } from '@emotion/core'
 
-beforeEach(() => {
-  // $FlowFixMe
-  document.head.innerHTML = ''
+const ClearCache = withEmotionCache((props, cache) => {
+  cache.sheet.flush()
+  return null
+})
```
mitchellhamilton: I'd rather create a new cache in each test and pass it down with the CacheProvider and clear everything in the head in beforeEach

PR #1144, opened 2018-12-28
Make imports make more sense in babel macros docs

jetpacmonkey:
**What**: Change `emotion/macro` to `@emotion/core` and remove redundant `css` import

**Why**: I'm pretty sure this was just a mistake, based on what's written around it

**How**: Just a docs change

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/docs/babel-macros.md
+++ b/docs/babel-macros.md
@@ -9,7 +9,7 @@ Most of Emotion's Babel Macros are available by adding `/macro` to the end of th
 ```jsx
 import styled from '@emotion/styled/macro'
 import css from '@emotion/css/macro'
-import { css, keyframes, injectGlobal } from 'emotion/macro'
+import { keyframes, injectGlobal } from '@emotion/core'
 ```
 
 > Note


docs/babel-macros.md
```diff
@@ -9,7 +9,7 @@ Most of Emotion's Babel Macros are available by adding `/macro` to the end of th
 ```jsx
 import css from '@emotion/css/macro'
-import { css, keyframes, injectGlobal } from 'emotion/macro'
```
Andarist, line 12: This was correct. We have 2 "versions" of Emotion:
- `@emotion/core` + `@emotion/styled`, those two are specific to React and don't work without it
- `emotion`, this is so-called "vanilla" Emotion which can be used in any app, it's completely framework-independent

So all 3 imports here are correct, even though they don't add up to a valid JavaScript when put into a single code snippet because some identifiers (such as `css`) are duplicated here.

PR #1152, opened 2019-01-01
Add dev warning about tagged template arguments being weirdly forwarded

Andarist:


packages/serialize/src/index.js
```diff
@@ -304,6 +309,15 @@ export const serializeStyles = function(
     stringMode = false
     styles += handleInterpolation(mergedProps, registered, strings, false)
   } else {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      strings.length !== args.length
+    ) {
+      console.warn(
+        `\`css\` got used in confusing manner - most likely you have forwarded all arguments from tagged template call to \`css\` with extra stuff appended to them.` +
+          ` This may cause parsing problems. Like in this example code snippet:\n${TAGGED_TEMPLATE_ARGS_WEIRDLY_FORWARDED_SNIPPET}`
```
mitchellhamilton: - this isn't necessarily being called from css, it could be styled or others
- confusing seems like a strange word choice
- `This may cause parsing problems` - this isn't really about parsing 
- the message doesn't say what people should do to fix it
- I don't trust tools to eliminate the code example outside of `process.env.NODE_ENV !== 'production'`

packages/serialize/src/index.js
```diff
@@ -304,6 +309,15 @@ export const serializeStyles = function(
     stringMode = false
     styles += handleInterpolation(mergedProps, registered, strings, false)
   } else {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      strings.length !== args.length
+    ) {
+      console.warn(
```
mitchellhamilton: This should use console.error so users get a stack trace.

packages/core/__tests__/css.js
```diff
@@ -188,3 +188,32 @@ test('child selector array', () => {
...
+      @media (min-width: 100px) {
+        ${css(...args, 'background-color: #fff;')};
+      }
+    `
+  }
+
+  const oldConsoleWarn = console.warn
+  // $FlowFixMe
+  console.warn = jest.fn()
+
```
mitchellhamilton: This test should use this helper. https://github.com/emotion-js/emotion/blob/master/scripts/test-utils/src/index.js#L11

Andarist:
Closing this as this is not a common mistake and I don't want to revive this PR now - it can always be done later.

PR #1166, opened 2019-01-10
support the as property in typescript

ssured:
**What**: typescript support for as

**Why**: not supported now

**How**: updated the typeing

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -39,7 +39,7 @@ export interface StyledOptions {
 }
 
 export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+  extends React.SFC<InnerProps & StyleProps & { theme?: Theme, as?: keyof JSXInEl}>,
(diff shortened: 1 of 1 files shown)


packages/styled-base/types/index.d.ts
```diff
@@ -39,7 +39,7 @@ export interface StyledOptions {
 }
 
 export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+  extends React.SFC<InnerProps & StyleProps & { theme?: Theme, as?: keyof JSXInEl}>,
```
Ailrun: A problem with this is you cannot use `as` in type-safe way.
For example,
```jsx
const X = styled('button')({
  width: '200px',
});

// Following does not emit an error with the typing.
<X as='a' value='...'>text</X>
// Following emits an error with the typing.
<X as='a' href='...'>text</X>
```

Andarist:
As far as I know `as` cannot be well-typed and I'm rather against providing unsafe typings (unless somebody figures out how to type this).

PR #1170, opened 2019-01-11
Upgrade preconstruct

mitchellhamilton:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


packages/emotion-theming/package.json
```diff
@@ -40,9 +40,11 @@
   "dependencies": {
     "@emotion/weak-memoize": "0.2.2",
     "hoist-non-react-statics": "^2.3.1",
-    "object-assign": "^4.1.1"
+    "object-assign": "^4.1.1",
+    "@babel/runtime": "^7.2.0"
```
Andarist: From what I've seen before it was only `extends` helpers that got used - which is fine, but in theory `object-assign` package should be used according to current configs, so maybe something is wrongly configured or there is a bug somewhere?

mitchellhamilton: I think it's used because it's a jsx spread rather than an object spread. I think i'm just gonna remove the object-assign stuff, while it's probably slightly better if people have react, it's probably better for the ecosystem if lots of packages use @babel/runtime.

Issue #839, opened 2018-09-06
Typescript definitions for @emotion/native


yonatanmn:
Hey, 
I'm building a RN app and wish to use @emotion/native, but since I'm using TS I'm having problems declaring correct typings.

krzysztofzuraw:
I can take look into it - is there anything before I jump into declaring types?

thefrontendwizard:
@krzysztofzuraw Are you still on to this? I'm interested in this too.

davidyeiser:
I'd be happy to help with this. (Though I've never done anything like it before.)

patsissons:
I was playing around with emotion in react native today, and ended up writing some brief but fairly accurate typings for the native package. I'm not confident i will have time to author a PR for this any time soon, so i'll post the effort on this issue in case it helps anyone else. Note that I added `@emotion/core` and `@emotion/styled-base` to `devDependencies` so i can reuse their existing type definitions.

The strategy here was to reuse as much existing typing as i could. This was easy for the `css` export since it's effectively the same for `react-native`. The `styled` export is a bit more complicated. I took the list of bound components and turned it into a giant string literal union. Then i created a new mapped type for mapping component names into styled components and joined that with the `CreateStyled` types from `styled-base`. This should cover nearly every use case for `react-native` as far as i can tell.

```tsx
// global.d.ts

declare module '@emotion/native' {
  import css from '@emotion/css';
  import {
    CreateStyled,
    CreateStyledComponentExtrinsic,
  } from '@emotion/styled-base';
  import React from 'react';
  import ReactNative from 'react-native';

  type StyledReactNativeComponents =
    | 'ActivityIndicator'
    | 'ActivityIndicatorIOS'
    | 'ART'
    | 'Button'
    | 'DatePickerIOS'
    | 'DrawerLayoutAndroid'
    | 'Image'
    | 'ImageBackground'
    | 'ImageEditor'
    | 'ImageStore'
    | 'KeyboardAvoidingView'
    | 'ListView'
    | 'MapView'
    | 'Modal'
    | 'NavigatorIOS'
    | 'Picker'
    | 'PickerIOS'
    | 'ProgressBarAndroid'
    | 'ProgressViewIOS'
    | 'ScrollView'
    | 'SegmentedControlIOS'
    | 'Slider'
    | 'SliderIOS'
    | 'SnapshotViewIOS'
    | 'Switch'
    | 'RecyclerViewBackedScrollView'
    | 'RefreshControl'
[...]

PR #1176, opened 2019-01-15
React Native improvements: correct components and TypeScript types

benhjames:
**What**:
- Added TypeScript types to `@emotion/native`
- Improved the list of components in `styled` to match the currently available components in React Native

**Why**:
For better type coverage for consumers of the package, and so the correct components can be used e.g. `styled.View` 🙂 

**How**:
- Added `index.d.ts` to `@emotion/native` which has type information for the exports: all the relevant emotion-ified React Native components and the `css` tagged template literal.
- Fixed list of components in `index.js`

**Checklist**:

- [ ] Documentation (N/A?)
- [x] Tests
- [x] Code complete


benhjames:
And fixes #839 👍

packages/native/types/index.d.ts
```diff
@@ -0,0 +1,67 @@
+// Definitions by: Ben James <https://github.com/benhjames>
+// TypeScript Version: 2.8
+
+import {
+  CreateStyled as BaseCreateStyled,
+  CreateStyledComponentExtrinsic
+} from '@emotion/styled-base'
```
Ailrun: This should not depend on `@emotion/styled-base` since package.json of this package does not have one.
(And please do not add the package, since it will add additional external dependencies)

derekstavis: What's the recommendation in this case, since those dependencies shouldn't be added? Is copying and pasting those types from `@emotion/styled-base` an OK strategy?

Ailrun: @derekstavis Although one adds duplicated codes, that would be a better approach.

jessep:
@benhjames are you planning on addressing the comments? would be great to use this

mrbrannan:
@Ailrun if @benhjames has no interest in taking this over the finish line I am interested in taking over #839

benhjames:
I'm afraid this isn't something I can continue working on - feel free to take over!

elektronik2k5:
@mrbrannan would you like to continue this? I'm using `@emotion/primitives` (which also lacks types) and would like to reuse the work done here.

Andarist:
@ndelangen I would want to wait for https://github.com/emotion-js/emotion/pull/1501 to land before proceeding with this. Would be great to have those definitions be rather similar.

Andarist:
Sorry that I have not been able to get to this sooner, but an overlapping PR implementing this ( https://github.com/emotion-js/emotion/pull/1634 ) has just been merged into `next` branch. This will be a part of the upcoming v11 release.

I would appreciate anyone interested in those typings to take a look at what got merged in and sharing their feedback with us.

PR #1190, opened 2019-01-23
Fixing 'Cannot read property 'key' of undefined' error

badrey:
**What**:
Fixing "Cannot read property 'key' of undefined" error when context provider's cache value is undefined

**Why**:
I am facing it in my project

> TypeError: Cannot read property 'key' of undefined
    at Object.insertStyles (/Users/user/misc/code/mobile/web/node_modules/@emotion/utils/dist/utils.cjs.dev.js:18:25)
    at render (/Users/user/misc/code/mobile/web/node_modules/@emotion/core/dist/core.cjs.dev.js:110:21)
    at /Users/user/misc/code/mobile/web/node_modules/@emotion/core/dist/core.cjs.dev.js:151:10
    at Object.children (/Users/user/misc/code/mobile/web/node_modules/@emotion/core/dist/core.cjs.dev.js:71:18)
    at ReactDOMServerRenderer.render (/Users/user/misc/code/mobile/web/node_modules/react-wildcat-handoff/node_modules/react-dom/cjs/react-dom-server.node.development.js:2446:55)
    at ReactDOMServerRenderer.read (/Users/user/misc/code/mobile/web/node_modules/react-wildcat-handoff/node_modules/react-dom/cjs/react-dom-server.node.development.js:2307:19)
    at Object.renderToString (/Users/user/misc/code/mobile/web/node_modules/react-wildcat-handoff/node_modules/react-dom/cjs/react-dom-server.node.development.js:2679:25)
    at serverRenderPromiseResult (/Users/user/misc/code/mobile/web/node_modules/react-wildcat-handoff/src/utils/serverRender.js:103:72)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)

**How**:
One-line change: null check replaced by not false check

**Checklist**
- [N/A] Documentation
- [N/A ] Tests
- [x] Code complete

### DIFF
--- a/packages/core/src/context.js
+++ b/packages/core/src/context.js
@@ -51,7 +51,7 @@ if (!isBrowser) {
     return (props: Props) => (
       <EmotionCacheContext.Consumer>
         {context => {
-          if (context === null) {
+          if (!context) {
             return (
               <BasicProvider>
                 {newContext => {


badrey:
It fails on typescript check but not because of my changes. It fails the same on current master

mitchellhamilton:
Could you create a reproduction of this happening?

badrey:
@Ailrun please review now

mitchellhamilton:
The cache context should never have a value of undefined so a reproduction would be great because it's likely that this is an environment issue rather than an emotion issue.

badrey:
@mitchellhamilton I upgraded the project to Babel 7 and still getting the error in certain scenarios. 
In the pull request I just replaced check for null to check for null or undefined. It happens during a page refresh. Could you accept the pull request. Or which ways of reproduction do I need to provide? Like a pet project where the issue is reproducible?

Andarist:
We would appreciate a simple repository with the issue reproduced. Just adding extra checks might only cover up true problem so we wouldnt like to check current runtime code without at least understanding why things dont work for you.

Dangoo:
> it's likely that this is an environment issue rather than an emotion issue.

Not sure if it helps, but I’m experiencing the same issue using `target: worker` with webpack in order to use it in a cloudflare worker.

Andarist:
@Dangoo could you prepare a repro case then? Should be reproducible without involving cloudflare workers etc, just use that setting in webpack.

Dangoo:
@Andarist Sure I can do that as soon as I find some time. Will post it here then.

Andarist:
@Dangoo if you manage to create a repro case then I would love to take a look at this. But for now I'm forced to close this because we don't have enough information to work on a fix. I suspect this is just a problem with tools configuration. Hope you understand.

PR #1200, opened 2019-01-29
Do not warn about using kebab-case for object keys being media queries

Andarist:
cc @clight-fastly 😉

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -88,8 +88,12 @@ if (process.env.NODE_ENV !== 'production') {
       }
     }
 
+    const processed = oldProcessStyleValue(key, value)
+    const isCssVariable = key.charCodeAt(1) === 45
+
     if (
-      key.charCodeAt(1) !== 45 &&
+      processed !== '' &&
+      !isCssVariable &&
       key.indexOf('-') !== -1 &&
       hyphenatedCache[key] === undefined
     ) {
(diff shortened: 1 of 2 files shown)


packages/serialize/src/index.js
```diff
@@ -89,7 +89,7 @@ if (process.env.NODE_ENV !== 'production') {
     }
 
     if (
-      key.charCodeAt(1) !== 45 &&
+      (key.charCodeAt(0) !== 64 /* @ */ || key.charCodeAt(1) !== 45) /* - */ &&
```
Andarist: im not sure what case is covered by this part `key.charCodeAt(1) !== 45) /* - */` as there is no test for this, is it supposed to handle css variables?

tkh44: I believe it is.

packages/serialize/src/index.js
```diff
@@ -89,7 +89,8 @@ if (process.env.NODE_ENV !== 'production') {
     }
 
     if (
-      key.charCodeAt(1) !== 45 &&
+      key.charCodeAt(0) !== 64 /* @ */ &&
```
mitchellhamilton: Rather than doing this extra check which won’t also work for nested selectors, you could move the oldProcessStyleValue call before this check and check if the value is not equal to an empty string

Andarist: what kind of nested selectors are we talking about?

mitchellhamilton: Something like this or any other selector with a hyphen.
```jsx
{
  ':last-of-type': null
}
```

Andarist: oh, right - gonna workaround this soon

Andarist:
@mitchellhamilton should be ready to merge now

PR #1201, opened 2019-01-31
[WIP] - Site Redesign

tkh44:
My attempt at reworking the design of the site.

**What**:
image


mitchellhamilton:
Some feedback:
- link styles
  - the color they change to on hover is so similar to the default color that it's barely noticeable
  - the link color isn't applied to inline code
  - not totally sure about this but maybe an underline on hover would be nice
- i really like the sidebar styling though maybe it could be sticky when you're scrolling and this i'm not super sure about but maybe the active link could be bold
- the anchors on headings that appear on hover are strangely aligned, it should probably be vertically centered or something
- community page is broken
- there are three different background colors that are slightly different dark shades: search, code and live code
- i really like the inline code styles
- this is something that has been there for a while but when selecting code, the selection color is different depending on the token. Applying this css might work well
```css
pre[class*="language-"]::selection, pre[class*="language-"] ::selection, .prism-code::selection, .prism-code ::selection {
    text-shadow: none;
    background: hsla(0,0%,100%,.15);
}
```
- page titles being smaller than headings seems strange

tkh44:
Thanks for the feedback. Here's the latest update.
image

mitchellhamilton:
I'm not totally sure about the lack of separation between the different sections on the page. It feels harder to focus on the content because there's no separation between content and navigation

I really like the links at the top with the color change on hover and the underline.

I really liked when the pink indicator thing on the left of the nav links was right on the edge of the page.

PR #1206, opened 2019-02-01
Add @emotion/core to recommended install script for @emotion/styled

Andarist:


docs/introduction.md
```diff
@@ -105,7 +105,7 @@ render(
 ```
 
 ```bash
-npm i @emotion/styled
+npm i @emotion/styled @emotion/core
```
Andarist: core is a peer dep for styled so it has to be installed too

Issue #1178, opened 2019-01-15
Problems surrounding SSR injection of <style> and unreliability of :first-child selectors


kylegach:
**Problem description:**

In emotion v10+, a new warning is thrown if you use `:first-child`, `:nth-child`, or `:nth-last-child` selectors in a styled component or css prop value. This is because, when using SSR, the `style` element is injected immediately above (prepended) the associated component. ([More details](https://github.com/emotion-js/emotion/issues/898).)

There are a number of problems with this approach:

> _Note: I made a [test case demo](https://codepen.io/anon/pen/BvMjeQ) to illustrate which selectors are unreliable._

1. In [our codebase](https://github.com/mineral-ui/mineral-ui/) and matching years of experience for myself and peers, the "first" selectors are used much more often than the "last". Making these selectors unreliable is a big blow to those expecting standard CSS features to work.
1. The suggested workaround of changing, e.g. `:first-child` to `:first-of-type`, only works if _all_ sibling elements are of the same type. This is probably fairly common for components like lists, grids, etc... But it's far from a guaranteed solution. It could also discourage the use of semantic markup by driving devs to "just use divs" for everything, since it's a simpler "fix".
1. The [list of unsafe selectors](https://github.com/emotion-js/emotion/blob/master/packages/cache/src/index.js#L208) is incorrect.
    - The adjacent sibling pattern, `* + [anything]`, is unreliable, but not listed.
      - This [type of selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls) may seem somewhat obscure, but it can be rather useful.
    - `:only-child` is unreliable, but not listed.
    - `:nth-last-child` is still reliable when prepending the `style` element and should not be listed.
1. Consumers not interested in SSR should be able to turn off the warning (#1105)

**Suggested solution:**

Let's tackle each problem, in turn:

1. "First" vs. "last" selectors (or prepending vs. appending the `style` element)

    If emotion injected the `style` element _after_ (appended) the associated component, then the following selectors would be unreliable: `:last-child`, `:nth-last-child`, `:only-child`.

    Compare that to the currently unreliable selectors: `* + [anything]`, `:first-child`, `:nth-child`, `:only-child`.

    By changing to appending the `style` element, emotion would no longer block use of the common adjacent sibling selector and the "last" (rather than "first") varieties of the selectors would be affected, which I would argue is a better tradeoff more in line with actual usage of each [citation needed].

1. "of-type" workaround

[...]

Issue #1105, opened 2018-12-13
Cannot turn off "potentially unsafe when doing server-side rendering" noise


RoystonS:
Emotion 10 is emitting a series of console warnings along the lines of:

```
The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to "first-of-type"
The pseudo class ":nth-child" is potentially unsafe when doing server-side rendering. Try changing it to "nth-of-type"
```

- `emotion` version: 10.0.4
- `react` version: 16.7.0-alpha2

**Problem description:**

1. I'm not using server-side rendering so these warnings aren't useful for me and are just console noise that means real problems will be masked.  Apart from running in a production build, I can't see any way to turn these warnings off right now.
1. The error message doesn't help me understand why these pseudo classes are a problem; I can't find them mentioned in the Emotion docs, or references on the Internet as to why these might be a problem.

**Suggested solution:**

1. Provide a way to turn off these warnings if users don't want them.
1. Provide a link, or some documentation, as to why these pseudos are problematic.

Andarist:
Leaving this here - https://github.com/emotion-js/emotion/issues/1059#issuecomment-444566635 as this issue is dedicated to this problem.

vajkri:
We are getting the warnings too using SSR. 
The styles targeting with `:nth-child()`, `:first-child` or `:last-child` don't even get applied server-side, so this causes elements to "jump" on page load when these styles get applied client-side.

brentertz:
Hi Folks,

The inability to disable these console messages is a blocker for us being able to ship the [Emotion 10 update to our component library](https://github.com/mineral-ui/mineral-ui/pull/816). As a library, these messages are also visible to all of our downstream consumers.

We do not wish to update the selector usage in our library as we do not believe Emotion's current SSR strategy to be a valid trade-off for breaking numerous commonly used CSS selectors, hence #1178. We are hopeful that an alternate approach can be found there as well, but in the mean time, simply would like to disable these console messages.

Is there anything that can be done here to move things forward?  We are willing to do the work, or collaborate, if you can provide some insight into the direction that you would like to take.

__Possible Solutions__

* Simply remove logging; instead improve and rely on SSR documentation
* Conditionally remove logging, e.g. ENV var or other opt out mechanism
* Only log when server side rendering
* Only log message once per page load
    * Unsure if this would avoid excess logging during unit tests though
* Remove the Stylis “linting” plugins from the default cache implementation and make them opt-in
[...]

PR #1209, opened 2019-02-01
[cache] Add ignore flag for the unreliable selectors warning

kylegach:
**What**:
1. Add the ability to ignore a selector's "potentially unsafe when doing server-side rendering" warning message by adding `/* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */` after the pseudo class
    - The flag is intentionally verbose (see [comments below](https://github.com/emotion-js/emotion/pull/1209#issuecomment-462035404))
2. Add tests for above

**Why**:

The inability to disable or ignore these warnings is an issue for some (#1105). Until a decision is made on #1178, it is likely worthwhile to provide a way to ignore them. This seemed like a lightweight way of doing so, that is nicely co-located with the existing warning logic.

**How**:

Modified the implementation of the stylis plugin used for the warning to check for the presence of the flag.

**Checklist**:

- [x] Documentation — Maintainers wish for this flag to be hard to discover, so docs were intentionally not updated
- [x] Tests
- [x] Code complete

### DIFF
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -201,21 +201,30 @@ let createCache = (options?: Options): EmotionCache => {
 
     stylis.use((context, content, selectors) => {
       switch (context) {
-        case 2: {
-          for (let i = 0, len = selectors.length; len > i; i++) {
-            // :last-child isn't included here since it's safe
-            // because a style element will never be the last element
-            let match = selectors[i].match(/:(first|nth|nth-last)-child/)
-            if (match !== null) {
-              console.error(
-                `The pseudo class "${
-                  match[0]
-                }" is potentially unsafe when doing server-side rendering. Try changing it to "${
-                  match[1]
-                }-of-type"`
+        case -1: {
+          const flag =
+            'emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason'
+          const unsafePseudoClasses = content.match(
+            /(:first|:nth|:nth-last)-child/g
+          )
+
+          if (unsafePseudoClasses) {
+            unsafePseudoClasses.forEach(unsafePseudoClass => {
+              const ignoreRegExp = new RegExp(
+                `${unsafePseudoClass}.*\\/\\* ${flag} \\*\\/`
               )
-            }
+              const ignore = ignoreRegExp.test(content)
+
+              if (unsafePseudoClass && !ignore) {
+                console.error(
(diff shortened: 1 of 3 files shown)


mitchellhamilton:
I'm not completely opposed to doing something like this but the way this works specifically would require us to also remove the property in production and I'm really against adding code in prod that isn't necessary in prod.

brentertz:
@mitchellhamilton @kylegach I'd suggest that we simply change it to use a CSS custom property, e.g. `--ignore`, so that it it valid syntax and does not need removed for production.

kylegach:
@mitchellhamilton — Glad to hear you're open to the idea. I updated the code (force-pushed over the first commit) and the PR description to reflect Brent's suggestion above.

mitchellhamilton:
Changing it to a custom property doesn't really solve the problem. There's nothing that will break with shipping an invalid css property to production and it's better than a custom property in a way since browsers will ignore the invalid property. The bigger problem is that an extra css property is there at all. 

I think it's possible to do with comments like this and since comments are already removed, it won't have any ramifications on production.

```js
css`
  &:first-child /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */ {
  }
`;
css({
  "&:first-child /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */": {}
});
```
Yes, I know it's long, that's intentional, it should be difficult for people to disable the warning. It's there to help people and fix a problem they could have so it should be be difficult to stop the warning because in most cases, people should fix their code.

kylegach:
@mitchellhamilton —

Oh! Thanks for the suggestion to place the comment _in_ the selector. I had tried comments first, but couldn't figure out how to make it work for style objects.

Agreed that it's a good idea to have an obnoxiously verbose flag.

Just updated the PR description and force-pushed again. Beyond changing the flag to a comment, the major changes from last review are no longer printing the compiled selector in the error (decided it wasn't that helpful) and a bunch more test cases for compound selectors.

Coolranch:
Did this flag make it in?  I'm using version 10.0.7 with the disable flag in my styles, but I'm still seeing the warning.

joshuaegclark:
Any updates on this getting merged?

leftmostgeek:
I'd like to see this change too.

kylegach:
Force-pushed one more time with the changes requested and updated the description accordingly. Should be good to go! Thanks for the help, @mitchellhamilton.

PR #1215, opened 2019-02-07
Implement Yarn Workspaces

joe-bell:
## What
Implement Lerna's [Yarn Workspaces option](https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces) for installing and linking local packages

## Why

While reading CONTRIBUTING I noticed the line regarding [Lerna's purpose in the repo](https://github.com/emotion-js/emotion/blob/789f94dc466e11b5ae285bdf1d2db33079d6713e/CONTRIBUTING.md#installation):

>NOTE:
>
>lerna is NOT used for installing packages. Only yarn is used. lerna is only used for publishing

If that's the case it seemed like a no-brainer to switch on [Lerna's Yarn Workspaces integration](https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces). This way Emotion contributors can fully leverage Yarn as a package manager.

Some more info on Yarn Workpaces can be found here: https://yarnpkg.com/lang/en/docs/workspaces/

I've used this integration for 3 Lerna repos and it works like a dream. You'll still only need to use the `lerna` command for odd tasks such as publishing.

### Stats

On my luxury Estonian hotel WiFi of **7.1Mbps** (measured on fast.com), I measured both the install times (with big differences) and some common tasks which didn't differ.

#### Install

Before (without Yarn Workspaces) - `203.67s`
screenshot 2019-02-07 at 19 53 26

With Yarn Workspaces -  **`116.19s`**
screenshot 2019-02-07 at 20 00 12

## How

By following the steps outlined in the [Lerna docs](https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces)

## Checklist

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

Oh and I also added a `clean` command as I found it handy, feel free to drop this commit if it's not necessary

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@
 - Run `yarn build` in the root directory to build the modules. (Required before publishing)
 - Run `yarn build PACKAGE_NAME ANOTHER_PACKAGE_NAME` to only build certain packages.
 - Run `yarn build:watch` to build packages on every change.
+- Run `yarn clean` to remove previously built modules and dependencies.
 
 ## Documentation Website Development
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,9 +8,7 @@
 - Run `yarn` in the repository's root directory to install everything you need for development.
 - Run `yarn build` in the root directory to build the modules.
 
-> NOTE:
->
-> lerna is **NOT** used for installing packages. Only yarn is used. lerna is only used for publishing
+> Note: Running `yarn` tells Lerna to install and link project packages as [Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). The `lerna` command itself is only used for publishing.
 
 ## Running Tests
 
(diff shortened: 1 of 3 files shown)


FezVrasta:
Tested and it works 👍

mitchellhamilton:
Thanks for the PR but this shouldn’t help anything. Yarn doesn’t do anything with the Lerna config, it’s only that Lerna can use yarn. People should be installing by running yarn, we already have a workspaces config in the root package.json. People should never use Lerna in this repo.

mitchellhamilton:
Thanks for the PR but this shouldn’t help anything. Yarn doesn’t do anything with the Lerna config, it’s only that Lerna can use yarn. People should be installing by running yarn, we already have a workspaces config in the root package.json. People should never use Lerna in this repo.

joe-bell:
Well that would explain why the build was so fast when I'm removing half the previous `workspaces` config…

My bad

PR #1220, opened 2019-02-10
Add import mapping to babel plugin

### LINKED ISSUE #1203: babel-plugin-emotion to support more packages
Hey emotion dev-team. Storybook maintainer here. Got a weird request..
Concerning https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion

Storybook uses emotion, and it’s great, but we’ve got a conflict:
If a user of storybook is also using emotion (happening more and more) potential conflicts of versions happen. When multiple version of emotion are in 1 frame, it mostly works, but theme contexts get “confused”.

So I came up with a solution so storybook would get control of the version inside itself (and addons) and users are in control of the emotion they are using. And the install of the user’s version would happen where they’d expect.

The solution was to wrap emotion packages by a package called `@storybook/theming`. It also happens to include the default themes.

So when I’m creating emotion components it looks like this:

```
import { styled } from '@storybook/theming';
```

Obviously the babel-plugin-emotion doesn’t detect this, and so it doesn’t do anything. Sad. For the most part it’s OK, except the autoLabel feature is sweeet.
But the babel-plugin is a requirement for certain features it seems. I’m getting this:
```
Component selectors can only be used in conjunction with babel-plugin-emotion
```

So sorry for the long intro.. here’s my question:
What can I do to get `babel-plugin-emotion` to work on 
`import { styled } from '@storybook/theming';` ?

mitchellhamilton:
Closes #1203 

This is just a sketch of what I'm thinking the api should look like right now, implementation coming soon.

cc @ndelangen wdyt of this API? (ignore the naming, it'll change, just the structure)


ndelangen:
Yeah this API would be awesome IMHO!

knpwrs:
I just spent a couple hours trying to solve babel+typescript+`@emotion/styled` before stumbling across this PR. This would be a great addition!

ndelangen:
@mitchellhamilton If you'd be able to spend the time, I'd love to assist with this, maybe do some pair programming. I'd like to learn more about this topic, and assist in the long term maintenance of this feature.

I'm in the Slack channel if you'll let me know when you're working on this, I can possibly assist and learn?

knpwrs:
Is there anything the community can do to help get this through? I think this would greatly improve the usability of Emotion with TypeScript.

ford04:
This also would be really useful with custom styled components that strongly type your app theme like described [here](https://emotion.sh/docs/typescript#define-a-theme). Current situation is that `babel-plugin-emotion` cannot create auto labels for my themed `styled` component so React Dev tools still shows `styled(div)` etc. .

sqal:
> I just spent a couple hours trying to solve babel+typescript+@emotion/styled

@knpwrs I have the same problem, have you found any workaround for this issue? or do we have to wait for this pull request to be merged?

knpwrs:
As far as I can tell we have to wait for this to be merged. My "workaround" for the time being is to just live without the features of the Babel plugin because I happen to value type safety more than the features, though I would love to have both.

mitchellhamilton:
For everyone curious about this, this definitely something we want to add but it's caused me to think about various APIs and optimizations which I'd like to change in emotion which will affect this API so I'd like to make those changes before adding this.

ford04:
Nice to hear that you are on it.
In the mean time I give up having an overview of my styled components in react dev tools 😂

(using themed custom styled)
image

Andarist:
@mitchellhamilton is there anything I could do to help move this forward? including doing other work - you have mentioned you want to prepare the groundwork for some optimizations first, could help with that

mitchellhamilton:
@Andarist I’ve been thinking even more and the changes I mentioned there aren’t things I want to do right now so the next thing to move this forward, the next thing to do is ensure that anything that re-exports jsx also exports `css` from @emotion/core and that the appropriate version of css is used. If you want to submit a PR against this branch, that would be great.

Also, other things that need to happen:
- remove instances option (the import mapping will solve the problem that that option was trying to solve)
- figure out good names for the things in the import mapping option.

Andarist:
@fforres if you want to help with that we would certainly appreciate the help. I was supposed to get to it, but I constantly can't find enough time to dive into it - being preoccupied with other OSS work. If you'd like to help you could join our Slack and ping me so we could discuss the details.

fforres:
Sounds good @Andarist 😄

packages/babel-plugin-emotion/src/index.js
```diff
@@ -22,6 +31,30 @@ let primitivesStyledMacro = createStyledMacro({
   isWeb: false
 })
 
+let transformersSource = {
+  emotion: vanillaTransformers,
+  '@emotion/css': {
+    default: coreCssTransformer
+  },
+  '@emotion/core': {
+    // jsx isn't included here because it's handled differently
```
Andarist: what does it mean that it's handled differently? would be good to point some point of reference or an explanation here

packages/babel-plugin-emotion/src/index.js
```diff
@@ -22,6 +31,30 @@ let primitivesStyledMacro = createStyledMacro({
...
+    css: coreCssTransformer
+    // TODO: maybe write transformers for keyframes and Global
```
Andarist: what could be handled but is not?

packages/babel-preset-css-prop/src/index.js
```diff
@@ -30,8 +30,8 @@ export default (
           sourceMap,
           autoLabel,
           labelFormat,
-          instances,
-          cssPropOptimization: true
+          cssPropOptimization: true,
```
Andarist: out of curiosity - why this even does exist? when one would not like to enable this optimization?

mitchellhamilton: Someone is using another library with a css prop. By default, we'll detect if `jsx` is imported and only do the transform then but that doesn't work with the auto-insertion so we have this option.

packages/babel-plugin-emotion/src/index.js
mitchellhamilton: Minifying styles for keyframes and Global

Andarist: K, seems interesting to me so I'm going to work on those when this PR lands.

ndelangen:
👏👏👏👏👏👏👏👏👏👏👏👏👏

earksiinni:
Has this made its way into the `@emotion/babel-plugin` packages that are currently available on NPM? I'm using `@emotion/babel-plugin@^11.0.0-next.12` and I still have trouble with custom `styled` imports in TS.

Andarist:
Yes, this should have been released quite some time ago under the `next` tag. If you experience any problems please create a repro case and report an issue.

PR #1224, opened 2019-02-11
move babel-plugin-emotion to devDependencies

dcousens:
Unless I have misunderstood


mitchellhamilton:
Thanks for the PR but babel-plugin-emotion has to be a dependency because it’s required by the babel macro

dcousens:
@mitchellhamilton could it be an optional dependency for those who use `'emotion/macro'`?
This dependency brings in the entire babel stack ...

Issue #1059, opened 2018-11-30
The docs have taken a sharp downhill turn with v10


jaredLunde:
The documentation for v10 needs some work. v10 isn't a small API change, it completely changes the way a user has to interact with the library. I have used `emotion` on all of my projects for over a year so it's not like I'm a newb to this library. If it's hard for me to understand just what is going on I can't imagine the hurdle for actual newcomers.

Given the obscure API that comes with this version (`@emotion/core` vs `emotion`, `jsx` as an used import?This pragma is supposed to be more clear/easy to use than a babel plugin? ), it would be nice if each of these new packages had their own documentation, since apparently they're all completely different. As it stands you have them all randomly strewn about with notes like 'you only need this if...'

Grant you I'm coming at this from a place of general negativity because I've disagreed with this direction from the start (gave a thumbs down on your initial proposal for example), but those feelings are compounded by the fact that the migration guide isn't at all clear. It makes suggestions about moving from `emotion` to `@emotion/core` but still says elsewhere in the docs you can still use `emotion`. OK. Well, when should I use each? Why should I use `@emotion/core` over `emotion`? What am I missing out on by not doing so?

I really hope this issue with the docs gets ironed out. I'd help, but clearly I don't understand the API as it is so I don't have anything to offer.  The only concrete suggestions I have are that you create completely separate sites for `emotion` and `@emotion/core` and highlight whatever the differences are somewhere in addition to explaining more clearly what `jsx` is for. Some more clarity about the new `css` prop usage would also be nice.

tkh44:
This is a completely valid criticism. I think we can address a few items to make this better for our users.

- Link to the v9 docs at the top

- Introduction needs to be rewritten and should proabably reflect the docs we had for the old doc site as @staylor mentions above. A simple rundown of the features without all the caveats. We can save those for the detailed docs on the nav bar.

- Remove the ambiguity between the react and non-react versions of emotion. We should have a dedicated doc to the non react version linked on the sidebar.

- #1049 should be merged and the docs should show this as the recommended way to set up the css prop with the manual pragma being the fallback.

	* This will require more explanation for certain cases where it could cause some confusion.

- Any docs referencing previous versions of emotion should be moved to the `migration guide` (e.g. https://emotion.sh/docs/ssr)

- Next.js, Gatsby, and Puppeteer sections from the SSR docs should be their own pages with links directly from the sidebar. This should be done even if, for example, the Next.js docs just say, "It just works".

[...]

PR #1231, opened 2019-02-14
Update readme to clarify that jsx comment and unused import can be removed

cheapsteak:
**What**: Update readme to clarify/assure that the jsx comment and unused import can be removed

**Why**: Spoke with a seasoned dev evaluating switching to SC because "emotion ruined their api" and pointed to the Quickstart code sample. Also this [tweet](https://twitter.com/larsgraubner/status/1095952530180329474) from another developer rightfully frustrated just enough to decide against this. I myself initially was worried about the same thing and took a bit of clicking around to confirm that the scope of change isn't as large as I was worried it'd be

Also in https://github.com/emotion-js/emotion/issues/1059#issuecomment-455013709
> Maybe it's not the right place to ask, but, in v10, is it possible to get rid of `jsx` import?

The jsx pragma cruft is unnecessarily scaring people away

**Checklist**:

- [ ] Documentation NA
- [ ] Tests NA
- [ ] Code complete NA

Somewhat related to https://github.com/emotion-js/emotion/issues/1059

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ let SomeComponent = props => {
 
 ### Do I Need To Use the Babel Plugin?
 
-The babel plugin is not required, but enables some optimizations and customizations that could be beneficial for your project.
+The babel plugin is not required, but it removes the need for both the `/** @jsx jsx */` comment and the unused jsx import, and enables some optimizations and customizations that could be beneficial for your project.
 
 Look here 👉 _[emotion babel plugin feature table and documentation](https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion)_
 


cheapsteak:
Thanks for the review/approval!
image
Netlify integrations failing, is this something I can resolve on my end?

README.md
```diff
@@ -58,7 +58,7 @@ let SomeComponent = props => {
 
 ### Do I Need To Use the Babel Plugin?
 
-The babel plugin is not required, but enables some optimizations and customizations that could be beneficial for your project.
+The babel plugin is not required, but it removes the need for both the `/** @jsx jsx */` comment and the unused jsx import, and enables some optimizations and customizations that could be beneficial for your project.
```
mitchellhamilton: This isn’t true of the babel plugin, it’s only not necessary when using @emotion/babel-preset-css-prop.

cheapsteak: Ah good catch
Since the preset includes the plugin, would it make sense to replace this line to reference the preset instead?

cheapsteak: Maybe something like image ? 

Still linking to the plugin because the readme for the preset doesn't contain the feature table. Perhaps it'd make sense to extract that out into a separate md file?

PR #1240, opened 2019-02-23
Add labels for function declarations

mitchellhamilton:
**What**:
Add labels for function declarations

**Why**:
So that function components using the css prop written as function declarations get labels. Currently this generally already works because of the stack trace parsing but there are cases when it will not work, e.g. bundlers rename variables for module concatenation or etc. so this means even in those cases it will still work. The one potential problem is that it will add labels anytime a css call(or css prop ) is used inside a function declaration which could be annoying for css calls inside utility functions. Maybe the solution could be to only add labels if the name for a function is in pascal case(so it's very likely a react component)? Thoughts?

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/babel-plugin-emotion/__tests__/__fixtures__/function-declaration.js
+++ b/packages/babel-plugin-emotion/__tests__/__fixtures__/function-declaration.js
@@ -1,0 +1,13 @@
+// @flow
+import * as React from 'react'
+import { jsx } from '@emotion/core'
+
+function Logo(props) {
+  return (
+    <a
+      css={{
+        display: 'block'
+      }}
+    />
+  )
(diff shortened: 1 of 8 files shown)


Andarist:
> Maybe the solution could be to only add labels if the name for a function is in pascal case(so it's very likely a react component)? Thoughts?

Seems like a reasonable heuristic as component's name have to be pascal case in order to make them usable with JSX transform, so 98% are written like this and not many other functions should share this naming scheme.

tkh44:
LGTM. Pascal case reasoning should cover most use-cases.

Issue #1208, opened 2019-02-01
jest-emotion No support for CSS pseudo-elements and pseudo-classes


Tokery:
- `jest-emotion` version: 10.0.6

**Relevant code:**

I am looking to create tests that can add modifiers to the class selector, like seen below. (From jest styled components)
```javascript
const Button = styled.button`
  @media (max-width: 640px) {
    &:hover {
      color: red;
    }
  }
`

test('it works', () => {
  const tree = renderer.create(<Button />).toJSON()
  expect(tree).toHaveStyleRule('color', 'red', {
    media: '(max-width:640px)',
    modifier: ':hover',
  })
})
```

As far as I'm aware, this doesn't yet exist.

Werter12:
@Tokery Actually, the PR with this functionality was merged https://github.com/emotion-js/emotion/pull/1110. I am going to add documentation for it soon. So, you can use the following syntax:
`
  expect(tree).toHaveStyleRule('color', 'red', {
    media: '(max-width:640px)',
    target: ':hover',
  })
`

Tokery:
Ah, I see. Thanks for the info.

stephenh:
I know this is a closed issue, but I can't get this to work with a non-trivial selector, i.e.:

```
<div css={ { "&:hover > *": { color: "red" } } }> ... </div>

expect(...).toHaveStyleRule("color", "red", { target: "&:hover > *" }
```

Doesn't work, and I've tried various approaches of putting spaces between/not-between `&` / `:` / `>` / `*`, including/not including `&`, etc., and can't seem to get anything to work.

I set a debug point inside of `toMatchStyleRule` and it really looks like the `> *` is not coming back at all for the `target` to later pick it up.

@Werter12 apologies for the ~year later lazy ask on a closed PR, but the docs are still light on this more esoteric selectors; are there any obvious ideas I'm missing? Thanks!

Andarist:
@Werter12 if you set up a simple repro for the problem I would be able to take a quick look at it (as in - debug it and assess if you have passed arguments incorrectly or if maybe we should change something in Emotion's `toMatchStyleRule`)

Andarist:
@stephenh  if you set up a simple repro for the problem I would be able to take a quick look at it (as in - debug it and assess if you have passed arguments incorrectly or if maybe we should change something in Emotion's `toMatchStyleRule`)

stephenh:
@Andarist I played with this a bit and figured out it does basically work, i.e. for this:

[...]

PR #1242, opened 2019-02-24
Add docs for jest-emotion toHaveStyleRule additional options

Werter12:
I've noticed that some functionality of `toHaveStyleRule` matcher is not clear for developers https://github.com/emotion-js/emotion/issues/1208 . So this PR intended to tell about additional options which developer can provide for `toHaveStyleRule` matcher like `target` and `media`.
- [x] Documentation

### DIFF
--- a/packages/jest-emotion/README.md
+++ b/packages/jest-emotion/README.md
@@ -94,14 +94,67 @@ import styled from '@emotion/styled'
 expect.extend(matchers)
 
 test('renders with correct styles', () => {
-  const H1 = styled.h1`
-    float: left;
+  const Svg = styled('svg')`
+    width: 100%;
   `
 
-  const tree = renderer.create(<H1>hello world</H1>).toJSON()
+  const Div = styled('div')`
+    float: left;
+    height: 80%;
+    &:hover {
+      width: 50px;
+    }
+    ${Svg} {
+      fill: green;
+    }
+    span {
+      color: yellow;
+    }
+    @media screen and (max-width: 1200px) {
+      font-size: 14px;
+    }
+  `
+
+  const tree = renderer
+    .create(
+      <Div>
+        <Svg />
+        <span>Test</span>
+      </Div>
+    )
+    .toJSON()
 
   expect(tree).toHaveStyleRule('float', 'left')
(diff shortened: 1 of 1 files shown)


packages/jest-emotion/README.md
```diff
@@ -94,14 +94,67 @@ import styled from '@emotion/styled'
...
+    .toJSON()
 
   expect(tree).toHaveStyleRule('float', 'left')
-  expect(tree).not.toHaveStyleRule('color', 'hotpink')
+  expect(tree).not.toHaveStyleRule('height', '100%')
+})
+```
+
+You can provide additional options for `toHaveStyleRule` matcher.  
+`target` - option helps to specify target where style rule should be found.
```
mitchellhamilton: This should mention that's it's a css selector.

packages/jest-emotion/README.md
```diff
@@ -94,14 +94,67 @@ import styled from '@emotion/styled'
...
+
+1.  Pseudo-class.
```
mitchellhamilton: Could you remove the 1., 2. and etc.?

packages/jest-emotion/README.md
```diff
@@ -94,14 +94,67 @@ import styled from '@emotion/styled'
...
+expect(tree).toHaveStyleRule('color', 'yellow', { target: 'span' })
+```
+
+3. Another styled component.
+
+```
+expect(tree).toHaveStyleRule('fill', 'green', { target: `${Svg}` })
+```
+
+`media` - option allows to test style rule within media.
```
mitchellhamilton: This doesn't seem like a great description of what it does, something along the lines of 'specifies the media rule that the matcher should look for the property in'. or that worded better

packages/jest-emotion/README.md
Werter12: yes

packages/jest-emotion/README.md
Werter12: sure

packages/jest-emotion/README.md
Werter12: Ok

PR #1252, opened 2019-02-28
fix(typescript): Preserve props union type

eps1lon:
**What**:
- change `Omit` to be distributive over union types

**Why**:
- `withTheme` and `styled` loose union type informaton

**How**:
- use recommended pattern from https://github.com/Microsoft/TypeScript/issues/28339#issuecomment-467220238
- use `any` instead of `unknown` for backwards compatibility. If you don't support typescript 2.x we can use `unknown`

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Code complete N/A

- typescript tests were also failing on `master`
- similar fix was accepted for `styled-components`:DefinitelyTyped/DefinitelyTyped#33061

### DIFF
--- a/packages/emotion-theming/types/helper.d.ts
+++ b/packages/emotion-theming/types/helper.d.ts
@@ -10,6 +10,6 @@ export type PropsOf<C extends React.ComponentType<any>> = C extends React.SFC<
         : never)
     : never
 
-export type Omit<T, U> = Pick<T, Exclude<keyof T, U>>
+export type Omit<T, U> = T extends any ? Pick<T, Exclude<keyof T, U>> : never
 export type AddOptionalTo<T, U> = Omit<T, U> &
   Partial<Pick<T, Extract<keyof T, U>>>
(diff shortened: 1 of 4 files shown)


packages/emotion-theming/types/helper.d.ts
```diff
@@ -10,6 +10,6 @@ export type PropsOf<C extends React.ComponentType<any>> = C extends React.SFC<
         : never)
     : never
 
-export type Omit<T, U> = Pick<T, Exclude<keyof T, U>>
+export type Omit<T, U> = T extends any ? Pick<T, Exclude<keyof T, U>> : never
```
mitchellhamilton: I'm not a ts user but in the recommended pattern you linked to, it says `T extends unknown` rather than `T extends any`, should it be using unknown or is there a reason for using any?

eps1lon: As far as I know `T extends any` would not evaluate to true for every type so `unknown` is safer. However for objects it should always be true. It is used since `unknown` was only released in typescript 3.0. If your types don't support 2.x we might as well go full in.

Also
> use any instead of unknown for backwards compatibility. If you don't support typescript 2.x we can use unknown

mitchellhamilton: Oops, didn't see that! I'd rather keep compatibility so I'm good with this.

PR #1254, opened 2019-03-01
Add @emotion/* dependencies to babel-plugin-emotion READMEs

elliottsj:
**What**:

Add @emotion/* dependencies to babel-plugin-emotion / @emotion/babel-preset-css-prop READMEs

**Why**:

babel-plugin-emotion includes macros which may generate application code
that imports these packages, e.g.

    import _styled from "@emotion/styled-base";

Normally, this is not a problem when using popular package managers like
npm / yarn because '@emotion/styled-base' is a transitive
dependency of '@emotion/styled', which is hoisted to the root
node_modules folder, enabling it to be found by webpack / node:

    node_modules/
      @emotion/styled/
        ...
      @emotion/styled-base/
        ...

But when using strict package managers such as pnpm (or yarn pnp),
@emotion/styled-base cannot be resolved because it can only be resolved
by code inside the @emotion/styled package:

    node_modules/
      @emotion/styled/
        ...
        node_modules/
          @emotion/styled-base/
          ...

Adding these to the respective READMEs of both babel packages should
help users avoid this issue.

**How**:

Manually editing README.md files.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A

### DIFF
--- a/packages/babel-plugin-emotion/README.md
+++ b/packages/babel-plugin-emotion/README.md
@@ -97,14 +97,16 @@ const myStyles = /* #__PURE__ */ css(
 ## Installation
 
 ```bash
-yarn add --dev babel-plugin-emotion
+yarn add --dev babel-plugin-emotion @emotion/styled-base @emotion/native @emotion/primitives
 ```
 
 or if you prefer npm
 
 ```bash
-npm install --save-dev babel-plugin-emotion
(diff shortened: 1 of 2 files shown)


Andarist:
This will be fixed when we land https://github.com/emotion-js/emotion/pull/1575 . Closing it right now though to clean up the issue tracker.

Issue #1214, opened 2019-02-05
Regression in Emotion + jest-emotion


FezVrasta:
- `emotion` version: 10.0.7
- `react` version: 16.7.0

**Relevant code:**

```jsx
import React from 'react';
import styled from '@emotion/styled/macro';

export const Bar = styled.div`
  color: blue;
`;

export const Foo = styled(props => (
  <div {...props}>
    <Bar />
  </div>
))`
  color: red;

  ${Bar} {
      color: yellow;
  }
`;
```

**What you did:**

```
it('child style to be read by toHaveStyleRule', () => {
  expect(mount(<Foo />).find(Bar)).toHaveStyleRule('color', 'yellow');
});
```

**What happened:**

```
 FAIL  src/bug.test.js
  ✕ child style to be read by toHaveStyleRule (17ms)

  ● child style to be read by toHaveStyleRule

    Expected color to match:
      yellow
    Received:
      blue
```

**Reproduction:**

https://github.com/FezVrasta/emotion-jest-regression

run `yarn test`

**Problem description:**

This started happening after I upgraded Emotion 10 to the latest patch version, I'm sure it works with:

```
"@emotion/core": "^10.0.4",
"@emotion/styled": "^10.0.4",
"jest-emotion": "^10.0.2",
```

Emotion itself actually works, in my repro you can see in the snapshot that the correct class is generated:

```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`child style to be read by toHaveStyleRule 1`] = `
.emotion-2 {
  color: red;
}

.emotion-2 .emotion-1 {
  color: yellow;
}

.emotion-0 {
  color: blue;
}

<Foo>
  <div
    className="emotion-2 emotion-3"
  >
    <Bar>
      <div
        className="emotion-0 emotion-1"
      />
    </Bar>
  </div>
</Foo>
`;

```

but jest-emotion can't read it for some reason/

**Suggested solution:**

Not sure.

Andarist:
I see the problem in the code and I'll look into fixing this, but truth to be told I can't get it working with `jest-emotion@10.0.2` so I'm not sure if this has ever worked.

FezVrasta:
I noticed this because a lot of project tests broken after the upgrade, so I'm pretty confident somehow it used to work.

Andarist:
This is going to be fixed by https://github.com/emotion-js/emotion/pull/1604

Issue #1187, opened 2019-01-22
Cannot make 'jest-emotion' work with Emotion 10, CSS Prop and Enzyme


ianvieira:
- `emotion` version: @emotion/core@10.0.6
- `react` version: 16.6.1
- `jest` version: 23.5.0

**Relevant code:**

setupTestFrameworkScriptFile file:
```javascript
const jsonSerializer = require('enzyme-to-json/serializer');
const emotionSerializer = require('jest-emotion');

expect.addSnapshotSerializer(emotionSerializer);
expect.addSnapshotSerializer(jsonSerializer);
```
Component.js
```javascript
render() {
       const ratioStyles = css(generatedObjectWithStyles);

        return (
            <div className={ theme.placeholder } css={ [ratioStyles, placeholderStyle] } />
        );
}
```

Component.test.js
```javascript
test('should render with required props', () => {
        const wrapper = shallow(
            <Component { ...mockProps } />
        );

        expect(wrapper).toMatchSnapshot();
    });
```

**What you did:**
I added the `@emotion/babel-preset-css-prop` preset, the code is working as expected and added the snapshot serializer.

**What happened:**
Component snapshot (Shallow):
```javascript
 +     <div
    +       className="Theme__Placeholder"
    +       css={
    +         Array [
    +           Object {
    +             "map": undefined,
    +             "name": "ac6mem",
    +             "next": undefined,
    +             "styles": "padding-bottom:56.25%;",
    +           },
    +           Object {
    +             "map": undefined,
    +             "name": "1wdfeuq",
    +             "next": undefined,
    +             "styles": "
    +     background: #e6e6e6;
    +     border-radius: 0.4rem;
    + ",
    +           },
    +         ]
    +       }
          />

```

Component snapshot (Mount):
```javascript
<div
    +           className=""
    +           css={
    +             Array [
    +               Object {
    +                 "map": undefined,
    +                 "name": "ac6mem",
    +                 "next": undefined,
    +                 "styles": "padding-bottom:56.25%;",
    +               },
    +               Object {
    +                 "map": undefined,
    +                 "name": "1wdfeuq",
    +                 "next": undefined,
    +                 "styles": "
    +     background: #e6e6e6;
    +     border-radius: 0.4rem;
    + ",
    +               },
    +             ]
    +           }
    +         />

```

**Problem description:**
The CSS prop is returning the object instead of the className prop filled and the styles in a separated part, as in the docs.

Tried also setting manually the pragma, same result.

[...]

PR #1256, opened 2019-03-05
adds enzyme mount snapshot test using css from @emotion/core

petegleeson:
**What**:
The entire object returned by css is being serialized in snapshot tests. I have written a test that uses `mount` from `enzyme` and `css` from `@emotion/core`. This output is quite different from previous versions of emotion and what is described in the [snapshot testing docs](https://emotion.sh/docs/testing#writing-a-test).

Related to #1214 and #1187 

**Why**:
It would be great if what is serialised stays roughly equivalent between emotion 9 and 10. The value of `map` changes fairly regularly which is problematic for snapshot tests.  

**How**:
I haven't looked into how to fix this yet but this is an isolated reproduction of the problem. I am going to have a poke around but any guidance on where to go from here would be great.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

### DIFF
--- a/packages/jest-emotion/test/printer.test.js
+++ b/packages/jest-emotion/test/printer.test.js
@@ -1,9 +1,9 @@
 // @flow
 import 'test-utils/legacy-env'
-import React from 'react'
 import renderer from 'react-test-renderer'
 import prettyFormat from 'pretty-format'
-import { css, cx } from 'emotion'
+/** @jsx jsx */
+import { css, jsx, ClassNames } from '@emotion/core'
 import { createSerializer } from 'jest-emotion'
 import { ignoreConsoleErrors } from 'test-utils'
 
@@ -23,8 +23,8 @@ describe('jest-emotion with dom elements', () => {
   it('replaces class names and inserts styles into React test component snapshots', () => {
     const tree = renderer
       .create(
-        <div className={divStyle}>
-          <svg className={svgStyle} />
+        <div css={divStyle}>
+          <svg css={svgStyle} />
         </div>
       )
       .toJSON()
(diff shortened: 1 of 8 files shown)


packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { css, jsx } from '@emotion/core'
+import { createSerializer as createEnzymeSerializer } from 'enzyme-to-json'
+import { createSerializer } from 'jest-emotion'
+
+expect.addSnapshotSerializer(createEnzymeSerializer())
+expect.addSnapshotSerializer(createSerializer())
+
+test('enzyme mount test', () => {
+  const Greeting = ({ children }) => (
+    <div css={css({ backgroundColor: 'red' })}>{children}</div>
```
mitchellhamilton: Not related to this PR but you don't need the css call here, you can pass the style object directly to the css prop.

petegleeson: Ah yep, thanks for updating this

petegleeson:
Unfortunately no chance to look at this today. Will pick this back up tomorrow

packages/jest-emotion/test/printer.test.js
```diff
@@ -94,15 +100,23 @@ describe('jest-emotion with DOM elements disabled', () => {
...
+  let tree = renderer
+    .create(
+      <ClassNames>
+        {({ css: classNamesCss, cx }) => (
+          <div
+            className={cx([
+              'net-42',
+              'net',
+              classNamesCss`
+                color: darkorchid;
```
mitchellhamilton: This doesn’t need to use ClassNames, you can use the css prop here and pass a class name as well and they’ll get concatenated.

petegleeson:
@mitchellhamilton I did some manual testing with these changes in Atlaskit and the snapshots look good from my end 👍

PR #1260, opened 2019-03-07
Adds missing peer dependency in @emotion/styled

arcanis:
**What**:

`@babel/core` is added as a peer dependency.

**Why**:

`@emotion/styled-base` has a peer dependency on `@babel/core`, but this package doesn't provide it - which is [invalid](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0).

**Checklist**:
- [ ] Documentation - n/a
- [ ] Tests - n/a
- [ ] Code complete - n/a

### DIFF
--- a/packages/styled/package.json
+++ b/packages/styled/package.json
@@ -14,6 +14,9 @@
     "@emotion/styled-base": "^10.0.7",
     "babel-plugin-emotion": "^10.0.7"
   },
+  "peerDependencies": {
+    "@babel/core": "^7.0.0"
+  },
   "devDependencies": {
     "dtslint": "^0.3.0"
   },
--- a/packages/styled/package.json
+++ b/packages/styled/package.json
@@ -17,6 +17,11 @@
   "peerDependencies": {
     "@babel/core": "^7.0.0"
   },
+  "peerDependenciesMeta": {
+    "@babel/core": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "dtslint": "^0.3.0"
   },


Andarist:
Hm, this is indeed unfortunate. `@emotion/styled` has a dual purpose - it might be used directly as runtime library or it might be used as babel macro. Only the latter has a peer dep on `@babel/core` but this cannot be expressed accurately with current package.json metadata.

It would be incorrect (maybe not incorrect, but at least not developer friendly) to warn `@emotion/styled` users who never intend to use macro flavor about missing peer dep.

I admit this is some problem that ideally would be addressed - but I'm not sure what's the best way to fix it and provide good DX at the same time 🤔

arcanis:
Yarn supports [optional peer dependencies](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md); would that be ok if I was marking `@babel/core` as optional?

Andarist:
This will still warn users about missing peer dep when using managers which don't understand the meta property, right?

arcanis:
@Andarist Yep. I've tried to work with both npm and pnpm on this and while npm didn't make clear whether they would implement it or not (they considered making peer dependencies optional by default), pnpm emitted an intent to implement as well: https://github.com/yarnpkg/rfcs/pull/105#issuecomment-438259276

mitchellhamilton:
I'm a little confused about this, I can't see a peerDependency in `@emotion/styled-base`'s package.json
https://github.com/emotion-js/emotion/blob/ca3ad1c1dcabf78a95b55cc2dc94cad1998a3196/packages/styled-base/package.json#L1-L43

Andarist:
This is getting resolved in https://github.com/emotion-js/emotion/pulls with help of optional peer dependencies. Thank you for raising awareness about this issue @arcanis !

PR #1261, opened 2019-03-07
Uses require.resolve when calling @emotion/core

arcanis:
**What**:

This PR fixes Emotion with PnP.

**Why**:

The `@emotion/babel-preset-css-prop` plugin is listing `@emotion/core` in its configuration without resolving first using `require.resolve`. It causes the resolution to lose track of the real package making the request, causing access denied errors with PnP.

**How**:

- The package now uses `require.resolve` before sending the module name forward. It's a good practice for babel plugins to do this ([cf in Jest](https://github.com/facebook/jest/blob/master/packages/babel-preset-jest/index.js), for example).

- The `@emotion/core` peer dependency is now properly listed.

**Checklist**:

- [ ] Documentation - n/a
- [ ] Tests - n/a
- [ ] Code complete - n/a

### DIFF
--- a/packages/babel-preset-css-prop/__tests__/__snapshots__/index.js.snap
+++ b/packages/babel-preset-css-prop/__tests__/__snapshots__/index.js.snap
@@ -19,7 +19,7 @@ export let Button = props => {
 
       ↓ ↓ ↓ ↓ ↓ ↓
 
-import { jsx as ___EmotionJSX } from \\"@emotion/core\\";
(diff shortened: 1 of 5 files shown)


Andarist:
I strongly disagree about this (please correct me if I'm wrong though). This PR assumes that this babel plugin is used solely when building apps - which is not necessarily true. This can be used i.e. when building libraries and we can't insert resolved path in such case. Resolving should be left for the tools implementing node resolution algorithm - such as bundlers etc

arcanis:
Pushed a commit to update the snapshot and get a better sense of the changes. I agree that it's more invasive than I'd like, I'll try to find a different way.

PR #1273, opened 2019-03-14
adds failing test for prop containing components using css

petegleeson:
**What**: 👋 @danieldelcore and I found an edge case with snapshot tests. The emotion `ForwardRef` component is not being skipped in cases like:
```jsx
expect(
  <Greeting content={<div css={{ background: 'hi' }}>cool</Greeting>
).toMatchSnapshot();
```
In the above example the forwardRef component around `div` will be serialised. We will take a look at this tomorrow morning and update the PR.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete


mitchellhamilton:
Closed via #1280

PR #1283, opened 2019-03-23
feat: expose EmotionCacheContext

FezVrasta:
**What**:

It exposes the `EmotionCacheContext` from `@emotion/core`, like it already happens with the `ThemeContext` export.

**Why**:

`React#useContext` works only if the whole React Context is provided, right now we only expose `CacheProvider`, which is not enough.

**How**:

I made the bare minimum amount of changes needed to export the pre-existing `EmotionCacheContext`.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

### DIFF
--- a/packages/core/src/context.js
+++ b/packages/core/src/context.js
@@ -4,7 +4,7 @@ import * as React from 'react'
 import createCache from '@emotion/cache'
 import { isBrowser } from './utils'
 
-let EmotionCacheContext: React.Context<EmotionCache | null> = React.createContext(
+export let EmotionCacheContext: React.Context<EmotionCache | null> = React.createContext(
   isBrowser ? createCache() : null
 )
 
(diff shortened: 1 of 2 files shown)


Andarist:
Why is this needed? Is it purely to avoid HOC usage?

If we are going to proceed with this one I would much more like exposing this as a hook, we'd have much better control over usage that way
```js
export const useEmotionContext = () => React.useContext(EmotionCacheContext)
````

FezVrasta:
@Andarist it's to enable usage with Hooks, so that anybody can start experimenting with ways to use Emotion with Hooks rather than render-props or HOCs.

I'll make the change you suggested if you prefer, but I really wanted to make the minimum amount of decisions in order to expose this context.

Please let me know how you prefer to proceed.

mitchellhamilton:
Thanks for the PR but the cache context is intentionally not exposed. The cache context should only be consumed by `withEmotionCache` because it does more than just consume the value, on the server, `withEmotionCache` needs to wrap the element that's rendered with a provider.

Issue #1216, opened 2019-02-08
nest a class selector using the class generated with `css` in Emotion 10


albertstill:
In Emotion@9 the below code makes the anchor underlined:

```js
import React from "react";
import { css } from "emotion";

const link = css`
  color: hotpink;
`;

const paragraph = css`
  color: gray;

  .${link} {
    border-bottom: 1px solid currentColor;
  }
`;

export default class App extends React.Component {
  render() {
    return (
      <div>
        <p className={paragraph}>
          Some text with a<a className={link}> link</a>.
        </p>
      </div>
    );
  }
}
```

I appreciate this would work without any code your side because `css` returns a class string. However in Emotion@10 it seems the `css` function does not handle this:

```js
/** @jsx jsx */
import React from "react";
import { css, jsx } from "@emotion/core";

const link = css`
  color: hotpink;
`;

const paragraph = css`
  color: gray;

  .${link} {
    border-bottom: 1px solid currentColor;
  }
`;

export default class App extends React.Component {
  render() {
    return (
      <div>
        <p css={paragraph}>
          Some text with a<a css={link}> link</a>.
        </p>
      </div>
    );
  }
}
```

In Emotion@10 I'd have to walk into the object returned by `css` and assume the `css-` prefix.

```js
const paragraph = css`
  color: gray;

  .css-${link.name} {
    border-bottom: 1px solid currentColor;
  }
`;
```
I appreciate you are encouraging users not to do this sort of [thing](https://5bb1495273f2cf57a2cf39cc--emotion.netlify.com/docs/nested) in Emotion 9, but would I be right in thinking there would be nothing to lose in supporting this in Emotion 10's tagged template function?

This problem arose for me because I'm working on a [`sass-to-emotion`](https://github.com/DomainGroupOSS/sass-to-emotion) codemod, and because it's something that is supported in Sass there exits prod CSS that operates like above.

albertstill:
Just to clarify something, I appreciate the selectors are unique in emotion due to the hash and don't need to be nested for specificity reasons. A use case would be that we have conditional classes higher up the DOM tree that change styles of nested elements. For example:

```css
.is-small-agent {
   .agent-photo {
     height: 30px;
   }
}

.agent-photo {
   height: 50px;
 }
```

konstantin24121:
We also had this problem because we use nesting to change inner nodes styles in complicated components. This is very strange that SerializedStyles.name return only name without css- prefix

pleunv:
[...]

PR #1286, opened 2019-03-27
Document nested generated classnames

srph:
**What**:
Documents nested generate classnames

**Why**:
Coming from styled-components, I thought this was a basic use-case. When I looked up the docs, I thought there wasn't any way around this. I had to find a similar issue like #1216 to achieve this.

**How**:
N/A

**Checklist**:

- [x] Documentation
- [ ] Tests (N/A)
- [ ] Code complete (N/A)

### DIFF
--- a/docs/nested.md
+++ b/docs/nested.md
@@ -51,3 +51,36 @@ render(
   </div>
 )
 ```
+
+You can also nest generated classnames:
+
+```jsx
+// @live
+/** @jsx jsx */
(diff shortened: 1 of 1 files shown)


Andarist:
This is not something that we recommend to do. In fact, this can break whenever you wrap your app within a `<CacheProvider/>` as you would have to configure a custom `key` as well and that shouldn't be `"css"`.

PR #1289, opened 2019-03-28
The default export of react-native has been removed.

Josema:
The default export of react-native has been removed. We must import it this way in order to make emotion/native works with react-native-web.

Related: 
https://github.com/necolas/react-native-web/issues/1310


louisgv:
Seems like there wasn't any release for April. Latest release was on March 25th, 3 days behind this PR xd....

PR #1292, opened 2019-03-28
Allow jest-emotion matcher to match target by RegExp

alanctkc:
**What**:

The jest-emotion matcher `toHaveStyleRule()` takes an option `target` (added in #1110) that allows developers to target rules in nested components or pseudo-selectors.

It matches the target based on a simple string search and looks like this:

```javascript
expect(div).toHaveStyleRule('color', 'yellow', { target: 'a:hover' })
```

This change lets developers also match the `target` by RegExp:

```javascript
expect(div).toHaveStyleRule('color', 'yellow', { target: /a$/ })
```

**Why**:

In some instances, a simple string search is too naive to target a rule.

For example:

```javascript
const style = css`
  .apple {
    color: red;
  }

  a {
    color: yellow;
  }

  a:hover {
    color: black;
  }
`
```

In this case `{ target: 'a' }` would match all three selectors, potentially causing a failed assertion by matching the wrong rule. Matching with RegExp gives the developer the opportunity to narrow down on the appropriate selector.

**How**:

The underlying util `hasClassNames()` was updated to accept a string or RegExp target, and it does basic type checking on the param to determine whether to do a RegExp test or a string search.

**Checklist**:
- [x] Documentation
- [x] Tests
- [x] Code complete

**A little note**:

I'm coming in green to Flow, so you might double-check my `target?: string | RegExp` type annotation.

### DIFF
--- a/packages/jest-emotion/README.md
+++ b/packages/jest-emotion/README.md
@@ -105,6 +105,44 @@ test('renders with correct styles', () => {
 })
 ```
 
+Use `media` and `target` options to assert on rules within media queries and to target nested components, pseudo-classes, and pseudo-elements.
+
+```jsx
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { matchers } from 'jest-emotion'
+import styled from '@emotion/styled'
+
+// Add the custom matchers provided by 'jest-emotion'
+expect.extend(matchers)
+
+test('renders with correct link styles', () => {
+  const Container = styled.div`
+    font-size: 14px;
+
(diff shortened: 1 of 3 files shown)


alanctkc:
@mitchellhamilton Totally agree. Thanks for the merge for now! Is there any open issue or discussion regarding thoughts around API changes? Not sure if I'll get time to improve it, but I'd like to see it cleaned up enough that I may be able to dive into it at some point. Ergonomics could be a lot better.

PR #1299, opened 2019-03-31
Initial POC and RFC for plugins to modify styles as they are passed into emotion's css prop.

tkh44:
# ⚠️ ⚠️ Don't get too excited for this we still have a series of issues to work out. ⚠️ ⚠️

### DIFF
--- a/packages/core/__tests__/custom-cache.js
+++ b/packages/core/__tests__/custom-cache.js
@@ -5,6 +5,8 @@ import createCache from '@emotion/cache'
 import { jsx, CacheProvider } from '@emotion/core'
 import renderer from 'react-test-renderer'
 
+import styledSystemCss from '@styled-system/css'
+
 function render(ele) {
   return renderer.create(ele).toJSON()
 }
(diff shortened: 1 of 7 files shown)


packages/core/__tests__/custom-cache.js
```diff
@@ -111,3 +111,56 @@ test('emotion plugins - @styled-system/css', () => {
...
+  let cache = createCache({
+    emotionPlugins: [styledSystemCss]
+  })
+
+  let SomeComponentFromNpm = props => {
+    // let's say i'm an author of a library using emotion
+    // i want to apply padding: 2px to my element and that's it
+    return <p css={{ padding: 2 }}>i want this to have padding: 2px</p>
+  }
+  // as a consumer of the above component, i don't want to have to care that it uses emotion
```
Andarist: @mitchellhamilton how is this problem conceptually different from stylis plugins though? this still is some kind of contextual processing in both cases

mitchellhamilton: It isn't conceptually or technically different but it's very different practically. Stylis plugins generally do things like RTL, add syntax or etc. with the important constraint that it's less likely to remove/change how syntax is interpreted.

Andarist: Even with RTL plugin, this can break stuff somewhat easily if a component doesn't know about being wrapped in - because not everything should actually be flipped to the other side and using `/* @noflip */` is necessary in such situations - but the component in question doesn't handle it properly, because it might not be rtl-aware.

mitchellhamilton: Yeah, that's true. I think the main distinction is that stylis plugins are a much more low level thing which not a huge number of people will use whereas this is intended to be used much more widely.

Also, what this is operating on would break a lot of our current assumptions which would mean we couldn't make some of the optimizations we currently do. For example, we assume that `css({ color: 'green' })` is equivalent to `css('color:green;')`, this would mean that that isn't necessarily true.

Andarist:
Closing this for housekeeping purposes as this seems very stale at this point in time and no consensus has been reached when this was initially discussed. We can always reopen this (or just open a new PR/issue) whenever anyone feels like it.

PR #1300, opened 2019-04-02
only changes jsx pragma when a css prop is detected

petegleeson:
**What**:

**Why**:
👋We are currently upgrading our monorepo from emotion 9 to emotion 10. The migration is pretty close to done. One thing we noticed were the bundle sizes of all our packages increasing by the size of `emotion/core`. This increase was happening regardless of whether the package was using emotion at all.

Digging in a bit, it looks like the `@emotion/babel-preset-css-prop` was always changing the pragma and adding a `@emotion/core` import. We only want to include emotion when it is used.

**How**:
The current version of the `@emotion/babel-plugin-jsx-pragmatic` changes the pragma if any JSX is found in the file. I made it a bit more specific and only change the pragma when a `css` prop exists on a component.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete

I don't have full context of the impact of this change. Let me know if there are alternative ways of solving this issue. Thanks.

### DIFF
--- a/packages/babel-plugin-jsx-pragmatic/__tests__/__fixtures__/css-prop.js
+++ b/packages/babel-plugin-jsx-pragmatic/__tests__/__fixtures__/css-prop.js
@@ -1,0 +1,3 @@
+import React from 'react'
+
+export default () => <p css={{ backgroundColor: 'hotpink' }}>hello</p>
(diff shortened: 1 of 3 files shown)


petegleeson:
A quick look at these changes running locally in our repo seems to have fixed it

petegleeson:
Speaking about this with @mitchellhamilton we decided that this isnt the way to go. This fails to capture the usecase of spreading the css prop on a component. There are likely other cases as well. To support emotion in specific packages within a monorepo we are going to do that with specific babel config for those packages. Going to close this PR.

PR #1311, opened 2019-04-11
docs:fix ThemeProvider API jsdoc

CanRau:
**What**:
Think it should read
`Object|Function => Object`
Instead of
`Object|Object => Object`

Right?

**Checklist**:


mitchellhamilton:
Thanks for the PR but it’s currently correct since it’s a function that accepts an object and returns an object rather than a function which accepts a function and returns an object

CanRau:
Thanks for the response, you're right now I get the notation thanks for clarifying 🙏😃

PR #1324, opened 2019-04-23
fix: allow class declaration inside object destructuring

FezVrasta:
**What**:
With the current types, we can't do this:

```jsx
<ClassNames>
  {({ css, redText = css`color: red` }) => (
    <div className={redText}>I'm red</div>
  )}
</ClassNames>
```

**Why**:
My type change make the above syntax allowed.

**How**:
I just added a type wildcard that accepts any property that has a string as value.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete

The practice of defining classes inside the object destructuring is very useful to make the code more concise, without it we are forced to create a new code block, define the variable, and return the JSX with an explicit return.

### DIFF
--- a/packages/core/src/class-names.js
+++ b/packages/core/src/class-names.js
@@ -75,7 +75,7 @@ type Props = {
     cx: (...args: Array<ClassNameArg>) => string,
     theme: Object,
     // This allows to do `({ css, myClass = css`color: red;` }) => `
-    [string]: string,
+    [string]: string
   }) => React.Node
 }
 
--- a/packages/core/src/class-names.js
+++ b/packages/core/src/class-names.js
@@ -73,7 +73,9 @@ type Props = {
   children: ({
     css: (...args: Array<any>) => string,
     cx: (...args: Array<ClassNameArg>) => string,
-    theme: Object
+    theme: Object,
+    // This allows to do `({ css, myClass = css`color: red;` }) => `
+    [string]: string,
   }) => React.Node
 }
 


mitchellhamilton:
Thanks for the PR but I don't think this is something we want to do, `ClassNames` is intentionally hard and annoying to use because you shouldn't be using it very much. The css prop and styled are the API that should be used in 99% of cases when using `@emotion/core`.

FezVrasta:
I understand the rationale, but being a CRA user with no intention to put the custom pragma comment on top of every file, I can't use the `css` property and this forces me (and my team) to rely on `ClassNames` when needed.

Also, there are some legacy components that take a class name as property that can't be used at all with the `css` property ☹️

I beg you to reconsider your decision, but obviously it's up to you.

Cheers!

mitchellhamilton:
>  no intention to put the custom pragma comment on top of every file

I'd be interested to hear the rationale behind not wanting to include the pragma in every file that needs the css prop, I personally do it in many projects and I'm really happy with it. 

> Also, there are some legacy components that take a class name as property that can't be used at all with the css property ☹️ 

If the prop that it accepts is called `className`, you can still use the css prop and the component will get a class with the styles from the css prop.

FezVrasta:
> I'd be interested to hear the rationale behind not wanting to include the pragma in every file that needs the css prop, I personally do it in many projects and I'm really happy with it.

It's way too confusing, people will start wondering why their styles are not applied when they forget to include the pragma. We'll have custom pragmas for files where we don't use `css` anymore due to some refactor, and things will get messy...

> If the prop that it accepts is called `className`, you can still use the css prop and the component will get a class with the styles from the css prop.
I'm referring to legacy components that look for `overlayClassName` or stuff like this.

mitchellhamilton:
> It's way too confusing, people will start wondering why their styles are not applied when they forget to include the pragma.

You can use [the `jsx-import` ESLint rule in eslint-plugin-emotion](https://github.com/emotion-js/emotion/blob/master/packages/eslint-plugin-emotion/docs/rules/jsx-import.md) so people will know when they need it and it'll automatically add it for them.

> We'll have custom pragmas for files where we don't use css anymore due to some refactor, and things will get messy...

What's the harm in having the pragma in files that don't use the css prop?

> I'm referring to legacy components that look for overlayClassName or stuff like this.

👍  Thanks for the clarification, this is exactly the case that ClassNames was made for. Are components like that a really common occurrence? If they are a very common occurrence, you might be better off using the `emotion` package, especially since you're using CRA so you won't get benefits from zero config SSR.

FezVrasta:
> You can use [the `jsx-import` ESLint rule in eslint-plugin-emotion](https://github.com/emotion-js/emotion/blob/master/packages/eslint-plugin-emotion/docs/rules/jsx-import.md) [](/emotion-js/emotion/blob/HEAD@{2019-04-24T12:19:30Z}/packages/eslint-plugin-emotion/docs/rules/jsx-import.md) so people will know when they need it and it'll automatically add it for them.

No I can't with CRA 😓

> What's the harm in having the pragma in files that don't use the css prop?

If I ask my team to add a custom pragma when they use `css`, then I expect the pragma to be removed once the `css` property is not used anymore. Otherwise in case of refactor, migration to different styling system or anything, we'll be confused by why that pragma is needed. It's basically the same reason why you don't keep unused imports in your code.

> 👍 Thanks for the clarification, this is exactly the case that ClassNames was made for. Are components like that a really common occurrence? If they are a very common occurrence, you might be better off using the `emotion` package, especially since you're using CRA so you won't get benefits from zero config SSR.

No it's not very common. I can't use `emotion` because I use a custom cache provider.

Andarist:
I don't believe this change is really type-correct and it only aims to be a sugar thing which helps to avoid extra statements. In addition to that, it could even cause misusage - like this:
https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzAEMMMBTAB3wGV8AnAS2wHMwBeMACgLoC4weTZgEp2APjABvAL7ooAV2wZ8DHGChw4nAB79Ok1GDAAjftnkBbY2ToAaQ2ADaglgF1+L5qmmi2EgG5wDAAmwvyBIVKyqBpa+g7G9kZw+AAWNt6+EgZGJORUtIwsnCnpdMKZqEA

So it's not even giving a type safety. I have to decline this, hope you understand.

PR #1336, opened 2019-04-30
Add Option to Keep CSS Comments in babel-plugin-emotion. Fixes #1325

### LINKED ISSUE #1325: Comments are stripped out before being passed to Stylis plugins, which breaks the stylis-rtl RTL plugin
- `emotion` version: 10.0.7
- `react` version: 16.6.0

**Relevant code:**

```javascript
const myCss = css`
    color: #000;
    /* @noflip */
    padding-right: 100px;
`;
```

**What you did:**
I used the `stylis-rtl` plugin with the Emotion cache and added a `/* @noflip */` comment to prevent a css property from being flipped. 

**What happened:**
The css property was flipped even though I used the required `/* @noflip */` comment to prevent flipping in a certain case. I tested the plugin itself, as well as the underlying cssjanus library and they all work as expected when the comment is used. I noticed that the CSS comment was already stripped out of the source when it was passed into the plugin. 

**Reproduction:**
https://codesandbox.io/s/7j00kmqzmj

**Problem description:**
Emotion seems to strip out the CSS comments before passing the content to the stylis plugins. Some Stylis plugins rely on comments to control their behavior. By stripping out the comments before passing the css to the plugins, that functionality is broken. For example, the `stylis-rtl` plugins relies on a `/* @noflip */` comment, either before a rule declaration, or before a property declaration, to prevent flipping that value. This is important for complex UIs that have several use cases where certain flipping behavior needs to be disabled in rtl languages in order to protect the integrity of the UI.

**Suggested solution:**
The css comments should remain intact when passed into the Stylis plugins, since various plugins depend on certain comments to change their behavior.

Zenwolf:
_[edited to reflect current changes]_

**What**:

Fixes #1325. This PR adds a list of CSS comment values that, when found, keeps the CSS comment instead of removing it when using the `babel-plugin-emotion` package.

**Why**:

Some CSS libraries require the use of comments to control library functionality. Stylis plugins can use these libraries to provide support via Stylis that Emotion uses. One example of this is the [stylis-rtl plugin](https://github.com/FindHotel/stylis-rtl) and the [cssjanus library](https://github.com/cssjanus/cssjanus/).

The cssjanus lib requires the use of comments to control the rtl flipping behavior for individual style rules or properties. This is an example of the comment: `/* @noflip */`.

The babel plugin is very convenient for developers using `babel-preset-css-prop` for the `css` react prop by automatically inserting the jsx pragma: `/* @jsx jsx */`. It would be really helpful if developers using libraries that depend on comments can leverage the plugin instead of having to manually configure every file.

**How**:

In order to support retaining comments, I added an approved list of CSS comment values. When a standard CSS comment contains the special values, the plugin will keep the comment instead of removing it. All other comments are removed.

**Checklist**:
- [ ] Documentation
- [x] Tests -- Updated the comments fixture and snapshot with new test cases.
- [x] Code complete

Thank you!

### DIFF
--- a/packages/babel-plugin-emotion/src/utils/minify-utils.js
+++ b/packages/babel-plugin-emotion/src/utils/minify-utils.js
@@ -2,11 +2,6 @@
 // babel-plugin-styled-components
 // https://github.com/styled-components/babel-plugin-styled-components/blob/8d44acc36f067d60d4e09f9c22ff89695bc332d2/src/minify/index.js
 
-const CSS_ALLOWED_COMMENTS = [
-  // @noflip supports RTL via stylis plugin using cssjanus lib. If the comments
-  // that include this are stripped out, RTL will not work.
-  '@noflip'
-]
 const multilineCommentRegex = /\/\*[^!](.|[\r\n])*?\*\//g
 const lineCommentStart = /\/\//g
 const symbolRegex = /(\s*[;:{},]\s*)/g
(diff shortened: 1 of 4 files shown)


Zenwolf:
Who can point me in the right direction to add a test for this new babel plugin feature?

mitchellhamilton:
I’m not a fan of adding another option like this. I think the better solution would be to keep all block comments that contain `@`  and remove everything else.

After that change, you can add a fixture in packages/babel-plugin-emotion/__tests__/__fixtures__

Zenwolf:
> I’m not a fan of adding another option like this. I think the better solution would be to keep all block comments that contain `@` and remove everything else.
> 
> After that change, you can add a fixture in packages/babel-plugin-emotion/**tests**/**fixtures**

Thank you for the comments, @mitchellhamilton ! If you are not interested in the general option to keep comments, then this sounds easier to do. I'll update the PR as soon as I get the chance.

packages/babel-plugin-emotion/src/utils/minify-utils.js
```diff
@@ -2,6 +2,11 @@
 // babel-plugin-styled-components
 // https://github.com/styled-components/babel-plugin-styled-components/blob/8d44acc36f067d60d4e09f9c22ff89695bc332d2/src/minify/index.js
 
+const CSS_ALLOWED_COMMENTS = [
+  // @noflip supports RTL via stylis plugin using cssjanus lib. If the comments
+  // that include this are stripped out, RTL will not work.
+  '@noflip'
+]
```
mitchellhamilton: Sorry, let me clarify, I meant any block comment that contains `@` anywhere in the comment, they should be kept so `/* @whatever */` and etc. should be kept.

Zenwolf: Sure, no problem. I can make that change.

Zenwolf:
Hi @mitchellhamilton , is there any way to know when a new version of Emotion will be released that contains these changes? Or do I need to do something in order to create a release?

mitchellhamilton:
This change has been included in the past two versions of babel-plugin-emotion

Zenwolf:
@mitchellhamilton Thank you very much! I really appreciate your response.

PR #1357, opened 2019-05-21
docs: better issue templates

FezVrasta:
**What**:
Added (in my opinion) better and easier to follow issue templates, divided by categories.

**Why**:
Managing issues is difficult, a lot of people don't follow the template, probably because it's too long and articulated. Also, we get a lot of "questions" that would better fit on StackOverflow.

**How**:
Added 3 different issue templates

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A


.github/ISSUE_TEMPLATE/--feature-request.md
```diff
@@ -0,0 +1,25 @@
+---
+name: "\U0001F680Feature request"
+about: I have a suggestion (and might want to implement myself)
+title: ''
+labels: enhancement, needs triage
+assignees: ''
+---
+
+<!-- Consider opening a pull request instead: it’s a more productive way to discuss new features -->
```
mitchellhamilton: I don't totally agree with this, it means that someone might spend a lot of time implementing a feature only to be told, "no this doesn't make sense in emotion".
```suggestion
```

PR #1364, opened 2019-05-25
Adds missing peer dependencies

arcanis:
**What**:

This diff adds two peer dependencies to `@emotion/styled`: `@emotion/core`, and `react`.

**Why**:

The `@emotion/styled` package has a hard dependency on `@emotion/styled-base`, which itself has peer dependencies on `@emotion/core` and `react`. It means that `@emotion/styled` should list the peer dependencies as well if possible.

**How**:

I took the same ranges as the ones in `@emotion/styled-base`.

**Checklist**:

- n/a - Documentation
- n/a - Tests
- n/a - Code complete


arcanis:
👌

PR #1375, opened 2019-05-31
add dependency @emotion/core requested by @emotion/styled-base and @e…

marco-silva0000:
…motion/styled-base

There is a missing dependency on this worktree, as react and @emotion/core are listed as a peer dependencies of dependency @emotion/styled-base

this raises a warning when running with yarn2(berry) and will break PNP when using emotion as a dependency.

tagging @arcanis as requested in https://yarnpkg.com/en/docs/pnp/troubleshooting

### DIFF
--- a/packages/styled/package.json
+++ b/packages/styled/package.json
@@ -11,8 +11,10 @@
     "test:typescript": "dtslint types"
   },
   "dependencies": {
+    "@emotion/core": "^10.0.0",
     "@emotion/styled-base": "^10.0.10",
-    "babel-plugin-emotion": "^10.0.9"
+    "babel-plugin-emotion": "^10.0.9",
+    "react": ">=16.3.0"
   },
   "devDependencies": {
     "dtslint": "^0.3.0"


packages/styled/package.json
```diff
@@ -11,8 +11,10 @@
     "test:typescript": "dtslint types"
   },
   "dependencies": {
+    "@emotion/core": "^10.0.0",
     "@emotion/styled-base": "^10.0.10",
-    "babel-plugin-emotion": "^10.0.9"
+    "babel-plugin-emotion": "^10.0.9",
+    "react": ">=16.3.0"
```
FezVrasta: I would expect React to be a devDependency 🤔

marco-silva0000: The warning I got is as follows:
@emotion/styled@npm:10.0.11 doesn't provide @emotion/core@^10.0.0 requested by @emotion/styled-base@npm:10.0.10
@emotion/styled@npm:10.0.11 doesn't provide react@>=16.3.0 requested by @emotion/styled-base@npm:10.0.10

It's styled-base that lists react as a peer dependency, either this package needs to add it or it is there mistakenly..

Andarist: if those entries are needed for PnP they should be peer dependencies as in styled-base - https://github.com/emotion-js/emotion/blob/dc27db398b6ebb2cf009250e02fe3c6f6a3518c5/packages/styled-base/package.json#L27-L30

@arcanis PnP doesn't handle transitive peerDependencies?

arcanis: Nope, we don't. The rational is very technical, but to abstract it consider that all packages in the dependency tree are only ever instantiated once if they don't specify peer dependencies. This is a strong guarantee that we can make and that wasn't possible before PnP (because fs-based hoisting didn't allow it).

If we were to support transitive peer dependencies, we would have to break this promise and potentially instantiate packages without peer dependencies multiple times, and the user (or even the intermediate package!) would have no control over it.

So we decided to be strict and require peer dependencies to be listed at every level (which we were considering anyway together with the folks at pnpm).

Andarist: Cool, thanks for the explanation! @marco-silva0000 if you could change those to peers instead of regulars dependencies then we should be able to merge this.

mitchellhamilton: @emotion/styled already has these peerDependencies becasue of #1364, at the time this PR was opened, that wasn't released but now it has been so this issue should be fixed so I'm going to close this PR.

Issue #1293, opened 2019-03-29
babel-emotion v10 autoLabel not using object props (worked in v9)


mherodev:
- `emotion` version: ^10.0.6
- `react` version: ^16.7.0

**Relevant code:**

```javascript
const Styled = {
  MyDiv: styled.div(() => ({
    color: 'blue',
  })),
};

const MyComponent = () => <Styled.MyDiv>Hello</Styled.MyDiv>;
```

**What you did:**

We collect all of our Styled components into one large `Styled` object so we can separate our styles into their own files. 

**What happened:**

In the previous of Emotion (v9) this would result correctly in classnames like `css-r20001-MyDiv`, but in the latest version we're seeing `css-r20001-Styled` instead, and that's for EVERY component, so tracking down styles has become very arduous. 

**Reproduction:**

Use the above code alongside a babel config that contains:

```jsx
plugins: [['emotion', { autoLabel: true, sourceMap: true }]]
```

If there's a repo with babel-emotion setup I can fork, I'd love to create a repro repo.

mherodev:
Here's a more thorough example:

```jsx
import React from 'react';
import styled from '@emotion/styled';

import './App.css';

const Styled = {
  MyDiv: styled.div(() => ({
    color: 'blue',
    fontSize: '72px',
  })),
};

const MyDiv = styled.div(() => ({
  color: 'red',
  fontSize: '72px',
}));

const Nested = {
  MyDiv,
};

function App() {
  return (
    <div className="App">
      <Styled.MyDiv>Test Styled.MyDiv</Styled.MyDiv>  {/* css-17zrc7r-Styled */}
      <MyDiv>Test MyDiv</MyDiv>                       {/* css-f1i1dm-MyDiv  */}
      <Nested.MyDiv>Nested MyDiv</Nested.MyDiv>       {/* css-f1i1dm-MyDiv  */}
    </div>
  );
}

export default App;

```

I have this reproduced locally and can publish the code if someone would find it useful.

PR #1387, opened 2019-06-06
Feature/babel plugin emotion auto label object properties #1293

mherodev:
**What**:

Adds object property auto label support to babel-plugin-emotion.

**Why**:

If all styled components are nested within a shared object, they will received the same label despite being different components.

See related issue: https://github.com/emotion-js/emotion/issues/1293

**How**:

Added a single, final case before fallbacks that handles object properties.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

This behavior was already present on a previous version of babel-plugin-emotion, so this PR is fixing a regression.

Thanks!

### DIFF
--- a/packages/babel-plugin-emotion/src/utils/label.js
+++ b/packages/babel-plugin-emotion/src/utils/label.js
@@ -44,15 +44,16 @@ function getDeclaratorName(path, t) {
       p.isVariableDeclarator() ||
       p.isFunctionDeclaration() ||
       p.isFunctionExpression() ||
-      p.isArrowFunctionExpression()
+      p.isArrowFunctionExpression() ||
+      p.isObjectProperty()
   )
   if (!parent) {
     return ''
   }
 
+  // we probably have a css call assigned to a variable
+  // so we'll just return the variable name
   if (parent.isVariableDeclarator()) {
-    // we probably have a css call assigned to a variable
-    // so we'll just return the variable name
     return parent.node.id.name
   }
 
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-emotion/src/utils/label.js
```diff
@@ -65,6 +66,11 @@ function getDeclaratorName(path, t) {
     return ''
   }
 
+  // we could also have an object property
+  if (parent.isObjectProperty()) {
+    return parent.node.key.name || parent.node.key.value
```
Andarist: This suggests that there should be 2 tests added - one for each case.

mherodev: The fallback is purely speculative based on related code inside `getIdentifierName` -- I'll just remove it.

mherodev:
It occurs to me, I don't know how to target a package within a mono repo for install, e.g.

`npm install github:emotion-js/emotion.git#7bad392c3dccacb1d9adcaba63438d21030b9fd1`

Is that possible, or do I have to publish my fork to use these changes now?

Andarist:
That's unfortunately not possible right now, which is a real bummer - https://github.com/npm/npm/issues/2974

mherodev:
Thanks @Andarist -- Is there a release schedule?

Andarist:
No strict schedule - @mitchellhamilton has to do a release.

mherodev:
Thanks for the quick merge! @mitchellhamilton

PR #1390, opened 2019-06-07
Add bolt-check

mitchellhamilton:


packages/jest-emotion/package.json
```diff
@@ -23,12 +23,12 @@
     "@emotion/core": "^10.0.10",
     "dtslint": "^0.3.0",
     "emotion": "^10.0.9",
-    "enzyme-to-json": "^3.3.5",
+    "enzyme-to-json": "^3.2.1",
```
Andarist: why the downgrades? 🤔

package.json
```diff
@@ -222,5 +155,159 @@
       "post-commit": "git update-index --again",
       "pre-commit": "lint-staged"
     }
+  },
+  "dependencies": {
+    "bolt-check": "^0.3.0",
+    "@babel/helper-module-imports": "^7.0.0",
```
Andarist: shouldnt those be kept as devDeps? it doesnt matter much for the root package.json but seems like devDep has a better semantic for those

mitchellhamilton: Because it doesn't matter and dependencies is the default.

packages/jest-emotion/package.json
mitchellhamilton: It was done automatically because the root depended on `^3.2.1`. Since nothing is breaking, I'm not going to change it.

PR #1396, opened 2019-06-11
add CacheProvider SSR usage to docs

souporserious:
**What**:
Updated docs about using `CacheProvider` with SSR.

**Why**:
I ran into issues when using `CacheProvider` and SSR specifically with Gatsby. After chatting with @tkh44 on Slack we came to realize it was the fact that I didn't use my own server implementation using `createEmotionServer` and passing the same `cache` through.

**How**:
Added documentation describing what to be aware of when using `CacheProvider` + SSR. I'm happy to expand on this with a code snippet or more information if it seems like it would be more helpful.

### DIFF
--- a/docs/cache-provider.md
+++ b/docs/cache-provider.md
@@ -50,4 +50,4 @@ render(
 
 ## Using CacheProvider with Server-Side-Rendering
 
-Emotion works with SSR out of the box. However, when implmenting `CacheProvider` we must also create our own server implmentation using [`createEmotionServer`](https://emotion.sh/docs/create-emotion-server) and pass it the same `cache` to ensure styles are created and hydrated properly.
+Emotion works with SSR out of the box. However, when implementing `CacheProvider`, we must also create our own server implementation using [`createEmotionServer`](https://emotion.sh/docs/create-emotion-server) and pass it the same `cache` to ensure styles are created and hydrated properly.
--- a/docs/cache-provider.md
+++ b/docs/cache-provider.md
@@ -47,3 +47,7 @@ render(
   </CacheProvider>
 )
 ```
+
+## Using CacheProvider with Server-Side-Rendering
+
+Emotion works with SSR out of the box. However, when implmenting `CacheProvider` we must also create our own server implmentation using [`createEmotionServer`](https://emotion.sh/docs/create-emotion-server) and pass it the same `cache` to ensure styles are created and hydrated properly.


docs/cache-provider.md
```diff
@@ -47,3 +47,7 @@ render(
   </CacheProvider>
 )
 ```
+
+## Using CacheProvider with Server-Side-Rendering
+
+Emotion works with SSR out of the box. However, when implementing `CacheProvider`, we must also create our own server implementation using [`createEmotionServer`](https://emotion.sh/docs/create-emotion-server) and pass it the same `cache` to ensure styles are created and hydrated properly.
```
Andarist: [This README](https://github.com/emotion-js/emotion/tree/a46e2aa507d1da0e1b77cc648acc9cba2cd816e2/packages/create-emotion-server) has to be tweaked too as it mentions only one reason to use this API, but it also should mentioned what you have discovered here.

mitchellhamilton:
This shouldn’t be true and if it is then there’s a bug in emotion or your code. Do you have a reproduction and is this with next by any chance?

souporserious:
@mitchellhamilton my set up is pretty specific right now with Gatsby, but I'll try and get a small repo together to showcase it. It _seems_ to make sense why I need it my case, but maybe I'm just doing something wrong 😅.

I create my own server using my `cache` from `createCache`:

```js
export const {
  extractCritical,
  renderStylesToString,
  renderStylesToNodeStream,
} = createEmotionServer(cache)
```

And then in my `gatsby-ssr` file I import and use `renderStylesToString`:

```jsx

import { ThemeWrapper } from './src/components/Theme'

import { renderStylesToString } from '../packages/server'

export const replaceRenderer = ({
  bodyComponent,
  replaceBodyHTMLString,
  setHeadComponents,
}) => {
  replaceBodyHTMLString(
    renderStylesToString(
      renderToString(<ThemeWrapper>{bodyComponent}</ThemeWrapper>)
    )
  )
}
```

I also have gatsby-browser setup as well:

```jsx
import { ThemeWrapper } from './src/components/Theme'

export const wrapRootElement = ({ element }) => (
  <ThemeWrapper>{element}</ThemeWrapper>
)
```

Should I not have to do this? 🤔 If I don't use `createEmotionServer` it won't use my custom cache options that I set and it falls back to the default settings.

mitchellhamilton:
Gonna close this. @souporserious if you can get a reproduction, please open an issue.

Issue #469, opened 2017-11-21
(emotion-theming) do i use withTheme or not? 


rileylnapier:
reading this article:  it says we no longer need to use withTheme.
https://medium.com/@tkh44/emotion-8-9f892346d0af

![image](https://user-images.githubusercontent.com/7017640/33098955-fe198cda-cec3-11e7-9fa7-906689e257cf.png)

But clicking into the documentation of emotion-theming, it says to use it...
https://github.com/emotion-js/emotion/blob/master/packages/emotion-theming/README.md#api

sooo, do I use it or not?

rileylnapier:
or is that that i need to wrap components with withTheme, just not wrap styled.div``??

probablyup:
You don't need to use `withTheme()` on components created with `react-emotion` like:

```js
const Thing = styled.div`
  color: {p => p.theme.color};
`;
```

However if you want to consume theme properties in a normal React component, that's when you use the `withTheme()` HOC.

```jsx
class MyComponent extends React.Component {
  render() {
    return <div>{this.props.theme.text}</div>
  }
}

const ThemedMyComponent = withTheme(MyComponent);
```

kevinSuttle:
The wording [in the docs](https://emotion.sh/docs/emotion-theming#usage) is a bit confusing still. 

<img width="735" alt="Image 2019-06-04 at 10 43 43 AM" src="https://user-images.githubusercontent.com/95672/58888635-b7a8b400-86b5-11e9-9d68-dd5ed426c4bc.png">

Also, there are no examples of function components here. Is there a reason for that?

Andarist:
> Also, there are no examples of function components here. Is there a reason for that?

No, there is not - it's just the same with class component as it is with function ones.

kevinSuttle:
Thanks, @Andarist. Reason I ask is I was using with Flow, and it didn't pick up the Props. Can open a separate issue if you'd like. 

<img width="1673" alt="Image 2019-06-04 at 2 53 57 PM" src="https://user-images.githubusercontent.com/95672/58905990-fdc33f00-86d8-11e9-8831-59b59c20bccf.png">

Andarist:
Better to open a new issue then

PR #1397, opened 2019-06-12
Fix Flow type for withTheme

SimeonC:
**What**: Flow Typeing for `withTheme` isn't correctly altering props.

**Why**: As mentioned in [comment](https://github.com/emotion-js/emotion/issues/469#issuecomment-498804266) you can't use `withTheme` easily with flow prop types.

**How**: Switching the `React.ComponentType` to `React.AbstractComponent`. To do this also changed it to a normal function as that displays the flow types in a more consise and clear way.

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/packages/emotion-theming/__tests__/with-theme.js
+++ b/packages/emotion-theming/__tests__/with-theme.js
@@ -30,6 +30,7 @@ test(`withTheme(Comp) hoists non-react static class properties`, () => {
   const ComponentWithTheme = withTheme(ExampleComponent)
 
   expect(ComponentWithTheme.displayName).toBe('WithTheme(foo)')
+  // $FlowFixMe
   expect(ComponentWithTheme.someSpecialStatic).toBe(
     ExampleComponent.someSpecialStatic
   )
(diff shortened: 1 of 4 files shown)


SimeonC:
So there is a flow error present, and I'm not sure how to fix this as it's apparently a valid error according to the flow issue: https://github.com/facebook/flow/issues/7612

It looks like this kind of error is due to the `hoistNonReactStatics` not supporting AbstractComponent.
Not sure how I can deal with this error...

mitchellhamilton:
I'd be happy with a `$FlowFixMe`.

SimeonC:
OK, thanks for the confirmation. I've pushed up a fix with the `$FlowFixMe` included.

packages/emotion-theming/__tests__/with-theme.js
```diff
@@ -30,6 +30,7 @@ test(`withTheme(Comp) hoists non-react static class properties`, () => {
   const ComponentWithTheme = withTheme(ExampleComponent)
 
   expect(ComponentWithTheme.displayName).toBe('WithTheme(foo)')
+  // $FlowFixMe
```
FezVrasta: it would have made sense to add an explanation of why this is needed, for future reference.

I guess it's too late.

SimeonC: 'tis never too late! Made a new PR https://github.com/emotion-js/emotion/pull/1399

PR #1402, opened 2019-06-14
ESLint jsx-import rule: Add jsx specifier to existing import from @emotion/core when auto-fixing

ahutchings:
**What**: When auto-fixing a missing `jsx` import, the rule currently adds a new import declaration even when there is already an import declaration from @emotion/core. This change adds the `jsx` specifier to that import when it exists. 

**Why**: To prevent duplicate import declarations from @emotion/core.

**How**:

1. Assign the existing `emotionCoreNode` variable a value when the import is found, instead of only assigning it when the jsx specifier exists
2. Add the jsx specifier after the last specifier of the emotionCoreNode

**Checklist**:

- [ ] Documentation **N/A**
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/polite-lions-visit/changes.json
+++ b/.changeset/polite-lions-visit/changes.json
@@ -1,0 +1,4 @@
+{
+  "releases": [{ "name": "eslint-plugin-emotion", "type": "patch" }],
+  "dependents": []
+}
--- a/.changeset/polite-lions-visit/changes.md
+++ b/.changeset/polite-lions-visit/changes.md
@@ -1,0 +1,1 @@
+jsx-import rule: Add jsx specifier to existing import from @emotion/core when auto-fixing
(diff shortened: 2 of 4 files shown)


ahutchings:
Both cases throw an exception before and after this change due to the [jsx specifier check reading from specifier.imported](https://github.com/emotion-js/emotion/blob/ca3ad1c1dcabf78a95b55cc2dc94cad1998a3196/packages/eslint-plugin-emotion/src/rules/jsx-import.js#L28), which doesn't exist in either case. I can add tests/fixes for both.

In the `import * as` case, the inserted jsx pragma would need to reference the namespace, like `// @jsx Emotion.jsx`. Do you happen to know if that is supported by tooling? If it's questionable we can fall back to adding a duplicate import `import {jsx} from '@emotion/core'`.

mitchellhamilton:
`@jsx Emotion.jsx` is supported by tooling. If you could fix those cases, that would be great!

ahutchings:
Updated with test cases and fixes for the `import *` and default export scenarios. This is ready to go pending review.

Issue #991, opened 2018-11-03
PropsOf helper doesn't understand union prop types


ncknuna:
- `@emotion/styled` version: 10.0.0-beta.7 
- `react` version: 16.6.0 and 16.7.0-alpha.0
- `typescript` version: 3.1.6

Relevant code.

```ts
import * as React from 'react'
import { PropsOf } from '@emotion/styled-base/types/helper'

class UnionPropComponent extends React.Component<{ prop: string } | { otherProp: number }> { }

type props = PropsOf<typeof UnionPropComponent>
```

What you did:

I was trying to get `styled` to work with a React Component with Union types (specifically, Ant Design's Button component. Doc: https://ant.design/components/button/ ; types: https://github.com/ant-design/ant-design/blob/master/components/button/button.tsx#L65). 

What happened:

Type inference failed to infer the types correctly. I tracked it down to the `PropsOf` helper, which will infer `never` for even basic components with union PropTypes, like above ^.

Reproduction:
I tried to reproduce on codesandbox, but I couldn't manage to get the import of PropsOf to work correctly: https://codesandbox.io/s/xnrw73yow

I can reproduce pretty reliably on VSCode though.

Suggested solution:
Unfortunately, I don't think I know TypeScript well enough to suggest a solution :(

Andarist:
Sorry for not responding back sooner. TS types are being reworked right now in https://github.com/emotion-js/emotion/pull/1501 and this helper `PropsOf` is being removed as part of this PR, so this issue won't be actionable on our side anymore. After this PR gets merged in please check if you can handle your use case OK and if not please report a new issue then. Thank you for understanding.

JakeGinnivan:
@ncknuna @Andarist yep, pretty sure this will be fine. There is now a test for props with unions.

After it lands if that isn't the case it may be slightly different to the tests but this is the test:

```
interface PropsA {
  title: string
}
interface PropsB {
  content: string
}

type PropsAB = PropsA | PropsB
class A extends React.Component<PropsAB> {
  static getDerivedStateFromProps(props: PropsAB) {
    return null
  }

  render() {
    return null
  }
}

const B = styled(A)`
  color: red;
`
;<B title="test" />
;<B content="test" />
```

PR #1405, opened 2019-06-18
PropsOf<C> TypeScript utilities - Support defaultProps

ryanswanson:
**What**:
Upgrade the `PropsOf<C>` TypeScript utilities to support `defaultProps` by delegating to new JSX and React types.

**Why**:
The `PropsOf<C>` utilities in _@emotion/styled-base_ and _emotion/theming_ currently result in a TypeScript error when using `styled` or `withTheme` with a component that has a `defaultProp` that is made optional by defaulting the prop value, yet is required on the Props interface and no value is passed to the component instance.

```fsh
Error: /src/emotion/packages/styled-base/types/tests.tsx:288:24
ERROR: 288:24  expect  TypeScript@next compile error:
Property 'prop' is missing in type '{}' but required in type 'Props'.
ERROR: 297:21  expect  TypeScript@next compile error:
Property 'prop' is missing in type '{}' but required in type 'Props'.
```
See tests in this PR for both class-based and function-based component scenarios.

**How**:
The solution provided here uses some newer typings for JSX and React (requiring some TypeScript and _@types/react_ upgrades).

Complexity is reduced by adapting a library-provided interface that is parameterized by both a component and its props (`JSX.LibraryManagedAttributes<C, P>`) to our utility interface parameterized by only a component (`PropsOf<C>`).

The `React.ComponentProps<C>` type utility is the key that allows us to extract Props (P) from the component (C).

The `JSX.LibraryManagedAttributes<C, P>` type utility ensures `props` with associated `defaultProps` are made optional (also see `ReactManagedAttributes` and `Defaultize` utilities in _@types/react_).

The type parameter `C` is constrained to match that of `React.ComponentProps<C>`: `keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>`.

A primary benefit of adapting to standard interfaces is reduced type maintenance over time.  I haven't tested it, but also noticed that Issue https://github.com/emotion-js/emotion/issues/991 may be resolved by this change as well since union prop types should be supported by these standard interfaces.

**Note:** Upon running `yarn test:typescript` with a clean build from master, I encountered TypeScript errors related to `React.ReactNode`'s use of a [union with both null and undefined](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L192) included.  I have suppressed the error by turning off the `no-null-undefined-union` flag in tslint; however, you may prefer a different approach.

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/tasty-hornets-taste/changes.json
+++ b/.changeset/tasty-hornets-taste/changes.json
@@ -1,0 +1,7 @@
+{
+  "releases": [
+    { "name": "emotion-theming", "type": "patch" },
+    { "name": "@emotion/styled-base", "type": "patch" }
+  ],
+  "dependents": []
+}
--- a/.changeset/tasty-hornets-taste/changes.md
+++ b/.changeset/tasty-hornets-taste/changes.md
@@ -1,0 +1,1 @@
+PropsOf<C> TypeScript utilities - Support defaultProps by delegating to new JSX and React types.
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "@changesets/get-github-info": "^0.1.2",
     "@types/jest": "^23.0.2",
     "@types/node": "^10.11.4",
-    "@types/react": "16.3.18",
+    "@types/react": "^16.8.20",
     "babel-check-duplicated-nodes": "^1.0.0",
     "babel-core": "^6.26.3",
     "babel-eslint": "^8.2.3",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,19 +27,19 @@
     "test:typescript": "dtslint types"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.3",
     "@emotion/cache": "^10.0.9",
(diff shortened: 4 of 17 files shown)


ryanswanson:
Thank you @Andarist and @mitchellhamilton for the quick turn-around on this.  @ahutchings and I spent some extra time in making the PR comprehensive hoping it would be easier to review and accept.

Superb work on emotion as well!  Love such libraries ❤️

ryanswanson:
@Andarist - Do you mind sharing when you expect this to make it into a release and what semver level it will likely be?

Andarist:
Not sure, @mitchellhamilton is doing releases - usually, it doesn't take too much time for a new release to be prepared though.

I think it should qualify as patch/minor - we definitely don't consider typing changes to affect semver in the same way as runtime changes affect it.

ryanswanson:
Great!  Thank you

PR #1410, opened 2019-06-24
@emotion/cache readme: fix minor typos

tinymachine:
**What**:
Minor copyedits to readme for @emotion/cache
- Add missing comma before non-restrictive 'which'
- Use gerunds consistently in 'Primary use cases' section
- Fix [comma splices](https://en.wikipedia.org/wiki/Comma_splice) in 'Options' section
- Correct the spelling of 'dynamically'
- Use proper form of possessive 'its'

**Why**:
To align the readme copy with the quality and seriousness of this project's code.

**How**:
By keyboard, mainly! :)

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset  N/A

### DIFF
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -2,7 +2,7 @@
 
 ### createCache
 
-`createCache` allows for low level customization of how styles get inserted by emotion. It's intended to be used with the [`<CacheProvider/>`](https://emotion.sh/docs/cache-provider) component to override the default cache which is created with sensible defaults for most applications.
+`createCache` allows for low level customization of how styles get inserted by emotion. It's intended to be used with the [`<CacheProvider/>`](https://emotion.sh/docs/cache-provider) component to override the default cache, which is created with sensible defaults for most applications.
 
 ```javascript
 import createCache from '@emotion/cache'
@@ -21,7 +21,7 @@ export const myCache = createCache({
 
 - Setting a [nonce](#nonce-string) on any `<style/>` tag emotion creates for security purposes
 
-- Use emotion with a developer defined `<style/>` tag
+- Using emotion with a developer defined `<style/>` tag
 
 - Using emotion with custom stylis plugins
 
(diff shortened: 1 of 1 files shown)


packages/cache/README.md
```diff
@@ -37,12 +37,12 @@ A Stylis plugin or plugins that will be run by stylis during preprocessing. [Rea
...
+Allows changing Stylis' prefixing settings; defaults to `true`. It can be a boolean or a function to dynamically set which properties are prefixed. [More information can be found in Stylis' docs](https://github.com/thysultan/stylis.js#vendor-prefixing).
 
 ### key: string
 
-The prefix before class names, this defaults to `css`. It will also be set as the value of the `data-emotion` attribute on the style tags that emotion inserts and it's used in the attribute name that marks style elements in `renderStylesToString` and `renderStylesToNodeStream`. This is **required if using multiple emotion caches in the same app**.
+The prefix before class names; defaults to `css`. It will also be set as the value of the `data-emotion` attribute on the style tags that emotion inserts and it's used in the attribute name that marks style elements in `renderStylesToString` and `renderStylesToNodeStream`. This is **required if using multiple emotion caches in the same app**.
```
Andarist: a semicolon inside a sentence looks slightly unnatural (at least to me), could we replace it i.e. with "which"?

tinymachine: Yeah, semicolon is an unusual punctuation mark, so I can see how it looks unnatural. Using "which" there could lead to some ambiguity, so it might be better to just replace the semicolons with periods and start a new sentence.

I wonder if it would improve readability to format the option heading to incorporate the default, though?

So instead of...

> ### key: string

...format it as

> ### `key` (string, default: `"css"`)

I noticed the [babel-plugin-emotion readme](https://github.com/emotion-js/emotion/blob/master/packages/babel-plugin-emotion/README.md#sourcemap) handles this by splitting to separate lines, like this:

> ### `key`
> `string`, defaults to `"css"`

And then the rest of the description would follow in a subsequent paragraph.

Andarist: Sounds good as well, could you tweak the PR accordingly?

tinymachine: Happy to. I'll follow the babel-plugin-emotion style (separating the data type and default to their own line), unless I hear otherwise from you.

PR #1414, opened 2019-06-25
Fix for opaque keyframes as object values

### LINKED ISSUE #1259: object style css not working with animations
- `emotion` version: 10.0.7
- `react` version: 16.8.3

**Relevant code:**

```javascript
/** @jsx jsx */
import { jsx, css, keyframes } from "@emotion/core";
import ReactDOM from "react-dom";

const kf = keyframes({
  from: {
    color: "back"
  },
  to: {
    color: "white"
  }
});

const blinkString = css`
  animation-name: ${kf};
  animation-duration: 1s;
  animation-iteration-count: infinite;
  animation-direction: alternate;
`;

const blinkObject = css({
  animationName: kf,
  animationDuration: "1s",
  animationIterationCount: "infinite",
  animationDirection: "alternate"
});

function App() {
  return (
    <div>
      <div css={blinkString}>I blink!</div>
      <div css={blinkObject}>I should blink, but do not!</div>
    </div>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

**What you did:**

Used object style CSS with animations.

**What happened:**

No animation :(

**Reproduction:**

https://codesandbox.io/s/mqoow6yq1j

**Problem description:**

Seems like either there is a problem with animations when using object style or the docs need to be updated to show how to use animations with object style.

**Suggested solution:**

Not sure...

Andarist:
fixes #1259

Let's discuss the fix first, going to add changeset later.

### DIFF
--- a/packages/core/__tests__/keyframes.js
+++ b/packages/core/__tests__/keyframes.js
@@ -113,6 +113,75 @@ cases(
         )
       }
     },
+    'object with animationName and opaque keyframes as value': {
+      render: () => {
+        let animation = keyframes(css`
+          from {
+            color: green;
+          }
+          to {
+            color: hotpink;
+          }
+        `)
+
+        return (
+          <div
+            css={{
+              animationName: animation,
+              animationDuration: '1s'
+            }}
+          >
+            {animation.name}
+          </div>
+        )
+      }
+    },
+    'object with animationName and string keyframes as value': {
+      render: () => {
+        let animation = keyframes(css`
+          from {
+            color: green;
+          }
+          to {
+            color: hotpink;
+          }
+        `)
+
+        return (
+          <div
+            css={{
+              animationName: `${animation}`,
+              animationDuration: '1s'
+            }}
+          >
+            {animation.name}
+          </div>
+        )
+      }
+    },
+    'object with animation and opaque keyframes as value': {
+      render: () => {
+        let animation = keyframes(css`
+          from {
+            color: green;
+          }
+          to {
+            color: hotpink;
+          }
+        `)
+
+        return (
+          <div
+            css={{
+              animation,
+              animationDuration: '1s'
+            }}
+          >
+            {animation.name}
+          </div>
+        )
+      }
+    },
     'multiple keyframes object': {
       render: () => {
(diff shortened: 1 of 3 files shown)


packages/core/__tests__/keyframes.js
```diff
@@ -113,6 +113,74 @@ cases(
...
+              animationName: animation,
+              animationDuration: '1s'
+            }}
+          >
+            {animation.name}
+          </div>
+        )
+      }
+    },
+    'object with animationName and string keyframes as value': {
```
Andarist: this is a test covering previously uncovered (but working) situation

packages/serialize/src/index.js
```diff
@@ -32,7 +32,7 @@ let processStyleValue = (
     case 'animation':
     case 'animationName': {
       if (typeof value === 'string') {
-        value = value.replace(animationRegex, (match, p1, p2) => {
+        return value.replace(animationRegex, (match, p1, p2) => {
```
Andarist: just a short circuit

packages/serialize/src/index.js
```diff
@@ -264,12 +264,22 @@ function createStringFromObject(
...
-          string += `${key}{${handleInterpolation(
+          const interpolated = handleInterpolation(
             mergedProps,
             registered,
             value,
             false
-          )}}`
+          )
+          switch (key) {
+            case 'animation':
```
mitchellhamilton: This case will never happen

packages/core/__tests__/keyframes.js
```diff
@@ -113,6 +113,74 @@ cases(
...
+          }
+          to {
+            color: hotpink;
+          }
+        `)
+
+        return (
+          <div
+            css={{
+              animation
```
mitchellhamilton: This won't do anything, you need to at least also specify the duration of the animation. (hence the previous comment)

Andarist: I can add animationDuration here to reflect better real-world scenario, but unfortunately animation can contain just an animation name and it's a valid CSS:
https://codesandbox.io/s/sad-cookies-oj0wk

PR #1421, opened 2019-06-26
Disallow Theme parameter when it was already parametrized by using CustomStyled

Andarist:
followup to #1398, cc @Ailrun

### DIFF
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -53,7 +53,30 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
 }
 
 type ReactClassPropKeys = keyof React.ClassAttributes<any>
-export interface CreateStyledComponentBase<
+
+interface CreateStyledComponentBaseThemeless<InnerProps, ExtraProps> {
+  <
+    StyleProps extends Omit<
+      Overwrapped<InnerProps, StyleProps>,
+      ReactClassPropKeys
+    > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>,
+    Theme extends object = object
+  >(
+    ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
+  ): StyledComponent<InnerProps, StyleProps, Theme>
+  <
+    StyleProps extends Omit<
+      Overwrapped<InnerProps, StyleProps>,
+      ReactClassPropKeys
+    > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>,
+    Theme extends object = object
+  >(
+    template: TemplateStringsArray,
+    ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
+  ): StyledComponent<InnerProps, StyleProps, Theme>
+}
+
+interface CreateStyledComponentBaseThemed<
   InnerProps,
   ExtraProps,
   StyledInstanceTheme extends object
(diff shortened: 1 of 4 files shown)


packages/styled-base/types/index.d.ts
```diff
@@ -53,41 +53,68 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
...
-  >(
-    template: TemplateStringsArray,
-    ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
-  ): StyledComponent<InnerProps, StyleProps, Theme>
-}
-export interface CreateStyledComponentIntrinsic<
+> =
+  // this "reversed" condition checks if StyledInstanceTheme was already parametrized when using CreateStyled
+  object extends StyledInstanceTheme
+    ? {
```
Andarist: I think both cases here could be unified (the only difference is that 1 version should allow for second type parameter and the other one should have it already baked in), but I'm having slightly hard time to figure out how to do this properly right now

packages/styled-base/types/index.d.ts
```diff
@@ -53,41 +53,68 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
 }
 
 type ReactClassPropKeys = keyof React.ClassAttributes<any>
-export interface CreateStyledComponentBase<
+export type CreateStyledComponentBase<
```
Ailrun: Using `type` instead of `interface` will expose internal types instead of hide details.
For types like this (or types that are complex enough), using `type` will give you complex information when you trying to see the type in editors.

packages/styled-base/types/index.d.ts
```diff
@@ -53,41 +53,68 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
...
   Theme extends object
-> extends CreateStyledComponentBase<JSXInEl[Tag], ExtraProps, Theme> {}
-export interface CreateStyledComponentExtrinsic<
+> = CreateStyledComponentBase<JSXInEl[Tag], ExtraProps, Theme>
+export type CreateStyledComponentExtrinsic<
   Tag extends React.ComponentType<any>,
   ExtraProps,
   Theme extends object
-> extends CreateStyledComponentBase<PropsOf<Tag>, ExtraProps, Theme> {}
+> = CreateStyledComponentBase<PropsOf<Tag>, ExtraProps, Theme>
```
Ailrun: For these two, using type would be better though.

packages/styled-base/types/index.d.ts
Ailrun: One should allow, and one should not allow, so I think it's impossible to unify those cases.

packages/styled-base/types/index.d.ts
Andarist: interfaces cant be conditional though, that's why I've switched to type

Ailrun: How about using two interfaces for each conditional cases to hide complex overloadings?
I mean, just like
```
interface A { ... }
interface B { ... }
type CreateStyledComponentBase =
  cond ?
    A :
    B;
```

Andarist: Out of curiosity - could you specify when exactly (preferably with a simple demo using TS playground) this matters? From my experience interfaces & types are interchangeable nowadays, but maybe I haven't noticed smth.

Ailrun: @Andarist As far as I remember,
```
// file1
import { A } from "something" // A is type
import { B } from "otherthing" // B is interface

const x: A = ...
const y: B = ...

export { x, y }
```

```
// file2
import { x, y } from "file1"

x // IDE will give you internal definition of A
y // IDE will give you the name of B
```

I'm not sure it's still the case. If it's not, please let me know, I will be glad to hear that!

Andarist: I think this has changed, I often see a type alias instead of its content - anyhow we might use interface here as well, doesnt matter (for me). I've amended the PR.

PR #1432, opened 2019-07-12
Attempt to fix rewritten links

Andarist:
Just a showcase of my quick attempt to fix the situation - it fails on GraphQL query though:
<details>
<summary>Error</summary>

```
The GraphQL query from /Users/mateuszburzynski/Desktop/emotion/site/src/templates/doc.js failed.

Errors:
  path is not defined
  
  GraphQL request (5:5)
  4:   doc: markdownRemark(fields: {slug: {eq: $slug}}) {
  5:     htmlAst
         ^
  6:     frontmatter {
  
URL path:
  /docs/install
Context:
  {
    "slug": "install"
  }
Plugin:
  none
Query:
  query DocBySlug(
    $slug: String!
  ) {
    doc: markdownRemark(fields: {slug: {eq: $slug}}) {
      htmlAst
      frontmatter {
        title
      }
    }
    avatar: file(name: {eq: "emotion"}) {
      childImageSharp {
        resolutions(width: 96, height: 96) {
          src
        }
      }
    }
  }
```
</details>

The problem is with links to packages which have a scope in their name because their directories dont have this scope included, but docs URLs have it in the path and we have to map between those to make it work.

I.e. https://emotion.sh/docs/create-emotion#options links to https://emotion.sh/docs/cache#options but in fact it should link to https://emotion.sh/docs/@emotion/cache#options

cc @mitchellhamilton

### DIFF
--- a/site/docs-yaml.js
+++ b/site/docs-yaml.js
@@ -3,8 +3,21 @@ const path = require('path')
 const yaml = require('js-yaml')
 const packageYamlPath = path.resolve(__dirname, '../docs/docs.yaml')
 
-module.exports = () => {
+const loadDocsYaml = () => {
   const yamlString = fs.readFileSync(packageYamlPath).toString()
 
   return yaml.safeLoad(yamlString)
+}
+
+module.exports = loadDocsYaml
+
+module.exports.getPackagesDirs = () => {
+  const packages = loadDocsYaml().filter(({ title }) => title === 'Packages')[0]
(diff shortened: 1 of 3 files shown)


Andarist:
Got fixed by https://github.com/emotion-js/emotion/pull/1903 for the `next` branch. I hope to ship v11 soon and when this happens the main site will benefit from this fix.

PR #1443, opened 2019-07-24
Extract rather than inline styles in Global component when in compat mode

### LINKED ISSUE #1424: extractCritical doesn't extract css from <Global> components
**Current behavior:**

`extractCritical` doesn't extract css from `<Global>` components.

Some applications, like for example Google AMP, cannot have multiple `<style>` tags and/or cannot have `<style>` tags rendered in the `<body>`.

**To reproduce:**

1. Use `extractCritical` to extract the CSS from the React app
2. Inject that CSS in it's own <style amp-custom> in the <head>

**Expected behavior:**

The `<style amp-custom>` tag is fine, but there's an additional tag 
`<style data-emotion-css="xxx">` with the css from the `<Global>` component.

The CSS from `<Global>` should be extracted as well.

**Environment information:**

- `react` version: 16.8.6
- `emotion` version: 10.0.14

I'll add a codesandbox with the production soon (cloudflare seems down right now).

mitchellhamilton:
**What**:
Extract rather than inline styles in Global component when in compat mode

**Why**:
Closes #1424 

**How**:

**Checklist**:

- [ ] Documentation n/a
- [X] Tests
- [x] Code complete
- [x] Changeset


luisherranz:
To understand the PR I was taking a look at what `cache.compat` does, and I noticed that the last line of this comment is probably not right anymore:
https://github.com/emotion-js/emotion/blob/801727b3bb775d70b4d3b537b02c91dfb0891f57/packages/cache/src/index.js#L161-L163

mitchellhamilton:
Yep, that comment is wrong, updated it.

luisherranz:
Thanks @mitchellhamilton.

PR #1444, opened 2019-07-26
Fix "emotion" -> "@emotion/core" typo.

jonearley:


jonearley:
Found another content change tracked in #1445. #1445 also has this #1444 change.

mitchellhamilton:
See https://github.com/emotion-js/emotion/pull/1445/files#r310321327

PR #1445, opened 2019-07-26
Update outdated/deprecated testing reference.

jonearley:
Jest has deprecated `setupTestFrameworkScriptFile` in favor of setupFilesAfterEnv. Updating the docs to be current.


docs/testing.md
```diff
@@ -16,10 +16,10 @@ By default snapshots with emotion show generated class names. Adding [jest-emoti
...
 
-Add the snapshot serializer in your [`setupTestFrameworkScriptFile`](http://facebook.github.io/jest/docs/en/configuration.html#setuptestframeworkscriptfile-string) _or_ at the top of your test file.
+Add the snapshot serializer in your [`setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array) _or_ at the top of your test file.
 
 ```javascript
-import * as emotion from 'emotion'
+import * as emotion from '@emotion/core'
 import { createSerializer } from 'jest-emotion'
 
 expect.addSnapshotSerializer(createSerializer(emotion))
```
mitchellhamilton: Thanks for the PR, it looks like this doc wasn't updated when some changes to jest-emotion were made and this is quite inaccurate. Could you change this section to be more like what's at https://github.com/emotion-js/emotion/tree/master/packages/jest-emotion#snapshot-serializer?

Andarist: Not sure what kind of a change is desired here - should we maybe just duplicate the intro from the linked docs?

docs/testing.mdx
```diff
@@ -16,10 +16,10 @@ By default snapshots with emotion show generated class names. Adding [jest-emoti
 npm install --save-dev jest-emotion
 ```
 
-Add the snapshot serializer in your [`setupTestFrameworkScriptFile`](http://facebook.github.io/jest/docs/en/configuration.html#setuptestframeworkscriptfile-string) _or_ at the top of your test file.
+Add the snapshot serializer in your [`setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array) _or_ at the top of your test file.
 
 ```javascript
-import * as emotion from 'emotion'
+import * as emotion from '@emotion/core'
```
DarkPurple141, line 22: Typescript complains for me using this line as the emotion object with `createSerializer`. (although it also complains with the old syntax). Is `createSerializer` typed properly?

Andarist: It should be typed correctly - could you provide a repro case?

DarkPurple141: Related: 
Screen Shot 2020-06-11 at 8 40 12 pm

DarkPurple141: it's reproducible in this repo: https://github.com/DarkPurple141/jest-emotion-typescript-minimal/blob/master/setup.ts

Andarist: Please always provide the issue **reproduced**. I'm really trying my best to provide help quickly and the linked file doesn't look like the one on your screenshot. I had to see a working file without any error and then jump here back to decipher what you were exactly trying to do. Context switches hurt and slow down.

This proposed change is incorrect - you shouldn't pass emotion to the `createSerializer`. It only accepts those 2 options:
https://github.com/emotion-js/emotion/blob/42714ccc080e74b5a2e44be97b0238c14c78a11a/packages/jest-emotion/README.md#options

I don't think you really should create your own serializer, just use the builtin one.

DarkPurple141: That pattern is from the [offical docs](https://emotion.sh/docs/testing#installation) which is what this PR is ultimately about tackling.

Andarist: Right, I will handle correcting it soon.

PR #1446, opened 2019-07-27
Amended docs to highlight there are 2 approaches to SSR

### LINKED ISSUE #1178: Problems surrounding SSR injection of <style> and unreliability of :first-child selectors

JakeGinnivan:
Also do not warn about nth child selectors if using direct API integrations.

Opening this as a discussion on how to fix #1178. 

I have a question though which will help me finish this PR off properly, I really do not understand the section in the docs with the CacheProvider for migration, if you do this you end up with 2 versions of the cache, the one which is created automatically and the one I create. This would not work with the fix I have implemented afaik?

// @mitchellhamilton

### DIFF
--- a/docs/ssr.md
+++ b/docs/ssr.md
@@ -2,6 +2,10 @@
 title: 'Server Side Rendering'
 ---
 
+Server side rendering in emotion 10 has two approaches, each with their pros and cons. In general approach 1 just works, but impacts usage of certain CSS selectors like nth-child. Approach 2 requires more work to integrate but does not limit your css usage.
+
+## Approach 1 - Adjacent styles
+
 Server side rendering works out of the box in Emotion 10 and above if you're only using `@emotion/core` and `@emotion/styled`. This means you can call React's [`renderToString`](https://reactjs.org/docs/react-dom-server.html#rendertostring) or [`renderToNodeStream`](https://reactjs.org/docs/react-dom-server.html#rendertonodestream) methods directly without any extra configuration.
 
 ```jsx
@@ -11,26 +15,25 @@ import App from './App'
 let html = renderToString(<App />)
 ```
 
-## Using Emotion 10 with the old SSR APIs
-
-It's also possible to use emotion 10 with the SSR APIs for vanilla Emotion. It should only be used for compatibility and migration purposes.
+The rendered output will insert a `<style>` tag above each element with styles for example
 
 ```jsx
-import { renderStylesToString } from 'emotion-server'
-import { cache } from 'emotion'
-import { CacheProvider } from '@emotion/core'
-import { renderToString } from 'react-dom/server'
+const MyDiv = styled('div')({ fontSize: 12 })
+<MyDiv>Text</MyDiv>
 
-let element = (
-  <CacheProvider value={cache}>
-    <App />
-  </CacheProvider>
-)
+// Will render as
 
-let html = renderStylesToString(renderToString(element))
+<style>.css-21cs4 { font-size: 12 }</style>
+<div class="css-21cs4">Text</div>
 ```
 
-## API
+This approach also does not require explicit hydration on the client.
+
+_Warning_ this approach can interfere with nth child selectors as it is inserting additional elements into your markup. You will get a warning if you use such selectors when using approach 1.
+
+## Approach 2 - API Integration
+
+It's also possible to use emotion 10 with the SSR APIs for vanilla Emotion. It requires more work but does not have the limitations of the first approach.
 
 ### renderStylesToString
 
@@ -148,3 +151,22 @@ if (!document.getElementById('root').hasChildNodes()) {
 > Note:
 >
 > The `sheet.speedy` call has to be run before anything that inserts styles so it has to be put into it's own file that's imported before anything else.
+
+## Migration
+
+...
+
+```jsx
+import { renderStylesToString } from 'emotion-server'
+import { cache } from 'emotion'
+import { CacheProvider } from '@emotion/core'
+import { renderToString } from 'react-dom/server'
+
+let element = (
+  <CacheProvider value={cache}>
+    <App />
+  </CacheProvider>
+)
+
+let html = renderStylesToString(renderToString(element))
+```
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -207,7 +207,7 @@ let createCache = (options?: Options): EmotionCache => {
             /(:first|:nth|:nth-last)-child/g
           )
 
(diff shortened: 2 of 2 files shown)


JakeGinnivan:
More than happy to have a quick skype/slack call to discuss the changes.

JakeGinnivan:
Hey @mitchellhamilton thoughts on this approach.

Would need a hand doing tests.

mitchellhamilton:
I'm mostly fine with the code change(there is technically the problem that users of renderStylesTo(String|NodeStream) won't see the message when it's valid for them but I think that's acceptable). For tests, write something like https://github.com/emotion-js/emotion/blob/3672d4044bbbe93b955adca5d588444e1bab983e/packages/styled/__tests__/styled-dom.js#L28-L54 in https://github.com/emotion-js/emotion/blob/master/packages/core/__tests__/compat/server.js
 
I'm uneasy with the docs changes. The primary recommendation should still be to use zero config SSR and not use the given selectors because while this is causing problems for some people, a much larger group are not having problems and by foregrounding this issue so significantly, it means people who just want to know how to get started with emotion have to make even more decisions. Using the emotion-server APIs should be aimed at people who have significant codebases using these selectors and don't want to migrate or want to delay migrating, it shouldn't be recommended for new code(if people really want to use it that way, they can still do that, I'm not aiming to stop that).

While, yes, this is restricting what consumers of emotion can do with css, I don't think that's inherently a bad thing, a significant part of the point of emotion and other css in js libraries is to exactly that by stopping people from writing class names themselves and IMO this is a similar thing, by applying a constraint, we're able to create something that is helpful is a huge way(I know the impact will differ from person to person but personally it's been hugely helpful for me).

JakeGinnivan:
Appreciate the response. 

> The primary recommendation should still be to use zero config SSR and not use the given selectors because while this is causing problems for some people, a much larger group are not having problems and by foregrounding this issue so significantly

I don't mind pushing people towards the zero-config but they should be aware of the trade offs they are making.

> While, yes, this is restricting what consumers of emotion can do with css, I don't think that's inherently a bad thing

I kinda disagree with this point, CSS in JS gives a heap of stuff but nth selectors are really really useful part of CSS and it was really unexpected for it to just not work. @mandymichael may be able to elaborate on this better than I can

davidkpiano:
> While, yes, this is restricting what consumers of emotion can do with css, I don't think that's inherently a bad thing

A styling SSR strategy that does not render styles **exactly the same** as it would without SSR is an SSR strategy that 🚫 **does not work** 🚫. Full stop.

It doesn't matter if it works for most people/apps and not for others. If it doesn't work for 100% of enabled use-cases, is not a workable strategy.

If https://github.com/emotion-js/emotion/issues/1178 won't be resolved another way (which it should be), I strongly believe this _needs_ to be included in the documentation, or hours will be wasted trying to figure out an alternate SSR strategy or refactoring their styles in a less-than-desirable way (all the ways listed in #1178).

JakeGinnivan:
I have opened https://github.com/emotion-js/emotion/pull/1483 to get the warning fixed, will make another pass at the docs and remove code changes in this PR tomorrow

JakeGinnivan:
Going to open a new PR

PR #1450, opened 2019-07-30
Split dependencies and devDependencies correctly

constgen:
**What**:
As `babel-plugin-emotion` is not used in runtime it should be listed in `"devDependencies"`.

**Why**:
There are several reasons.
**The first.** The documentation of this plugin recommends to install it to Dev Dependencies https://www.npmjs.com/package/babel-plugin-emotion

```
npm install --save-dev babel-plugin-emotion
```
**The second.** In my case it produces the common issue with ESLint when there are several instancies of @babel/types (https://github.com/babel/babel-eslint/issues/681#issuecomment-493526294)

_This is my output of `npm list @babel/types` where it is clear why it happens in my project._

```
+-- @atomspace/eslint@3.0.2
| `-- @neutrinojs/eslint@8.3.0
|   `-- babel-eslint@8.2.6
|     +-- @babel/traverse@7.0.0-beta.44
|     | +-- @babel/generator@7.0.0-beta.44
|     | | `-- @babel/types@7.0.0-beta.44
|     | +-- @babel/helper-function-name@7.0.0-beta.44
|     | | +-- @babel/helper-get-function-arity@7.0.0-beta.44
|     | | | `-- @babel/types@7.0.0-beta.44
|     | | +-- @babel/template@7.0.0-beta.44
|     | | | `-- @babel/types@7.0.0-beta.44
|     | | `-- @babel/types@7.0.0-beta.44
|     | +-- @babel/helper-split-export-declaration@7.0.0-beta.44
|     | | `-- @babel/types@7.0.0-beta.44
|     | `-- @babel/types@7.0.0-beta.44
|     `-- @babel/types@7.0.0-beta.44
`-- react-select@3.0.4
  `-- @emotion/css@10.0.14
    `-- babel-plugin-emotion@10.0.14
      `-- @babel/helper-module-imports@7.0.0
        `-- @babel/types@7.5.5
```
`react-select` is installed as `"dependency"`. So it should not install anything related to Babel

**How**:
Update dependenciea in `pacakge.json` of `@emotion/css` module

**Checklist**:

- [ ] Documentation (N/A)
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -12,11 +12,11 @@
   },
   "dependencies": {
     "@emotion/serialize": "^0.11.8",
-    "@emotion/utils": "0.11.2",
-    "babel-plugin-emotion": "^10.0.14"
+    "@emotion/utils": "0.11.2"
   },
   "devDependencies": {
-    "dtslint": "^0.3.0"
+    "dtslint": "^0.3.0",
+    "babel-plugin-emotion": "^10.0.14"
   },
   "publishConfig": {
     "access": "public"


Andarist:
> The first. The documentation of this plugin recommends to install it to Dev Dependencies https://www.npmjs.com/package/babel-plugin-emotion

The recommendation is about consumers - like you. In case of our core package this is different - we need to install this dependency for our consumers (so this doesn't require any extra setup) for our babel macros to work.

constgen:
Ok. Do I understand right that `@emotion/css` should be in `"devDependencies"` of my dependencies to not be installed in my project? Because this is what exactly breaks ESLint. You know, ESLint never resolves paths to its plugins and parsers. It always consumes modules by names from the very top level. 
To fix this I expect `react-select` not to install `@babel/helper-module-imports`. In your point of view, in which repository should this be fixed?

Andarist:
> Ok. Do I understand right that @emotion/css should be in "devDependencies" of my dependencies to not be installed in my project

Only if it's really a dev dep of your dependency. That means - if it's not used at runtime by your dependency.

I must say that I don't fully understand the problem at hand - but it really seems to be an issue of ESLint rather than emotion or your dependency using emotion. Multiple versions of the same package are "expected" to pop up in your node_modules and ESLint should handle this gracefully.

Also - maybe upgrading your babel-eslint would fix this for you? I see you are on v8 which is using some outdated babel dependencies.

constgen:
Before I cancel this PR, can you definitely say that these changes are not safe and will brake everything?

Andarist:
Yes, this change here is a breaking change for macro users.

PR #1477, opened 2019-08-20
jest-emotion: Fix snapshots when using fragments & deep mode

### LINKED ISSUE #1463: jest-emotion: TypeError: node.getAttribute is not a function
**Current behavior:**

On some specific React components/trees, the jest-emotion serializer throws an error: `TypeError: node.getAttribute is not a function`.

If I comment out the `jest-emotion` serializer from `setupTests.js` the snapshot is written properly.

**To reproduce:**

1. Clone https://github.com/quid/refraction
2. `cd refraction`
3. `yarn`
4. `git checkout 6b7a3f0746510779191d5547f357d9afca99227b`
5. `cd packages/expandable-table`
6. `yarn test`

**Expected behavior:**

A snapshot should be written.

**Environment information:**

- `react` version: 16
- `emotion` version: 10.0.15

liamcmitchell-sc:
**What**: Fix snapshots when using fragments & deep mode

**Why**: To fix #1463

**How**: Modifying getNodes to handle arrays

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset 

This required adding a deep serializer. I couldn't think of a cleaner way to have both deep and shallow serializers in the same test file.


packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -98,3 +111,20 @@ test('enzyme test with prop containing css element with other label', () => {
   )
   expect(tree).toMatchSnapshot()
 })
+
+test('enzyme test with prop containing css element in fragment', () => {
+  useDeepSerializer = true
```
Andarist: couldn't matcher be extended differently (or used differently?) to avoid this beforeEach etc dance?

liamcmitchell-sc: The trouble is that matchers are global in the test file and there is no API for them to be removed.

If it was added within this test there is a danger that someone will add another test after and not intend for it to use the deep serializer.

If tests need to use both modes, they have to be switched by modifying the test function and that requires a before/afterEach to reset the mode between tests.

If the existing tests can be converted to using deep mode, we don't have the problem. I just didn't want to touch existing tests.

Andarist: cant a custom matcher use snapshots API? we could maybe just extend expect with separate `.toMatchDeepSnapshot` or smth like this, not sure if this is possible though

liamcmitchell-sc: Doesn't look any easier.

Andarist: Wouldn't using such approach - https://jestjs.io/docs/en/expect#custom-snapshot-matchers - be a cleaner solution?

FezVrasta: @Andarist does it mean someone should have knowledge of what's a deep snapshot? I honestly had no clue of why my tests were failing, I just did expect them to work.

Andarist: It certainly shouldn't crash. I'm not using enzyme, so I might be missing something here, but... I'm not talking here about adding some extra public API - which mode of the serializer you are using in your app is a separate thing from how our tests here are structured. And for the latter, we could just extend jest with 2 different matchers to have clean APIs for testing both modes of that enzyme serializer used in combination with emotion.

liamcmitchell-sc: The problem is serialization, not matchers. Serialization relies on globally installed serializers, regardless of how they are used in matchers.

If we don't want to touch the global serializer, we cannot test using `expect().toMatchSnapshot()`. Instead we need to figure out the internal jest serializer API, create a local serializer and test with something like `expect(localSerialize(wrapper)).toBe('x')`.

It only needs to be added to a custom matcher if it needs to be reused.

Andarist: I've create a proof of concept of what I had in mind - you can check it out here: https://github.com/liamcmitchell-sc/emotion/pull/1

Andarist:
I've decided to use my approach for testing that deep mode to stop stalling this PR. This ultimately boils down to personal preference, both approaches work. Hope you don't mind.

Thank you for diving into this and fixing the problem!

liamcmitchell-sc:
No worries, thanks for merging.

Issue #1354, opened 2019-05-20
TypeError: Cannot read property 'value' of undefined


amykapernick:
- `emotion` version: `10.0.9` (was an issue on 9, upgraded to attempt to troubleshoot
- `react` version: `16.8.6
- `gatsby` version: `2.1.15`

**Relevant code:**
`/src/components/header/style.js`
```javascript

export const Nav = styled('nav')`
	display: flex;
	flex-direction: row;
	justify-content: space-between;
	height: auto;
	overflow-x: scroll;
	-webkit-overflow-scrolling: touch;
	max-width: 100vw;
	z-index: 20;

	&:after {
		animation: ${scrolling} 2s linear infinite;
		color: ${props => props.theme.colours.primary.main};
		/* color: ${props => props.theme.colours.primary.main}; */
		content: '\00BB';
		font-size: 2em;
		position: absolute;
		right: 10px;
		bottom: 13%;
	}

	@media (min-width: ${props => props.theme.values.screens.small}) {
		display: flex;
		flex-direction: row;
		margin-left: auto;
		padding: 0 10px;
		overflow-x: visible;

		&:after {
			content: normal;
		}
	}

	@media (min-width: ${props => props.theme.values.screens.medium}) {
		margin-left: 0;
		margin-top: 3vw;
	}

	@media (min-width: ${props => props.theme.values.screens.large}) {
		margin-top: 1vh;
		padding: 40px 20px 10px;
	}

	/* @media (min-width: ${props => props.theme.values.screens.large}) {
		margin-top: 1vh;
		padding: 40px 20px 10px;
	} */
`

```

**What you did:**
I did nothing, tried to do some work on an existing project and immediately got this error when running `npm start`. Then tried upgrading to `v10` to see if that fixed the issue but still got this error.

Troubleshooting with this file, I commented everything out until it built, then uncommented until I ran into an issue, found the issue was 2 lines.
```color: ${props => props.theme.colours.primary.main};``` and ``` @media (min-width: ${props => props.theme.values.screens.large}) {```

When I tried re-writing the lines it works, but styles after the commented lines don't come through. (eg. on line 107, the colour line I re-wrote is fine, but the rest of it doesn't come through). If I try moving or re-writing those lines it still throws an error.

**What happened:**

```
./src/components/header/style.js
Module build failed (from ./node_modules/gatsby/dist/utils/babel-loader.js):
TypeError: Cannot read property 'value' of undefined
[...]

PR #1478, opened 2019-08-21
Add warnings about using illegal escape sequances

Andarist:
More context - https://github.com/emotion-js/emotion/issues/1354#issuecomment-517914040


Andarist:
Not sure why flow is failing on CI, my PR shouldn't affect that.

PR #1483, opened 2019-08-27
fix: Remove warn when using nth selector with extract critical

JakeGinnivan:
**What**: Removing the nth selector warn when using extract critical. I am separating the docs changes because this change will be useful separate to improving the docs.
Relates to #1178

**Why**: The warning is incorrect

**How**: 

**Checklist**:
- [x] Documentation (NA)
- [x] Tests
- [x] Code complete
- [x] Changeset 

Failing test before the change:
image

### DIFF
--- a/.changeset/ten-bulldogs-kick/changes.json
+++ b/.changeset/ten-bulldogs-kick/changes.json
@@ -1,0 +1,7 @@
+{
+  "releases": [
+    { "name": "@emotion/cache", "type": "patch" },
+    { "name": "emotion-server", "type": "patch" }
+  ],
+  "dependents": []
+}
--- a/.changeset/ten-bulldogs-kick/changes.md
+++ b/.changeset/ten-bulldogs-kick/changes.md
@@ -1,0 +1,1 @@
+nth selector will no longer warn when using extract critical
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -211,7 +211,7 @@ let createCache = (options?: Options): EmotionCache => {
             /(:first|:nth|:nth-last)-child/g
           )
 
-          if (unsafePseudoClasses) {
+          if (unsafePseudoClasses && cache.compat !== true) {
             unsafePseudoClasses.forEach(unsafePseudoClass => {
               const ignoreRegExp = new RegExp(
                 `${unsafePseudoClass}.*\\/\\* ${flag} \\*\\/`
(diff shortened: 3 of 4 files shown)


JakeGinnivan:
@mitchellhamilton any ideas on the failing test?

mitchellhamilton:
The element needs to be wrapped in a CacheProvider

Andarist:
`emotion-server` is using the default cache by default though, so it shouldn't be necessary to run those tests against a custom cache (which can be provided by CacheProvider), right?

mitchellhamilton:
The `emotion` package's cache should be provided to the cache provider
```jsx
import { cache } from 'emotion'
import { CacheProvider } from '@emotion/core'

<CacheProvider value={cache}><TheThing/></CacheProvider>
```

JakeGinnivan:
The tests in packages/emotion-server/test/index.test.js don't wrap in cache provider, so it works. I would have assumed that if you use the built in extractCritical that emotion would just use the default cache for everything?

Just wondering why renderToString doesn't need the cache provider?

I could try and fix this issue too if I understand more

mitchellhamilton:
The components in those tests are wrapped in CacheProviders, https://github.com/emotion-js/emotion/blob/ed1eb899f25b7f6aaec004967f8c7a50f3406541/packages/emotion-server/test/util.js#L109

JakeGinnivan:
Ah I see, should we throw if it's not wrapped in a cache provider?

I kinda think it should be one or another. Either it just uses the default cache if one isn't provided, or throw saying a cache has not been provided?

And how does the default work? Or it doesn't rely on a cache?

JakeGinnivan:
Thanks for your help @mitchellhamilton 

I have wrapped in the cache provider. Would love to understand my questions above though if you have time. It would be useful context when writing the updated docs taking in your feedback

JakeGinnivan:
Hey @mitchellhamilton how is this looking now? Would be great to get it released, it is blocking our upgrade to emotion 10 due to heaps of warnings :)

Once it's merged I can work on updating the docs

mitchellhamilton:
> Ah I see, should we throw if it's not wrapped in a cache provider?
> 
> I kinda think it should be one or another. Either it just uses the default cache if one isn't provided, or throw saying a cache has not been provided?
> 
> And how does the default work? Or it doesn't rely on a cache?

It does use a default cache when one isn't provided but it will write the styles in inline style tags so if you want to use it with emotion-server you need to provide the same cache that emotion-server uses

PR #1489, opened 2019-09-02
Add support for custom values

miklschmidt:
**What**:
Adds support for custom values in your css.

**Why**:
This allows you to make utility classes for handling things such as colors or units.
A handy usecase for this feature is using the `color` package, making color manipulation a breeze.
```js
import Color from 'color';

class MyColorUtility extends Color {
  valueOf() {
    return this.toString('rgb');
  }
}
const MyDiv = styled.div({
  background: new MyColorUtility(new Color("#fff").darken(.4).fade(.5)),
});
```

My specific use case is a lot more complicated, involving theme color variations, runtime color editing and luminance based color switching through global css variables.

**How**:
This feature was achieved by checking for objects in `createStringFromObject` and whether those object return a string from 'valueOf()'. The resulting string is then used as the property value.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset 

 
This feature might be covered by #1299?

I'm not very familiar with flow, the types probably needs updating. I'm well versed in typescript, so if this request gets accepted, i'll fix that part as well as the documentation.

I tried generating a changeset, but i'm not really sure the output is correct, so i didn't commit it.

Let me know if there's anything i should do differently!

### DIFF
--- a/packages/core/__tests__/__snapshots__/css.js.snap
+++ b/packages/core/__tests__/__snapshots__/css.js.snap
@@ -47,6 +47,16 @@ exports[`handles camelCased custom properties in object styles properly 1`] = `
 />
 `;
 
+exports[`handles valueOf on custom object values 1`] = `
+.emotion-0 {
+  color: #F00;
+}
+
+<div
+  className="emotion-0"
+/>
+`;
+
 exports[`label in css call 1`] = `
 .emotion-0 {
   color: hotpink;
--- a/packages/core/__tests__/css.js
+++ b/packages/core/__tests__/css.js
@@ -219,3 +219,25 @@ test('handles camelCased custom properties in object styles properly', () => {
 
   expect(tree.toJSON()).toMatchSnapshot()
 })
+
+test('handles valueOf on custom object values', () => {
+  class ColorUtil {
+    color = null
+
+    constructor(color: string) {
+      this.color = color
+    }
+    valueOf() {
+      return this.color
+    }
+  }
+
+  const tree = renderer.create(
+    <div
+      css={{
+        color: new ColorUtil('#F00')
+      }}
+    />
+  )
+  expect(tree.toJSON()).toMatchSnapshot()
+})
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -269,20 +269,36 @@ function createStringFromObject(
             )};`
           }
         } else {
+          let customValue = value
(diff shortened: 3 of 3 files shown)


Andarist:
Why cant u just stringify ur objects on ur own instead of relaying on the emotion to do that for u?

miklschmidt:
@Andarist because that rules out composition and programatic style manipulation. Preferably you don't want to stringify before sending off the variables to emotion.

Andarist:
Could you give an example of what do you mean? In a given example:
```js
const MyDiv = styled.div({
  background: new MyColorUtility(new Color("#fff").darken(.4).fade(.5)),
});
```
I don't see how you would like to manipulate it further as you hand out that object directly to emotion here, so don't see how stringifying in a place like this wouldn't be appropriate.

miklschmidt:
@Andarist i'm sorry, that was a very contrived example which doesn't explain my point very well, my bad, i'll try again.

Imagine MyColor being Color with valueOf() returning a color string (in my case a css variable).
**styles/colors.js**
```js
export const danger = MyColor('#f00');
export const success = MyColor('#0f0');
export const info = MyColor('#0f0');
```

**styles/buttons.js**
```js
import {success, danger, info} from './colors';
export const normalButtonColorMixin = {
  background: success,
  borderColor: success.darken(.2)
}
```

**button.js**
```js
import {normalButtonColorMixin} from './styles/buttons';

export const Button = styled.button(normalButtonColorMixin, {
  fontSize: 12,
  padding: '4px 7px',
  color: normalButtonColorMixin.luminosity() > 0.5 ? 
         normalButtonColorMixin.background.darken(.2) :
         normalButtonColorMixin.background.lighten(.2)
})

```

This way, if i change success in `styles/colors.js`, all my button variations will still look right. This is still fairly simplified, but i hope it clarifies my use case :)

miklschmidt:
Just happened to come across issue #292, this would solve that as well, without having designers dealing with .toString() at the right time :)

Andarist:
I encourage you to use theming or your custom wrapper around `styled`. It's first time somebody has asked for something like this and this doesn't warranty including this into core lib yet - especially that this has some costs in bundle size & runtime speed.

miklschmidt:
A custom wrapper was my fallback, i wrote it yesterday. Theming unfortunately doesn't solve anything. But thanks for your feedback, i will close this!

PR #1499, opened 2019-09-11
Add useTheme hook for emotion-theming

tkh44:
**What**:
Add `useTheme` hook to emotion-theming

**Why**:
Sometimes this is more convenient for theming than relying on the functional argument in the css prop. This is especially true with object styles or projects that have existing styles and are moving to theming. Instead of refactoring every css call to be a function this is a much simpler update.

**How**:

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


packages/emotion-theming/src/use-theme.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+import * as React from 'react'
```
mitchellhamilton: Doing 👇 will mean that emotion-theming won't break for people who aren't on `>=16.8.0`

```suggestion
import React from 'react'
```

docs/theming.mdx
```diff
@@ -10,6 +10,8 @@ npm install -S emotion-theming
 
 Add `ThemeProvider` to the top level of your app and access the theme with `props.theme` in a styled component or provide a function that accepts the theme as the css prop. The api is laid out in detail [in the documentation](https://emotion.sh/docs/emotion-theming).
 
+### With css prop
```
mitchellhamilton: The example here uses styled too

packages/emotion-theming/src/use-theme.js
Andarist: @mitchellhamilton `import * as React from 'react'` was always supported, this ain't nothing to do with react@16.8. And also - the wildcard is somewhat more correct than the default export. Or maybe you were referencing something I'm not aware of.

mitchellhamilton: I was thinking that it would fail for users who aren't on `>=16.8.0` because rollup would change `import * as React from 'react`' to `import { useContext } from 'react'` and that export wouldn't exist so it would fail. Though I guess since React is CJS, it wouldn't have failed? I'd still rather leave it as `import React from 'react'` though just in case someone uses an ESM bundle of React and aliased it or rollup users with rollup-plugin-commonjs because of the `namedExports` option or something like that.

Andarist: Interesting edge cases. I guess technically they are possible so yeah - let's keep the default for now.

PR #1501, opened 2019-09-12
Restructured types to greatly reduce number of types created during c…

### LINKED ISSUE #1496: Emotion 10 greatly increases number of types during TypeScript compilation
I am opening this to start a discussion, I plan on debugging it and trying to figure out why but upgrading to emotion 10 increased our project compilation time from ~30 seconds to ~70seconds. Any ideas/help would be great though.

Got these stats by just running `yarn tsc --diagnostics`

**Current behavior:**
Emotion 9:
```
Files:          2237
Lines:        308745
Nodes:       1049322
Identifiers:  331530
Symbols:     1031128
Types:        232302
Memory used: 750896K
I/O read:      1.47s
I/O write:     0.00s
Parse time:    5.07s
Bind time:     1.42s
Check time:   22.07s
Emit time:     0.00s
Total time:   28.55s
```

Emotion 10:
```
Files:           2246
Lines:         313024
Nodes:        1056772
Identifiers:   333322
Symbols:      1393179
Types:         852685
Memory used: 1185914K
I/O read:       1.94s
I/O write:      0.00s
Parse time:     5.67s
Bind time:      1.34s
Check time:    52.39s
Emit time:      0.00s
Total time:    59.40s
```

**To reproduce:**

I wanted to rule out our codebase, so have created a repo project at https://github.com/JakeGinnivan/emotion-types-sandbox

Emotion 10, with 1 styled component:
https://github.com/JakeGinnivan/emotion-types-sandbox

```
Files:           110
Lines:         84080
Nodes:        285703
Identifiers:   99654
Symbols:       67179
Types:         11015
Memory used: 115850K
I/O read:      0.08s
I/O write:     0.00s
Parse time:    0.97s
Bind time:     0.45s
Check time:    0.68s
Emit time:     0.00s
Total time:    2.09s
```

Emotion 9
https://github.com/JakeGinnivan/emotion-types-sandbox/tree/emotion9

```
Files:           108
Lines:         83784
Nodes:        284068
Identifiers:   99089
Symbols:       63864
Types:          2793
Memory used: 111337K
I/O read:      0.10s
I/O write:     0.00s
Parse time:    0.92s
Bind time:     0.35s
Check time:    0.18s
Emit time:     0.00s
Total time:    1.46s
```

Baseline (without emotion):
```
Files:           104
Lines:         83593
Nodes:        283283
Identifiers:   98776
Symbols:       62954
Types:          1206
Memory used: 110732K
I/O read:      0.12s
I/O write:     0.00s
Parse time:    0.92s
Bind time:     0.40s
Check time:    0.12s
Emit time:     0.00s
Total time:    1.43s
```

Styled components:
```
Files:           111
Lines:         93595
Nodes:        303502
Identifiers:  105234
Symbols:       71068
Types:          4978
Memory used: 125100K
I/O read:      0.14s
I/O write:     0.00s
Parse time:    1.01s
Bind time:     0.41s
Check time:    0.26s
Emit time:     0.00s
Total time:    1.68s
```

Summary:
Baseline (just react): 1206 types (1.43s @ 110,732K memory used)
Baseline + 1 emotion 9 styled component: 2793 types (1.46s @ 111,337K memory used)
Baseline + 1 emotion 10 styled component: 11,015 types (2.09s @ 115,850K memory used)
Baseline + 1 styled-components 4.3.2: 4978 types (1.68s @ 125,100K memory used)

**Expected behavior:**

The emotion TypeScript types need to be simplified

**Environment information:**

- `react` version: 16.9.0
- `emotion` version: 10.0.17
- `typescript` version: 3.6.3

JakeGinnivan:
…ompilation

BREAKING CHANGE: There are a few breaking changes, see below

* withTheme now infers types properly and may require removing the manually specified generic parameter
* The Theme generic parameter has been removed from a number of types exported from `@emotion/styled` and `@emotion/styled-base`.
* Introduced new CreateThemedStyled type which is exported from emotion-theming, use type to create your own themed `styled` export. See updated docs
* WithTheme should be imported from emotion-theming
* CreateStyledComponentExtrinsic, CreateStyledComponentIntrinsic and CreateStyledComponentBase all have been replaced with CreateStyledComponent

Fixes #1496

**What**:
This PR simplifies a number of types to try and speed up compilation of TypeScript projects when using emotion

**Why**:
Our project over doubled compilation time when we upgraded from emotion 9 to emotion 10. This brings compilation time back down to emotion 9 times.

**How**:

I reduced complexity in the types by removing mapped typeds.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

Want to get initial feedback, I have consumed all these types in our large project and have gone through a few iterations of fixing issues already.


packages/serialize/types/index.d.ts
```diff
@@ -63,7 +61,19 @@ export interface ObjectInterpolation<MP>
   extends CSSPropertiesWithMultiValues,
     CSSPseudos<MP>,
     CSSOthersObject<MP> {}
-export type FunctionInterpolation<MP> = (mergedProps: MP) => Interpolation<MP>
+export interface FunctionInterpolation<MP> {
```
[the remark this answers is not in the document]
Andarist: I'm OK with forbidding functions returning functions, but seems like returning arrays should be allowed. Such arrays don't have to contain functions though, so maybe it's just a matter of splitting the declarations into two - `Interpolation` & `LeafInterpolation`. This way `ArrayInterpolation` could output different set of things and you could prevent the whole thing from being recursive. 

With TS supporting recursive type aliases soon such recursive patterns will be more and more common - so they are not inherently bad on their own 🤷‍♂ Seems like in here we might want to try to avoid them, but seems like a workaround for TS perf problems.

packages/emotion-theming/types/index.d.ts
```diff
@@ -14,17 +18,43 @@ export function ThemeProvider<Theme>(
...
+React.FC<Omit<P, 'theme'> & { theme?: any }>
 
 export interface EmotionTheming<Theme> {
   ThemeProvider(props: ThemeProviderProps<Theme>): React.ReactElement
-  withTheme<C extends React.ComponentType<any>>(
-    component: C
-  ): React.SFC<AddOptionalTo<PropsOf<C>, 'theme'>>
+  withTheme<P extends object>(
+    component: React.ComponentType<P & { theme: Theme }>
+  ): React.FC<Omit<P, 'theme'> & { theme?: any }>
```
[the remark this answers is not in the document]
Andarist: Cool - ping me once you refactor or come to some conclusions. I'll take another look then.

packages/emotion-theming/types/index.d.ts
```diff
@@ -14,17 +18,43 @@ export function ThemeProvider<Theme>(
...
 ): React.ReactElement
 
-/**
- * @todo Add more constraint to C so that
- * this function only accepts components with theme props.
- */
-export function withTheme<C extends React.ComponentType<any>>(
-  component: C
-): React.SFC<AddOptionalTo<PropsOf<C>, 'theme'>>
+export function withTheme<P extends object>(
```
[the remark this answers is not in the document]
Andarist: Right now they are not synced - take a look at `component` type and what both say about `theme`

packages/emotion-theming/types/index.d.ts
```diff
@@ -14,17 +18,43 @@ export function ThemeProvider<Theme>(
...
+  <Tag extends keyof JSX.IntrinsicElements, ExtraProps = {}>(
+    tag: Tag,
+    options?: StyledOptions
+  ): CreateStyledComponent<
+    JSX.IntrinsicElements[Tag],
+    ExtraProps & { theme: Theme }
+  >
 }
+
+export type WithTheme<P, T> = P extends { theme: infer Theme }
```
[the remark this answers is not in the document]
Andarist: Ok, I haven't noticed that this just moves it around.

packages/styled-base/types/index.d.ts
```diff
@@ -26,105 +24,38 @@ export {
...
+  ): StyledComponent<InnerProps, StyleProps>
+  <StyleProps extends object>(
     template: TemplateStringsArray,
-    ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
-  ): StyledComponent<InnerProps, StyleProps, Theme>
+    ...styles: Array<Interpolation<InnerProps & StyleProps & ExtraProps>>
+  ): StyledComponent<InnerProps, StyleProps>
 }
 
-interface CreateStyledComponentBaseThemed<
```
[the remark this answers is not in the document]
Andarist: There is currently an issue with TS types not being tested at all by the CI (I'm going to work on fixing this soon) - but there are some existing tests around this. Could you refactor them where appropriate & also check if everything that is being testes by them is covered now? 

We'd like to ensure at least that everything can still be expressed correctly - even if with different exported types.

Andarist:
@JakeGinnivan I must say that I like changes so far, seems cleaner in many places. This still needs some work and testing but looks really great at this point in time. I would like to applaud this effort and preparing such a nice PR 👏

JakeGinnivan:
Thanks @Andarist, I will keep working through all the feedback. I think things can be improved in a few places and I might try to write some tests around the types as well.

Likely will be a few days, have a busy weekend this weekend so haven't been able to do much on it.

I was also copying types between 3 projects so need to ensure all actually are in sync to test properly.

packages/styled-base/types/index.d.ts
```diff
@@ -26,105 +24,38 @@ export {
...
+export interface CreateStyledComponent<InnerProps, ExtraProps> {
+  <StyleProps extends object>(
+    ...styles: Array<Interpolation<InnerProps & StyleProps & ExtraProps>>
+  ): StyledComponent<InnerProps, StyleProps>
+  <StyleProps extends object>(
     template: TemplateStringsArray,
-    ...styles: Array<Interpolation<WithTheme<StyleProps, Theme>>>
-  ): StyledComponent<InnerProps, StyleProps, Theme>
+    ...styles: Array<Interpolation<InnerProps & StyleProps & ExtraProps>>
+  ): StyledComponent<InnerProps, StyleProps>
```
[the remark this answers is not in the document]
JakeGinnivan: There is a type to specify additional props on the CreateStyled function, like so:

```
const StyledComponent4 = styled('div')<{ noisy: boolean }>`
  width: 100px;
  color: ${props => props.noisy ? 'red' : 'blue'};
`
```

I'm not really sure where the ExtraProps stuff came in, but it was super handy to just use to add `theme` on.

If the above approach will cause issues (I saw a comment about some styled-components plugin having issues?) can I get more info so I can test it properly :)

packages/styled-base/types/index.d.ts
```diff
@@ -26,105 +24,38 @@ export {
...
    */
-  withComponent<NewTag extends keyof JSXInEl>(
+  withComponent<NewTag extends keyof JSX.IntrinsicElements>(
     tag: NewTag
-  ): StyledComponent<JSXInEl[NewTag], StyleProps, Theme>
-  withComponent<Tag extends React.ComponentType<any>>(
-    tag: Tag
-  ): StyledComponent<PropsOf<Tag>, StyleProps, Theme>
+  ): StyledComponent<JSX.IntrinsicElements[NewTag], StyleProps>
+  withComponent<Props extends object>(
```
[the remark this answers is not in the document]
Andarist: actually - after a thought - this should be `extends {}`, because `object` represents all non-primitive types, so this extends now is satisfied with things we don't want to allow

packages/emotion-theming/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: I can't figure out how `EmotionTheming` could be used. The only use would be if you could export a module as an interface. For example

```
declare module "emotion-theming" extends EmotionTheming<MyTheme> { }
```
There is a discussion about this feature at https://github.com/microsoft/TypeScript/issues/420

Going to delete the type for now, it isn't referenced in documentation and a google search turns nothing up.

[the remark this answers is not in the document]
Andarist: Seems like it could be imported and used regularly, but I agree that it doesn't make much sense to export this like that. You can always use `typeof` to get those types

packages/styled-base/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: Tests found an issue, have lost support for `defaultProps`, will restore this and fix up all the tests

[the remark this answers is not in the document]
JakeGinnivan: Will check how much putting the defaultProps support back in, seeing as that is a pattern which React is moving away from I am tempted to document a workaround by specifying the Prop types manually.

We gain improvements in type inference elsewhere, so I think it's worth the tradeoff.

To detect defaultProps we need to restore the generic type being the component, which means we cannot put constraints on the prop generic type.

So to sum up, I want to drop support for defaultProps, then put in the TypeScript docs on how to support it by using this syntax instead

`styled<PropsOf<MyComponent>>(MyComponent)`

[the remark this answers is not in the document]
Andarist: > To detect defaultProps we need to restore the generic type being the component, which means we cannot put constraints on the prop generic type.

Could you elaborate about those tradeoffs? Examples would be good, I'm not able to follow this PR super closely, bringing myself up to speed each time is quite time-consuming when analyzing previous & current types

In general, I would be OK with dropping `defaultProps` support out of the box if it gives us some concrete benefits. Not sure if such change is OK as part of v10 though and I wouldn't like to make any decisions about that on my own. cc @mitchellhamilton @Ailrun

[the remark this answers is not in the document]
JakeGinnivan: Lets take CreateStyled

```
export interface CreateStyled {
    <C extends React.ReactType<any>, ExtraProps = {}>(
        component: React.ComponentType<Props>,
        options?: StyledOptions
    ): CreateStyledComponent<PropsOf<C>, ExtraProps>

    <Tag extends keyof JSX.IntrinsicElements, ExtraProps = {}>(
        tag: Tag,
        options?: StyledOptions
    ): CreateStyledComponent<JSX.IntrinsicElements[Tag], ExtraProps>
}

vs

export interface CreateStyled {
    <Props extends object, ExtraProps = {}>(
        component: React.ComponentType<Props>,
        options?: StyledOptions
    ): CreateStyledComponent<Props, ExtraProps>

    <Tag extends keyof JSX.IntrinsicElements, ExtraProps = {}>(
        tag: Tag,
        options?: StyledOptions
    ): CreateStyledComponent<JSX.IntrinsicElements[Tag], ExtraProps>
}
```

The first the generic type is the component itself, so when we need to reference the props we need to Extract it using mapped typed + infer.

We have no generic constrains on Props in this scenario but functions like `withTheme` should.

That said, I just checked and using variant 2 in our project does not increase build times or types much at all. I might kick this discussion down the road and just support it.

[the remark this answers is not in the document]
JakeGinnivan: Oops, made the change to CreateStyled, not CreateThemedStyled. 

I have some compilation issues now.

For my the signature is way easier to work with. If the type inference doesn't work it's really hard to bail out. 

image

I have the default React props, but it's failed to extract the props properly. In our project I could remove a bunch of manually specified types and rely on inference with the latter. 

It also is super easy to give the user a type which allows them to grant. Let me work on going with the latter and updating the docs, then we can discuss. It *feels* like a better way to go but that's just me leaning towards simplicity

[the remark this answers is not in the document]
JakeGinnivan: So if you want to support defaultProps / propTypes you need to do something like this:

`styled<PropsOf<typeof ReactClassComponent0>>(ReactClassComponent0)({})`

This allows us to remove workaround functions like this monster from our codebase

```
/**
 * The `typeSafeWrappedStyledComponent` fn can be used to wrap exisitng styled components and add more props to them.
 * This would usually work natively (see https://emotion.sh/docs/typescript), however emotion doesn't handle
 * union (|) types well with props, so you can use this function to do it instead:
 *
 * typeSafeWrappedStyledComponent(MyComponent)<MyAdditionalProps>(
 *  ({ myAdditionalProp, myOrigionalProp }) => (...)
 * )
 *
 * Emotion also does not type their wrapped components to pass on props to the wrapper. This allows
 * us to do that in a typesafe manner
 */

export const typeSafeWrappedStyledComponent = <C extends WrappedComponentBase>(
    Component: C,
    options?: StyledOptions
) => {
    type Props = GetComponentProps<C>
    type ReactClassPropKeys = keyof React.ClassAttributes<any>
    type Styles<T> = Interpolation<WithTheme<Omit<T, ReactClassPropKeys>, Theme>>[]

    return <AdditionalProps extends {}>(...styles: Styles<Props & AdditionalProps>) => {
        const WrappedComponent = React.forwardRef<any, Props & AdditionalProps>((props, ref) => {
            const newProps = { ...props, ref }
            const ToRender = Component as React.FC<Props & AdditionalProps>
            return <ToRender {...newProps} />
        })

        WrappedComponent.displayName = 'StyledComponentWithMixedProps'
        return styled(WrappedComponent, options)(...styles)
    }
}
```

JakeGinnivan:
I'm also trying to write down as I go. https://medium.com/@jakeginnivan/improving-emotion-10s-typescript-types-622e2d172f6f

This is slowing me down, sorry about that but it's helping me make sure the changes are more deliberate.

Andarist:
@JakeGinnivan Don't worry - it's not like we want to rush this. You are doing a tremendous effort here only because you are motivated to do so on your own - what I'm trying to say, I don't expect anything from you, it's OSS. However, I'm of course grateful for all your efforts 👍 

The writeup itself is a great idea and can't wait to dive into it - I'm going to be busy in the following days, but I promise to look at it and on your latest commits afterwards.

packages/css/types/tests.ts
```diff
@@ -28,11 +28,3 @@ css`
   position: relative;
   top: ${'20px'};
 `
-
-// $ExpectError
-css(() => 'height: 300px;')
```
JakeGinnivan: These no longer fail, not sure that is an issue? It's a FunctionInterpolation returning a SerialisedInterpolation isn't it?

packages/emotion-theming/types/tests.tsx
```diff
@@ -54,7 +55,7 @@ const ThemedCompWithDefault = withTheme(CompCWithDefault)
 const {
   ThemeProvider: TypedThemeProvider,
   withTheme: typedWithTheme
-} = emotionTheming as emotionTheming.EmotionTheming<Theme>
+}: emotionTheming.EmotionTheming<Theme> = emotionTheming
```
JakeGinnivan: Found a test which used `EmotionTheming`, have added it's usage to the docs

packages/styled-base/types/helper.d.ts
```diff
@@ -7,5 +7,8 @@ export type PropsOf<
   C extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>
 > = JSX.LibraryManagedAttributes<C, React.ComponentPropsWithRef<C>>
 
+// We need to use this version of Omit as it's distributive (Will preserve unions)
```
JakeGinnivan: I went to use the inbuilt, but it is different so i added a comment

packages/styled-base/types/tests.tsx
```diff
@@ -82,22 +82,15 @@ const Canvas1 = styled('canvas', {
...
- * to be able to declare extra props without using
- * `` styled('button')<ExtraProps>`...` ``, which does not supported in
- * styled-component VSCode extension.
- * If your tool support syntax highlighting for `` styled('button')<ExtraProps>`...` ``
- * it could be more efficient.
- */
-const Button2 = styled<'button', PrimaryProps>('button')`
-  fontsize: ${5}px;
+
+const Button2 = styled('button')<PrimaryProps>`
```
JakeGinnivan: I have updated the test to use this overload instead. I didn't have any issues with the styled components extension in VSCode. All works great

packages/emotion-theming/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: I've removed this, unsure what it did and there are no tests. Can restore

packages/styled-base/types/tests.tsx
JakeGinnivan: Docs have also been updated to be inline with this style

packages/styled-base/types/tests.tsx
```diff
@@ -217,6 +214,7 @@ const StyledClass1 = StyledClass0.withComponent(ReactClassComponent1)
 declare const ref1_0: (element: ReactClassComponent1 | null) => void
 declare const ref1_1: (element: ReactClassComponent0 | null) => void
 declare const ref1_2: (element: HTMLDivElement | null) => void
+// $ExpectError
```
JakeGinnivan: This is now an error because column doesn't exist on the new component, only the original

packages/emotion-theming/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: Sorted out all the typings around this

packages/styled-base/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: Have got all the existing tests and a few new ones passing. Afaik no issues. Resolving this comment as the current signatures support refs / default props etc.

packages/styled-base/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: Have sorted this out am pretty sure.

packages/styled-base/types/index.d.ts
```diff
@@ -135,16 +66,16 @@ export type CreateStyledComponentExtrinsic<
  * If your tool support syntax highlighting for `` styled('button')<ExtraProps>`...` ``
  * it could be more efficient.
  */
-export interface CreateStyled<Theme extends object = any> {
```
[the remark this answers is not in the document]
JakeGinnivan: Have restored this now, comment no longer valid

JakeGinnivan:
@Andarist would love a hand fixing the jest-emotion failure with `dtslint types`

I think it's because jest has it's own types in the /build folder which are not the global types.

$ dtslint types
Error: /Users/jakeg/_code/emotion/packages/jest-emotion/types/index.d.ts:5:47
ERROR: 5:47   expect  TypeScript@next compile error: 
Namespace 'global.jest' has no exported member 'ExpectExtendMap'.
ERROR: 25:9   expect  TypeScript@next compile error: 
Namespace 'global.jest' has no exported member 'SnapshotSerializerPlugin'.
ERROR: 26:26  expect  TypeScript@next compile error: 

/Users/jakeg/_code/emotion/packages/jest-emotion/types/test.ts:37:1
ERROR: 37:1   expect  TypeScript@next compile error: 
Cannot find name 'expect'.
ERROR: 38:1   expect  TypeScript@next compile error: 
Cannot find name 'expect'.

frankwallis:
> @Andarist would love a hand fixing the jest-emotion failure with `dtslint types`
> 
> I think it's because jest has it's own types in the /build folder which are not the global types.
> 

I found that removing the `typeRoots` entry from `jest-emotion/types/tsconfig.json` fixes the errors with the `jest` namespace not being found

JakeGinnivan:
Thanks for that @frankwallis. That did the trick

packages/serialize/types/index.d.ts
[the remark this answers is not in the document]
Andarist: @JakeGinnivan thoughts about this one?

packages/serialize/types/index.d.ts
```diff
@@ -76,7 +86,7 @@ export type Interpolation<MP = undefined> =
   | SerializedStyles
   | ArrayInterpolation<MP>
   | ObjectInterpolation<MP>
-  | Equal<MP, undefined, never, FunctionInterpolation<MP>>
```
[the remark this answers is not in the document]
Andarist: ok, nvmd then - gonna resolve this comment

packages/css/types/tests.ts
Andarist: I think this should fail (not sure 100% yet), because of this:
https://github.com/emotion-js/emotion/blob/a835b89ce8448c590d931ea03e21e506e245c927/packages/serialize/src/index.js#L197
You are right about what this currently is, but seems like css and styled should have slightly different Interpolation type. The pattern here should be allowed for styled, but not so much for css

packages/styled-base/types/tests.tsx
```diff
@@ -146,16 +137,17 @@ const Button5 = styled<typeof ReactSFC0, PrimaryProps>(ReactSFC0)(props => ({
...
   display: flex;
-  flexdirection: ${props => props.column && 'column'};
+  flex-direction: ${props => props.column && 'column'};
 `
 ;<Container0 column={false} />
 // $ExpectError
 ;<Container0 />
 
 const Container1 = Container0.withComponent('span')
-;<Container1 column={true} />
```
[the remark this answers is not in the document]
Andarist: Could you describe the change in more details? Were you able to swap out "old" intrinsic attributes and give the returned component the ones of `span`? Are other props (user-defined) still available after such swap?

packages/styled-base/types/tests.tsx
```diff
@@ -185,29 +175,36 @@ const Container4 = styled(ReactSFC2)<ContainerProps>(props => ({
...
 // $ExpectError
 ;<Container5 column={true} />
-// $ExpectError
-;<Container5 value={242} />
 
 // $ExpectError
 styled(ReactSFC2)<ReactSFCProps1>()
 
-/**
- * @todo
```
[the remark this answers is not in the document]
Andarist: This is wild guess - not based on the coe here, as I don't have time to study it right now - but maybe `Container0` and `ReactClassComponent2` have some overlapping props which are not compatible with each other?

packages/styled-base/types/tests.tsx
Andarist: I'm wondering if this is a correct change - while that prop is not existing on the new component the created interpolations are being carried over to the new component and that prop could be used in previous interpolations, right? In such a situation this could fail miserably at runtime, right?

In general withComponent & as are really difficult to type correctly due to their dynamic(-ish) semantics.

Andarist:
I've left few comments, but didn't have time yet to go through the code - just did a round over comments and such.

packages/styled-base/types/tests.tsx
JakeGinnivan: Oh, right. I get that, I have to build up an intersection of the previous component props.

Nice catch!

packages/styled-base/types/tests.tsx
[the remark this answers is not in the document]
JakeGinnivan: Will expand tests and add comments

packages/css/types/tests.ts
JakeGinnivan: Switching the type from Interpolation to CSSInterpolation in the css module fixed the tests.. I didn't change any of this so can only assume it was failing before?

packages/serialize/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: I had a play with LeafInterpolation, I like the idea but it ended up being more complex it needed.

Interpolations are the next part of my blog post i'm writing so may end up with an improvement here after, I just wanted to get tests all running before making too many changes in this arrea

packages/styled-base/types/tests.tsx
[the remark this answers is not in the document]
JakeGinnivan: Nice, have restored and extended the note in the todo.

packages/css/types/tests.ts
Andarist: Might be as TS tests were not run on CI. Would be great to fix it here though.

packages/styled-base/types/tests.tsx
JakeGinnivan: Fixed this now.

I have restructured the generic type params in StyledComponent and related types to make it really clear and the change allows ComponentProps from a normal component to get transferred while dropping tag props

packages/css/types/index.d.ts
```diff
@@ -14,6 +18,6 @@ export { Interpolation, SerializedStyles }
 
 export default function css(
   template: TemplateStringsArray,
-  ...args: Array<Interpolation>
+  ...args: Array<CSSInterpolation>
```
JakeGinnivan: @Andarist this change fixed the test where `css` was too permissive. Let me know if you forsee any issues

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
 export interface StyledOptions {
   label?: string
   shouldForwardProp?(propName: string): boolean
   target?: string
 }
 
-export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+/**
+ * @typeparam ComponentProps  Props which will be included when withComponent is called
```
JakeGinnivan: @Andarist do these params make sense, and happy for tsdocs?

packages/styled-base/types/tests.tsx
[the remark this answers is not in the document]
JakeGinnivan: This is all sorted now, and tests added.

JakeGinnivan:
@Andarist those comments were super useful. This is feeling pretty close now.

Need to do another round of testing on my own projects, then probably pretty close.

packages/emotion-theming/types/index.d.ts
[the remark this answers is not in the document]
Andarist: This has been handled right now, right?

packages/css/types/tests.ts
Andarist: Cool!

packages/styled-base/types/tests.tsx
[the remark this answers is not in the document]
Andarist: Superb 👌

packages/css/types/index.d.ts
Andarist: Should be fine 👍

packages/serialize/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: I will take a look in the next few days.

In terms of breaking changes, I’ve removed a bunch of them. Should I rewrite the commits and drop back to more correct commit messages. Most of the initial ones no longer apply.
I could even close down this PR and we start fresh?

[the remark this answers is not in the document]
JakeGinnivan: I will take a look in the next few days.

In terms of breaking changes, I’ve removed a bunch of them. Should I rewrite the commits and drop back to more correct commit messages. Most of the initial ones no longer apply.
I could even close down this PR and we start fresh?

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
+ * @typeparam SpecificComponentProps  Props which will *not* be included when withComponent is called
+ */
+export interface StyledComponent<ComponentProps, SpecificComponentProps>
```
Andarist: this requires 2 type arguments now, isn't `ComponentProps` enough for majority of use cases? especially that `SpecificComponentProps` are somewhat purely to support `withComponent` which in general is a discouraged API?

In other words - WDYT about adding a default `{}` to  `SpecificComponentProps`?

JakeGinnivan: Default makes sense, will add

packages/serialize/types/index.d.ts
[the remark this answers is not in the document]
Andarist: Nah, keep the commits - they show the progression here. We can always squash this whole thing when merging to master, but the reference to this PR and its history will be nicely kept.

JakeGinnivan:
@Andarist I have looked into the interpolation side of things.

The `Equal<MP, undefined, never, FunctionInterpolation<MP>>` basically meant if the merged props were undefined then you couldn't use the function interpolations. I can't see how this would work because we infer props on all code paths, and at worst it would be an empty object. So now I think this is fine to remove as if the user tried to use those props, they won't be too much use. Not worth the complexity.

Other than that, I honestly don't have enough context to make meaningful changes in this area.

packages/serialize/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: The only thing I could see which may improve the types in this space is creating specific types for the interpolation path vs the css object path. But honestly not sure if there is much difference here. 

If you have any ideas around cleaning up I can have a go, but I think the gains were mainly in cleaning up the CreateStyled related functions

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
   target?: string
 }
 
-export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+/**
+ * @typeparam ComponentProps  Props which will be included when withComponent is called
+ * @typeparam SpecificComponentProps  Props which will *not* be included when withComponent is called
+ */
+export interface StyledComponent<ComponentProps, SpecificComponentProps = {}>
```
JakeGinnivan: Type change:

`StyledComponent<InnerProps, StyleProps, Theme extends object>` to `StyledComponent<ComponentProps, SpecificComponentProps = {}>`

Theme no longer explicitly a generic parameter on StyledComponent, simply use `StyledComponent<{ theme: Theme }>` instead. The second generic argument (`SpecificComponentProps`) are props which will not transfer to the next component when `withComponent` is called.

Migration example:

`StyledComponent<Props, {}, MyTheme>` -> `StyledComponent<Props & { theme?: MyTheme }>`

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
+  /**
+   * @typeparam AdditionalProps  Additional props to add to your styled component
+   */
+  <AdditionalProps extends object>(
     template: TemplateStringsArray,
-    ...styles: Array<Interpolation<WithTheme<StyleProps, StyledInstanceTheme>>>
-  ): StyledComponent<InnerProps, StyleProps, StyledInstanceTheme>
-}
-
-export type CreateStyledComponentBase<
```
JakeGinnivan: Change log:

`CreateStyledComponentBase`, `CreateStyledComponentBaseThemeless` and `CreateStyledComponentBaseThemed` have been merged into `CreateStyledComponent`.

To migrate:

`CreateStyledComponentBaseThemed<P, P1, MyTheme>` -> `CreateStyledComponent<P & { theme?: MyTheme }, { }, { theme: MyTheme }>`

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
- * If your tool support syntax highlighting for `` styled('button')<ExtraProps>`...` ``
- * it could be more efficient.
+ * This function accepts a React component or tag ('div', 'a' etc).
+ *
+ * @example styled(MyComponent)({ width: 100 })
+ * @example styled(MyComponent)(myComponentProps => ({ width: myComponentProps.width })
+ * @example styled('div')({ width: 100 })
+ * @example styled('div')<Props>(props => ({ width: props.width })
  */
-export interface CreateStyled<Theme extends object = any> {
```
JakeGinnivan: Change log: 

CreateStyled functions no longer take a second `ExtraProps` argument. Instead move it to after the create styled call. For example

`styled<typeof MyComponent, ExtraProps>(MyComponent)({})` -> `styled(MyComponent)<ExtraProps>({})`

packages/css/types/index.d.ts
```diff
@@ -14,6 +18,6 @@ export { Interpolation, SerializedStyles }
...
 ): SerializedStyles
-export default function css(...args: Array<Interpolation>): SerializedStyles
+export default function css(...args: Array<CSSInterpolation>): SerializedStyles
```
JakeGinnivan: Change log:

`css` function has been restricted to prevent passing of invalid types

packages/emotion-theming/types/index.d.ts
```diff
@@ -3,30 +3,43 @@
...
+) => React.ReactElement
 
-export function useTheme<Theme>(): Theme
-
-/**
- * @todo Add more constraint to C so that
- * this function only accepts components with theme props.
- */
-export function withTheme<C extends React.ComponentType<any>>(
+export type withTheme<Theme = any> = <
```
JakeGinnivan: Change log:

withTheme can now have the Theme type specified when calling it. For example `withTheme<MyTheme>(MyComponent)`

Breaking change, if you were specifying the ComponentType you will need to remove the generic parameter. Recommend following example setup in the TypeScript docs under theming section

packages/emotion-theming/types/index.d.ts
```diff
@@ -3,30 +3,43 @@
...
+export const useTheme: useTheme
 
 export interface EmotionTheming<Theme> {
-  ThemeProvider(props: ThemeProviderProps<Theme>): React.ReactElement
-  withTheme<C extends React.ComponentType<any>>(
-    component: C
-  ): React.SFC<AddOptionalTo<PropsOf<C>, 'theme'>>
+  ThemeProvider: ThemeProvider<Theme>
+  withTheme: withTheme<Theme>
+  useTheme: useTheme<Theme>
```
JakeGinnivan: Change log:

useTheme added to EmotionTheming interface and can now create your own closed variation of withTheme. More information in the docs under the theming section.

packages/serialize/types/index.d.ts
```diff
@@ -63,7 +61,20 @@ export interface ObjectInterpolation<MP>
...
+    | null
+    | undefined
+    | boolean
+    | number
+    | string
+    | ComponentSelector
+    | Keyframes
+    | SerializedStyles
+    | ArrayInterpolation<MP>
+    | ObjectInterpolation<MP>
```
JakeGinnivan: Change log:

Function interpolation can no longer return another function directly.

packages/styled-base/types/index.d.ts
JakeGinnivan: Fix: Now union prop type on React components are preserved when creating the Styled component

JakeGinnivan:
Ok @Andarist. I think I'm done. All the tests have been fixed or updated and I've now added the changelog, updated docs etc. The change log comments were just for me to collect them for writing the actual changelog. Am going to resolve them now as it would be better to review the actual changelog.

The break changes are actually now quite minimal, and I think mostly it will be removing manually specified generic arguments (this fixed most errors in our codebase). And it also picked up a few type issues in our codebase too, so the types have actually improved.

These types still greatly reduce the number of types during compilation and also build time for us. ~850k types down to ~230k types.

packages/styled-base/types/index.d.ts
JakeGinnivan: Done

packages/emotion-theming/types/index.d.ts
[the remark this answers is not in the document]
JakeGinnivan: Yes. It's required because we want theme to be optional on the resulting component, but Props of components will have it required.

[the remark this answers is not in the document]
JakeGinnivan: There are tests which will fail if we remove this

JakeGinnivan:
I published my blog post, it covers some of the initial direction. I think it's interesting even if this doesn't land.

https://medium.com/@jakeginnivan/understanding-and-improving-emotion-10s-typescript-types-622e2d172f6f

packages/emotion-theming/types/index.d.ts
```diff
@@ -3,30 +3,43 @@
...
   theme: Partial<Theme> | ((outerTheme: Theme) => Theme)
   children?: React.ReactNode
 }
 
-export function ThemeProvider<Theme>(
-  props: ThemeProviderProps<Theme>
-): React.ReactElement
-
-export function useTheme<Theme>(): Theme
+export interface ThemeProvider<Theme = any> {
```
Andarist: shouldn't this be both restricted to `{}` and defaulted to it? `any` doesn't seem right - especially that there are places where we spread themes:
https://github.com/emotion-js/emotion/blob/1b3791d4c458838a097c815a38108c9719023b92/packages/emotion-theming/src/theme-provider.js#L30

packages/styled-base/types/index.d.ts
```diff
@@ -1,5 +1,5 @@
 // Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
-// TypeScript Version: 3.1
+// TypeScript Version: 3.2
```
Andarist: why the bump?

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
 } from '@emotion/serialize'
 
-export { ComponentSelector, Interpolation }
-
-type JSXInEl = JSX.IntrinsicElements
-
-export type WithTheme<P, T> = P extends { theme: infer Theme }
-  ? P & { theme: Exclude<Theme, undefined> }
-  : P & { theme: T }
+export { ComponentSelector, Interpolation, PropsOf, Omit }
```
Andarist: does `Omit` have to exported from here?

Andarist:
I just have resolved all pending comments and noticed some minor things while doing so, so I've commented on them. Going to make a full round review in following days though - it's quite a lot to take in at once 😅

JakeGinnivan:
Really appreciate the great reviews. Will address the comments when I can!

.changeset/long-apes-admire/changes.md
```diff
@@ -0,0 +1,38 @@
...
+  **Breaking change:** Generic argument changed, if you were specifying the ComponentType you will need to remove the generic parameter. Recommend following example setup in the TypeScript docs under theming section
+
+- `css` function has been restricted to prevent passing of invalid types
+- `CreateStyled` functions no longer take a second `ExtraProps` argument. Instead move it to after the create styled call. For example
+
+  `styled<typeof MyComponent, ExtraProps>(MyComponent)({})`  
+  to  
+  `styled(MyComponent)<ExtraProps>({})`
+
+- `CreateStyledComponentBase`, `CreateStyledComponentBaseThemeless` and `CreateStyledComponentBaseThemed` have been merged into `CreateStyledComponent`.
```
Andarist: I'm not entirely sure, but at least `CreateStyledComponentBaseThemeless` and `CreateStyledComponentBaseThemed` weren't meant for public usage, so we can avoid mentioning them here to avoid the unnecessary mental overhead for readers. Not sure about `CreateStyledComponentBase` - that would have to be checked.

.changeset/long-apes-admire/changes.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Improvements
+
+- useTheme added to EmotionTheming interface and can now create your own closed variation of withTheme. More information in the docs under the theming section.
+- Union types as props are better supported and should be inferred properly
+- Build times should be reduced significantly in larger projects.
+
+## Breaking changes
+
+- Function interpolation can no longer return another function directly
```
Andarist: more out of curiosity than from wanting to allow it - does this change have any impact? as we've noticed that interpolation~ type still stays recursive, even when support for this one got dropped

.changeset/long-apes-admire/changes.md
```diff
@@ -0,0 +1,38 @@
...
+- withTheme can now have the Theme type specified when calling it. For example `withTheme<MyTheme>(MyComponent)`
+
+  **Breaking change:** Generic argument changed, if you were specifying the ComponentType you will need to remove the generic parameter. Recommend following example setup in the TypeScript docs under theming section
+
+- `css` function has been restricted to prevent passing of invalid types
+- `CreateStyled` functions no longer take a second `ExtraProps` argument. Instead move it to after the create styled call. For example
+
+  `styled<typeof MyComponent, ExtraProps>(MyComponent)({})`  
+  to  
+  `styled(MyComponent)<ExtraProps>({})`
```
Andarist: is this 100% equivalent? could a codemod be written to ease the migration?

docs/typescript.mdx
```diff
@@ -39,11 +39,20 @@ To make the css prop work with pure TypeScript (without babel plugin) you need t
 
 ```tsx
 /** @jsx jsx */
-import { css, jsx } from '@emotion/core'
-```
+import { jsx } from '@emotion/core'
 
+<div css={{ background: 'black' }} />
+```
 As a result you may be not able to use react fragment shorthand syntax - `<></>`, but still you can use `<Fragment></Fragment>`.
```
Andarist: we could mention here that this is actually a TS compiler limitation - in Babel it's possible to configure jsx pragma and jsxFrag pragma separately

docs/typescript.mdx
```diff
@@ -39,11 +39,20 @@ To make the css prop work with pure TypeScript (without babel plugin) you need t
...
 
+You can still use the css helper and pass the className yourself.
+
+```tsx
+import { css } from '@emotion/core'
+
+<div className={css({ background: 'black' })} />
```
Andarist: this won't work

Either use `css` from `emotion` package or pass the resulting opaque object to `css` prop instead of the `className` (depends on the intention of this added section)

packages/create-emotion/types/tslint.json
```diff
@@ -15,6 +15,7 @@
             "check-typecast",
             "check-type-operator",
             "check-preblock"
-        ]
+        ],
+        "no-null-undefined-union": false
```
Andarist: what's the reason behind disabling this?

packages/emotion-theming/types/tests.tsx
```diff
@@ -11,12 +13,11 @@ declare const theme: Theme
 
 interface Props {
   prop: boolean
-  theme: Theme
 }
-declare const CompSFC: React.SFC<Props>
-declare class CompC extends React.Component<Props> {}
+declare const CompSFC: React.FC<Props>
```
Andarist: the declared identifier could be renamed as well

packages/emotion-theming/types/tests.tsx
Andarist: The added usage example uses `as` assertion, this doesn't. It would be better to keep consistency across those two places. I'm wondering though - why this here (without `as`) is allowed? Like TS can't guarantee that some kind of unknown Theme parameter (the one in `emotionTheming` here) is assignable to a "concrete" Theme defined here 🤔

packages/memoize/types/index.d.ts
```diff
@@ -1,3 +1,3 @@
-type Fn<T> = (key: string) => T
+export type Fn<T> = (key: string) => T
```
Andarist: is this export needed? it's not that useful publicly, just an internal shorthand so the `memoize` declaration can stay cleaner

packages/emotion-theming/types/tslint.json
```diff
@@ -1,21 +1,22 @@
...
+      "check-module",
+      "check-rest-spread",
+      "check-type",
+      "check-typecast",
+      "check-type-operator",
+      "check-preblock"
+    ],
+    "no-null-undefined-union": false,
+    "no-unnecessary-generics": false,
+    "callable-types": false
```
Andarist: Could we unify all of existing tslint.json configs?

packages/serialize/types/index.d.ts
```diff
@@ -63,7 +61,20 @@ export interface ObjectInterpolation<MP>
   extends CSSPropertiesWithMultiValues,
     CSSPseudos<MP>,
     CSSOthersObject<MP> {}
-export type FunctionInterpolation<MP> = (mergedProps: MP) => Interpolation<MP>
+
+export interface FunctionInterpolation<MP> {
```
Andarist: MP looks enigmatic, could we rename it to MergedProps or to something more meaningful (if you have a better name than MergedProps for this)?

packages/serialize/types/tests.ts
```diff
@@ -16,36 +16,33 @@ const testObjectInterpolation1: ObjectInterpolation<undefined> = {
 }
 
 // $ExpectType SerializedStyles
-serializeStyles({}, [])
+serializeStyles([], {})
```
Andarist: why the arguments order has changed here and at other places in this file? 😄

packages/styled-base/types/helper.d.ts
```diff
@@ -5,7 +5,7 @@ import * as React from 'react'
  */
 export type PropsOf<
```
Andarist: how this type is different from "builtin" `ComponentProps`?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6d7624e50fcad42e22a1cfddcea4bc7fa03bc29d/types/react/index.d.ts#L756-L760
Also - what about that suggestion there that we should prefer using `ComponentPropsWithRef` or `ComponentPropsWithoutRef` appropriately over `ComponentProps`?

packages/styled-base/types/index.d.ts
```diff
@@ -12,10 +12,9 @@
  * a style of that component.
  */
 
-import { ComponentSelector, Interpolation } from '@emotion/serialize'
 import * as React from 'react'
-
-import { Omit, Overwrapped, PropsOf } from './helper'
+import { ComponentSelector, Interpolation } from '@emotion/serialize'
+import { PropsOf, Omit } from './helper'
```
Andarist: could we rename this Omit to DistributiveOmit? right now it's super easy to miss that this is not the same as builtin one

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
 }
 
-interface CreateStyledComponentBaseThemed<
-  InnerProps,
-  ExtraProps,
-  StyledInstanceTheme extends object
+/**
+ * @typeparam ComponentProps  Props which will be included when withComponent is called
+ * @typeparam SpecificComponentProps  Props which will *not* be included when withComponent is called
+ * @typeparam StyleProps  Params passed to styles but not exposed as React props. These are normally library provided props
```
Andarist: would it even be possible to build a wrapper over `styled` API or smth that would actually use this? 🤔

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
-  <
-    StyleProps extends Omit<
-      Overwrapped<InnerProps, StyleProps>,
-      ReactClassPropKeys
-    > = Omit<InnerProps & ExtraProps, ReactClassPropKeys>
-  >(
+  /**
+   * @typeparam AdditionalProps  Additional props to add to your styled component
+   */
+  <AdditionalProps extends object>(
```
Andarist: this should `extends {}` from what I understant

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
-    : CreateStyledComponentBaseThemed<
-        InnerProps,
-        ExtraProps,
-        StyledInstanceTheme
+    ...styles: Array<
+      Interpolation<
+        ComponentProps &
+          SpecificComponentProps &
+          AdditionalProps &
+          AdditionalProps
```
Andarist: `AdditionalProps & AdditionalProps` ? 😅

packages/styled-base/types/index.d.ts
```diff
@@ -24,127 +23,88 @@ export {
...
+ *
+ * @example styled(MyComponent)({ width: 100 })
+ * @example styled(MyComponent)(myComponentProps => ({ width: myComponentProps.width })
+ * @example styled('div')({ width: 100 })
+ * @example styled('div')<Props>(props => ({ width: props.width })
  */
-export interface CreateStyled<Theme extends object = any> {
-  <Tag extends React.ComponentType<any>, ExtraProps = {}>(
-    tag: Tag,
+export interface CreateStyled<Theme = any> {
```
Andarist: I think I've mentioned this elsewhere, but better to repeat myself than to miss this. 

This should both restrict Theme to be `extends {}` and provide a `{}` as default rather than `any`

packages/styled-base/types/tests.tsx
```diff
@@ -193,20 +190,35 @@ const Container5 = Container3.withComponent(ReactSFC2)
 // $ExpectError
 ;<Container5 value={242} />
 
-// $ExpectError
-styled(ReactSFC2)<ReactSFCProps1>()
-
 /**
  * @todo
  * I wish we could raise errors for following two `withComponent`s.
+ * The product a component type which is invalid, but you need to consume the component
```
Andarist: is this grammatically correct?

packages/styled-base/types/tests.tsx
```diff
@@ -296,3 +310,73 @@ const StyledFCWithDefaultProps = styled(FCWithDefaultProps)`
...
+class C extends React.Component<TaggedPropsAB> {
+  render() {
+    return null
+  }
+}
+
+const D = styled(C)`
+  color: red;
+`
+;<D tag="a" title="test" />
```
Andarist: Would it be possible having such `tag` being any component and having other acceptable props being based on it?  This would probably require conditional types - but maybe we could cover `as` prop?

packages/styled/types/tests.tsx
```diff
@@ -1,35 +1,42 @@
...
 styled.div
-// $ExpectType CreateStyledComponentIntrinsic<"svg", {}, any>
+// $ExpectType CreateStyledComponent<{ theme?: any; }, SVGProps<SVGSVGElement>, { theme: any; }>
 styled.svg
 
 {
-  // $ExpectType CreateStyledComponentIntrinsic<"svg", { bar: string }, { themed: "black" }>
-  styled.div<{ bar: string }, { themed: 'black' }>`
+  // $ExpectType StyledComponent<{ theme?: any; } & { bar: string; }, DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>>
+  styled.div<{ bar: string }>`
```
Andarist: the intention for this "test" was to check if the given Theme (when it was a separate parameter) is available in interpolations on props, as with new typings theme is "just a prop" I guess there is no need for that original intention to be anyhow preserved 

👍

packages/weak-memoize/types/index.d.ts
```diff
@@ -1,4 +1,6 @@
-type UnaryFn<Arg, Return> = (arg: Arg) => Return
+// TypeScript Version: 2.2
+
+export type UnaryFn<Arg, Return> = (arg: Arg) => Return
```
Andarist: similarly as at another place - does it have to be exported?

packages/create-emotion/types/tslint.json
JakeGinnivan: Interpolations are `undefined | null`. The linting was failing. I can remove one of them if you want?

docs/typescript.mdx
JakeGinnivan: Nice, have updated this and added a comment about importing from the correct location.

.changeset/long-apes-admire/changes.md
JakeGinnivan: Not really, I could revert this change.

packages/weak-memoize/types/index.d.ts
JakeGinnivan: Added the export because the linting rules were failing. Will remove the export and disable that linting rule

packages/styled-base/types/index.d.ts
JakeGinnivan: I could export a different type from the `styled` package which is compatible with the old syntax?

packages/styled-base/types/index.d.ts
JakeGinnivan: I thought it would be preferable than duplicating it like it was before. It is used by emotion-theming

.changeset/long-apes-admire/changes.md
JakeGinnivan: Honestly if you are following the docs you shouldn't really be using any of those types directly. Happy to just remove this line.

packages/emotion-theming/types/tslint.json
JakeGinnivan: I'd prefer that to be a separate PR to be honest. Unfortunately the tests being broken has caused this PR to get a little out of scope.

packages/serialize/types/index.d.ts
JakeGinnivan: Done, don't have a better name.

packages/styled-base/types/helper.d.ts
JakeGinnivan: ComponentProps appears it doesn't take into account default props.

Using ComponentPropsWithRef instead of ComponentProps causes this test to fail:
`const AdditionallyStyledDiv = themedStyled(StyledDiv)({})`

Finally, without changing ComponentPropsWithRef to just ComponentProps `const StyledClass0 = styled(ReactClassComponent0)({})` has a ref property which is not correct.

As far as I can see any changes around these props will cause some tests to fail.

.changeset/long-apes-admire/changes.md
JakeGinnivan: In this scenario yes, there may be scenarios i've missed. 

Related to pain of upgrading, have you thought about moving the types to definitely typed. That way people could upgrade the types independent of the library.

docs/typescript.mdx
JakeGinnivan: Done

packages/styled-base/types/index.d.ts
JakeGinnivan: Done

packages/serialize/types/tests.ts
JakeGinnivan: They got reordered in #1236, but the tests were not updated.

packages/emotion-theming/types/tests.tsx
JakeGinnivan: Done, and renamed a few other instances

packages/styled-base/types/tests.tsx
JakeGinnivan: Heh, yeah. Terrible English, fixed.

packages/styled-base/types/tests.tsx
JakeGinnivan: Not sure that is possible. It would require the type to change based on one of it's keys.

packages/styled/types/tests.tsx
JakeGinnivan: Removed the outer assertion, left the inner one.

packages/styled-base/types/index.d.ts
JakeGinnivan: In this case I think `any` is correct. The other instance I need to check. `Theme` is still used as the type for the `theme` key.

JakeGinnivan: Or does emotion restrict theme to be an object?

JakeGinnivan: Going to restore the `extends object` though

packages/styled-base/types/index.d.ts
JakeGinnivan: Heh, sorted

packages/styled-base/types/index.d.ts
JakeGinnivan: Have updated and added a few more contraints around

packages/emotion-theming/types/index.d.ts
JakeGinnivan: Interesting. I can see how we could restrict on ThemeProvider, but for most other scenarios I think it needs to be `any` unless emotion definitely ensures it will pass an empty object for theme and never `undefined`

packages/emotion-theming/types/tests.tsx
JakeGinnivan: Yeah odd. Must be because of the default of `Theme` to `any` that it's assignable. At any rate, have synced this.

packages/memoize/types/index.d.ts
JakeGinnivan: Sorted

JakeGinnivan:
I think I have answered or fixed all your comments.

We possibly could export some Compat types which have the same generic params as the previous to ease migration. I didn't find the upgrade too hard, in fact it cleaned up our code quite a bit because emotion correctly inferred types in most places.

packages/emotion-theming/types/index.d.ts
Andarist: Yes - this is given. The empty object is a default:
 https://github.com/emotion-js/emotion/blob/eabb4609729464fb37d58fa02d370caa7eca511f/packages/core/src/context.js#L17

.changeset/long-apes-admire/changes.md
Andarist: I think those complex ones can be removed from here - as they were never intended to be used publicly.

.changeset/long-apes-admire/changes.md
Andarist: Maybe let's revert it for now if it doesn't have any meaningful impact - less breaking changes and technically more in line with runtime semantics

.changeset/long-apes-admire/changes.md
Andarist: Do u know how the development on DefinitelyTyped works? Who can merge? Who makes the decisions. How to propose a change? And how to land such a big refactor as this? 

Migrating them there would certainly make our lives here easier, but I'm wondering what about quality for types there? Especially for a complex library like emotion?

packages/create-emotion/types/tslint.json
Andarist: both are needed, so we need to disable this lint rule, like you have done already 👍

packages/emotion-theming/types/tslint.json
Andarist: Sure, no problem.

packages/styled-base/types/helper.d.ts
Andarist: > ComponentProps appears it doesn't take into account default props.

Could you elaborate more on that?

packages/styled-base/types/index.d.ts
Andarist: Hm, not sure if we are on the same page - I'm wondering purely about how useful that third param is. Like I don't see immediately how one could use it. Is there any test covering this? Could you point me to it?

packages/styled-base/types/index.d.ts
Andarist: As mentioned elsewhere - we kinda assume in runtime this is an object (and provide default), so I think `extends {}` would be correct. If you don't agree - let's discuss more, I might be wrong.

packages/styled-base/types/tests.tsx
Andarist: Let's not try it out for now then 😉

Andarist:
> We possibly could export some Compat types which have the same generic params as the previous to ease migration. I didn't find the upgrade too hard, in fact it cleaned up our code quite a bit because emotion correctly inferred types in most places.

I'm wondering if most of this wouldn't be codemodable? That way we could automate migration for most cases.

JakeGinnivan:
@Andarist I am speaking at DDD Sydney next week, so prepping for that has been my focus. I've never done a codemod before and I'm not sure how well it would work.

I don't think I will be able to pick this back up for at least a week, it's super close afaik, just need to make your suggested change around theme.

It would be awesome if you could get this over the line, otherwise we can catch up once I'm back in a week or so and try to close it out.

Andarist:
> @Andarist I am speaking at DDD Sydney next week, so prepping for that has been my focus.

No problem, sounds cool - would love to watch your talk. If you get a recording please share a link!

> I've never done a codemod before and I'm not sure how well it would work.

I could help with writing one - the question is if we can easily translate all old typings to new ones or if it cannot be done automatically and manual migration is needed?

> It would be awesome if you could get this over the line, otherwise we can catch up once I'm back in a week or so and try to close it out.

I'd prefer to wait for you as I'm busy as well with other OSS work at the moment and when it comes to TS I don't feel too confident to make those changes myself right now. I know my TS OK, mapped types and other quirky stuff included, but still struggling with things from time to time - so would much better prefer to have you supervise this work (please re-check my reasoning behind left comments, I might be wrong somewhere) as you are more experienced with that.

packages/emotion-theming/types/index.d.ts
JakeGinnivan: Finally sorted!

.changeset/long-apes-admire/changes.md
JakeGinnivan: These have been simplfied

.changeset/long-apes-admire/changes.md
JakeGinnivan: Done

.changeset/long-apes-admire/changes.md
JakeGinnivan: Yeah not sure, I think the TypeScript team merge. But they are working on automation around this using the bot.

It automatically will ping people who authored the types I think as well. Just a bit of an idea, not 100% how it would all work

packages/styled-base/types/helper.d.ts
JakeGinnivan: If a class component has `defaultProps` on it, PropsOf will end up with those props being optional because they were provided by default props

packages/styled-base/types/index.d.ts
JakeGinnivan: Here is where it's used

image

From the outside `{ theme?: Theme }` will be included with withComponent, but the intrinsic props won't, then `{ theme: Theme }` is not optional when consuming props inside the interpolations.

packages/styled-base/types/index.d.ts
JakeGinnivan: Done!

packages/styled-base/types/index.d.ts
JakeGinnivan: Without it I get this:

Error: /Users/jakeg/_code/emotion/packages/styled-base/types/tests.tsx:332:18
ERROR: 332:18  expect  Compile error in typescript@3.1 but not in typescript@3.2.
Fix with a comment '// TypeScript Version: 3.2' just under the header.
Argument of type 'typeof A' is not assignable to parameter of type '"symbol" | "object" | "a" | "abbr" | "address" | "area" | "article" | "aside" | "audio" | "b" | "base" | "bdi" | "bdo" | "big" | "blockquote" | "body" | "br" | "button" | "canvas" | ... 154 more ... | "view"'.
  Type 'typeof A' is not assignable to type '"view"'.
ERROR: 355:18  expect  Compile error in typescript@3.1 but not in typescript@3.2.
Fix with a comment '// TypeScript Version: 3.2' just under the header.
Argument of type 'typeof C' is not assignable to parameter of type '"symbol" | "object" | "a" | "abbr" | "address" | "area" | "article" | "aside" | "audio" | "b" | "base" | "bdi" | "bdo" | "big" | "blockquote" | "body" | "br" | "button" | "canvas" | ... 154 more ... | "view"'.
  Type 'typeof C' is not assignable to type '"view"'.
ERROR: 361:1   expect  Expected an error on this line, but found none.

.changeset/long-apes-admire/changes.md
JakeGinnivan: Regarding codemods, I hope not too many people are using the former syntax. Most of these breaking changes would be when people couldn't get the existing types to infer their props properly and had to be explicit. Or due to the type structures you needed to specify more generic args before.

JakeGinnivan:
@Andarist sorry it took me so long to get back to this. Let me know if you spot anything else or have any more questions.

I am good for the next few weeks hopefully if these types get released to tag me in any issues and I will try and answer/resolve.

My wife is due early December so can't promise quick turn around time as that date gets closer.

kristojorg:
Is there any way to install from a specific branch when using a monorepo like this? I really need these type changes but am struggling to get them installed. Any recommendations?

JakeGinnivan:
@kristojorg the way I have done it is copy the .d.ts files into a node_modules folder in the `src` folder of my project.

They will take over from the inbuilt.

Like this https://github.com/JakeGinnivan/emotion-types-sandbox/tree/master-types/src/node_modules

JakeGinnivan:
@kristojorg if you give them a go, please let me know if you hit any issues!

kristojorg:
@JakeGinnivan Great trick, thank you! I've done that and these types seem to be resolving the issues I was having using Emotion and Theme-UI. Thanks for your work here!

.changeset/long-apes-admire/changes.md
```diff
@@ -0,0 +1,31 @@
...
+
+- `StyledComponent` type no longer supports the third generic `Theme` paramter. Instead add the `theme` prop to the first `Props` argument. For example:
```
Andarist: nit: paramter -> parameter

docs/typescript.mdx
```diff
@@ -39,10 +39,21 @@ To make the css prop work with pure TypeScript (without babel plugin) you need t
...
-import { css, jsx } from '@emotion/core'
+import { jsx } from '@emotion/core'
+
+<div css={{ background: 'black' }} />
 ```
 
 As a result you may be not able to use react fragment shorthand syntax - `<></>`, but still you can use `<Fragment></Fragment>`.
+This is a limitation of the TypeScript compiler not being able to independently specify jsx pragma and jsxFrag pragma.
+
+You can still use the css helper and pass the className yourself (ensure you are importing from the `@emotion` package, not `@emotion/core`).
```
Andarist: I'm not sure if this should be included in this sections - seem a little bit off-putting, especially given that we don't mention this in non-TS sections (as far as i know). 

And also slight nit - `@emotion` should be `emotion`

Andarist:
> @Andarist sorry it took me so long to get back to this. Let me know if you spot anything else or have any more questions.

Looks great, left just some nit-picking comments regarding docs.

> I am good for the next few weeks hopefully if these types get released to tag me in any issues and I will try and answer/resolve.

I'll discuss with Mitchell how we want to proceed with this one and I'll let you know.

> My wife is due early December so can't promise quick turn around time as that date gets closer.

Congrats! I'm in a very similar situation as well, just for mid-December 😉 No worries though - we'll hopefully manage to handle issues.

Andarist:
@JakeGinnivan Could you merge master to this branch and change the merge base to https://github.com/emotion-js/emotion/pull/1600 ?

docs/typescript.mdx
elektronik2k5: I think that this is a crucial bit of information and should be included both here - and in the non TS docs. "Off putting" up front is better than broken and surprising in hindsight.

JakeGinnivan:
> @JakeGinnivan Could you merge master to this branch and change the merge base to #1600 ?

Done

docs/typescript.mdx
JakeGinnivan: Not sure about that @Andarist, `@emotion/core` is mentioned everywhere else. I personally don't use the css prop so I pulled this info out of somewhere when I noticed it. It felt like something which should be in the docs

hsz:
Is there any chance that it can go live before v11?

Andarist:
@hsz not really, this is somewhat a breaking change and as we are preparing v11 right now it’s just better to include this there. We are going to release a prerelease version with those types and without any breaking changes (other than that). So you should be able to switch to beta v11 line easily if you want to use those typings.

elektronik2k5:
Sorry for the (somewhat) off topic: is there a list of goals or an umbrella issue for v11 anywhere? A plan or timeline? All I was able to find was the[ v11 milestone](https://github.com/emotion-js/emotion/milestone/1) tagged issues.

Andarist:
@elektronik2k5 there is no such umbrella ticket at the moment, but it will be probably created soon

@JakeGinnivan thanks for working on this so hard, I'm going to merge this into v11 right now and comment here in some time, once this gets published to npm (under `next` tag)

Andarist:
This has just been published as 11.0.0-next.1

fabien0102:
Hi!

I'm not totally sure if it was addressed somewhere before, but when I update to the `@next` version with typescript in strict mode, since `props.theme` become optional, nothing compile, in a very nasty way 😕 (a lotttttttt of errors, really ^^) 

I've isolated the usecase in this codesandbox -> https://codesandbox.io/s/charming-nobel-7nbh9
It's basically the documentation implementation with typescript in strict mode:

image

This seams to be related to this PR, so I would like to know if it's an issue (and I will open a real issue) or if I'm missing something obvious 😅 

Thanks for this great library! Really looking forward the v11 😍

Andarist:
As you have used `CreateStyled` I believe this shouldn't happen right now. @JakeGinnivan could you take a look at this one?

JakeGinnivan:
Yeah, this is a bug for sure. Will fix and add some tests.

JakeGinnivan:
Thanks for the repro @fabien0102 , we don't use that particular overload in our codebase so I missed it.

PR up.
https://github.com/emotion-js/emotion/pull/1632

Kadrian:
How to use `CreateStyled` with a theme now? It doesn't work like `export styled as CreateStyled<Theme>` anymore, because `CreateStyled` isn't generic anymore.

Andarist:
You don't have to use CreateStyled to provide Theme type, you can just augment the Theme interface as described here: https://deploy-preview-1600--emotion.netlify.com/docs/typescript#define-a-theme

Kadrian:
Aha! That's pretty cool. But that **only** works for me when wrapping html tags directy. 

When I'm wrapping an existing component, the theme is `any` again, see: https://codesandbox.io/s/friendly-agnesi-ufbpb

Bildschirmfoto 2020-03-26 um 18 23 43

Should I open a new ticket? Great work with the library by the way!

Andarist:
Yes, you can open a new ticket. Please include a repro case in it.

PR #1503, opened 2019-09-12
Add TS types to util packages - hash, memoize, weak-memoize

Andarist:


packages/weak-memoize/types/tests.ts
```diff
@@ -0,0 +1,36 @@
...
+
+declare class Qwe {
+  answer: number
+}
+
+// $ExpectType Array<Foo>
+weakMemoize((arg: Foo) => [arg])({ bar: 'xyz' })
+
+// $ExpectType string[]
+weakMemoize((arg: string) => [arg])('foo')
```
JakeGinnivan: @Andarist this test is failing currently, it conflicts with the test on line 19 which expects a string arg to be an error.

What is the intended way for this to work then i can fix

Andarist: this line can just get removed

PR #1508, opened 2019-09-16
Use hooks for theme context inside styled

danielhusar:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/styled-base/src/index.js
+++ b/packages/styled-base/src/index.js
@@ -83,99 +83,95 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
     }
 
     const Styled: any = withEmotionCache((props, context, ref) => {
-      return (
-        <ThemeContext.Consumer>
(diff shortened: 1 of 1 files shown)


Andarist:
emotion@10 is compatible with react@>=16.3 now, so we can't use hooks in core APIs. This has to wait for emotion@11

FezVrasta:
@Andarist is there a roadmap for emotion@11? Do you want to reserve it for a big API update (like it happened from 10 to 11), or would it be okay to cut a major just to jump on the hooks train?

Andarist:
From what I know (I'm not a decisive person here) it's OK to cut a new release to jump on the hooks train - just some leftover cleanups to do before it can be released, but unfortunately, due to time limitations, there is no one working actively on it right now.

PR #1509, opened 2019-09-17
Add type definition for useTheme from emotion-theming

XeeD:
**What**:
This PR is adding TypeScript type definition for `useTheme`hook.

**Why**:
The type definition was completely missing which prevents us from using `useTheme` in a TypeScript project.

**How**:
Added the definition + attempted to write a test for it. Let me know if it looks OK.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/dirty-owls-remain/changes.md
+++ b/.changeset/dirty-owls-remain/changes.md
@@ -1,0 +1,1 @@
+Add TypeScript type definition for the useTheme hook in emotion-theming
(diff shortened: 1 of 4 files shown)


.changeset/dirty-owls-remain/changes.md
```diff
@@ -0,0 +1 @@
+Add type for useTheme hook in emotion-theming
```
Andarist: could you mention here that this is about TS types?

XeeD: Sure, I've rephrased this to be clearer. Can you take a look?

PR #1514, opened 2019-09-21
feat: improve type definition of styled

laysent:
**What**:
Fix a potential issue of TypeScript definition in emotion. See issue https://github.com/ant-design/ant-design/issues/18868 for an example.

**Why**:
When a component has any static lifecycle defined and when this component has required props, `React.ComponentType<any>` will not be enough. `React.ComponentType<any>` has static definition of `GetDerivedStateFromProps<any, {}>`, while a component with required props might have a static API looks like `GetDerivedStateFromProps<{ title: string >, {}>`, types will not be assignable and TypeScript will throw out an error.

**How**:
Explicitly defines how props of this given component should look like. 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Changeset

### DIFF
--- a/.changeset/spicy-dogs-shop/changes.json
+++ b/.changeset/spicy-dogs-shop/changes.json
@@ -1,0 +1,4 @@
+{
+  "releases": [{ "name": "@emotion/styled", "type": "patch" }],
+  "dependents": []
+}
--- a/.changeset/spicy-dogs-shop/changes.md
+++ b/.changeset/spicy-dogs-shop/changes.md
@@ -1,0 +1,1 @@
+improve TypeScript definition
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -47,7 +47,7 @@ export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
   withComponent<NewTag extends keyof JSXInEl>(
     tag: NewTag
   ): StyledComponent<JSXInEl[NewTag], StyleProps, Theme>
-  withComponent<Tag extends React.ComponentType<any>>(
+  withComponent<Tag extends React.ComponentType<PropsOf<Tag>>>(
     tag: Tag
   ): StyledComponent<PropsOf<Tag>, StyleProps, Theme>
 }
@@ -120,7 +120,7 @@ export type CreateStyledComponentIntrinsic<
   Theme extends object
 > = CreateStyledComponentBase<JSXInEl[Tag], ExtraProps, Theme>
 export type CreateStyledComponentExtrinsic<
-  Tag extends React.ComponentType<any>,
+  Tag extends React.ComponentType<PropsOf<Tag>>,
   ExtraProps,
   Theme extends object
 > = CreateStyledComponentBase<PropsOf<Tag>, ExtraProps, Theme>
@@ -136,7 +136,7 @@ export type CreateStyledComponentExtrinsic<
  * it could be more efficient.
  */
 export interface CreateStyled<Theme extends object = any> {
-  <Tag extends React.ComponentType<any>, ExtraProps = {}>(
+  <Tag extends React.ComponentType<PropsOf<Tag>>, ExtraProps = {}>(
     tag: Tag,
     options?: StyledOptions
   ): CreateStyledComponentExtrinsic<Tag, ExtraProps, Theme>


Andarist:
There is a PR aiming to restructure typings quite a lot - https://github.com/emotion-js/emotion/pull/1501 . Could you check if maybe your issue is already solved in that PR?

laysent:
> There is a PR aiming to restructure typings quite a lot - #1501 . Could you check if maybe your issue is already solved in that PR?

Hi, I have tried from my end. It seems to work, but only when explicitly defined. Here is the code I tried:

```jsx

interface PropsA {
  title: string;
}
interface PropsB {
  content: string;
}
declare type Props = PropsA | PropsB;
class A extends React.Component<Props> {
  render() {
    return null;
  }
}

const B = styled<Props>(A)`
  color: red;
`;
```

The above code works. However, if changing the code to:

```jsx
const B = styled(A)`
  color: red;
`;
```

TypeScript will throw out error, as it seems TypeScript assumes the above statement is equivalent to `const B = styled<PropsA>(A)`

Might be an issue with TypeScript though.

JakeGinnivan:
I need to push up a bunch of changes in progress. I am extending the tests and trying to measure the impact of each type change. It's unfortunately quite a slow process.

Happy to jump on a call and start working on it together though!

JakeGinnivan:
I've pushed my latest set of changes, can you give it a spin now?

JakeGinnivan:
Ah, I have added your example as a test.

It's an issue with union types as a prop. Let me try roll your fix into my PR and see what it does to compilation time/number of types

JakeGinnivan:
Your change works great for this and doesn't make any different to perf in a quick test.

Will roll it into my changes if that's ok?

laysent:
Hi @JakeGinnivan , I can confirm that the issue I described in comment has been fixed.
Unfortunately, the actual issue that I wish to fix in this PR hasn't been fixed in #1501. Here is the code to reproduce it:

```jsx

interface PropsA {
  title: string;
}
interface PropsB {
  content: string;
}
declare type Props = PropsA | PropsB;
class A extends React.Component<Props> {
  static getDerivedStateFromProps(props: Props) {
    return null;
  }
  render() {
    return null;
  }
}

const B = styled(A)`
  color: red;
`;
```

The issue caused by `C extends React.ComponentType<any>` defined in TypeScript file.`React.ComponentType<any>` has `getDerivedStateFromProps(props: any)`, which `getDerivedStateFromProps(props: Props)` cannot be assigned to, since `Props` has required field in it.

And your are correct, my version of fix can't handle with Component where `defaultProps` are defined.

In this case, I would suggest that we use your previous version instead? That version requires one to explicitly pass in props type sometimes, but it at least works for both version.

laysent:
@JakeGinnivan Finally find a solution for all this (hopefully)
The reason why my solution does not work with `defaultProps` is: suppose that the component has props `{ a: boolean }` and defaultProps is also `{ a: boolean }`. It turns out that `PropsOf<C>` will result in `{ a?: boolean }`, which is correct by definition, but it's not what `React.ComponentType` wishes to have. What `React.ComponentType` wishes to have is simply `{ a: boolean }`, the original one, not the one modified due to declaration of defaultProps.
I have created a PR on your end, please check and see if that works: https://github.com/JakeGinnivan/emotion/pull/1

JakeGinnivan:
@laysent that is great, I will merge that in once I get all the other tests green.

Unfortunately they have not been running for a while so I am having to hunt through PRs to find when things broke so I know how to fix things properly. It's pretty time consuming, hope to finish it off in the next few days!

JakeGinnivan:
@laysent got the tests green, merged your change and then made the same change on the themed and withComponent versions of that function too.

All looks good.

JakeGinnivan:
@laysent found an issue and need to do other things now. But this now fails:

```
const StyledDiv = themedStyled('div')({})
const AdditionallyStyledDiv = themedStyled(StyledDiv)({})
```

I am thinking I might just put a 3rd overload in which has the original signature, it seems to then work for both.

laysent:
@jakegavin Hi, I have create another PR, hopefully to fix this: https://github.com/JakeGinnivan/emotion/pull/2
I believe the issue is: `const StyledDiv = themedStyled('div')({})` does not have `ref` defined, thus when using `ComponentPropsWithRef`, `propTypes` defined in `React.ComponentType` will mis-match. Using `ComponentPropsWithoutRef` works with this issue.

JakeGinnivan:
@Andarist I think this Pr can be closed. The changes are all merged into mine

hsz:
@jakegavin 
Any eta regatrding #1501 ? 
Thanks for your work by the way!

JakeGinnivan:
@hsz I have just rebased and am working through the remaining feedback. Want to finish off tonight or tomorrow

Andarist:
Closing in favor of #1501

PR #1526, opened 2019-10-01
babel-plugin-emotion: Fix `autoLabel` crash on `css` used in expression assigned to an object pattern

ahutchings:
I am experiencing an exception when running babel-plugin-emotion against a particular usage of emotion. I've added a failing fixture with a (somewhat) reduced test case - it can probably be reduced further once the root cause is identified.

So far, I've found that something is going wrong in the getDeclaratorName function  in utils/label, resulting in it returning null. I plan to investigate this further, but wanted to open an issue in advance to having a fix in case someone has suggestions on how it should be fixed.

```
TypeError: Cannot read property 'charAt' of undefined

      126 |   let declaratorName = getDeclaratorName(path, t)
      127 |   // if the name starts with _ it was probably generated by babel so we should ignore it
    > 128 |   if (declaratorName.charAt(0) === '_') {
          |                      ^
      129 |     return ''
      130 |   }
      131 |   return declaratorName

      at charAt (packages/babel-plugin-emotion/src/utils/label.js:128:22)
```


Andarist:
Thank you for investigating this and providing a failing test case!

Looking into the code we assume right now that if `css` call is contained within VariableDeclarator then it's `id` is a simple Identifier:
https://github.com/emotion-js/emotion/blob/a835b89ce8448c590d931ea03e21e506e245c927/packages/babel-plugin-emotion/src/utils/label.js#L59-L63
but in your case it's actually a ObjectPattern.

Assuming my guess is right then the simplified case would be something like:
```
const [a, b] = [css`color: hotpink;`, css`background-color: #000;`]
```
I like your test though as it showcase way better a real-life scenario in contrary to my "reduced" case which is completely artificial. So I would keep yours - it's simple enough.

ahutchings:
Thanks for the feedback @Andarist. I've updated the PR with a fix and changeset. I believe this is ready to go pending review.

PR #1532, opened 2019-10-03
InterpolationTheme should pass its generic info on

nmccready:
Interpolation<T> accepts types and it should be forwarded from InterpolationTheme.

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -47,11 +47,13 @@ export interface GlobalProps<Theme> {
  */
 export function Global<Theme = any>(props: GlobalProps<Theme>): ReactElement
 
-export function keyframes(
+export function keyframes<Theme = any>(
   template: TemplateStringsArray,
-  ...args: Array<Interpolation>
+  ...args: Array<Interpolation<Theme>>
 ): Keyframes
-export function keyframes(...args: Array<Interpolation>): Keyframes
+export function keyframes<Theme = any>(
+  ...args: Array<Interpolation<Theme>>
+): Keyframes
 
 export interface ArrayClassNamesArg extends Array<ClassNamesArg> {}
 export type ClassNamesArg =
@@ -63,8 +65,11 @@ export type ClassNamesArg =
   | ArrayClassNamesArg
 
 export interface ClassNamesContent<Theme> {
-  css(template: TemplateStringsArray, ...args: Array<Interpolation>): string
-  css(...args: Array<Interpolation>): string
+  css(
+    template: TemplateStringsArray,
+    ...args: Array<Interpolation<Theme>>
+  ): string
+  css(...args: Array<Interpolation<Theme>>): string
   cx(...args: Array<ClassNamesArg>): string
   theme: Theme
 }
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -35,8 +35,8 @@ export function withEmotionCache<Props, RefType = any>(
 export const jsx: typeof createElement
 
 export type InterpolationWithTheme<Theme> =
-  | Interpolation
-  | ((theme: Theme) => Interpolation)
+  | Interpolation<Theme>
+  | ((theme: Theme) => Interpolation<Theme>)
 
 export interface GlobalProps<Theme> {
   styles: InterpolationWithTheme<Theme>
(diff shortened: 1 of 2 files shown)


packages/core/types/index.d.ts
```diff
@@ -47,11 +47,13 @@ export interface GlobalProps<Theme> {
  */
 export function Global<Theme = any>(props: GlobalProps<Theme>): ReactElement
 
-export function keyframes(
+export function keyframes<Theme = any>(
```
nmccready: Here down also had the same issue.

packages/core/types/index.d.ts
```diff
@@ -35,8 +35,8 @@ export function withEmotionCache<Props, RefType = any>(
 export const jsx: typeof createElement
 
 export type InterpolationWithTheme<Theme> =
-  | Interpolation
-  | ((theme: Theme) => Interpolation)
+  | Interpolation<Theme>
```
nmccready: Original problem.

Andarist:
Could u discuss those changes under - https://github.com/emotion-js/emotion/pull/1501 . We’d also need test cases for this - when current types fail? How do u plan on utilizing the change?

nmccready:
> Could u discuss those changes under - #1501 . We’d also need test cases for this - when current types fail? How do u plan on utilizing the change?

Is this really worth discussing there? This is a fix and not meant to be a refactor like #1501 is.  The current utilization of `Interpolation` with the theme is incorrect. 

I'll add the test cases.

Andarist:
> Is this really worth discussing there? 

I believe it might be - the other PR is a comprehensive rework of the current typings and it would be good for it to be aware of every change and uncovered use cases.

> I'll add the test cases.

that would be great 👍

packages/core/types/index.d.ts
Andarist: Is it possible to access Theme inside `keyframes`? not sure how this can be done in vanilla JS

packages/core/types/tests.tsx
```diff
@@ -123,3 +125,10 @@ interface TestTheme1 {
     )
   }}
 </ClassNames>
+
+// issue 1532 https://github.com/emotion-js/emotion/pull/1532
```
nmccready: Test added.

nmccready:
> rework of the current typings

Right, and that is exactly my point. That would be a breaking change or minor / major bump. This is not as it is merely a patch. So while the other ticket could potentially benefit and piggy back of these changes (if not already done), I don't think it's necessary for me to dive into the refactor. Please consider merging this and releasing a patch.

Andarist:
I'll look at this later - but you haven't added any test for keyframes change so far, so I'm still unsure if it's the correct change.

nmccready:
Sorry, forgot about the keyframes. Will do.

packages/core/types/tests.tsx
```diff
@@ -123,3 +124,62 @@ interface TestTheme1 {
...
+}
+
+// BEGIN NON TEMPLATE ANIMATIONS
+// without restricting less
+;<LessComponent
+  less={keyframes<AnimationTheme>({ from: { top: 0 }, to: { top: 20 } })}
+/>
+
+// FAILS IN TYPESCRIPT 2.8 only due to <LessComponent<AnimationTheme> not being supported
+// ;<LessComponent<AnimationTheme>
```
nmccready: If you want to do a breaking minor / major change to drop support for typescript 2.8, then these tests would pass.

packages/core/types/tests.tsx
```diff
@@ -123,3 +124,62 @@ interface TestTheme1 {
...
+
+interface AnimationTheme {
+  from: { top: number }
+  to: { top: number }
+}
+
+// BEGIN NON TEMPLATE ANIMATIONS
+// without restricting less
+;<LessComponent
+  less={keyframes<AnimationTheme>({ from: { top: 0 }, to: { top: 20 } })}
```
Andarist: I don't see how `AnimationTheme` is relevant here, `keyframes`' both inputs and return value are not dependant on it at all. 

Maybe the fix you are after here is somewhere else? Might be that `args` passed to `keyframes` are wrongly types and shouldn't use `Interpolation` type at all. I'm not 100% sure if this is the correct answer - this is the direction I would investigate right now, because parametrizing `keyframes` with `Theme` doesn't make much sense to me.

Andarist:
We've reworked those types in the upcoming v11. Some of the changes from this PR were not needed/wrong - but only caused by the fact that we had wrong types used at various places.

Sorry that we couldn't get this done sooner, I just hope that v11 will fix your problem entirely and provide a much better TS experience overall. Stay tuned for the release - or if you want, you can start using it now as we publish `next` versions to npm pretty often (there are not many breaking changes beside TS typings)

Issue #1513, opened 2019-09-20
Everything hot pink after 'Global Styles'!


shiatsumat:
**Description:**
It is cool that when in the article 'Global Styles' virtually everything including the header and the navigation bar turns into hot pink.
However, even more curiously and awkwardly, that effect applies also to every article after 'Global Styles'.

**Documentation links:**
All the articles after 'Global Styles' https://emotion.sh/docs/globals, i.e.
https://emotion.sh/docs/keyframes, https://emotion.sh/docs/ssr, https://emotion.sh/docs/with-props, ....

For example, the article 'Keyframes' https://emotion.sh/docs/keyframes looks like this.
<img width="994" alt="Screenshot 2019-09-20 11 41 52" src="https://user-images.githubusercontent.com/529217/65295210-a66bb080-db9b-11e9-87cc-9cc3cc74f330.png">

AntonNiklasson:
If this is not already being worked on I could take a stab at it 👋

Andarist:
@AntonNiklasson you are more than welcome to take a stab at this. If you need any help you can ping me on our Slack. Most likely we should purge all global styles on navigation.

AntonNiklasson:
I'm not able to reproduce this. Not on the live site, and not in my local development environment. 🤷‍♂

shiatsumat:
That sounds strange. I experienced this 'hot pink' phenomenon on Google Chrome (version 77.0.3865.90), Safari (version 13.0.1) and Firefox (version 69.0.1) on MacBook Mojave (version 10.14.6).

AntonNiklasson:
Alright, this is weird. I was actually able to reproduce it now. Loading the site straight into https://emotion.sh/docs/globals makes the `hot pink` stick around when navigating. If I reload the page on something like `https://emotion.sh/docs/introduction`, navigate the "Global styles" and then to some other page, the global styles are purged.

I'll look into it a bit further.

shiatsumat:
Ah, I have also experienced that.

zslabs:
I've run into this pretty randomly as well.

Maybe not as "flashy", but could we just extract the `.some-class` styles from the example into `<Global />` and target just that selector to help stop the bleeding onto other pages?

mitchellhamilton:
> Maybe not as "flashy", but could we just extract the .some-class styles from the example into <Global /> and target just that selector to help stop the bleeding onto other pages?

This sounds good to me.

PR #1536, opened 2019-10-04
Limit scope of <Global /> example

zslabs:
**What**:

Resolves https://github.com/emotion-js/emotion/issues/1513

Limits scope of `<Global />` example on the documentation site to get around an issue where the `<Global />` component is bleeding to other documentation pages.

**Why**:

I believe this may be a SSR dependent issue; where `<Global />` is not technically removed from the DOM in the same way it is via client-side on component unmount. I'm not sure if this is a bug or just an implementation detail; would also be happy to add a note about this on the page as well if we felt it was warranted.

**How**:

Updated doc example.

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset N/A


docs/globals.mdx
```diff
@@ -6,28 +6,20 @@ Sometimes you might want to insert global css like resets or font faces. You can
...
-      styles={css`
-        * {
-          color: hotpink !important;
-        }
-      `}
-    />
     <Global
       styles={{
         '.some-class': {
+          color: 'hotpink !important',
```
mitchellhamilton: Could you move this back to the string one but make it target `.some-class`(I want there to still be a string example)

docs/globals.mdx
```diff
@@ -6,28 +6,20 @@ Sometimes you might want to insert global css like resets or font faces. You can
...
           fontSize: 50,
           textAlign: 'center'
         }
       }}
     />
-    <div className="some-class">
-      Everything is hotpink now!
-    </div>
+    <div className="some-class">Everything is hotpink now!</div>
```
mitchellhamilton:
```suggestion
    <div className="some-class">This is hotpink now!</div>
```

zslabs:
Updated this one from your comments. Thanks as always!

PR #1537, opened 2019-10-05
Add missing /* #__PURE__ */ annotation to creation of EmotionCssPropInternal

Andarist:
We could automate adding those with my https://github.com/Andarist/babel-plugin-annotate-pure-calls - but not sure how do you feel about it, so I've added this inline for now.


mitchellhamilton:
> We could automate adding those with my https://github.com/Andarist/babel-plugin-annotate-pure-calls - but not sure how do you feel about it, so I've added this inline for now.

I'd rather add them manually just in case things break

PR #1538, opened 2019-10-05
Push source maps & labels to cooked/raw arrays in TS-transpiled tagged template expressions

Andarist:
This outputs code as if source maps were added to tagged template expression itself - instead of pushing new arguments to `__makeTemplateObject` and causing strings & values length mismatch.

Fixes issue reported in https://github.com/emotion-js/emotion/issues/1354#issuecomment-534793492


Andarist:
@mitchellhamilton good catch, handled that as well and added tests

Issue #1495, opened 2019-09-10
Accept objects as className in styled components


benmerckx:
**Current behavior:**

Class names passed to styled components that are not of type string are not passed to React.
````js
const myclass = {toString: () => 'myclass'}
const Comp = styled.div``
<Comp className={myclass} /> // => no "myclass" in output
````

See: https://github.com/emotion-js/emotion/blob/66cda64128631790b81e3c9df273a972358ea593/packages/styled-base/src/index.js#L102-L108

As React accepts objects and runs `toString` implicitly this would be expected to work (hence marked as bug).

**To reproduce:**

See above

**Expected behavior:**

Example should render `<div class="_emotionserializedclass_ myclass" />`

Andarist:
This has been fixed in https://github.com/emotion-js/emotion/pull/1545

PR #1545, opened 2019-10-11
Accept objects as className in styled components

jgroszko:
**What**: Accept objects as className in styled components

**Why**: Matches react behavior per issue #1495 

**How**: Add a special case for objects that don't already have a 'styles' property

**Checklist**:

- [ N/A ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/afraid-dolls-speak.md
+++ b/.changeset/afraid-dolls-speak.md
@@ -1,0 +1,6 @@
+---
+'@emotion/styled-base': minor
+'@emotion/styled': minor
+---
+
+Accept objects as className in styled components
--- a/.changeset/afraid-dolls-speak.md
+++ b/.changeset/afraid-dolls-speak.md
@@ -3,4 +3,4 @@
 '@emotion/styled': patch
 ---
 
-Accept objects as className in styled components
+Accept objects as `className` in styled components (they are stringified) to match React behavior
--- a/.changeset/afraid-dolls-speak.md
+++ b/.changeset/afraid-dolls-speak.md
@@ -1,6 +1,6 @@
 ---
 '@emotion/styled-base': patch
-'@emotion/styled': minor
+'@emotion/styled': patch
 ---
 
 Accept objects as className in styled components
--- a/.changeset/afraid-dolls-speak.md
+++ b/.changeset/afraid-dolls-speak.md
@@ -1,5 +1,5 @@
 ---
-'@emotion/styled-base': minor
+'@emotion/styled-base': patch
 '@emotion/styled': minor
 ---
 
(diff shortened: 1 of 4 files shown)


packages/styled-base/src/index.js
```diff
@@ -106,6 +106,17 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
                 props.className
               )
             }
+            if (
```
Andarist: IMHO this should just be else if to the preceding if statement, like this:
```js
if (typeof props.className === 'string') {
  // ...
} else if (props.className != null) {
  className += props.className
}
```

jgroszko: Ahhh yes that does make sense

packages/styled-base/src/index.js
```diff
@@ -105,7 +105,17 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
                 classInterpolations,
                 props.className
               )
+            } else if (
```
Andarist: Maybe that wasn't clear - but in my last comment I've also included alternative solution to the "problem", the condition of this else if should be just `props.className != null` and the body of this shouldn't call `getRegisteredStyles`, like this:
```js
if (typeof props.className === 'string') {
  // ...
} else if (props.className != null) {
  className += `${props.className} `
}
```

jgroszko: Ahhh okay, I see what you mean

Andarist:
Seems like you have uncovered a mistake in tests. This line:
https://github.com/emotion-js/emotion/blob/a07f8a73d9b2a567dbb4e68c58d2c304ae9ea833/packages/styled/__tests__/styled.js#L395
should use `css` prop and not `className`.

.changeset/afraid-dolls-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/styled-base': minor
```
Andarist:
```suggestion
'@emotion/styled-base': patch
```

.changeset/afraid-dolls-speak.md
```diff
@@ -0,0 +1,6 @@
...
+'@emotion/styled': minor
```
Andarist:
```suggestion
'@emotion/styled': patch
```

.changeset/afraid-dolls-speak.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Accept objects as className in styled components
```
Andarist:
```suggestion
Accept objects as `className` in styled components (they are stringified) to match React behavior
```

Andarist:
Thanks!

PR #1547, opened 2019-10-11
Fix `composition of nested pseudo selectors` test

Andarist:
I'm actually not sure if this is how it should get fixed - because I'm not sure what the exact intention of this test is. Seems to me that the applied fix might be incorrect, because this doesn't any composition - just that 2 classes can be applied to the same element. This is what a "sibling" test does though:
https://github.com/emotion-js/emotion/blob/c3f0bc101833fff1ee4e27c7a730b821a7df4a15/packages/styled/test/composition.test.js#L142
So maybe actually both should be adjusted to use `jsx` + `css` prop?

### DIFF
--- a/packages/styled/__tests__/__snapshots__/styled.js.snap
+++ b/packages/styled/__tests__/__snapshots__/styled.js.snap
@@ -141,21 +141,33 @@ exports[`styled composition based on props 2`] = `
 `;
 
 exports[`styled composition of nested pseudo selectors 1`] = `
-.emotion-0 {
+.emotion-1 {
   font-size: 2rem;
   padding: 16px;
 }
 
-.emotion-0:hover {
+.emotion-1:hover {
   color: blue;
 }
 
-.emotion-0:hover:active {
+.emotion-1:hover:active {
   color: red;
 }
 
+.emotion-0:hover {
+  color: pink;
+}
+
+.emotion-0:hover:active {
+  color: purple;
+}
+
+.emotion-0:hover.some-class {
+  color: yellow;
+}
+
 <button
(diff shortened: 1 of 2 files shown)


Andarist:
I've wanted to tweak this according to the suggestion, but I've encountered rather a bizarre situation.

This test: https://github.com/emotion-js/emotion/blob/c3f0bc101833fff1ee4e27c7a730b821a7df4a15/packages/styled/test/composition.test.js#L142 is using `className` prop with emotion-generated class name and it still manages to "flatten" the final applied class name because of both emotion and styled using the same cache instance (even though 2 are created).

After some digging, I've found out that this is the culprit:
https://github.com/emotion-js/emotion/blob/1bb3efe399ddf0f3332187f3c751fbba9326d02c/scripts/test-utils/compat-render-json.js#L12

So it generally leads me to the conclusion that probably tests using this thing often are not testing how people are using the library by default. 

Could you describe what's the exact purpose of this compat-render-json?

mitchellhamilton:
compat-render-json exists because it was a quick way to migrate all the tests to Emotion 10.

Andarist:
Should we try to remove it now and fix the failing tests then?

mitchellhamilton:
If you want to, I don't mind, it's not really causing any problems so I wouldn't say it's a very high priority thing

Andarist:
Gonna look into it then - I hope it won't take much time 😅 This tripped me quite a bit now as I was not aware of this thing and had to debug this for some time and even after I've came to conclusion what must have happened I still had to figure out how this actually happens, looked through babel stuff etc.

PR #1548, opened 2019-10-11
Accept objects as `className` in jsx-created components (they are stringified) to match React behavior

Andarist:
Followup to https://github.com/emotion-js/emotion/pull/1545


Andarist:
Once PRs get approved can we merge them right away? Or do you prefer to keep them open for some reason?

mitchellhamilton:
You can merge them right away when they're approved, the usual reason why I don't is because I'm waiting on CI and get distracted

Andarist:
Thanks for the reply! Roger that 😉

PR #1553, opened 2019-10-20
Add dev warning about keyframes being interpolated into plain strings.

Andarist:
Arguably the preferred way to write this would be to wrap the thing in `css` call which can handle interpolated `keyframes` without falling back into regexp extraction. 

It's slightly weird to me though that this is allowed when using object styles but it is not when using plain strings. I can think of an argument for this (object values being shorter and not needing `css` wrapping beside this one thing), but this looks inconsistent as part of the API. So if we decide not to merge this because of performance reasons we should at least add a dev warning about this.

EDIT:// changed the PR to introduce a dev warning instead of regexp matching all strings

### DIFF
--- a/packages/core/__tests__/warnings.js
+++ b/packages/core/__tests__/warnings.js
@@ -2,7 +2,7 @@
 /** @jsx jsx */
 import 'test-utils/next-env'
 import css from '@emotion/css'
-import { jsx, Global } from '@emotion/core'
+import { jsx, Global, keyframes } from '@emotion/core'
 import renderer from 'react-test-renderer'
 
 // $FlowFixMe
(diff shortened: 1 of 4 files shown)


mitchellhamilton:
I think I'd rather have a dev warning about this because running this on all string styles. It's fine when it's only for the values of a couple properties on object styles but for every string style, I'd rather avoid it in production.

Andarist:
@mitchellhamilton adjusted

packages/core/__tests__/warnings.js
```diff
@@ -206,3 +206,39 @@ test('unterminated comments', () => {
...
+  )
+  expect(console.error.mock.calls).toMatchInlineSnapshot(`
+    Array [
+      Array [
+        "\`keyframes\` output got interpolated into plain string, please wrap it with \`css\`.
+
+    Instead of doing this:
+
+    const animation0 = {from,to{color:green;}50%{color:hotpink;}}
+    const animation1 = {from{transform:rotate(0deg);}to{transform:rotate(360deg);}}
```
mitchellhamilton: The keyframes should probably be wrapped in a keyframe call.

Andarist: right, right, had that in mind, forgot to put it into the code 😅 fixed!

PR #1565, opened 2019-10-23
All options work with unejected create-react-app

pcowgill:
**What**: Noting that all options work with an unejected `create-react-app` app.

**Why**: Devs new to emotion who want to use it with a zero-configuration project will be steered away from trying the "preferred" APIs for emotion with the current docs.

**How**: Simple edits to the markdown file

**Checklist**:

- [x] Documentation
- [n/a] Tests
- [n/a] Code complete
- [n/a] Changeset 

Feel free to call out any wording changes that could gate merging. Thanks!


pcowgill:
Think we can merge this one?

pcowgill:
@mitchellhamilton @Andarist Whenever you have time (I'm sure you're very busy!), I'd be curious for your thoughts on the changes in this PR. Thanks!

docs/introduction.mdx
```diff
@@ -20,7 +20,7 @@ The [emotion](https://www.npmjs.com/package/emotion) package is framework agnost
 
 - Requires no additional setup, babel plugin, or other config changes.
 
-- Works in situations where configuration is restricted or not possible such as with [Create React App](https://facebook.github.io/create-react-app)
+- Works in situations where configuration is restricted or not possible such as with [Create React App](https://facebook.github.io/create-react-app). However, it should be noted that with the current version of `Create React App` as of October 2019, the below methods also work out of the box for both the default and `--typescript` configurations.
```
Andarist: I'm not sure if this change is needed. Could you explain in what way it was confusing before the change?

docs/introduction.mdx
```diff
@@ -61,6 +61,8 @@ The ["@emotion/core"](https://www.npmjs.com/package/@emotion/core) package requi
 
 - Best when using React with a build environment that can be configured.
 
+- Also works well even if using React in a nonconfigurable build environment like an unejected `Create React App` app. This is true for the current version of `Create React App` as of October 2019. This method and the one below both work out of the box for the default and `--typescript` configurations of a new `Create React App` app.
```
Andarist: I think I would just merge this with the previous bullet point, smth like:
```
Best when using React with a build environment that can be configured, but also works...
```

And I'm not sure how the second part of this point:
> This is true for the current version of `Create React App` as of October 2019. This method and the one below both work out of the box for the default and `--typescript` configurations of a new `Create React App` app.

is relevant.

PR #1566, opened 2019-10-23
Upgrade flow

Andarist:
Q: do we want to keep using flow?

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -398,6 +398,7 @@ export const serializeStyles = function(
   let name = hashString(styles) + identifierName
 
   if (process.env.NODE_ENV !== 'production') {
+    // $FlowFixMe SerializedStyles type doesn't have toString property (and we don't want to add it)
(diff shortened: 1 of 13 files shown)


packages/serialize/src/index.js
```diff
@@ -398,6 +398,7 @@ export const serializeStyles = function(
   let name = hashString(styles) + identifierName
 
   if (process.env.NODE_ENV !== 'production') {
+    // $FlowFixMe SerializedStyles type doesn't have toString property (and we don't want to add it)
```
Andarist: I've encountered this problem while consuming emotion in flow-based project, this has lead me to upgrading flow here, BUT, weirdly, this error doesn't show up when running upgraded flow in our repository. The error makes sense though as this has been typed to return exact object shape - so not sure why this is not caught in our repo, but still it was worth fixing.

FezVrasta:
Yes please! Keep Flow! It’s the only sound type system we have

PR #1569, opened 2019-10-25
Add dev hint about css object (generated by Babel) being stringified by accident

Andarist:
Followup to https://github.com/emotion-js/emotion/pull/1219 . I haven't noticed before that Babel can generate this object too, so it bypasses the runtime serialization and thus it doesn't have this warning attached to it.


packages/babel-plugin-emotion/__tests__/source-maps/__snapshots__/index.js.snap
```diff
@@ -129,6 +136,7 @@ const SomeComponent = styled.div({
       ↓ ↓ ↓ ↓ ↓ ↓
 
 import _styled from \\"@emotion/styled-base\\";
+var _EMOTION_STRINGIFIED_CSS_ERROR__ = \\"You have tried to stringify object returned from \`css\` function. It isn't supposed to be used directly (e.g. as value of the \`className\` prop), but rather handed to emotion so it can handle it (e.g. as value of \`css\` prop).\\";
```
mitchellhamilton: Shouldn't this be a function?

Andarist: Yes, it should - stupid mistake 🤦‍♂

packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
```diff
@@ -91,10 +98,29 @@ export let transformExpressionWithStyles = ({
...
+          )
+          state.emotionStringifiedCssId = uid
+          const cssObjectToString = t.functionDeclaration(
+            uid,
+            [],
+            t.blockStatement([
+              t.returnStatement(t.stringLiteral(CSS_OBJECT_STRINGIFIED_ERROR))
+            ])
+          )
+          cssObjectToString._compact = true
```
mitchellhamilton: What's this?

Andarist: A hint for babel generator, in this case it doesnt matter much - can remove it if you prefer to omit this. 

It basically tells generator to print the statement in a single line, so instead of
```js
function a() {
  return 42
}
```
it prints
```
function a() { return 42 }
```

Babel is using this flag when inserting helpers.

PR #1570, opened 2019-10-25
fix: refactor styled-base Flow types to work again

### LINKED ISSUE #1515: styled components typed as "any"
**Current behavior:**

Right now the styled components are annotated with `any`, I was wondering why this decision was made and if it's something you'd be interested in fixing? I'm referring to the Flow types.

**Expected behavior:**

I would expect the generated styled components to have a `React$ComponentType` type rather than `any`.

**Environment information:**

- `react` version: 16.9.0
- `emotion` version: 10.0.7

FezVrasta:
**What**:
work needed to fix #1515

**Why**:
right now styled components are typed as `any`

**How**:
refactored the Flow types of styled-base to make sure they return the proper types

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


packages/styled-base/src/utils.js
```diff
@@ -1,9 +1,31 @@
...
 
 export type Interpolations = Array<any>
 
+export type StyledOptions = {
+  label?: string,
+  shouldForwardProp?: string => boolean,
+  target?: string
+}
+
+export type StyledComponent = {
```
FezVrasta: I'm not sure this is the best way to define this type, but I'm unable to extend the built-in component types.

packages/styled-base/tests/flow.js
```diff
@@ -0,0 +1,8 @@
+// @flow
```
FezVrasta: This file is meant to be checked by Flow, not to be run by Jest.

Andarist: could you rename this directory to smth like `flow-tests`? `tests` is too generic and we probably want to keep flow-related type tests in a separate directory - so its name should be intuitive

packages/styled-base/src/utils.js
Andarist: yeah - had the same issue when I've tried to handle this once I've found the culprit, couldnt just add `withComponent` to `StatelessFunctionComponents` and others

packages/styled-base/src/index.js
```diff
@@ -221,7 +217,7 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
       )(...styles)
     }
     return Styled
-  }
+  }: CreateStyledComponent)
```
Andarist: is this extra cast needed? Styled is being returned already so it should be an indicator of the containing function's type. Maybe it's just me but this looks a little bit strange

packages/styled-base/src/utils.js
```diff
@@ -1,9 +1,31 @@
...
+  (Interpolations): React$Node,
+  withComponent: (
+    nextTag: ElementType,
+    nextOptions?: StyledOptions
+  ) => StyledComponent,
+  displayName: string,
+  defaultProps: any,
+  __emotion_real: StyledComponent,
```
Andarist: IMHO those `__emotion` properties should be hidden from the user. Instead of adding them here to the declared type we can just cast Styled to any when assigning those.

Andarist:
What do you mean by "part of the work needed to fix #1515"? What has to be done to consider that ticket done other than this PR?

packages/styled-base/src/utils.js
FezVrasta: I’d argue that Flow types are not only for the consumers, they should help also the Emotion maintainers to write tipe safe code, so “any” should never be used

packages/styled-base/src/index.js
FezVrasta: No I just added it for clarity, I can remove it

packages/styled-base/src/utils.js
Andarist: those are literally used in a single file and are not supposed to be used outside of this, so I would say it's better to hide them from the outside world to avoid misusage

FezVrasta:
I still got “any” out from the styled package, but maybe it’s just some local issue with the monorepo

Andarist:
> I still got “any” out from the styled package, but maybe it’s just some local issue with the monorepo

Could you try building & linking packages to your repro case? That way you could check how this would behave if this would get published

packages/styled-base/src/utils.js
FezVrasta: Now we expose a type without them, but internally we keep still type check them, so we get the best of both worlds.

.changeset/red-chefs-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled-base': patch
```
Andarist: this probably should include `@emotion/styled` package as well as they both are tightly coupled to each other

packages/styled-base/src/utils.js
```diff
@@ -1,9 +1,32 @@
...
 
 export type Interpolations = Array<any>
 
+export type StyledOptions = {
+  label?: string,
+  shouldForwardProp?: string => boolean,
+  target?: string
+}
+
+export type StyledComponent<P> = React$StatelessFunctionalComponent<P> & {
```
Andarist: interesting, did u manage finally to extend this builtin type? 🤔 Also - shouldn't we refer to this as `React.StatelessFunctionalComponent`?

FezVrasta:
I updated the types, now StyledComponent is a polymorphic type (like React$StatelessFunctionalComponent) so that you can use it as follows:

```js
type Props = { color: string };
const Foo = styled<Props>`
`;
```

This has the advantage that it makes it easy to catch errors inside your styled components, but has the disadvantage of requiring the type in all cases.

I would like to make it optional, so that it fallbacks to `any`, but I can't seem to figure out how to do it. Ideas?

packages/styled-base/src/utils.js
```diff
@@ -1,9 +1,32 @@
...
+  defaultProps: P,
```
Andarist: is this correct? shouldnt `Partial<P>` equivalent from flow be used here?

FezVrasta: Yeah probably this should be `$Shape<P>`

packages/styled-base/src/utils.js
FezVrasta: Yeah I don't know what happened, it works now 😂

I can change it, but just FYI, internally, `React.StatelessFunctionalComponent` aliases to `React$StatelessFunctionalComponent`

Andarist:
> I would like to make it optional, so that it fallbacks to any, but I can't seem to figure out how to do it. Ideas?

Isn't requiring type at all times (as its hard to infer it) better than falling back to any?

packages/styled-base/src/utils.js
Andarist: Yeah - I know that, but `React$StatelessFunctionalComponent` looks like accessing something private :p

FezVrasta:
> Isn't requiring type at all times (as its hard to infer it) better than falling back to any?

Yes it's absolutely better, but it's kind of a breaking change, because now all the untyped styled components will require the annotation. Also, if I don't expect to assign any property to my styled component and I plan to use it only internally of some larger component I may not want to define a type for each of them, like it happens when I define a internal component `props => <div {...props} />` and Flow simply infers its type rather than asking me to define it.

packages/styled-base/src/utils.js
FezVrasta: I wasn't unable to use `$Shape<P>` because that would then get assigned to `P` and throw a type error.

It looks like Flow doesn't support `defaultProps`, I typed it as `any` for now :-(

https://github.com/facebook/flow/issues/1660

FezVrasta:
I guess I hit a road blocker... Flow doesn't seem to support type parameters on tagged template strings 😢 

So that means we aren't able to do:

```
const Foo = styled.div<Props>`
  color: red;
`;
```

https://github.com/facebook/flow/issues/7527

FezVrasta:
I managed to make the polymorphic type argument optional by using function overload, now I just need to add back the extra static properties and I'd say we are good to go!

packages/styled/src/index.js
```diff
@@ -5,7 +5,7 @@ import { tags } from './tags'
 // bind it to avoid mutating the original function
 const newStyled = styled.bind()
 
-tags.forEach(tagName => {
+tags.forEach((tagName: string) => {
```
Andarist: shouldnt this be easily inferrable?

Andarist:
@FezVrasta let me know when you are done with this PR - I'll do a code review then and we'll be able to merge this in

packages/styled/src/index.js
FezVrasta: It does't infer it for some reason, I see it marked as `any`

FezVrasta:
I don't understand why if I write the following code in `styled-base/src/index.js` Flow is happy about the missing function type annotation, but if I do the same in a different file it complains :-(

```
const styled = createStyled('div')

// OK
const Ok = styled<{ color: string }>({ color: props => props.color })

const Ko = styled({ color: props => props.color })
```

packages/styled/src/index.js
FezVrasta: Got it, `tags.js` didn't have the flow comment header

FezVrasta:
Oh damn it, Flow is right...

The test fails because I'm trying to export `styled({ color: props => props.color })`, if I just assign it without exporting it everything works fine.

FezVrasta:
Ok @Andarist I think we are ready for a final review.

Notable unresolved/unaddressed issues:

1. Flow doesn't support generics on tagged templates
2. I haven't added type checking on the `${props => props.color}` calls inside CSS declarations (I don't know how they are called)

packages/styled-base/src/index.js
```diff
@@ -204,7 +202,7 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
         ) {
           return 'NO_COMPONENT_SELECTOR'
         }
-        // $FlowFixMe
+        // $FlowFixMe: coherce undefined to string
```
Andarist: nit: it's "coerce"

Andarist:
@FezVrasta thanks!

bradringel:
So, this change has caused our flow error count to jump up to around 800 after getting this version bump in our project. Any suggestions for how to adopt this type of change in a large project. We're also using `emotion-theming`, so a lot of the times the props for our styled components come from the theme context, and many of our styled components are created with the tagged template literal version of styled.  This was a rough change to introduce in a patch version bump

Andarist:
@bradringel could you reduce your failing call sites to small repro cases? when fixing those we could include them as "flow tests".

bradringel:
@Andarist sure. A super basic one like this fails:

```
//@flow

export const UserName = styled.div`
  color: ${props => props.theme.colors.primary};
`;
```
Fails with the error 
image


typing the component with `any` fixes it, but that's a lot of components to go add `any` to:

```
//@flow

export const UserName:any = styled.div`
  color: ${props => props.theme.colors.primary};
`;
```
doesn't have an error

Andarist:
Isn't the problem here that previously such component had "props" typed to `any` implicitly and it was type-unsafe to use "props"? The situation here, even if it causes more flow errors, gets improved in this regard. This could be treated as a bug fix.

bradringel:
That's true, it was implicitly unsafe, but now we need to go and fix 800 of these types of errors. And because flow doesn't support generics for tagged template literals, which is how a lot of those components are written, they're all just going to get typed to `any`. 

This is technically a safer way to write those components, but is a big change to the way we need to interact with this library, and was introduced in a patch version.

armount:
Let me preface this by saying we are fans of Emotion - we use it all over our project.  So, please don't take this as an attack, I'm just looking for advice for suggested usage in Flow situations.  I could be missing something really easy (it wouldn't be the first or last time), but it feels like Flow users are in for a wild time if no version of tagged templates is supported.

Is your position that we should be doing this instead?
```
// some utility file having these types
export const Div = styled('div');
export const Span = styled('span');

// some other file
export const MyComponent = Div<Props>({ ... props based object literal here... });
```

Our whole project uses both of the tagged template forms below (I realize, unsafe), and both do not pass flow checks if I try to type the 'styled' call with a generic.  We export them because we like to separate all of our styled components which form our 'building blocks' from the complex components that use them instead of lumping them together in the same file. Using the tagged template versions of these calls let us write css-like implementations instead of JSON-like structures, which was also really appealing:

```
export const MyComponent  = styled('div')` ...css style props here...`;
export const MyComponent = styled.div` ...css style props here...`;
```

I feel like the answer, "yes, you can use tagged template forms of the calls but you will be forced to type them all to keep Flow from complaining"  is a lot different than "you can't use this supported feature even if you type it unless you want to live with Flow complaining until they support tagged templates".  The former is a lot less work than the latter, even if both are time consuming to some degree in a large project. That's where I agree with @bradringel that this feels like a major change, not a minor version bump.  I think I'll wait to hear what you suggest and pin to the previous version in the meantime.  Based on the answer, we'll plan to make these changes as part of tech debt after our release deadline is met.

Thanks for taking the time to read and respond.  I do not imagine being a maintainer of a popular OSS project is easy - much respect.

FezVrasta:
I made some further improvements to the Flow types in this PR https://github.com/emotion-js/emotion/pull/1588

It’s currently in the next branch.

I’m from mobile but once I get back on my computer I’ll try to provide a more complete answer

Andarist:
I see that on the `next` branch this is supported:
https://github.com/emotion-js/emotion/blob/923ded01e2399a242206d590f6646f13aba110e4/packages/styled/flow-tests/index.js#L18-L20
So you should be able to work around Flow not accepting generics on tagged template calls. I'm not sure if this supported with types on `master` though, could you check this for me?

I understand the inconvenience caused by this and I'm truly sorry for that. I believe that it's not practical to tie types to semver closely though (when it makes sense to use we try to do that though, in example we've pushed TS types refactoring to `next` so it doesn't affect latest releases). It's just super hard to do - with 2 type systems that evolve over time and make breaking changes between their versions.

> That's true, it was implicitly unsafe, but now we need to go and fix 800 of these types of errors.

If those were previously unsafe and typed to any I could help you writing a codemod that would just make those explicit in your code.

PR #1572, opened 2019-10-25
Fixed issue with cache created in render crashing SSRed site

### LINKED ISSUE #1468: Issue when using CacheProvider in Gatsby (SSR)
**Current behavior:**

I'm building a gatsby website with emotion. I have a custom ui library with a wrapper that looks something like this:

```jsx
const MyWrapper = ({ children, key = "stl" }) => {
  const myCache = createCache({ key })
  // I also have a custom container setup and some global styles but this is enough to reproduce the issue
  return <CacheProvider value={myCache}>{children}</CacheProvider>
}
```

With the follow landing page on my gatsby application:
```jsx
  <MyWrapper>
    <div
      css={css`
        color: green;
      `}
    >
      Hello world!
    </div>
  </MyWrapper>
```

I get the following error when I use the built version of the application: 
Screenshot 2019-08-11 at 13 38 33

If you disable javascript, the page loads and looks fine so I'm guessing something is going wrong when rehydrating. It looks like it's happening in react-dom so I don't think it's related to gatsby. Feel free to close the issue if you think otherwise.

Maybe related: https://github.com/emotion-js/emotion/pull/1396

Thank you for your help 🙏 

**To reproduce:**

1. `git clone https://github.com/kombucha/gatsby-emotion-bug.git`
2. `yarn`
3. `yarn serve` (this will build and serve the gatsby website, ie not development mode)
4. Go to http://localhost:9000/. You should see an error.

**Expected behavior:**

It should rehydrate the application without error.

**Environment information:**

- `react` version: 16.9
- `emotion` version: 10.0.15
- `gatsby-plugin-emotion` version: 4.1.2

Andarist:
fixes #1468 

This is currently just a proof-of-concept. I'd like to hear out the review before proceeding with this further.

Code like this:
```js
const MyWrapper = ({ children, key = "stl" }) => {
  const myCache = createCache({ key })
  // I also have a custom container setup and some global styles but this is enough to reproduce the issue
  return <CacheProvider value={myCache}>{children}</CacheProvider>
}
```
is crashing SSRed site at runtime because React gets confused a lot by this as most likely it had already chance to recognize style tags as being rendered, but we move those styles inside of render (so during rehydration).

This is not a backward-compatible solution yet, because I had to change the rendered data attribute and the change affects more than one package so it's not self-contained. Thoughts on that? We can try to handle both cases at runtime (old and new one)

**A little bit more detailed explanation**

The issue was caused by server/client mismatch - React has started rendering (and thus seen the current state of the DOM with style tags rendered besides regular components). Moving style tags to a final container (document.head by default) happen inside `createCache`, so it was too late in the case of a cache being created in the render. At least that was what I've concluded from all of this.

I've tried to recreate repro case on CodeSandbox but I have no luck - all I can get is a mismatch error: https://codesandbox.io/s/jolly-jang-idu3r . I've also tried using production react-dom there but it didn't change anything.

I'm not entirely sure how React treats mismatched content but it seems like it tries to patch things up in this scenario - and as it has seen style as part of the rendered output (and that it's actually first child) it tries to:
- add a copy of the "regular" child (a div with "Hello world" in this example), so if we pause with a debugger when React commit its mutation effects we can see for that moment that there are 2 the same children visible
- removes the "seen" style tag - and it crashes on this step as it has already been moved by a side effect in `createCache` and the style tag has changed a parent (surprisingly for React)


packages/cache/src/index.js
```diff
@@ -41,6 +41,14 @@ let getServerStylisCache = isBrowser
       }
     })
 
+let ssrStyles
+if (isBrowser) {
+  ssrStyles = document.querySelectorAll(`style[data-emotion]`)
+  Array.prototype.forEach.call(ssrStyles, (node: HTMLStyleElement) => {
+    document.head.appendChild(node)
```
Andarist: we need to move those before `render`, head seems like a "safe place"

packages/cache/src/index.js
```diff
@@ -68,14 +76,17 @@ let createCache = (options?: Options): EmotionCache => {
...
+        inserted[attrib[i]] = true
+      }
 
-    Array.prototype.forEach.call(nodes, (node: HTMLStyleElement) => {
-      const attrib = node.getAttribute(`data-emotion-${key}`)
-      // $FlowFixMe
-      attrib.split(' ').forEach(id => {
-        inserted[id] = true
-      })
       if (node.parentNode !== container) {
```
Andarist: we might need to reparent here what already got moved previously to head if the container is a custom one

Andarist:
Ok, I'll try later adding backward compatibility and some tests. Thanks for the feedback.

mitchellhamilton:
I think we might want to wait for v11 to do this because otherwise we have to decide between elements having different attributes which would be bad for users expecting them to have the attributes they have right now or writing both attributes which would mean larger SSR payloads.

Andarist:
Yeah, you are probably right. I've created v11 milestone and added this and a bunch of other stuff to it. The list is certainly not complete yet, I need to go through all issues first before I call it complete - but if you have anything you'd like to add there, please feel free to do so 😉

JakeGinnivan:
I am not sure this is a good idea, it makes emotion modules not pure because they effect things outside of themselves. This will affect tree shaking etc. I don't know how the automatic hydration works now though, so this may already be an issue.

Also wouldn't the example cause issues due to re-creating the cache on each render. If you wanted to support this scenario in addition to below you need to memoize createCache based on the key.

I have a few ideas on how this should be resolved

## 1. Introduce a factory to create MyWrapper
This doesn't affect emotion at all.

``` jsx
const createMyWrapper = key => {
  const myCache = createCache({ key })
  const MyWrapper = ({ children }) => {
    return <CacheProvider value={myCache}>{children}</CacheProvider>
  }
  return MyWrapper
}

const MyWrapper = createMyWrapper('stl')

export default () => (
  <MyWrapper>
    <Component />
  </MyWrapper>
)
function Component() {
  return <div css={css`
        color: green;
      `}>
    Hello world!
  </div>;
}
```

## 2. Update the hydrate function in create-emotion to do the hoisting to head
This also requires you rendering the `<style data-emotion={key} />`. It currently gets created as a side effect in `insert` => createStyleElement

Best place for this seems to be inside jsx.js, you could render an additional style element then mark it as inserted on the cache?

Then hydrate could look something like this:

```
function hydrate() {
  if (typeof document === 'undefined') {
    return
  }
  const prefixNodes = document.querySelectorAll('style[data-emotion]')
  Array.prototype.map.call(prefixNodes, prefixNode => prefixNode.getAttribute(`data-emotion`))
    .forEach(prefix => {
      const nodes = document.querySelectorAll(`style[data-emotion-${prefix}]`)
      Array.prototype.forEach.call(nodes, node => {
        document.head.appendChild(node)
      })
    })
}
```

JakeGinnivan:
Example code for option 2

```
const MyWrapper = ({ children, key = "stl" }) => {
  // This one won't work (if you comment out the other one)
  const myCache = React.useRef(() => createCache({ key }))
  // console.log(myCache, cacheLookup)

  return <Fragment>
    <style data-emotion={key} />
    <CacheProvider value={myCache.current}>{children}</CacheProvider>
  </Fragment>
}

function hydrate() {
  if (typeof document === 'undefined') {
    return
  }
  const prefixNodes = document.querySelectorAll('style[data-emotion]')
  Array.prototype.map.call(prefixNodes, prefixNode => prefixNode.getAttribute(`data-emotion`))
    .forEach(prefix => {
      const nodes = document.querySelectorAll(`style[data-emotion-${prefix}]`)
      Array.prototype.forEach.call(nodes, node => {
        document.head.appendChild(node)
      })
    })
}

hydrate()
export default () => (
  <MyWrapper>
    <Component />
  </MyWrapper>
)
function Component() {
  return <div css={css`
        color: green;
      `}>
    Hello world!
  </div>;
}
```

Andarist:
> I am not sure this is a good idea, it makes emotion modules not pure because they effect things outside of themselves. 

This definitely is a downside of this. 

> This will affect tree shaking etc.

Not that much, the side effect is short and concise - but yeah, in general this eager call cannot be tree shaken. 

> I don't know how the automatic hydration works now though, so this may already be an issue.

Automatic hydration happens at the moment inside of the `createCache` call - here:
https://github.com/emotion-js/emotion/blob/ed0a460a4f5c58ccf56c27c8a11d85a507c85055/packages/cache/src/index.js#L68-L83

For this to work automatically each request has to have its own unique cache because we need to track which exact rules are needed for the current SSR output and we want to avoid duplication when the same rule gets used multiple times. This cache gets created~ automatically here (it's created by BasicProvider):
https://github.com/emotion-js/emotion/blob/ed0a460a4f5c58ccf56c27c8a11d85a507c85055/packages/core/src/context.js#L58-L72

> Also wouldn't the example cause issues due to re-creating the cache on each render.

So as described above - this already happens (on the server-side with automatic SSR).

JakeGinnivan:
> Also wouldn't the example cause issues due to re-creating the cache on each render.

Understand on the server, it's only a single pass. But what about on the client? If the user is calling createCache inside a render method, I can't see any caching of the return value anywhere?

Andarist:
> Understand on the server, it's only a single pass. But what about on the client? If the user is calling createCache inside a render method, I can't see any caching of the return value anywhere?

The client is using different strategy - it uses plain default cache from context's default value and doesn't recreate it on each render (when not customizing cache etc).

JakeGinnivan:
Yeah, but the issue was creating a custom cache with a different key during render?

mitchellhamilton:
A consumer using a custom cache on the client should cache it themselves with `useMemo`

JakeGinnivan:
Not sure where we are at.

But IMO I don't think this change is needed. The pattern used in the original issue is broken for a few reasons. 

I think you should just add some docs that caches should not be created during rendering if using SSR. There are heaps of ways to solve the issue once you put that constraint on.

mitchellhamilton:
Why is the pattern in the original issue broken besides the fact that the creation of the cache wasn’t memoized? 

It’s also important to call out that we need to move some style tags on initial load always anyway because of the default cache so this doesn’t really cause any more code to run

JakeGinnivan:
> Why is the pattern in the original issue broken besides the fact that the creation of the cache wasn’t memoized?

Memoizing the cache creation still causes the original issue because style tags are being removed during a render.

The default cache works because this is done when you load the library, not during initial hydration. 

I don't mind this change, as long as it doesn't impact extract critical

mitchellhamilton:
This PR moves all the tags to the head before render so that’s not an issue since at most, the styles tags will be moved from the head to somewhere else and importantly they won’t be in the tree managed by React so it won’t cause any problems.(though in most cases they’ll stay in the head)

JakeGinnivan:
The docs would just need to include memoizing the cache (or building this in), and instructions on how to hydrate if that isn't done automatically.

Andarist:
I think we all here understand the core issue (based on recent comments) - but some older comments look a little bit confused, so I've included a little bit more thorough explanation in the original post.

After some thinking about this, I'm more convinced now that we shouldn't merge this. This requires some side-effectful code being executed in a top-level scope during initialization and I would consider this to be an anti-pattern. This can never be eliminated by minification tools - which is not that big of a deal but indicates a code smell. The whole thing of moving style tags around is already a hack (a clever one) and I don't think we need to make it even hackier. Its purpose is to provide a 0 config SSR, but when one creates a custom cache it already is past the "0 config" point. It would, of course, be better if things would just work under all possible circumstances - but in here we really have received a single report about this in 1 year in which emotion@10 is out, so it seems like documentation around this can be enough.

We could try to think how to warn people at runtime about this better - one thing that comes to my mind that we could track in dev jsx calls (so basically rendering) and `createCache` calls to determine if `createCache` has been called during render or before it. 

Other than that we could also propose a solution for this as it seems like a use case itself is valid. It might in some edge case scenarios be easier to create cache during render (where one has access to props etc) rather than per request (although it's always possible to create it per request). A solution would be to just move those style tags "manually" to head or any other destination container of choosing (we could make it easier by exposing a helper for this - which we would use ourselves internally as well)

mitchellhamilton:
We already have to move style elements for the default cache to the head though so this only means that we have to move some more style elements so I don't really see the problem?

Andarist:
What we do currently is a little bit hacky (although clever) - but it's at least contained within a call that returns something. This has potential (mostly theoretical) of being removed by DCE. 

What I don't like the most after sitting on this PR for a bit that it's guesswork - we try to move **everything** into the head and later we try to correct ourselves when this is not what was supposed to happen. The current solution is at least always targeted from the start.

Andarist:
@mitchellhamilton any more thoughts on this one? How do we want to proceed?

mitchellhamilton:
I'm still of the opinion that we should do this.

> What we do currently is a little bit hacky (although clever) - but it's at least contained within a call that returns something. This has potential (mostly theoretical) of being removed by DCE.

I don't think this really matters since we will always create a default cache so we're always going to be running this code

> What I don't like the most after sitting on this PR for a bit that it's guesswork - we try to move everything into the head and later we try to correct ourselves when this is not what was supposed to happen. The current solution is at least always targeted from the start.

While, yes, it will sometimes be wrong, I think it's good enough because in most cases the container option is used for iframes rather than a custom container in the same page so it won't matter

Andarist:
> I don't think this really matters since we will always create a default cache so we're always going to be running this code

The problem I see is about transitive dependencies. Imagine a situation like this - a component library built without emotion but with its select component built on top of `react-select` which in turn depends on `@emotion/core`. And a consumer of that component library importing just a Button - ideally in this situation no `react-select` & `@emotion/core` code should land in the final bundle. Right now top-level side effects of `@emotion/core` will have to land in it though and this particular thing can't be optimized away with `#__PURE__`. Technically, maybe, `sideEffects: false` in our package.json could help with that but I don't trust this particular hint - its semantics are not set in stone nor described comprehensively anywhere and it's easy for some tool to implement it differently than existing tools which could break us.

**Maybe** we could wrap this side effect inside a `createDefaultContext` which could be annotated with `#__PURE__` and would only be retained in the bundle if emotion gets used. I still rather believe it would be better to split those responsibilities though (and require a user to call an extra function upfront) - it's a little bit too magical for me that a "create*" method would perform such a side effect.

mitchellhamilton:
I suppose we could do the `createDefaultContext` thing but I don't really think this is a really a big issue, if `@emotion/core` is imported by yourself or a dependency it is acceptable for some code to be loaded. For your example, I would say that the real issue is that a Button and Select are exported by the same module.

Aside - Is your primary concern about tree shaking or is there another concern?

Andarist:
> I suppose we could do the createDefaultContext thing but I don't really think this is a really a big issue, if @emotion/core is imported by yourself or a dependency it is acceptable for some code to be loaded. For your example, I would say that the real issue is that a Button and Select are exported by the same module.

Partially it's true - I just believe that we could use such module structures that it wouldn't matter. The tree-shaking is a really nice goal - so consumers don't have to worry about this kind of stuff. It's not perfect though - especially given dynamic JS natures, but we, as module authors, can help it do its job in a perfect way. It's not always possible - but if we can and we know how those things work then we should strive for a structure allowing it to work.

> Aside - Is your primary concern about tree shaking or is there another concern?

After thinking about it, I'm also not a fan of the overall approach - it seems too magical too me. I don't intend to die on this hill alone though, so if you believe it's good etc we could proceed with it. I rather regret that I've implemented this in the first place though 😉 If we are going to proceed with it I would also need to know if we want to do those hoops with `createDefaultContext` or is it too much for your taste?

mitchellhamilton:
I'd still like to do this. I don't mind much about the createDefaultContext thing, if you want to do it, sure, if not, that's fine too.

packages/core/src/global.js
```diff
@@ -72,7 +72,7 @@ export let Global: React.AbstractComponent<
     return (
       <style
         {...{
-          [`data-emotion-${cache.key}`]: serializedNames,
+          [`data-emotion-global`]: `${cache.key} ${serializedNames}`,
```
Andarist: had to differentiate this with `-global` so it doesnt get rehydrated into "main" sheet

packages/core/src/global.js
```diff
@@ -119,14 +111,14 @@ export let Global: React.AbstractComponent<
...
+
       let sheet: StyleSheet = ((sheetRef.current: any): StyleSheet)
-      if (sheet.tags.length) {
-        // if this doesn't exist then it will be null so the style element will be appended
-        let element = sheet.tags[sheet.tags.length - 1].nextElementSibling
-        sheet.before = ((element: any): Element | null)
+      sheet.before = cache.sheet.tags.length ? cache.sheet.tags[0] : null
+      cache.insert(``, serialized, sheet, false)
+
+      return () => {
```
Andarist: refactored those 2 hooks a little bit, as discussed briefly on Slack, an extra careful review is advised here

Andarist:
@mitchellhamilton I've roughly adjusted the code. Please review with extra care as this might be considered a major breaking change - it changes all `data-emotion`~ attributes.

packages/core/src/context.js
```diff
@@ -5,15 +5,27 @@ import { useContext, forwardRef } from 'react'
...
 
-let EmotionCacheContext: React.Context<EmotionCache | null> = React.createContext(
+const createCacheContext = () => {
+  if (isBrowser) {
+    const ssrStyles = document.querySelectorAll(`style[data-emotion]`)
+    // get SSRed styles out of the way of React's hydration
+    // document.head is a safe place to move them to
+    Array.prototype.forEach.call(ssrStyles, (node: HTMLStyleElement) => {
+      ;((document.head: any): HTMLHeadElement).appendChild(node)
+    })
```
mitchellhamilton: I think this should happen inside `@emotion/cache` since otherwise this will be broken for `emotion`/to be renamed to `@emotion/css`

packages/core/__tests__/__snapshots__/global.js.snap
```diff
@@ -3,40 +3,40 @@
...
-    html{background-color:hotpink;}
+    @-webkit-keyframes animation-ocj8pk{from,to{color:green;}50%{color:hotpink;}}
   </style>
   <style
-    data-emotion="css-global"
+    data-emotion="css"
   >
     
-    h1{-webkit-animation:animation-ocj8pk 1s;animation:animation-ocj8pk 1s;}
+    @keyframes animation-ocj8pk{from,to{color:green;}50%{color:hotpink;}}
```
mitchellhamilton: This reordering of things worries me, could we revert the global refactoring since from what I understand, it's solving a bug or anything, only size/code niceness, right?

packages/core/src/context.js
Andarist: How could it affect vanilla `emotion`? This thing here only fixes moving "inline" style tags (produced by zero-config SSR) - vanilla `emotion` is not affected by this as it doesn't produce such style tags, so it should be safe to have this in `@emotion/core`

packages/core/__tests__/__snapshots__/global.js.snap
Andarist: This change could have fixed a slight bug with `sheet.before` - but I need to give it slightly more thought. Probably better to move this out of this PR so I've reverted this change for now.

packages/core/src/context.js
mitchellhamilton: You can use `renderStylesToString` and `renderStylesToNodeStream` with vanilla emotion which both produce inline style tags

Andarist:
After some more thinking I have one more concern about this (although it's not strictly exclusive to this change, it relates to it).

Because of one of the latest changes - https://github.com/emotion-js/emotion/pull/1696/files#diff-af18f1f9e18306ed3fc3d28811507a87R5 - cache/sheet **owns** its SSRed styles, which is a good thing because they are just another styles, only inserted to the document in the other point in time (SSR). Given this fact, we shouldn't allow for a situation where a single style tag is owned by multiple sheets/caches and right now it happens by default - because the same `key` (`css`) is used for 2 default caches (one in `emotion`, one in `@emotion/core`). We cannot escape associating keys with sheets/caches, CSS is global and such - but we should at least try to warn users when 2 caches are created with the same `key`.

I see 2 options how we could handle this:
- those 2 default caches should have different keys
- there should be a single default cache with `css` key, used by both libraries - but this would force us to implement couple of stuff differently (like `.compat` handling etc)

mitchellhamilton:
I'm sure I really understand the problem with style tags being "owned" by multiple sheets/what problems it would cause?

Andarist:
Flushing is the problem. One cache might call flush and it might remove styles inserted by a different one.

mitchellhamilton:
I think that those problems are acceptable since flushing is a pretty rare thing and hijacking styles into a custom container shouldn't be a problem since you'll only use a cache with a custom container in one of two scenarios:

- an iframe which means you're not server rendering and it's a different document
- a custom cache in the same page with different stylis plugins, or etc in which case you should be using a custom key

mitchellhamilton:
Idea: what if we made key a required option on the cache

Andarist:
Yes, this is pretty much my thinking - with the addition of introducing dev-only warnings about duplicate keys (maybe just for browser envs as on the server caches are recreated each time). That would also kinda imply that maybe we should use different keys for `@emotion/react` and `@emotion/css`.

Andarist:
@mitchellhamilton Ok, I've rebased and reimplemented this slightly. Would appreciate a re-review.

Andarist:
@mitchellhamilton good catch with that options argument! anything else we'd like to change here? or is this ready to merge?

Andarist:
I've added changesets & I'm merging this in, but please take a look at them.

PR #1575, opened 2019-10-27
Move @emotion/styled-base to @emotion/styled/base

mitchellhamilton:
**What**:

Move @emotion/styled-base to @emotion/styled/base

**Why**:

So that package authors who are using @emotion/styled and babel-plugin-emotion don't need to also add @emotion/styled-base to their dependencies.(if they didn't, their packages would not work in more strict package managers/bundlers)

**How**:

Use a Preconstruct entrypoint

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset


FezVrasta:
That’s great! Thanks

packages/styled/base/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "main": "dist/styled.cjs.js",
+  "module": "dist/styled.esm.js",
+  "umd:main": "dist/styled.umd.min.js",
+  "browser": {
+    "./dist/styled.cjs.js": "./dist/styled.browser.cjs.js",
```
Andarist: i'm not entirely sure about it right now - but I vaguely remember at least some tools ignoring this alias field for extra entries, and that I had to include those redirects in root package.json

Andarist:
@mitchellhamilton I think this is nearly done, just surprisingly some flow errors popped up. I've asked @FezVrasta for help with this.

PR #1580, opened 2019-10-28
Warn about  being used as object style's key

### LINKED ISSUE #1097: emotion macro returns undefined when used as selector
when using the babel macro for `@emotion/styled/macro` the element returns an `undefined` selector it should either throw if it cannot be found or return the right selector.

e.g.

``` js
import styled from '@emotion/styled/macro';

const ListItem = styled.li();
const List = styled.ul({
    [ListItem]: {
        color: 'green'
    }
});
```

Andarist:
closes #1097

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -13,6 +13,9 @@ Because you write your CSS inside a JavaScript string you actually have to do do
 You can read more about this here:
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#ES2018_revision_of_illegal_escape_sequences`
 
+const UNDEFINED_AS_OBJECT_KEY_ERROR =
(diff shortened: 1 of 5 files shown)


packages/serialize/src/index.js
```diff
@@ -290,10 +293,7 @@ function createStringFromObject(
           (registered == null || registered[value[0]] === undefined)
         ) {
           for (let i = 0; i < value.length; i++) {
-            string += `${processStyleName(key)}:${processStyleValue(
-              key,
-              value[i]
-            )};`
+            string += `${key}:${processStyleValue(key, value[i])};`
```
mitchellhamilton: Why is processStyleName not being used here?

Andarist: Oh, I was experimenting with code changes - didn't mean to actually commit this. Gonna revert this back in a moment and gonna add a test for this.

Andarist:
@mitchellhamilton done with the changes

PR #1588, opened 2019-10-29
Make StyledComponent polymorphic

### LINKED ISSUE #1587: Make StyledComponent type polymorphic
I started to use the new Flow types into my project, and I figured out it would make sense to make `StyledComponent` a polymorphic type, so that we can do this:

```typescript
const Foo = styled<Props>(props => <div {...props} />)`
`;

const Bar = styled<Props>('div')`
`;
```

Which would allow to:

1. type `props` inside the styled component logic
2. allow to workaround the lack of tagged template support of Flow

Unfortunately I doubt my Flow knowledge is good enough to implement this without breaking the existing types, which make sense because we also want to allow:

```typescript
const Foo = styled.div<Props>`
`;
```

What do you think?

FezVrasta:
**What**:

implements/fixes #1587

**Why**:

read #1587

**How**:

`StyledComponent` is now polymorphic (just for the `CreateStyled` function)

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/packages/styled-base/src/utils.js
+++ b/packages/styled-base/src/utils.js
@@ -10,17 +10,17 @@ export type StyledOptions = {
   target?: string
 }
 
-export type StyledComponent<P> = StatelessFunctionalComponent<P> & {
+export type StyledComponent<Props> = StatelessFunctionalComponent<Props> & {
   defaultProps: any,
   toString: () => string,
   withComponent: (
     nextTag: ElementType,
     nextOptions?: StyledOptions
-  ) => StyledComponent<P>
+  ) => StyledComponent<Props>
 }
 
-export type PrivateStyledComponent<P> = StyledComponent<P> & {
-  __emotion_real: StyledComponent<P>,
+export type PrivateStyledComponent<Props> = StyledComponent<Props> & {
+  __emotion_real: StyledComponent<Props>,
   __emotion_base: any,
   __emotion_styles: any,
   __emotion_forwardProp: any
@@ -39,15 +39,15 @@ export const getDefaultShouldForwardProp = (tag: ElementType) =>
     ? testOmitPropsOnStringTag
     : testOmitPropsOnComponent
 
-export type CreateStyledComponent = <P>(
+export type CreateStyledComponent = <Props>(
   ...args: Interpolations
-) => StyledComponent<P>
+) => StyledComponent<Props>
 
 export type CreateStyled = {
-  <P>(
+  <Props>(
     tag: ElementType,
     options?: StyledOptions
-  ): (...args: Interpolations) => StyledComponent<P>,
+  ): (...args: Interpolations) => StyledComponent<Props>,
   [key: string]: CreateStyledComponent,
   bind: () => CreateStyled
 }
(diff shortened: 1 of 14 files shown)


Andarist:
> Do you think is it alright or there's something off?

I'm not sure - I don't know Flow's philosophy. The fact that a fragment there changes the situation there is kinda wild for me 😱 Maybe this is something you could ask Flow maintainers? Not sure how often they respond nowadays though.

FezVrasta:
I was able to reproduce the same issue with just native React types:

https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gEoCmAhgMZ6RayqgDkWJ5tA3IoqbAHYDOeAwlxzFknQlgCCoALyh0VdN2kA+UAB4AJsgBuoAN4A6Q3NgKAvqGBLW7Lr1ADOQkWIBC02fMVSVG7XsP6xmYWVmw4AJ7ohKAACp7uuqCkABbI0OoMnABcBIw4ACQAcrDq0abWiIQAHpi4SbZ4APIA1u4AFEHcOXEm3ACUympKqg5OohJKup36KWkZhJymqsCjwuPiw5bW1bV4HDx4ANKw7Z3dngPeaqvOWC6T07PpmUsrgmuuVkA

I asked the same question on SO, maybe someone knows more:
https://stackoverflow.com/questions/58626420/component-requires-type-annotation-only-when-used-as-root-of-exported-component

Andarist:
Sure, just let me know when this is ready. Thanks for working on this!

FezVrasta:
hopefully CI should pass now, this lint rule that prevents semicolons is killing me...

Andarist:
> hopefully CI should pass now, this lint rule that prevents semicolons is killing me...

Isn't precommit hook just fixing this for you automatically?

> I would like to add a page in the docs where I explain how the Emotion Flow types work, any idea where would it make sense to put it?

I would say that the best place is just under TypeScript in the Tooling section

FezVrasta:
@Andarist I added the new documentation page, feel free to merge the PR if you think it's okay.

FezVrasta:
> > hopefully CI should pass now, this lint rule that prevents semicolons is killing me...
> 
> Isn't precommit hook just fixing this for you automatically?

In the case I hit, Prettier needed to run several times to properly format the code (`razzle/index.js`), and for some reason it doesn't like the fact we had some empty lines between two code blocks...

packages/styled-base/src/utils.js
```diff
@@ -11,7 +10,7 @@ export type StyledOptions = {
   target?: string
 }
 
-export type StyledComponent<P> = React.StatelessFunctionalComponent<P> & {
+export type StyledComponent<P> = StatelessFunctionalComponent<P> & {
```
mitchellhamilton: I know this isn't from this PR but could we rename `P` to `Props`(and the same thing in the other places we have `P`)?

FezVrasta: `P` is the same name used by the built-in Flow types for `StatelessFunctionalComponent`, are you sure you want to give it a different name in our implementation?

mitchellhamilton: Yes

FezVrasta: Updated

PR #1589, opened 2019-10-30
Improved docs around SSR apis

### LINKED ISSUE #1178: Problems surrounding SSR injection of <style> and unreliability of :first-child selectors

JakeGinnivan:
Fixes #1178

**What**: 2 main changes
1) Improve the docs around SSR to make it clear there is an advanced integration which does not break some CSS features.
2) ~The hydrate function is not on custom created caches which is required to remove the nth child warning when using the default cache.~ I've now removed this change, I found out about using the `data-emotion-css` attribute to automatically hydrate

**Why**:
Because it should be the users decision what trade offs they would like, do you want a really simple integration with some CSS limitations, or do you want to perform a more complex integration and remove the CSS limitations.

For us it is not acceptable to not use the nth child selectors.

**How**:

Copied the hydrate function from the create-emotion module into the createCache function

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


JakeGinnivan:
Not sure how code coverage has decreased, I have just updated the docs

JakeGinnivan:
@mitchellhamilton have circled back around to this. Hopefully this change is a good compromise between showing people there is a more advanced integration which removes the CSS limitations, but prevents usage of the streaming APIs.

mitchellhamilton:
I made a couple wording changes and a little change to the code example, notably removing the client section because it's not necessary when using the default cache. Let me know what you think about the changes.

I know I've been hesitant about this but you're right that we need to call this out, thanks for this, the feedback is really appreciated. ❤️

JakeGinnivan:
@mitchellhamilton unfortunately the client part is still needed, otherwise you get warnings in your console about unsafe selectors :(

mitchellhamilton:
👍 added it back

jooj123:
Thanks @JakeGinnivan @mitchellhamilton - very clear now 👍

JakeGinnivan:
Thanks @mitchellhamilton, I think this will help a lot of people.

PR #1592, opened 2019-10-31
feat: add Flow types to is-prop-valid

FezVrasta:
**What**:

Added Flow types to `@emotion/is-prop-valid`

**Why**:

Packages depending on this will end up with a generic `any` type instead of the correct type.

**How**:
Added `// @flow` headers

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/shiny-bottles-itch.md
+++ b/.changeset/shiny-bottles-itch.md
@@ -1,0 +1,5 @@
+---
+'@emotion/is-prop-valid': minor
+---
+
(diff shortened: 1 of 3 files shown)


FezVrasta:
Would it make sense to add an ESLint rule to require all the javascript files to have the`// @flow` header (or the `// @noflow` one)

PR #1599, opened 2019-11-02
Serialize functions put as object style values

Andarist:


packages/serialize/src/index.js
```diff
@@ -255,6 +254,74 @@ function handleInterpolation(
...
+      if (
+        key === 'NO_COMPONENT_SELECTOR' &&
+        process.env.NODE_ENV !== 'production'
+      ) {
+        throw new Error(
+          'Component selectors can only be used in conjunction with babel-plugin-emotion.'
+        )
+      }
+
+      if (Array.isArray(value) && typeof value[0] === 'string') {
```
Andarist: I've removed `&& (registered == null || registered[value[0]] === undefined)` from here. If it was needed - please share an example so I can add a test for this.

mitchellhamilton: It's for cases like this

```jsx
import {css} from 'emotion'

css({':hover': [css({color:'green'})]})
```

mitchellhamilton:
I’m kinda leaning towards not doing this, the current behaviour is working fine and this isn’t really a thing that people are asking for

Andarist:
@mitchellhamilton I'm also not that crazy about this one, I'd only like to achieve consistency between what can be used and what cannot be. As we discussed before currently wrapping a function in an array (at the same position) suddenly makes those functions serializable, as in:
```js
   const H1 = styled.h1({
      fontSize: [props => props.fontSize]
    })
```

Any ideas about how we could improve consistency other than merging this PR?

mitchellhamilton:
Just checked making it an array doesn't make it work, it produces this css:
```css
.css-z30v7q fontSize{100;}
```

So I feel like the answer is to leave it because this isn't causing any problems.

PR #1600, opened 2019-11-02
v11

### LINKED ISSUE #1606: Emotion 11
We're planning to do a major version of Emotion soon, we don't have a definite timeline but it should be out before the end of the year, most likely earlier.

We don't currently have any massive API changes planned but there will be some package renaming which should be possible to migrate with codemods alone. (This isn't anything like the v9 -> v10 change)

The big changes that we are currently doing:
- Package renaming (#1635)
- Use hooks internally which means a better tree in React DevTools and a smaller bundle size (#967)
- Rewrite TypeScript definitions (#1501)
- Allow mapping of imports in the babel plugin (#1220) 

We'll update this issue for larger planned changes. For smaller changes, please see the [v11 milestone](https://github.com/emotion-js/emotion/milestone/1). The issues on the milestone aren't necessarily going to be addressed in v11 but we'll consider them.

We are already doing pre-releases which will be available on the `next` dist tag on npm.

mitchellhamilton:
IMPORTANT NOTE: This PR should not be merged right now.

Any PRs that should be in v11 should be based against this PR(the `next` branch). Once a PR is merged into the `next` branch, the Changesets release action will open PRs as it does for master doing prereleases which can be merged into this branch and will then be published to the `next` tag on npm.

closes #1606


Gersom-NL:
```javascript
const Foobar = styled.div`
    background: ${p => p.theme.colors.brand};
`;
```

This code errors because `p.theme` is optional. Is this intended behaviour? Should we always use the non-null assertion operator like this `p.theme!.colors.brand`?

Fyi, I'm currently using this to manage my theme typing:

`styled.d.ts`
```typescript
declare module '@emotion/styled' {
    import { CreateStyled } from '@emotion/styled/types/index';
    import { theme } from '~/theme';

    const styled: CreateStyled<typeof theme>;

    export * from '@emotion/styled/types/index';
    export default styled;
}
```

Gersom-NL:
Also, I get this error whenever I start NextJS dev server:

```
[ error ] ERROR in /.../node_modules/@emotion/styled/types/base.ts
20:3 Cannot re-export a type when the '--isolatedModules' flag is provided.
    18 |
    19 | export {
  > 20 |   ArrayInterpolation,
       |   ^
    21 |   CSSObject,
    22 |   FunctionInterpolation,
    23 |   ObjectInterpolation
```

Andarist:
As to the theme - currently, it should be typed to `any` from what I know, but this is being worked on in different PRs. So in the near future, you will be able to opt-in to extending `Emotion.Theme` with whatever type you like. Could you prepare a repro case for your problem? I could take a look at this then, cc @JakeGinnivan 

As to `isolatedModules` - I'm not sure how this works exactly. From what I know this flag is useful when actually writing code in TS but wanting to transpile with, in an example, Babel. I'm not sure if our files should use the same restriction - did it work for you in emotion@10? The types were put roughly in the same files, but files got moved around, so maybe this somehow affected how TS tries to compile them. A repro case for this would be highly appreciated as well.

Gersom-NL:
Yes, typing my theme does work in v10 and the `isolatedModules` error is also a regression from v11. This is probably a babel specific issue, since that is what NextJS is using internally.

Andarist:
@Gersom-NL I believe this has been fixed by @frankwallis in #1627. This fix should be available soon in a new version (just merged in releasing PR - so the release flow is running just now)

sneridagh:
@Andarist How is v11 doing? Any idea on the roadmap? Are they any blockers? Thanks!

Andarist:
As you can see [here](https://github.com/emotion-js/emotion/milestone/1) - there is only a single issue blocking us. I will be looking into it soon-ish. Other than that we'd like to support ESM in new node versions so we need to provide `pkg.json#exports` for all packages - that will be implemented in preconstruct which we are using to build our packages and from what I understand Mitchell has started working on that one a few days ago.

ivanbanov:
Thank you for merging!

image

PR #1604, opened 2019-11-04
Jest emotion/improvements

Andarist:
Match rules in declarations with component used as a selector and take specificity into account when matching styles

fixes https://github.com/emotion-js/emotion/issues/1214

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1282,6 +1282,16 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@emotion/babel-preset-css-prop@^10.0.5":
+  version "10.0.23"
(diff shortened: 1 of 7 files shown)


FezVrasta:
Thanks!

yarn.lock
```diff
@@ -1282,6 +1282,16 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@emotion/babel-preset-css-prop@^10.0.5":
```
Andarist: im not sure why our babel plugins were added to the lock 🤔

FezVrasta: It usually happens if some dependency is referencing a version different from latest, or when yarn shits its pants (it happens often)

Andarist: actually I've found why this has happened right now with a help of `yarn why` - those are dependencies of `gatsby-plugin-emotion`, so it makes sense that they have been added

PR #1607, opened 2019-11-05
Ensure material UI regression is fixed

### LINKED ISSUE #1167: [@emotion/styled 10] TypeError using styled with @material-ui/core components after upgrade to v10
- `typescript` version: 3.2.2
- `emotion` version: 10.0.6
- `react` version: 16.7.0
- `material-ui` version: 3.8.2

**Relevant code:**
```javascript
const StyledTypography = styled(Typography)(({ theme }) => ({
  backgroundColor: theme.palette.secondary.main,
}));
// ...
<StyledTypography variant="h6" //...
 ^^^^^^^^^^^^^^^^
```
**What you did:** With Typescript and using @material-ui/core 3.7 or 3.8, try to upgrade from Emotion 9 to Emotion 10.

**What happened:** Using @material-ui/core with the upgraded Emotion 10 cause a TypeError when styling any material-ui component with styled function. 

**Error log:** 
```
Type error: Type '{ children: ReactNode[]; variant: "h6"; }' is not assignable to type '(IntrinsicAttributes & TypographyProps & ClassAttributes<Component<TypographyProps, any, any>> & Pick<(TypographyProps & ClassAttributes<Component<TypographyProps, any, any>>) | (TypographyProps & Attributes), "title" | ... 258 more ... | "innerRef"> & { ...; } & { ...; }) | (IntrinsicAttributes & ... 4 more ... & {...'.
  Type '{ children: ReactNode[]; variant: "h6"; }' is not assignable to type 'IntrinsicAttributes & TypographyProps & Attributes & Pick<(TypographyProps & ClassAttributes<Component<TypographyProps, any, any>>) | (TypographyProps & Attributes), "title" | ... 258 more ... | "innerRef"> & { ...; } & { ...; }'.
    Type '{ children: ReactNode[]; variant: "h6"; }' is missing the following properties from type 'Pick<(TypographyProps & ClassAttributes<Component<TypographyProps, any, any>>) | (TypographyProps & Attributes), "title" | "color" | "hidden" | "style" | "children" | ... 254 more ... | "innerRef">': style, classes, className, headlineMapping, innerRef  TS2322
```

**Reproduction:** A GitHub repository is provided next with an example (based on a simple material-ui component, Typography. Also, there is a CodeSandbox with the same code, but it hides the problem and works.
- [GitHub](https://github.com/sergiohgz/mui38-emotion10)
- [CodeSandbox](https://codesandbox.io/s/92o7w22lzy)

**Problem description:** When using `styled` from `@emotion/styled` with any component from `@material-ui/core`, the second part of the args contains a type error, so there is no way to use. Styling HTML components works well (`div`, `span`, etc) or any component created by user works like a charm. Also, this happens with template literal syntax.

**Suggested solution:** Currently, I tried to type `styled(Typography)<any>({})` but with this solution we will lose the types in the style object, so this isn't a solution at all, only the worst bypass I should never thought.

JakeGinnivan:
Fixes #1167

### DIFF
--- a/packages/styled/types/tests-materialui.tsx
+++ b/packages/styled/types/tests-materialui.tsx
@@ -1,0 +1,107 @@
+import * as React from 'react'
+import styled from '@emotion/styled/base'
+
+// Needs to be inlined to test older versions of TypeScript
+type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>
+
+/**
+ *
+ * Everything below is types extracted from MaterialUI to verify a v9 -> v10 regression (#1167)
+ *
+ */
+export type ThemeStyle =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | 'h6'
+  | 'subtitle1'
+  | 'subtitle2'
+  | 'body1'
+  | 'body2'
+  | 'caption'
+  | 'button'
+  | 'overline'
+
+export type ClassNameMap<ClassKey extends string = string> = Record<
+  ClassKey,
+  string
+>
+export interface StyledComponentProps<ClassKey extends string = string> {
+  /**
+   * Override or extend the styles applied to the component.
+   */
+  classes?: Partial<ClassNameMap<ClassKey>>
+  innerRef?: React.Ref<any>
+}
+export type StandardProps<
+  C,
+  ClassKey extends string,
+  Removals extends keyof C = never
+> = Omit<C, 'classes' | Removals> &
+  StyledComponentProps<ClassKey> & {
+    className?: string
+    ref?: C extends { ref?: infer RefType } ? RefType : React.Ref<unknown>
+    style?: React.CSSProperties
+  }
+
+type Alignment = 'inherit' | 'left' | 'center' | 'right' | 'justify'
+
+type Style = ThemeStyle | 'srOnly'
+export interface TypographyProps
+  extends StandardProps<React.HTMLAttributes<HTMLElement>, TypographyClassKey> {
+  align?: Alignment
+  color?:
+    | 'initial'
+    | 'inherit'
+    | 'primary'
+    | 'secondary'
+    | 'textPrimary'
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
Up to you if you want to merge this test

Andarist:
I will gladly merge this - although I would prefer moving this test to a separate file. When we add more stuff at the bottom it will be hard to see where this test actually starts and where it ends.

Additionally, I see a problem on CI with older versions of TS not recognizing builtin Omit (rightfully so). Is it possible to limit TS versions being tested on per test file basis? I had problems with this recently in https://github.com/emotion-js/emotion/pull/1575 . If it's not possible then just inline Omit into this file.

JakeGinnivan:
Will sort out over next few days.

JakeGinnivan:
Not sure what is going on with this build error =/ It's happening locally but I can't see why

Andarist:
Seems like it has problem extracting `TypographyProps`. Not sure if this is worth fighting if it only affects "older" versions of TS. OTOH the situation here doesn't seem overly convoluted so maybe it's worth investigating why this fails - I'd suggest trying to bisect that test to see when it starts to fail. If we'd know that we could maybe assess how many use cases this might affect.

We could also try to only test this with newer versions of TS but I'm not sure how exactly this can be achieved on per-file basis. I've found such comment in dtslint repo - https://github.com/microsoft/dtslint/blob/971f5e2f90285646c97cb7b4a2b1b8b5a555cfdc/src/util.ts#L101 . So seems like it's possible to run against some version, but in here we'd ideally want to run those for all versions since a particular version 🤔 

If I find time today for this - I'll try bisecting as well.

JakeGinnivan:
I tried adding the comment, it didn't do much :(

It does work inside the .d.ts files, but not the test files. Not sure it's worth it?

Andarist:
Because it seems to be hard to set up a test for this I don't think there is much point in keeping this open now. The good thing is that we got a confirmation that the issue got fixed.

PR #1609, opened 2019-11-05
Updated types to support global Theme definition

### LINKED ISSUE #1197: Typescript support for theme when using css prop
- `emotion` version: 10.0.6
- `react` version: 16.8.0-alpha.1

**Relevant code:**

```javascript
<span
  css={theme => ({
    fontWeight: theme.font.weight.light,
    position: "absolute",
    right: "8px",
    top: "52%"
  })}
>
```

**What you did:**
Created a theme using using the `emotion-theming` package

**What happened:**
No typescript support for the provided theme object.

**Reproduction:**

**Problem description:**

From the [documentation](https://emotion.sh/docs/typescript#define-a-theme), it appears the current typescript support allows you to define the theme object shape when using `styled()`, but not when using the css prop. Currently (and as confirmed by the documentation), the object is typed as `any`.

**Suggested solution:**

Not sure if possible, but hopefully provide some way of defining a theme type via typescript generics, even when using the css prop. It appears the type is hardcoded for the css prop to `any` [here](https://github.com/emotion-js/emotion/blob/master/packages/core/types/index.d.ts#L86)

joltmode:
**What**:
Fixes #1197 

**Why**:
The current typings do not support a way of providing theme types globally. The re-exports can only go so far, and the `styled` themed re-export actually tampers with babel plugin.

**How**:
I added `Theme` export via a global namespace as suggested by @JakeGinnivan so that it can be overriden with:

```
// emotion-theme.d.ts
declare global {
  namespace Emotion {
    export interface Theme {
      your: 'theme',
      comes: 'here'
    }
  }
}
```

And updated _all_ usages of previous default `any` to use the namespace one.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset 

I don't know how to add test cases for this.
I tried by adding a `globalTests.tsx` containing actual tests and `emotion-theme.d.ts` containing the declaration alongside the respective tests. But all my efforts to prevent `emotion-theme.d.ts` global declaration from bleeding into un-overriden usecase tests failed. If the global theme declaration bleeds into un-overriden usecase tests, it breaks the tests there. Probably for the same reasons as [can be seen in styled-components typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c9e13869e9ef4449184316c43ab407842860d41f/types/styled-components/index.d.ts#L407-L414):

```typescript
/**
 * This interface can be augmented by users to add types to `styled-components`' default theme
 * without needing to reexport `ThemedStyledComponentsModule`.
 */
// Unfortunately, there is no way to write tests for this
// as any augmentation will break the tests for the default case (not augmented).
// tslint:disable-next-line:no-empty-interface
export interface DefaultTheme {}
```

~I'm opening this as a draft because I lack documentation/changeset, and want some initial reviews.~  
Updated to ready, because... how would one review it otherwise?

Also, what does `Code complete` mean in the checklist?


joltmode:
@JakeGinnivan any reason not to go with `styled-components` approach?

- Where we'd `export interface DefaultTheme {}` in `@emotion/core`
- use as default for other packages via `import { DefaultTheme } from '@emotion/core'` 
- and the user could extend via:
  ```
  // emotion-theme.d.ts
  import '@emotion/core'
  
  declare module '@emotion/core' {
     export interface DefaultTheme {
      your: 'theme',
      comes: 'here'
    }
  }
  ```

I feel this one as the more cleaner approach. The one thing I do dislike, is that it's tied to `@emotion/core` namespace, which actually, has nothing to do about theming.

packages/core/types/index.d.ts
```diff
@@ -77,25 +77,30 @@ export interface ClassNamesProps<Theme> {
...
   interface DOMAttributes<T> {
-    css?: InterpolationWithTheme<any>
+    css?: InterpolationWithTheme<Emotion.Theme>
   }
 }
 
 declare global {
+  namespace Emotion {
+    // tslint:disable-next-line: no-empty-interface strict-export-declare-modifiers
+    export interface Theme {}
```
joltmode: @JakeGinnivan any reason not to go with `styled-components` approach\*?

- Where we'd `export interface DefaultTheme {}` in the module type scope
- use as default for other packages via `import { DefaultTheme } from '@emotion/core'`
- and the user could extend via:
  ```typescript
  // emotion-theme.d.ts
  import '@emotion/core'
  
  declare module '@emotion/core' {
     export interface DefaultTheme {
      your: 'theme',
      comes: 'here'
    }
  }
  ```
I feel this one as the more cleaner approach. The one thing I do dislike, is that it's tied to `@emotion/core` namespace, which actually, has nothing to do about theming.

\* The question is not limited to Jake, just that his recommendation on the issue posting was to go this route.

JakeGinnivan: Only reason I suggested what I did is it's the one I am familiar with. Express uses this to extend the Request/Response objects.

I can't say I have enough of an understanding of the styled components approach to sum up trade offs right now.

JakeGinnivan: Something which springs to mind, is the global approach may not work with babel typescript compilation?

joltmode: @JakeGinnivan hmm, I cannot seem to find any reason why it wouldn't work with babel. In either way it would need to load an additional type declaration from local codebase. 

As long as babel supports declaration merging and namespaces - _which I think it does, otherwise it's not in compliance with TS and would beat the purpose_ - either of solutions should work.

My question to you (@JakeGinnivan) was more on performance implications (though, I don't see much difference, but I'm not a TS pro by any means) and for everyone about the overall DX we expose.

mitchellhamilton: My 2c on this is that we should do this one because it doesn't matter much and thinking in modules feels better.

```tsx
declare module '@emotion/core' {
  export interface DefaultTheme {
    your: 'theme',
    comes: 'here'
  }
}
```

> The one thing I do dislike, is that it's tied to @emotion/core namespace, which actually, has nothing to do about theming.

This isn't really true, the ThemeContext lives in `@emotion/core` and you can even use that without ever using `emotion-theming`

packages/core/types/index.d.ts
```diff
@@ -24,6 +24,8 @@ export {
   ObjectInterpolation
 } from '@emotion/css'
 
+export type AnyIfEmpty<T extends object> = keyof T extends never ? any : T
```
JakeGinnivan: I think this was the sort of type which caused the number of types and build time to increase dramatically during compilation.

It would be worth trying to see how many types are created before/after during a compilation with --diagnostics

JakeGinnivan:
More general question.

Why don't we just remove the `Theme` generic parameter from everywhere and just reference `Emotion.Theme` or the module version directly?

It would remove a heap of generic params and complexity from the types?

JakeGinnivan:
Nice work getting the ball rolling on this, I think this will be an awesome change. I had a go at just removing the generic params for theme from everywhere?

Thoughts:

https://github.com/joltmode/emotion/pull/1

Also, for compat we could just use documentation to fill the gap (I thought I posted this somewhere, but can't find it. Oh well).

```
declare global {
  export interface Theme extends Record<string, any> {
  }
}
```

JakeGinnivan:
After doing the above, I wonder if there is much value in the `emotion-theming` package at all.

IMO v11 could completely remove it, you just move the ThemeProvider, hook and HoC into core?

packages/core/types/index.d.ts
Andarist: @mitchellhamilton that's true that `ThemeContext` can be used directly, but is it even a good thing? I understand that wrapping "enforces" you to learn some APIs - but they relate to regular APIs so much, that it's hardly noticeable. And also wrapping (by wrapping I mean `useTheme` and `ThemeProvider` helpers) gives some better encapsulation - it defines a clear way of using a particular thing and helps us with providing any extra functionality, babel transforms or error/warning messages in those - while this becomes harder to do when exporting `ThemeContext` directly.

Having those things inside a single package would make it easier to install those things and if somebody doesn't want to use those exports tree-shaking should just do its job at pruning those.

mitchellhamilton:
I'm good with moving the theming APIs to core. @Andarist want to create a PR?

Andarist:
@mitchellhamilton sure, gonna handle this today.

Andarist:
@joltmode I've just merged in the PR which has moved theming APIs into `@emotion/core`. Could you rebase this branch and incorporate suggestions from https://github.com/joltmode/emotion/pull/1 ?

Andarist:
@joltmode will you be able to work on this in the near future? If not - that's perfectly fine, we'd only appreciate a heads up so somebody could take over this.

joltmode:
@Andarist yes, will strive to push this forward during the weekend.

Andarist:
Looking quite good!

As discussed briefly before - if it's possible we'd prefer to keep that `Emotion.Theme` as part of the `@emotion/core` module - rather than in the global namespace.

joltmode:
Yup, looking to update that one.

Currently working on fixing tests, forgot to run before pushing. 

Was a pretty nasty rebasing, loads of changes inbetween! Quite possibly this was the toughest merge conflict in my development life so far. 🥇

packages/styled/types/tests.tsx
```diff
@@ -24,25 +24,13 @@ styled.svg
...
-
-  return {}
-})
-;<StyledDiv2 />
-
 // Can override theme optionally on prop
-;<StyledDiv2 theme={{ themeProp: '' }} />
+// TODO: Fix types for this usage
+// $ExpectError
+;<StyledDiv theme={{ primary: '' }} />
```
joltmode: This here currently errors, because when defining theme this way, one has to implement for the full spec of Theme.

_I'm a little tired to think of solutions from my end, can chime in tomorrow. Just exposing this issue for discussion._

docs/typescript.mdx
```diff
@@ -248,34 +248,33 @@ const App = () => (
 ### Define a Theme
```
joltmode: The whole docs on Theme usage with TS feels massively outdated now, and I feel like it should be reworded altogether.

_Exposing for discussion._

Andarist: Right - I feel like we can handle it in a separate PR though. Ideally, such docs would present the current solution with its tradeoffs and incorporate stuff from this https://github.com/emotion-js/emotion/pull/973

packages/core/types/tests-theming.tsx
```diff
@@ -102,34 +72,12 @@ typedWithTheme((props: { value: number }) => null)
...
-const StyledDiv2 = themedStyled.div({})
-;<StyledDiv2 />
-// $ExpectError
-;<StyledDiv2 theme={{ primary: 0, secondary: 0 }} />
-
-export type StyleDefinition<T = {}> = Interpolation<WithTheme<T, Theme>>
-export type ObjectStyleDefinition<T = {}> = ObjectInterpolation<
-  WithTheme<T, Theme>
->
+export type StyleDefinition = Interpolation
```
Andarist: we shouldn't export types from test files

packages/core/types/tests-theming.tsx
```diff
@@ -139,7 +87,4 @@ const style2: ObjectStyleDefinition = {
 }
 
 // Can use ThemeProvider
-;<ThemeProvider theme={{ prop: 'val' }} />
-;<TypedThemeProvider theme={{ primary: '', secondary: '' }} />
-// $ExpectError
-;<TypedThemeProvider theme={{ nope: 'string' }} />
+;<ThemeProvider theme={{ primary: 'val' }} />
```
Andarist: How this works now? Is `{primary: string}` compatible with builtin (empty) `Theme`?

Andarist: ah, you extend it in the other file, cool 👍

packages/serialize/types/index.d.ts
```diff
@@ -56,17 +59,18 @@ export type Keyframes = {
...
     CSSPseudos<MP>,
     CSSOthersObject<MP> {}
 
-export interface FunctionInterpolation<MergedProps> {
+export interface FunctionInterpolation<MergedProps = { theme: Theme }> {
   (mergedProps: MergedProps): Interpolation<MergedProps>
 }
 
-export type Interpolation<MergedProps = undefined> =
+export type Interpolation<MergedProps = { theme: Theme }> =
```
Andarist: actually I think that `Interpolation` is also used in `css` definition and in there you don't have access to `Theme`, so the previous version with default being `undefined` was correct for `css`'s case

packages/serialize/types/index.d.ts
```diff
@@ -79,7 +83,7 @@ export type Interpolation<MergedProps = undefined> =
   | ObjectInterpolation<MergedProps>
   | FunctionInterpolation<MergedProps>
 
-export function serializeStyles<MP>(
+export function serializeStyles<MP = { theme: Theme }>(
```
Andarist: same here - `serializeStyles` gets called by `css`

packages/styled/types/tests.tsx
Andarist: Not sure if this is incorrect though. If `StyledDiv` depends on `Theme` we can't know on which parts of the `Theme` it actually depends, so technically putting an incomplete theme here could be a mistake.

Andarist:
@JakeGinnivan would appreciate you taking a look at this as well

packages/styled/types/tests.tsx
joltmode: Yeah, with a clear mind today, I totally agree - didn't really think it through yesterday.

packages/serialize/types/index.d.ts
joltmode: Yeah, this and the `serializeStyles` stuff should be better viewed by @JakeGinnivan. Not on the same train of thought as he is on with these.

I also invited you @JakeGinnivan as a member to my repo, so it's not mandatory to have to PR.

Andarist: > > css definition and in there you don't have access to Theme

> Just for my understanding, why is this the case?

I've meant `css` function - you can't do `css(() => /**/)`. From what remember such function just gets stringified. `css` call is being executed without any notion of context, you can hoist it wherever you want etc - so it has no access to theme or props. 

A different story is the `css` props - where you can get access to the builtin Theme.

> Looking at the other types elsewhere we could actually do another round of simplification if we could add theme into these types directly

What kind of simplification do you have in mind? I've noticed now that we have Interpolation, CSSInterpolation and even InterpolationWithTheme and I think they might be used sometimes in wrong contexts (especially the Interpolation type). I need to draw out the dependency tree of those types somewhere for myself to analyze this further 😅

docs/typescript.mdx
Andarist: Yes, this is crucial and I plan to get this done before v11 👍

packages/styled/types/base.d.ts
```diff
@@ -73,7 +73,10 @@ export interface CreateStyledComponent<
   <AdditionalProps extends {} = {}>(
     ...styles: Array<
       Interpolation<
-        ComponentProps & SpecificComponentProps & StyleProps & AdditionalProps
+        ComponentProps &
+          SpecificComponentProps &
+          StyleProps &
+          AdditionalProps & { theme: Theme }
```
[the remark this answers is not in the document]
Andarist: `{ theme: Theme }` gets always passed in as `StyleProps`, so we should be able to remove this from here, but when I try to do that then I get such errors:
```
Error: /emotion/packages/styled/types/tests-base.tsx:156:32
ERROR: 156:32  expect  TypeScript@next compile error: 
Object is possibly 'undefined'.

/emotion/packages/styled/types/tests.tsx:19:7
ERROR: 19:7    expect  Expected type to be:
  { theme?: Theme | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { bar: string; } & { theme: Theme; }
got:
  { theme?: Theme | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { bar: string; }
```

Still investigating this, but right now I don't see why this fails.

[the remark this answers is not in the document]
Andarist: While I don't understand why this fails - it might actually be a proper thing to specify this here instead of passing it from the above level in `StyleProps`. After all - `theme` is **always** available in interpolations, so it shouldn't be possible to "cancel" it out by passing StyleProps without it?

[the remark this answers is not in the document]
Andarist: Ok - I know why this thing has been failing:
https://github.com/joltmode/emotion/blob/464922628c9197988ac7ae38d33b84b6eb6d63d8/packages/styled/types/base.d.ts#L89-L93
This doesn't include `StyleProps`. It should right?

Regardless of that - my previous question stands. Should we pass Theme as part of StyleProps from the above or should we just pass it directly to Interpolation?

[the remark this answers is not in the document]
Andarist: Ok, I've decided to move `{ theme: Theme }` inline where parameters are being passed to Interpolation inside styled components. Seems the most appropriate solution - but I'm happy to discuss this further, maybe I've missed something? 

More than that - I'm right now kinda questioning if we need StyleProps at all, what do you think?

stramel:
@Andarist I'm not sure if this is causing a problem or if it's just misleading.

I noticed it seems to be pulling a test base typing instead the interface from the main library.

image

Andarist:
Interesting, could you prepare a runnable repro case (a simple repository)? That would help me a lot to jump into investigating this. It might be also worth reporting an issue about this. Notice though that main library typing is picked up correctly (you can see that on the right of your screenshot - first item on the list is the one from the main library), the problem is though that the test typings are also picked up.

stramel:
I will do my best to create a repo

JakeGinnivan:
I wonder if we should just not include the test files @Andarist?

Something like this
image

Andarist:
Yeah, that would be the simplest fix. Prepared a PR for that - https://github.com/emotion-js/emotion/pull/1702

PR #1615, opened 2019-11-07
test: styled Flow tests

FezVrasta:


packages/styled/flow-tests/base.js
```diff
@@ -1,23 +1,17 @@
 /* eslint-disable no-unused-vars */
 // @flow
-import * as React from 'react'
+import React from 'react'
 import createStyled from '../src/base'
 import type { CreateStyledComponent, StyledComponent } from '../src/utils'
 
-export const valid: CreateStyledComponent = createStyled('div')
+const valid: StyledComponent<empty> = createStyled('div')({
```
FezVrasta: Don't ask me how it used to work before, but `CreateStyled` return type is not `CreateStyledComponent`, it's `(...args: Interpolations) => StyledComponent<Props>` (defined inline).

To make the test simpler, I call `createStyled` to generate `(...args: Interpolations) => StyledComponent<Props>`, and then I call it again, to generate a Styled Component. Then I check the returned type is `StyledComponent`.

The result is pretty much the same.

packages/styled/flow-tests/base.js
```diff
@@ -1,23 +1,17 @@
...
+  color: 'red'
+})
 
 // $FlowExpectError: we can't cast a StyledComponent to string
-export const invalid: string = createStyled('div')
+const invalid: string = createStyled('div')({ color: 'red' })
```
FezVrasta: Same thing here, I test the final StyledComponent type instead of the intermediate type.

packages/styled/flow-tests/base.js
```diff
@@ -1,23 +1,17 @@
...
 
 const styled = createStyled('div')
-type Props = { color: string }
-// prettier-ignore
-const Div = styled<Props>({ color: props => props.color })
```
FezVrasta: Again, no idea how it used to work... `createStyled()` doesn't return a polymorphic type, only `createStyled.*` are polymorphic types.

PR #1620, opened 2019-11-08
[typescript] Add `T` param to Matchers interface

spudly:
This fixes typescript type definitions when using jest-emotion with jest 24.

I simply added a parameter to the Matchers interface in the typescript types file for jest-emotion.

In Jest 24 they added a second generic parameter to the `Matchers` interface. Typescript requires interfaces to have the same parameters, so jest-emotion needs to add it also.

See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/39982

**Checklist**:
- [N/A] Documentation
- [N/A] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/mean-deers-confess.md
+++ b/.changeset/mean-deers-confess.md
@@ -2,4 +2,4 @@
 'jest-emotion': major
 ---
 
-Added the `T` parameter to the `Matchers` interface in the typescript definitions for jest-emotion
+Added the `T` parameter to the `Matchers` interface in the TypeScript definitions to make this module compatible with `@types/jest@^24.0.20`.
--- a/.changeset/mean-deers-confess.md
+++ b/.changeset/mean-deers-confess.md
@@ -1,0 +1,5 @@
+---
+'jest-emotion': major
+---
+
+Added the `T` parameter to the `Matchers` interface in the typescript definitions for jest-emotion
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
     "@mdx-js/react": "^1.0.27",
     "@preconstruct/cli": "^1.1.1",
     "@testing-library/react": "^9.3.2",
-    "@types/jest": "^23.0.2",
+    "@types/jest": "^24.0.22",
     "@types/node": "^10.11.4",
     "@types/react": "^16.9.11",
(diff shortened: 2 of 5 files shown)


Andarist:
@spudly if this is a requirement (to parameter count to match) - isn't this a breaking change (sort of) for people using older version of jest (or its typings)?

spudly:
Yes, good point. I guess I'll have to change the changeset...

Andarist:
You can target `next` branch with this PR instead - adjust the changeset to be of type major + upgrade `@types/jest` deps.

JakeGinnivan:
Does the dependency on "@types/jest" need to be bumped too?

Andarist:
Yes, that would require updating as well.

@spudly will you have time to address the code review comments?

spudly:
Made the requested changes. Ready for review again.

spudly:
sorry, had conflicts. resolved.

Andarist:
Thanks for this! I should merge this soon.

As a point of reference for anyone interested in this - `@types/jest` got this second param just very recently (17 days ago) in v24.0.20. https://www.npmjs.com/package/@types/jest/v/24.0.20

cc @JakeGinnivan just as a confirmation - such change is OK as we should get ourselves up-to-date and there is no way to support both, right? I've tried searching if this can somehow be overloaded to match both, but couldn't find a way.

.changeset/mean-deers-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'jest-emotion': major
+---
+
+Added the `T` parameter to the `Matchers` interface in the typescript definitions for jest-emotion
```
Andarist:
```suggestion
Added the `T` parameter to the `Matchers` interface in the TypeScript definitions to make this module compatible with `@types/jest@^24.0.20`.
```

Andarist: I've wanted to commit this suggestion through GitHub interface - but seems like I don't have sufficient permissions to push to your fork.

You can check how to allow this for the future PRs here - https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork . But at the moment just commit this suggestion with a click on a button.

PR #1621, opened 2019-11-09
Upgrade Babel-related deps

### LINKED ISSUE #1614: Upgrade Babel-related deps
I'd like to upgrade them as one Babel helper doesn't get correctly inserted with a reference to `@babel/runtime` package. 

Wanted to just do it now, but don't want to get conflicts so I'd like to do it after landing https://github.com/emotion-js/emotion/pull/1575 . Leaving this here solely as a reminder for myself.

Andarist:
I don't think this deserves a changeset, it changes the dist files a little bit - but it's a micro optimization.

closes #1614


scripts/babel-preset-emotion-dev/src/index.js
```diff
@@ -11,8 +11,10 @@ module.exports = () => {
       require.resolve('babel-plugin-fix-dce-for-classes-with-statics'),
       require.resolve('@babel/plugin-transform-flow-strip-types'),
       require.resolve('babel-plugin-codegen'),
-      require.resolve('@babel/plugin-proposal-object-rest-spread'),
-      require.resolve('@babel/plugin-transform-runtime'),
+      [
+        require.resolve('@babel/plugin-transform-runtime'),
+        { version: require('@babel/runtime/package.json').version }
```
Andarist: this allows inserting references to `@babel/runtime` for helpers which were not available yet in 7.0.0-beta.0 release

scripts/babel-preset-emotion-dev/src/index.js
```diff
@@ -11,8 +11,10 @@ module.exports = () => {
       require.resolve('babel-plugin-fix-dce-for-classes-with-statics'),
       require.resolve('@babel/plugin-transform-flow-strip-types'),
       require.resolve('babel-plugin-codegen'),
-      require.resolve('@babel/plugin-proposal-object-rest-spread'),
```
Andarist: this was not using loose, I've removed it because it's already included nowadays in `@babel/preset-env` which in turn uses `loose` here, so it saves few bytes because we don't duplicate `objectSpread2` helper across 2 packages (they were bundled in as it was a new helper) and we just reuse regular `extends` helper

Issue #1272, opened 2019-03-13
TypeScript props and HTML attributes collisions


kshahani:
- `emotion` version: 10.0.0
- `react` version: ^16

**Relevant code:**

```javascript
type Props = {
  color: string | string [];
}

const MyComponent = styled('div')<Props>`
  color: pink;
`;
```

**What you did:**

I want a prop that coincides with the name of an attribute on the HTMLElement type (i.e.: `color` exists on a `div` element as a type `string`, but I want a prop called `color` which could be either a `string` or a `string[]`). For a little bit of more context, this is how `styled-system` works with its exported style prop `color` and type `ColorProps`.

**What happened:**

![image](https://user-images.githubusercontent.com/22452366/54301061-9774c780-4594-11e9-8e63-a8d25f642121.png)

The relevant debugging information is `Type 'string | string[]' is not assignable to type 'string'`.

**Reproduction:**

Simply copy the code snippet above into your editor.

**Problem description:**

I'm rendered unable to type this component without renaming that prop, which I really don't want to do because that breaks the styled-system contract.

**Suggested solution:**

No ideas! Any help welcome.

travisbloom:
I'm struggling with this as well. Trying to find if there's a way of omitting some of the default html props so we can override.

kshahani:
Our workaround is:

```
const MyComponent = styled<'div', Props>('div')`
  color: pink;
`;
```

...which allows the `ExtraProps` to override the attributes on a `div` element.

maximtmartin:
@kshahani are you sure this works?

When I try this I'm still getting typescript errors:


`type ColorProps = string | Partial<Record<Breakpoint, string> & { _: string; }>;`
`
type ButtonProps = ColorProps 
`

`const Button = styled<'button', ButtonProps>('button', {
  shouldForwardProp: prop => isPropValid(prop) || prop === 'routed'
})(...)`

usage:

`
 <Button
      color={{ _: 'green.2' }}
    >
   Test
    </Button>
`

Ts error:


> Type '{ _: string; }' is not assignable to type 'string | (string & Partial<Record<Breakpoint, string> & { _: string; }>) | undefined'.
>   Type '{ _: string; }' is not assignable to type 'string & Partial<Record<Breakpoint, string> & { _: string; }>'.
>     Type '{ _: string; }' is not assignable to type 'string'.
> 
> 23       color={{ _: 'green.2' }}
> 
>   node_modules/@types/react/index.d.ts:1682:9
[...]

PR #1624, opened 2019-11-11
Created overload of create styled which supports shouldForwardProp be…

JakeGinnivan:
**What**: A potential solution to #1272

**Why**: Because shouldForwardProps is a runtime solution to something which is not supported in the type system

**How**: I have created an additional overload in the create styled function which forces shouldForwardProps to be a custom type guard. 

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

If we like this solution I can add docs

### DIFF
--- a/.changeset/polite-tips-attend/changes.md
+++ b/.changeset/polite-tips-attend/changes.md
@@ -1,0 +1,15 @@
+Added createStyled overload to handle when shouldForwardProp is a custom type guard for intrinsic props
+
+As an example, if you want to override the type of the `color` prop:
+
+```ts
+export const Box = styled('div', {
+  shouldForwardProp: (
+    propName
+  ): propName is Exclude<keyof JSX.IntrinsicElements['div'], 'color'> =>
+    propName !== 'color'
+})<{ color: Array<string> }>(props => ({
+  color: props.color[0]
+}))
+;<Box color={['green']} />
+```
--- a/.changeset/polite-tips-attend/changes.md
+++ b/.changeset/polite-tips-attend/changes.md
@@ -1,4 +1,4 @@
-Added createStyled overload to handle when shouldForwardProp is a custom type guard for intrinsic props
+Added `CreateStyled` overload to handle when `shouldForwardProp` is a custom type guard for intrinsic props.
 
 As an example, if you want to override the type of the `color` prop:
 
--- a/packages/styled/types/tests.tsx
+++ b/packages/styled/types/tests.tsx
@@ -51,3 +51,17 @@ const Container2 = styled.div<{ test: number }>(props => ({
 const Container3 = styled.div(({ theme }) => ({
   width: theme.width
 }))
+
+// This example shows using shouldForwardProps to define a prop
+// who's type clashes with an intrinsic prop.
+// It makes use of a custom type guard on shouldForwardProp to exclude color
+
+export const Box = styled('div', {
+  shouldForwardProp: (
+    propName
+  ): propName is Exclude<keyof JSX.IntrinsicElements['div'], 'color'> =>
+    propName !== 'color'
+})<{ color: Array<string> }>(props => ({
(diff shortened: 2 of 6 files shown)


packages/styled/types/base.ts
```diff
@@ -90,12 +100,24 @@ export interface CreateStyledComponent<
 export interface CreateStyled<Theme extends {} = any> {
   <C extends React.ComponentType<React.ComponentProps<C>>>(
     component: C,
-    options?: StyledOptions
+    options?: StyledOptions<PropsOf<C>>
   ): CreateStyledComponent<PropsOf<C> & { theme?: Theme }, {}, { theme: Theme }>
 
+  <
```
JakeGinnivan: I added a new overload to prevent a breaking change of forcing the shouldForwardProp to be a type guard

Andarist:
Does this cover the use case described in #1272? Is it possible to create a styled HTML element with custom `shouldForwardProp` in such a way that we prevent forwarding a certain prop X and thus we can specify a type for that X prop that is different than the one defined on `JSX.IntrinsicElements[some_type]`.

Not sure what the best real-life example for this is but let's assume `div` & `color` prop - which has currently a `string` type - but we'd like to make it an array of strings.

JakeGinnivan:
Yeah it will do.

``` ts
export const Box = styled('div', {
  shouldForwardProp: (propName): propName is Omit<JSX.IntrinsicElements['div'], 'color'> => propName !== 'color' 
})(
)
```

Whatever the logic is for filtering out the props, they need to create a type guard of the same time. It's harder for omitting like above. Overall I think it's a good improvement over the casting

Andarist:
Could you add this as a test case?

JakeGinnivan:
@Andarist done

packages/styled/types/tests.tsx
```diff
@@ -51,3 +51,17 @@ const Container2 = styled.div<{ test: number }>(props => ({
...
 }))
+
+// This example shows using shouldForwardProps to define a prop
+// who's type clashes with an intrinsic prop.
+// It makes use of a custom type guard on shouldForwardProp to exclude color
+
+export const Box = styled('div', {
+  shouldForwardProp: (
+    propName
+  ): propName is Exclude<keyof JSX.IntrinsicElements['div'], 'color'> =>
```
Andarist: this is somewhat an example of how TS is unsound 😅 Not a big deal here though - it's just an example

Andarist:
Could you add a changeset for this? And ideally a short note in the docs?

packages/styled/types/tests.tsx
JakeGinnivan: Yeah, it would be great if TypeScript could generate a type guard signature based on the type narrowing inside a function like this. May be worth raising an issue?

JakeGinnivan:
@Andarist done, also added overload for normal components too

.changeset/polite-tips-attend/changes.md
```diff
@@ -0,0 +1,15 @@
+Added createStyled overload to handle when shouldForwardProp is a custom type guard for intrinsic props
```
Andarist:
```suggestion
Added `CreateStyled` overload to handle when `shouldForwardProp` is a custom type guard for intrinsic props.
```

Andarist: I've wanted to commit this suggestion through GitHub interface - but seems like I don't have sufficient permissions to push to your fork.

You can check how to allow this for the future PRs here - https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork . But at the moment just commit this suggestion with a click on a button.

Andarist:
Unfortunately, you need to resolve conflicts on this one.

JakeGinnivan:
@Andarist rebased

PR #1627, opened 2019-11-12
v11: fix for typescript definitions when using isolated modules

frankwallis:
Renaming:
`styled/types/base.ts` to `styled/types/base.d.ts`
`styled/types/helper.ts` to `styled/types/helper.d.ts`

Because when importing `@emotion/styled/base` with isolatedModules turned on, TypeScript was giving a compiler error:

```
ERROR: 21:3   expect  TypeScript@next compile error: 
Cannot re-export a type when the '--isolatedModules' flag is provided.
```

This happens because when transpiling a re-export with isolatedModules turned on TypeScript does not know if the exported item is a type or a runtime object and so cannot elide re-exported types.

By renaming the files to be definition files, this restriction no longer applies.

**Checklist**:
- [ ] Documentation - N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset - N/A

I have not created a changeset as this PR is against `next` and will go out with the next prerelease.
Thanks for all your hard work, really looking forward to the improvements in v11.
Let me know if you need me to make any more changes or create a changeset.


Andarist:
Hm, I would say this deserves a changeset as well - we are releasing prereleases often and seeing this in a changelog might help somebody. We will probably have to clean up the generated changelog before releasing final v11 anyway (to keep there only things that actually were done and to avoid noisy stuff).

Thanks for the PR! I wouldn't know that this has such effect without investigating this for quite a bit of time probably, so this is a great time-saver ❤️

packages/styled/types/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "baseUrl": "../",
     "forceConsistentCasingInFileNames": true,
+    "isolatedModules": true,
```
Andarist: do you see any downsides of using this in here? I can't think of any, cc @JakeGinnivan

frankwallis: `isolatedModules` will cause the compilation to emit errors if the code is not able to be transpiled in single-file mode (ie. babel or ts.transpile). So it will raise errors for doing type re-exports or using const enums. AFAIK all other errors will also be emitted so I think it is safe to keep it turned on and it may even be worth adding it to the other configuration files.

frankwallis: Some info about the impact here: https://github.com/microsoft/TypeScript/issues/16351

Andarist: > it may even be worth adding it to the other configuration files.

Yeah, I was thinking about this as well.

.changeset/odd-insects-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled': minor
```
Andarist:
```suggestion
'@emotion/styled': patch
```

.changeset/odd-insects-cough.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+fix for typescript error importing @emotion/styled/base when using isolatedModules
```
Andarist:
```suggestion
Fix for TypeScript error when importing `@emotion/styled/base` in combination with `isolatedModules` option.
```

Andarist: I've wanted to commit this suggestion through the GitHub interface - but it seems like I don't have sufficient permissions to push to your fork.

packages/styled/types/tsconfig.json
JakeGinnivan: Makes sense to me 👍

.changeset/odd-insects-cough.md
frankwallis: done, thanks

packages/styled/types/tsconfig.json
frankwallis: I can add that to the PR if you want when I get home this evening.

Andarist: Merged this one in as we can add this flag in other configs in separate PR. Thanks for the fix!

PR #1628, opened 2019-11-12
Move theming APIs to core

Andarist:
- [x] adjust docs
- [x] decide if we want to export ThemeContext from `@emotion/core`
- [x] decide if we want to keep or remove `withTheme`
- [x] changeset
- [x] prepare a deprecation entry point for `emotion-theming`

cc @mitchellhamilton @JakeGinnivan @joltmode


docs/typescript.mdx
```diff
@@ -241,8 +241,8 @@ type Theme = {
 export default styled as CreateStyled<Theme>
 
 // You can also create themed versions of all the other theme helpers and hooks
-const { ThemeProvider, withTheme, useTheme } = emotionTheming as emotionTheming.EmotionTheming<Theme>
-export { ThemeProvider, withTheme, useTheme }
+const { ThemeProvider, useTheme } = { ThemeProvider, useTheme } as EmotionTheming<Theme>
```
joltmode: The `ThemeProvider` and `useTheme` are lacking as imports, rendering the documentation a little incomplete.

JakeGinnivan:
What is the go with the remove theming RFC? If the plan is to migrate theming to hooks then maybe that should be the play instead?

docs/typescript.mdx
Andarist: good point, gonna fix together with other docs stuff 👍

Andarist:
> What is the go with the remove theming RFC? If the plan is to migrate theming to hooks then maybe that should be the play instead?

We believe in everything that is written in the RFC. The bare context seems to make more sense because it solves multiple problems - and the only downside is that you have to import and call the hook to get access to the context. We have decided to keep the theme in the core, at least for the time being, because v11 focus is to migrate internals to hooks and doing some minor cleanups. People would most likely get upset about removing the built-in theme. Ideally, I'd like to incorporate those docs from the RFC into published docs - just the tone will be a little bit different (after all we are not removing theme right now). Out of curiosity - what do you think about arguments presented there and how would you feel about built in theme removal?

JakeGinnivan:
I have no issues with theme being removed actually which is why I asked. I thought the RFC was pretty reasonable. Then again the planned simplification of theme also is nice. 

Just thinking that if the plan is to remove theming then it probably doesn't make sense to make these breaking changes now to just have users break again with v12 or whatever it is when theming is removed entirely. 

I do like the approach of documenting for libraries to to theming using hooks as documented in that RFC. It removes the conflicts issue and keeps theme in core for normal usage?

mitchellhamilton:
I'm thinking what we might want to do is export is exactly:

```jsx
export let ThemeContext = React.createContext({})
export let useTheme = () => useContext(ThemeContext)
```

I think this would be really good because it would have parity with the "create your own theme" approach so rather than thinking about theming as there being two approaches, there would only be the one approach but we would have a "default" one that you can use.

JakeGinnivan:
At that point you can't just destructure `theme` from props like you can do now, and you would still need to introduce the global theme type etc for these types.

Unsure including those 2 lines would provide much value over documenting them in the docs.

mitchellhamilton:
I still mean that we would pass the theme to styled interpolations and the css prop. I more mean we shouldn't have withTheme and a ThemeProvider component that merges the theme with the parent in the component.

joltmode:
While this wouldn't affect me, but @JakeGinnivan has a good point in here when thinking about the wider ecosystem:

> Just thinking that if the plan is to remove theming then it probably doesn't make sense to make these breaking changes now to just have users break again with v12 or whatever it is when theming is removed entirely.

JakeGinnivan:
I also use the merging capability of ThemeProvider a fair bit, so if we just exported the context then you would have to do the merging yourself.

We could also keep the ThemeProvider but it has an optional prop for the context to use? 

Though this is seeming like maybe we should just keep what we have, simplify as much as we can and then docs for how libraries should do theming to not conflict with the main app

packages/core/types/theming.d.ts
```diff
@@ -21,26 +12,13 @@ export interface ThemeProvider<Theme extends {} = any> {
   (props: ThemeProviderProps<Theme>): React.ReactElement
 }
 
-export type withTheme<Theme extends {} = any> = <
```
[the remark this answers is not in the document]
Andarist: I've removed it as part of this PR - but the decision about it is not final. It seems like in hooks world this is not needed. I know class components can't use hooks but they can do smth like that:
```js
// this example is not generic, but also `withTheme` tends to be used sparingly, so this might not be a problem
const WithTheme = props => {
  const theme = useTheme()
  return <SomeClass theme={theme} {...props}/>
}
```

So I don't know if we should keep builtin `withTheme`.

Andarist:
> Just thinking that if the plan is to remove theming then it probably doesn't make sense to make these breaking changes now to just have users break again with v12 or whatever it is when theming is removed entirely.

I don't mind keeping `theme` builtin indefinitely, but mostly because I think this is what people want and that removing it would lead to an unnecessary negative backlash. Major versions are here to introduce some breaking changes though, so I don't think that breaking people "again" is necessarily that bad. Moving this at a slower pace means more gradual migration than all at once - which might be in turn better. cc @mitchellhamilton 

> I do like the approach of documenting for libraries to to theming using hooks as documented in that RFC. It removes the conflicts issue and keeps theme in core for normal usage?

That's the current status of this, especially given that we don't know if we ultimately want to remove the builtin theme or not.

> I more mean we shouldn't have withTheme and a ThemeProvider component that merges the theme with the parent in the component.

I agree that `withTheme` seems unnecessary (OTOH it shouldn't cost us that much to keep it 🤷‍♂ ), but I'm not sure about `ThemeProvider`. Even though this abstraction is thin it is really convenient - I also don't believe this is some new API that has to be learned much, because it works really similarly to `setState`. The API doesn't do much, but what it does seems valuable to me.

> We could also keep the ThemeProvider but it has an optional prop for the context to use?

I kinda like that 👍

mitchellhamilton:
> I don't mind keeping theme builtin indefinitely

Agreed.

---

@JakeGinnivan and others, would you be able to share real world usages of ThemeProvider - especially usages that aren't at the top level?

Andarist:
For example smth like <Inverse/> which parses current theme and provides inverted colors for its subtree.

JakeGinnivan:
> I agree that withTheme seems unnecessary (OTOH it shouldn't cost us that much to keep it 🤷‍♂ ), but I'm not sure about ThemeProvider. Even though this abstraction is thin it is really convenient - I also don't believe this is some new API that has to be learned much, because it works really similarly to setState. The API doesn't do much, but what it does seems valuable to me.

Agree, I think my comment was directed at removing the typescript but not the actual implementation. I think we should remove both.

JakeGinnivan:
As for real world, I don't have anything which changes colours.

Here are some examples:

``` ts
export const thewest: Theme = {
    kind: Product.TheWest,
    fonts: fonts.thewest,
    margins: metrics.thewest.margins,
    cards: metrics.thewest.cards,
    componentMetrics: metrics.thewest.componentMetrics,

    // This following properties are provided through route resolution in provide-route-resolved-data-to-theme.tsx
    fallbackImages: undefined as any,
    section: 'default' as 'default',
    sectionValues: sections.thewest.default,
}

<ThemeProvider theme={theme}>
  <App />
</ThemeProvider>
```

Then inside the routing we have:

``` ts
 <ThemeProvider
            theme={{
                fallbackImages,
                section,
                sectionValues,
            }}
        >
            {children}
        </ThemeProvider>
```

This is how we do
https://7news.com.au/
vs
https://7news.com.au/sunrise

``` ts
export const OverrideThemeSection: React.FC<Props> = props => {
    const { children, section } = props
    return <ThemeProvider theme={{ section }}>{children}</ThemeProvider>
}
```

And this allows us to theme a particular part of a page in a theme from another section of the site.

docs/typescript.mdx
JakeGinnivan: If we do the global theme I think a lot of the work around this area will be wasted. Just something to keep in mind

packages/core/types/theming.d.ts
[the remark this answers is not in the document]
JakeGinnivan: We have pretty much migrated to hooks, but it has taken time. I think at a minimum you should add in the upgrade notes a snippet to polyfill it.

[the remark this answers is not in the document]
Andarist: Sure, a migration path for each change is going to be documented. I think this one even could be automated with codemods, just need to consider people wrapping components with multiple HOCs at once.

mitchellhamilton:
Okay, let's keep ThemeProvider and we might as well keep withTheme as well since it can be tree-shaken

Andarist:
This is pretty ready to be merged in. A final review would be appreciated.

Other than that there is only a single thing I'd like to decide on - should we export "raw" `ThemeContext` from the `@emotion/core` package?

mitchellhamilton:
> Other than that there is only a single thing I'd like to decide on - should we export "raw" ThemeContext from the @emotion/core package?

Yes because people might want to use `ThemeContext.Consumer` or `ThemeContext.Provider` directly.

JakeGinnivan:
> Yes because people might want to use ThemeContext.Consumer or ThemeContext.Provider directly.

Why would you use ThemeContext over ThemeProvider out of interest, that would mean there is two ways to do the same thing, one just has more features?

JakeGinnivan:
> > Yes because people might want to use ThemeContext.Consumer or ThemeContext.Provider directly.
> 
> Why would you use ThemeContext over ThemeProvider out of interest, that would mean there is two ways to do the same thing, one just has more features?

To clarify, over ThemeProvider *and* useTheme.

.changeset/weak-islands-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/core': major
+'emotion-theming': major
+---
+
+`emotion-theming` has been removed and all its exports were moved to `@emotion/core` package. Please import them like this `import { useTheme, ThemeProvider } from '@emotion/core'` from now on.
```
mitchellhamilton:
```suggestion
`emotion-theming` has been removed and all its exports were moved to `@emotion/core` package. Please import them like this `import { useTheme, ThemeProvider, withTheme } from '@emotion/core'` from now on.
```

packages/emotion-theming/src/index.js
```diff
@@ -1,4 +1,3 @@
-// @flow
-export { default as ThemeProvider } from './theme-provider'
-export { default as withTheme } from './with-theme'
-export { default as useTheme } from './use-theme'
+throw new Error(
+  "`emotion-theming` has been removed and all its exports were moved to `@emotion/core` package. Please import them like this `import { useTheme, ThemeProvider } from '@emotion/core'`."
```
mitchellhamilton:
```suggestion
  "`emotion-theming` has been removed and all its exports were moved to `@emotion/core` package. Please import them like this `import { useTheme, ThemeProvider, withTheme } from '@emotion/core'`."
```

mitchellhamilton:
For ThemeProvider, not wanting the merging behavior.

For useTheme, using `ThemeContext.Consumer` and `static contextType = ThemeContext` in a class component

PR #1631, opened 2019-11-13
chore(ci): GH Actions to yarn frozen-lockfile

maraisr:
What:
GitHub actions yarn install should be running the `frozen-lockfile` flag.

Why:
Because the lockfile _might_ not be listened too.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
         run: npm install --global yarn
 
       - name: Install Dependencies
-        run: yarn
+        run: yarn --frozen-lockfile
 
       - name: Create Release Pull Request or Publish to npm
         uses: changesets/action@master


Andarist:
This is true that it would be better to use this flag, I had in the past problems with it though. It was caused by optional, platform-dependent, dependency - mainly `fsevents`. Do you know how to workaround this? The problem was that my team was using MacOS and CI was running on Linux.

maraisr:
That is very interesting, and I never knew that was an issue. So maybe it is 🤷🏼‍♂️. However, im in a similar boat as you. My team and I and Mac and 2 Windows devs. And our CI runs Ubuntu. And have never had issues. 

I mean at the end of the day, its not like there is a "windowsDependencies", its all just dependencies. So that map would be, and probably should be the same cross environments. Binaries, is a postinstall script, which isn't part of the lockfile. So if fsevents need to be installed, afiak that is a node-gyp thing, which would still install/not-install per platform.

But Im more than happy to run some more analysis on this issue, if youre wanting to err on the side of caution. But in saying that, i've never given the flag a second thought. All my CI's run it.

Andarist:
> That is very interesting, and I never knew that was an issue. So maybe it is 🤷🏼‍♂️. However, im in a similar boat as you. My team and I are Mac and 2 Windows devs. And our CI runs Ubuntu. And have never had issues.

Hm, maybe I've misinterpreted something - didn't have time for that, had to regenerate lock file anyway and didn't add that flag back. Gonna try again if this shouldn't be a problem.

PR #1634, opened 2019-11-14
adding types for @emotion/native

patsissons:
**What**:

adds some types for `@emotion/native`

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset


Andarist:
Thank you for working on this! From what I can see those are based on emotion@10 typings - would you be willing to convert them to what we have currently on the next branch for `@emotion/styled`?

patsissons:
@Andarist the types are based on the `next` branch? am i missing something?

Andarist:
For example there are no equivalents for `ThemelessReactNativeStyledComponent` & `ThemedReactNativeStyledComponent` in https://github.com/emotion-js/emotion/blob/ad77ed24b4bfe62d6c80f0498cd7e76863e2f28e/packages/styled/types/base.d.ts

patsissons:
it's possible the naming is just confusing. Themeless just means that the `Theme` type is empty and the field is then removed from the downstream types.

Andarist:
I think this still interacts a little bit different with Theme. On `next` branch there are no conditional tests against Theme generic - it just gets passed along into appropriate types as argument.

Looking also for example at this:
https://github.com/emotion-js/emotion/blob/ad77ed24b4bfe62d6c80f0498cd7e76863e2f28e/packages/styled/types/base.d.ts#L134-L141
and in Theme shows up twice - one is optional (passed as ComponentProps) and one is required (passed as StyleProps). I guess this might be the difference linked to my first statement in this comment - you don't allow at the moment a `theme` prop which overrides context `theme`.

patsissons:
The conditional types were meant as an improvement to these types. i can completely remove the `theme` field from type system visibility (for style props) if it provides no value. I did have a mistake in my earlier types that is now corrected to allow overriding `theme` on styled components regardless of whether a contextual theme is present.

Andarist:
I would prefer to keep those types structured in the very same way as `@emotion/styled` types are structured - because quite frankly we don't personally use `@emotion/native`. The main focus of emotion is web and changes are made primarily for related packages. Having those in sync would at least make it a lot easier to backport changes from `@emotion/styled` to this package.

I'm not yet entirely sure, but it's also very likely that Theme will get reworked quite a bit (but not focus on this as part of this PR - unless you want to wait till we merge changes to `@emotion/styled`). We plan to leverage interface merging so people can just augment built-in type with their typed Theme. This should simplify things quite a bit - the work is being done as part fo this PR https://github.com/emotion-js/emotion/pull/1609 if you want to take a look.

patsissons:
ok, i reduced the types to just be analogies to the default `styled` types. We can't really share any types because of the react native stylesheet type (not without altering the `styled` types at least).

Andarist:
> We can't really share any types because of the react native stylesheet type (not without altering the styled types at least).

That's fine - sharing probably would make them more complex than they need to be.

Thanks for doing this! I'll probably merge this in over the weekend or shortly after that.

patsissons:
Shall I move this out of draft?

Andarist:
@patsissons yeah, you can do that. I'm in the middle of aligning those types just a little bit more to what we have in `@emotion/styled` now. (I'm in example including support for `shouldForwardProp`)

patsissons:
sure thing, let me know if you want me to make any other changes.

Andarist:
I've pushed out my changes - the work is not 100% done yet though. I would highly appreciate your review on this one.

From what I understand `@emotion/native` accepts a different set of CSS properties than "web" version and thus it doesn't quite make sense to reuse types from `@emotion/serialize` because they assume web. I've created copies of those which I have needed to make this work here. 

Unfortunately, types which I have used to create `ObjectInterpolation `  conflict with each other, so the created type is invalid:
https://github.com/emotion-js/emotion/blob/c3d48f9ebaaa75e0a0e9bce69d1fae774dfe31d7/packages/native/types/base.d.ts#L14-L18

I'm not sure what's the best way to work around this here. Any ideas?

Other than that tests need some work - I haven't really touched them, so they are using a type that doesn't exist anymore. Would you be willing to work on making those tests work?

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,206 @@
+// Definitions by: Pat Sissons <https://github.com/patsissons>
+// TypeScript Version: 3.4
+
+import { ComponentPropsWithoutRef, ComponentType } from 'react'
+import * as RN from 'react-native'
+
+type ReactNative = typeof RN
+
+export type ReactNativeStyle = ReturnType<ReactNative['StyleSheet']['flatten']>
+export type ObjectInterpolation = RN.ViewStyle | RN.TextStyle | RN.ImageStyle
```
patsissons: this type union should be fine instead of the polymorphic version that was being used previously.

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,206 @@
...
+  ForwardedProps extends keyof Props = keyof Props
+> {
+  shouldForwardProp?(propName: PropertyKey): propName is ForwardedProps
+}
+
+export interface StyledOptions<Props> {
+  shouldForwardProp?(propName: PropertyKey): boolean
+}
+
+export interface StyledWithComponent<ComponentProps extends {}> {
```
patsissons: open to a better name for this

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,206 @@
...
+    },
+    {},
+    { theme: Theme }
+  >
+
+  <Component extends ComponentType<ComponentPropsWithoutRef<Component>>>(
+    component: Component,
+    options?: StyledOptions<ComponentPropsWithoutRef<Component>>
+  ): CreateStyledComponent<
+    ComponentPropsWithoutRef<Component> & { theme?: Theme },
```
patsissons: i wasn't sure why the custom component variant is using an intersection for component props rather than specific props like the react native components are below.

packages/native/types/base.d.ts
Andarist: Seems like it's hard for us to create strict typings for those if we want to have types similar to those for the web version. From what I understand this will indeed work somewhat OK, but will allow incorrect object styles to be set on a particular component (in example it will allow text styles to be used on a view component). To keep things simple, I would say it's OK. WDYT?

Andarist:
I see you have restructured typings a little bit. I'm not saying that those changes are bad, most of them are probably good - but also many of them could be considered rather cosmetic. I've intentionally written them like I've written them, so they could mirror what we have in for a web version.

Truth to be told `@emotion/native` has been created by a community member, we've merged it in and all - as it works. We don't have time to focus on it though in the same way as we focus on the web version. That's why I think it's crucial for us to have those types as close as possible to each other so we can sync them in the future. If we diverge them now it will just be way harder for us to maintain this as we are not using `@emotion/native` in any projects on our own.

If you believe that some changes you have made are particularly good we could evaluate if they could be backported to `@emotion/styled` as well - that could be done in followup PRs though.

packages/native/types/base.d.ts
Andarist: cc @JakeGinnivan, that was based on the types for `@emotion/styled`

packages/native/types/base.d.ts
patsissons: I'll take another pass to see if I can improve it, but I think that may not be avoidable

patsissons:
Most of the cosmetic/name changes were to help me trace how the types were being expanded so I could ensure all types were being forwarded through properly. Removal of the `PropsOf` helper and changing `Tag` to `Component` is because native has no concept of jsx tags, so it's specific to react native. If there were some other changes you would like to discuss I'm happy to explain any decisions and revert if that's desired.

packages/native/types/base.d.ts
```diff
@@ -3,25 +3,21 @@
 
 import { ComponentPropsWithoutRef, ComponentType } from 'react'
 import * as RN from 'react-native'
-import { PropsOf } from '@emotion/core'
```
patsissons: the `ComponentProps` construct in the official React typings should be the standard for props extraction on new typing efforts, it differs slightly from the `PropsOf` in core but in ways that produce more accurate props (specifically for `react-native`). In the case of `react-native`, i can't think of a reason the `ref` prop would be used in styles so i went with the `WithoutRef` variant, though i'm totally willing to stand down on that assumption if there is any reason whatsoever to use `ref` in styles (my understanding is that it's only used for native prop manipulation).

packages/native/types/base.d.ts
```diff
@@ -3,25 +3,21 @@
...
-    RN.TextStyle,
-    RN.ImageStyle {}
+export type ObjectInterpolation = RN.ViewStyle | RN.TextStyle | RN.ImageStyle
 
 export interface FunctionInterpolation<MergedProps> {
   (mergedProps: MergedProps): Interpolation<MergedProps>
 }
 
-export type Interpolation<MergedProps = undefined> =
+export type Interpolation<MergedProps = unknown> =
```
patsissons: `unknown` is a more descriptive default for these `MergedProps` than `undefined`

packages/native/types/base.d.ts
```diff
@@ -32,7 +28,7 @@ export type Interpolation<MergedProps = undefined> =
   | ArrayInterpolation<MergedProps>
   | FunctionInterpolation<MergedProps>
 
-type ReactNativeTags =
+export type ReactNativeComponentNames =
```
patsissons: I was originally going to use `ReactNativeStylableComponentNames` but that seemed overly verbose.

packages/native/types/base.d.ts
```diff
@@ -73,9 +69,11 @@ type ReactNativeTags =
   | 'ViewPagerAndroid'
   | 'WebView'
 
-export type ReactNativeElements = {
-  [Tag in ReactNativeTags]: PropsOf<ReactNative[Tag]>
-}
+export type ReactNativeComponents = Pick<ReactNative, ReactNativeComponentNames>
```
patsissons: `ReactNativeElements` is a very popular `react-native` design system, so to be less confusing I renamed this to a more descriptive `ReactNativeComponents`. Additionally, converting the mapped type description into the more compact (and equivalent) `Pick` construct.

packages/native/types/base.d.ts
```diff
@@ -73,9 +69,11 @@ type ReactNativeTags =
...
+
+export type ReactNativeComponentProps<
+  ComponentName extends ReactNativeComponentNames
+> = ComponentPropsWithoutRef<ReactNativeComponents[ComponentName]>
```
patsissons: I noticed i was needing to extract the props out quite frequently, so i built a handy extractor construct.

packages/native/types/base.d.ts
```diff
@@ -89,21 +87,26 @@ export interface StyledOptions<Props> {
...
+export interface StyledWithComponent<ComponentProps extends {}> {
+  withComponent<
+    Component extends ComponentType<ComponentPropsWithoutRef<Component>>
+  >(
+    component: Component
+  ): StyledComponent<ComponentProps & ComponentPropsWithoutRef<Component>>
+  withComponent<ComponentName extends ReactNativeComponentNames>(
+    component: ReactNativeComponents[ComponentName]
+  ): StyledComponent<ComponentProps, ReactNativeComponentProps<ComponentName>>
+}
```
patsissons: these `withComponent` signatures were pulled out into their own interface so that we can use a type union below for `StyledComponent`.

packages/native/types/base.d.ts
```diff
@@ -89,21 +87,26 @@ export interface StyledOptions<Props> {
...
-> extends React.FC<ComponentProps & SpecificComponentProps> {
-  withComponent<C extends React.ComponentType<React.ComponentProps<C>>>(
-    component: C
-  ): StyledComponent<ComponentProps & PropsOf<C>>
-  withComponent<Tag extends keyof ReactNativeElements>(
-    tag: Tag
-  ): StyledComponent<ComponentProps, ReactNativeElements[Tag]>
-}
+> = ComponentType<ComponentProps & SpecificComponentProps> &
+  StyledWithComponent<ComponentProps>
```
patsissons: It doesn't make much sense to type `StyledComponent` as an extension to a functional component. Technically it should be typed as a `ForwardRefExoticComponent` (and more generally, `NamedExoticComponent`). This might be a more appropriate choice (i'll give it a try in a moment).

packages/native/types/base.d.ts
```diff
@@ -149,45 +152,58 @@ export interface CreateStyledComponent<
...
     { theme?: Theme },
-    ReactNativeElements[Tag],
+    ReactNativeComponentProps<ComponentName>,
     { theme: Theme }
   >
 }
 
-declare const styled: CreateStyled
-export default styled
+export const styled: CreateStyled
```
patsissons: as per the code in `base.js` the `styled` member is a named export, and there is no default export in that module.

packages/native/types/base.d.ts
patsissons: an added note, we may still need/want to `import '@emotion/core'` here so that the side effects of core are automatically applied when importing `native`.

packages/native/types/tslint.json
```diff
@@ -1,20 +1,24 @@
...
+      "check-module",
+      "check-rest-spread",
+      "check-type",
+      "check-typecast",
+      "check-type-operator",
+      "check-preblock"
+    ],
+    "no-null-undefined-union": false,
+    "strict-export-declare-modifiers": false
+  }
```
patsissons: these rules were copied from `core` and then pared down a bit for rules that didn't need to be disabled. we grew `"callable-types": false` to deal with the `FunctionInterpolation` definition, and the two rules at the bottom to deal with `type Interpolation` and `type ReactNative = typeof RN` respectively.

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,227 @@
...
+import {
+  ComponentPropsWithoutRef,
+  ComponentType,
+  NamedExoticComponent
+} from 'react'
+import * as RN from 'react-native'
+
+type ReactNative = typeof RN
+
+export type ReactNativeStyle = RN.ViewStyle | RN.TextStyle | RN.ImageStyle
```
patsissons: I noticed the old `ReturnType<typeof RN.StyleSheet.flatten>` was actually producing an incorrect result (not sure why) of just `ViewStyle`. So I'm just manually defining the union instead now.

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,227 @@
...
+
+export type ReactNativeComponentProps<
+  ComponentName extends ReactNativeComponentNames
+> = ComponentPropsWithoutRef<ReactNativeComponents[ComponentName]>
+
+export type ReactNativeStyleType<Props> = Props extends {
+  style?: RN.StyleProp<infer StyleType>
+}
+  ? StyleType
+  : ReactNativeStyle
```
patsissons: this construct lets us extract the specific shape of a `react-native` style prop, and support custom components too.

packages/native/types/base.d.ts
patsissons: `NamedExoticComponent` works appropriately here so i've gone with that choice.

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * @typeparam ComponentProps  Props which will be included when withComponent is called
+ * @typeparam SpecificComponentProps  Props which will *not* be included when withComponent is called
+ * @typeparam StyleProps  Params passed to styles but not exposed as React props. These are normally library provided props
+ */
+export interface CreateStyledComponent<
+  ComponentProps extends {},
+  SpecificComponentProps extends {} = {},
+  StyleProps extends {} = {},
+  StyleType = ReactNativeStyle
```
patsissons: I found that the `StyleType`, while properly inferred, was allowing an `ImageStyle` to be used as a `TextStyle` but not the other way around. Feels like a bug in the TypeScript inference logic because i couldn't see a relation between the two, but i couldn't get a better answer as to why (i left some comments in the test file).

packages/native/types/base.d.ts
```diff
@@ -0,0 +1,227 @@
...
+  >
+
+  <Component extends ComponentType<ComponentPropsWithoutRef<Component>>>(
+    component: Component,
+    options?: StyledOptions<ComponentPropsWithoutRef<Component>>
+  ): CreateStyledComponent<
+    ComponentPropsWithoutRef<Component> & { theme?: Theme },
+    {},
+    { theme: Theme },
+    ReactNativeStyleType<ComponentPropsWithoutRef<Component>>
```
patsissons: as long as the custom component defines a `style?: StyleProp<T>`, the `T` will get inferred and applied to the template interpolation types.

packages/native/types/index.d.ts
```diff
@@ -0,0 +1,48 @@
+// Definitions by: Pat Sissons <https://github.com/patsissons>
+// TypeScript Version: 3.4
+
+import '@emotion/core'
```
patsissons: LMK if this makes sense. I think it's required and is best imported here on the api surface.

packages/native/types/index.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  ...args: Array<Interpolation>
+): ReactNativeStyle
+export function css(...args: Array<Interpolation>): ReactNativeStyle
+
+export type StyledComponents<Theme extends {} = any> = {
+  [ComponentName in ReactNativeComponentNames]: CreateStyledComponent<
+    { theme?: Theme },
+    ReactNativeComponentProps<ComponentName>,
+    { theme: Theme },
+    ReactNativeStyleType<ReactNativeComponentProps<ComponentName>>
```
patsissons: all styled `react-native` components should now be strongly typed against their specific style prop shape.

patsissons:
I ran into some composition issues testing out these types with a more complicated setup. The `react-native` styles sometimes override the definitions of certain style fields which can cause grief if you want to share some common styles and compose them downstream. I updated the `css` function signature to support optionally explicit typing to solve this, as well as implicit typing if the function is used with simple style objects (though these won't be type inferred at definition time). This appears to be working properly in my project, so hopefully it's a worthwhile fix.

Andarist:
@patsissons sorry that I haven't yet reviewed this, I hope to do it this week

packages/native/types/index.d.ts
Andarist: Actually I'm not sure what this was supposed to do - especially in a definition file, but it doesn't matter as I import `Theme` from `@emotion/core` here now, so it's in here one way or another

packages/native/types/base.d.ts
Andarist: this makes sense, cc @JakeGinnivan - shouldn't we do the same for web types?

packages/native/types/base.d.ts
Andarist: wouldnt we be able to do the same without pulling out into its own interface?

packages/native/types/base.d.ts
Andarist:
```ts
type True = RN.ImageStyle extends RN.TextStyle ? true : false
type False = RN.TextStyle extends RN.ImageStyle ? true : false
```

Seems like ImageStyle is a subtype of TextStyle

packages/native/types/index.d.ts
```diff
@@ -0,0 +1,50 @@
...
+  Interpolation,
+  ObjectInterpolation,
+  ReactNativeStyle,
+  StyledComponent,
+  StyledOptions
+} from './base'
+
+export function css<StyleType extends ReactNativeStyle = ReactNativeStyle>(
+  template: TemplateStringsArray,
+  ...args: Array<Interpolation>
```
Andarist: I'm not sure if this should use `Interpolation`. For instance, it doesn't make sense for `css` to accept functions - for the web we've split this into two similar, but separate, Interpolation and CSSInterpolation. I feel like the same should be done here.

Andarist:
@patsissons sorry that I haven't reviewed this sooner but had plenty on my plate. Let me know if you are still interested in working on this one (not much left to do).

packages/native/types/index.d.ts
patsissons: i think it was just to make sure any ambient types were automatically pulled in, but you're right that since we're pulling in `Theme` now it's already there.

packages/native/types/base.d.ts
patsissons: haha interesting, i'm not going to dig further since it's already difficult to understand why that may be.

packages/native/types/base.d.ts
patsissons: I think you're right, we don't need this to be its own interface any more (i think it was useful before when i needed to use a union).

packages/native/types/index.d.ts
patsissons: makes sense to me, i can add those suggestions in pretty quickly

patsissons:
> @patsissons sorry that I haven't reviewed this sooner but had plenty on my plate. Let me know if you are still interested in working on this one (not much left to do).

No worries, i'm in the same boat. I've addressed your review suggestions and things looks reasonable to me right now. My mind isn't as deep into these types as it was a few weeks ago so maybe spot check the changes i made. Otherwise i think i'm satisfied with these types for an MVP release, i'll try and pull them into a project later today to give them a broader test.

packages/native/types/base.d.ts
Andarist: Similarly on web input is a subtype of div. It's not like those types form such hierarchy, but rather that div is so basic, that it has no special props associated with it and everything that can be used on it can be used on input as well (+ input has some specific props for it). And because TS type system is structural rather than nominal this forms implicit supertype-subtype relationship.

packages/native/types/base.d.ts
```diff
@@ -67,10 +67,32 @@ export type ReactNativeStyleType<Props> = Props extends {
...
 export type ObjectInterpolation<
   StyleType extends ReactNativeStyle = ReactNativeStyle
 > = StyleType
 
+export interface ArrayCSSInterpolation<
+  MergedProps,
+  StyleType extends ReactNativeStyle = ReactNativeStyle
+> extends Array<CSSInterpolation<MergedProps, StyleType>> {}
+
+export type CSSInterpolation<
```
Andarist: I don't think that `MergedProps` props are actually used anyhow inside CSSInterpolation, it only seems to be passed around, but not utilized. This is good though - with `css` you don't get access to any props at all. This means though that we can just drop this typeparam here.

patsissons: yup, absolutely. a bit of blind refactoring lands you with useless type params 😄

patsissons:
@Andarist are we good to ship? 🚢

packages/native/types/base.d.ts
resolritter: @patsissons instead of hardcoding the names like this, what do you think of:

```typescript
type ReactNativeComponents = Pick<ReactNative, {
      [Key in keyof ReactNative]: ReactNative[Key] extends ComponentType ? Key : never
  }[keyof ReactNative]>
type ReactNativeComponentNames = keyof ReactNativeComponents
```

I'm using it like this:
```typescript
  type ActivityIndicatorProps = ComponentPropsWithoutRef<ReactNativeComponents["ActivityIndicator"]>
```

And it says the type is `type ActivityIndicatorProps = RN.ActivityIndicatorProps`, which seems about right.

patsissons: the hardcoded names are specifically to represent the [hardcoded component exports](https://github.com/emotion-js/emotion/blob/master/packages/native/src/index.js#L8-L57) in the native module. It's possible that the two sets are the same, but i wanted the styles to accurately match the module, at least for this first release.

PR #1636, opened 2019-11-15
Update Jest Matchers Generics

azizhk:
**What**:
Jest's Typings have been updates and Matchers now have a new Generic.
So just updating that @types/jest

**Why**:
Because jest version is 24 and @types/jest version should also be 24

**Checklist**:

- [ ] Documentation - NA
- [ ] Tests - NA
- [x] Code complete
- [x] Changeset

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "@mdx-js/mdx": "^1.1.0",
     "@mdx-js/react": "^1.0.27",
     "@testing-library/react": "^8.0.2",
-    "@types/jest": "^23.0.2",
+    "@types/jest": "^24.0.23",
     "@types/node": "^10.11.4",
     "@types/react": "^16.8.20",
     "babel-check-duplicated-nodes": "^1.0.0",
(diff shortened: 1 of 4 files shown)


Andarist:
Thanks for doing this. However, this is sort of a breaking change for existing `jest-emotion` users and because we are working on a new major version (v11) it seems reasonable to batch this change together with other breaking stuff. In fact, the similar change has been already merged into v11 branch. You can check out the changelog and stuff here - https://github.com/emotion-js/emotion/pull/1600 and you could even install those beta versions to benefit from the fix, up to this point in time there were not many breaking changes merged into v11 so it should be fairly easy to migrate (and even when we introduce breaking changes we plan to provide codemods to help you migrate automatically). Just remember to pin your versions if you decide to do so.

azizhk:
Oh cool, no worries.

We are actually still stuck on v9 😅
Just picked up the task of upgrading to v10.
Have patched my changes for now using [patch-package](https://www.npmjs.com/package/patch-package)
https://github.com/anarock/pebble/pull/274/files#diff-16916b32ece077af6364c7d374ab1be1

PR #1642, opened 2019-11-18
Added basic support for accepting custom `shouldForwardProp` option in React Native related packages.

### LINKED ISSUE #900: shouldForwardProp in styled component

Andarist:
closes #900

There is an open PR for adding TS types for `@emotion/native` - https://github.com/emotion-js/emotion/pull/1634 - and I'd like to keep those types in sync with what we have for `@emotion/styled` and to make it easier I had to implement basic support for this here.

### DIFF
--- a/packages/primitives-core/src/styled.js
+++ b/packages/primitives-core/src/styled.js
@@ -1,23 +1,35 @@
 // @flow
 import * as React from 'react'
-import { testAlwaysTrue, pickAssign, interleave } from './utils'
+import { interleave } from './utils'
 import { ThemeContext } from '@emotion/core'
 import { createCss } from './css'
 
-let defaultPickTest = prop => prop !== 'theme'
+let testOmitPropsOnComponent = prop => prop !== 'theme'
 
-type options = {
+type CreateStyledOptions = {
   getShouldForwardProp: (cmp: React.ElementType) => (prop: string) => boolean
+}
+
+type StyledOptions = {
+  shouldForwardProp?: (prop: string) => boolean
 }
 
 export function createStyled(
   StyleSheet: Object,
-  { getShouldForwardProp = () => defaultPickTest }: options = {}
+  {
+    getShouldForwardProp = () => testOmitPropsOnComponent
+  }: CreateStyledOptions = {}
 ) {
   const css = createCss(StyleSheet)
 
-  return function createEmotion(component: React.ElementType) {
-    let pickTest = getShouldForwardProp(component)
+  return function createEmotion(
+    component: React.ElementType,
+    options?: StyledOptions
+  ) {
+    let shouldForwardProp =
+      options && options.shouldForwardProp
+        ? options.shouldForwardProp
+        : getShouldForwardProp(component)
 
     return function createStyledComponent(...rawStyles: *) {
       let styles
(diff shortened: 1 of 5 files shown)


packages/primitives-core/src/styled.js
```diff
@@ -31,22 +43,33 @@ export function createStyled(
       // do we really want to use the same infra as the web since it only really uses theming?
       // $FlowFixMe
       let Styled = React.forwardRef((props, ref) => {
-        let mergedProps = pickAssign(testAlwaysTrue, {}, props, {
-          theme: props.theme || React.useContext(ThemeContext)
-        })
+        let mergedProps = props
```
Andarist: those changes align the implementation to what we have in `@emotion/styled`

Andarist:
I'm going to just merge this in as native packages are not our main focus. I'm open to review even after merging - changes are rather trivial though, so 🤷‍♂

PR #1643, opened 2019-11-18
Relaxed types for shouldForwardProp

### LINKED ISSUE #1641: v11 types: new shouldForwardProp type is overly restrictive
**Current behavior:**

I have been trying out the new @next tags, and have run into a type error with code that previous type-checked and worked. I'm fairly certain this is due to the changes in https://github.com/emotion-js/emotion/pull/1624

The `shouldForwardProp` option is now inferred to being of type `(prop: keyof Props) => prop is ForwardedProps`. But this restricts `prop` to being one of the known props of an element type.

**To reproduce:**

See [the code sandbox](https://codesandbox.io/s/emotion-issue-template-7xt3k) for an example that used to work and now fails
  
**Expected behavior:**

I *think* that the argument to `shouldForwardProp` should always just be a `string`, since at the moment `styled` is called, we don't know what additional props the component might accept.

**Environment information:**

- Typescript 3.6.2
- Emotion libs @next (see codepen for exact versions)
- @types/react 16.9.4

JakeGinnivan:
This function needs to be able to filter a props for a generic argument of the resulting function.

Fixes #1641

**Checklist**:
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/cool-candles-lie.md
+++ b/.changeset/cool-candles-lie.md
@@ -1,0 +1,5 @@
+---
+'@emotion/is-prop-valid': minor
+---
+
+[TypeScript] Allow isPropValid to take any PropertyKey as an argument (instead of just string)
--- a/.changeset/cool-candles-lie.md
+++ b/.changeset/cool-candles-lie.md
@@ -2,4 +2,4 @@
 '@emotion/is-prop-valid': minor
 ---
 
-[TypeScript] Allow isPropValid to take any PropertyKey as an argument (instead of just string)
+Allow `isPropValid` to take any `PropertyKey` as an argument (instead of just `string`).
--- a/.changeset/dull-carrots-enjoy.md
+++ b/.changeset/dull-carrots-enjoy.md
@@ -4,4 +4,3 @@
 
 Relaxed types for shouldForwardProp
 
-This function needs to be able to filter a props for a generic argument of the resulting function.
--- a/.changeset/dull-carrots-enjoy.md
+++ b/.changeset/dull-carrots-enjoy.md
@@ -1,0 +1,7 @@
+---
+'@emotion/styled': patch
+---
+
+Relaxed types for shouldForwardProp
+
+This function needs to be able to filter a props for a generic argument of the resulting function.
--- a/.changeset/dull-carrots-enjoy.md
+++ b/.changeset/dull-carrots-enjoy.md
@@ -2,5 +2,5 @@
 '@emotion/styled': patch
 ---
 
-Relaxed types for shouldForwardProp
+Relaxed types for `shouldForwardProp` as it needs to be able to filter props for a generic argument of the resulting function.
 
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -31,13 +31,13 @@ export interface FilteringStyledOptions<
   ForwardedProps extends keyof Props = keyof Props
 > {
   label?: string
-  shouldForwardProp?(propName: keyof Props): propName is ForwardedProps
+  shouldForwardProp?(propName: PropertyKey): propName is ForwardedProps
   target?: string
 }
 
 export interface StyledOptions<Props> {
   label?: string
-  shouldForwardProp?(propName: keyof Props): boolean
+  shouldForwardProp?(propName: PropertyKey): boolean
   target?: string
 }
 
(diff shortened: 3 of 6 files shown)


packages/styled/types/tests-base.tsx
```diff
@@ -98,6 +99,22 @@ const Canvas1 = styled('canvas', {
 ;<Canvas0 id="id-should-be-filtered" />
 ;<Canvas1 />
 
+const styledWithForwardedExtraProp = styled('div', {
+  // this used to work, but now complains because 'priority' is not a prop of 'div'
+  shouldForwardProp: prop => prop !== 'priority' && isPropValid(prop)
+})
+
+type Priority = 'info' | 'warning' | 'error'
+interface Theme {
```
Andarist: this seems unused

packages/styled/types/tests-base.tsx
```diff
@@ -98,6 +99,22 @@ const Canvas1 = styled('canvas', {
 ;<Canvas0 id="id-should-be-filtered" />
 ;<Canvas1 />
 
+const styledWithForwardedExtraProp = styled('div', {
+  // this used to work, but now complains because 'priority' is not a prop of 'div'
```
Andarist: the comment is relevant as a reason for this PR - it will look really odd though in the code after this gets merged in

JakeGinnivan: Fixed

packages/styled/types/tests-base.tsx
JakeGinnivan: Removed

packages/styled/types/base.d.ts
```diff
@@ -31,13 +31,13 @@ export interface FilteringStyledOptions<
   ForwardedProps extends keyof Props = keyof Props
 > {
   label?: string
-  shouldForwardProp?(propName: keyof Props): propName is ForwardedProps
+  shouldForwardProp?(propName: PropertyKey): propName is ForwardedProps
```
Andarist: you are not extracting anything based on this, couldnt we just use a simple `string` here?

JakeGinnivan:
```
A type predicate's type must be assignable to its parameter's type.
  Type 'ForwardedProps' is not assignable to type 'string'.
    Type 'keyof Props' is not assignable to type 'string'.
      Type 'string | number | symbol' is not assignable to type 'string'.
        Type 'number' is not assignable to type 'string'.ts(2677)
```

is the error I get

Andarist: Ah, so it has to match to ` ForwardedProps extends keyof Props`, interesting. Gonna give it a thought later, but probably we'll keep PropertyKey if that's the case.

JakeGinnivan: I can replace `keyof Props` with `Extract<keyof Props, string>` to only extract strings which works. It just makes everything look more complex.

.changeset/cool-candles-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/is-prop-valid': minor
+---
+
+[TypeScript] Allow isPropValid to take any PropertyKey as an argument (instead of just string)
```
Andarist:
```suggestion
Allow `isPropValid` to take any `PropertyKey` as an argument (instead of just `string`).
```

.changeset/dull-carrots-enjoy.md
```diff
@@ -0,0 +1,7 @@
+---
+'@emotion/styled': patch
+---
+
+Relaxed types for shouldForwardProp
```
Andarist:
```suggestion
Relaxed types for `shouldForwardProp` as it needs to be able to filter props for a generic argument of the resulting function.
```

.changeset/dull-carrots-enjoy.md
```diff
@@ -0,0 +1,7 @@
...
+
+This function needs to be able to filter a props for a generic argument of the resulting function.
```
Andarist:
```suggestion
```

.changeset/cool-candles-lie.md
Andarist: I've wanted to commit this suggestion through the GitHub interface - but it seems like I don't have sufficient permissions to push to your fork.

PR #1648, opened 2019-11-22
Support shallow rendering

### LINKED ISSUE #949: toHaveStyleRule does not work with Emotion 10
- `emotion` version: 10.0.0-beta.5
- `react` version: 16.5.2

Relevant code.

```javascript
const EmotionComp = () => (
  <div
    css={css`
      width: 500px;
    `}
  />
);

describe("emotion-issue", () => {
  it("should have width", () => {
    const wrapper = shallow(<EmotionComp />);
    expect(wrapper).toHaveStyleRule("width", expect.anything());
  });
});
```

What you did:
Try to test styled component with `toHaveStyleRule` matcher and enzyme.

What happened:
The matcher cannot find `className`.

Reproduction:
See `jsx.spec.js` and `styled.spec.js` in the following link.
https://codesandbox.io/s/woo7jr1515

Problem description:
Since Emotion 10 uses React>=16.3 context API, we cannot get `ActualComponent` just with `.dive()` method.
https://github.com/emotion-js/emotion/blob/98a40a0002c11c63d7f4a28a6c62b1e4d0c4d7eb/packages/jest-emotion/src/utils.js#L3-L27

Suggested solution:
Add more condition for enzyme case so that we can handle `Context.Consumer` case.

ajs139:
**What**:
Fix #949, support Shallow Rendering

**Why**:
While using shallow rendering is generally considered unwise, there are projects that wish to convert from CSS to CSS-in-JS and that already use shallow rendering in their tests.

**How**:
The `jest-emotion` package was updated to dive into shallow rendered components to trigger the addition of the styles to the DOM to support `toHaveStyleRule`. Snapshots are supported by adding classNames to the shallow-rendered object by leveraging the `css` prop.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

The snapshot support is less than ideal, but it is much improved over the existing behavior. For example, it ensures that the `map` props aren't included in the snapshot. If this PR is accepted, it is probably a good idea to still recommend against shallow rendering in general, and noting that YMMV with shallow rendered snapshots.


packages/jest-emotion/test/matchers.test.js
```diff
@@ -67,9 +67,11 @@ describe('toHaveStyleRule', () => {
       const wrapper = enzyme[method](<Component />)
       expect(wrapper).toHaveStyleRule('color', 'red')
       expect(wrapper).not.toHaveStyleRule('width', '100%')
-      const svgNode = wrapper.find('svg')
-      expect(svgNode).toHaveStyleRule('width', '100%')
-      expect(svgNode).not.toHaveStyleRule('color', 'red')
+      if (method !== 'shallow') {
```
ajs139: If we're using a shallow render, we won't be able to find the svg element, as it will be wrapped - this is expected behavior, and inline with other CSS-in-JS libs.

Andarist:
Thanks for preparing a PR! ❤️ I'll try to review this over the weekend.

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -4,31 +4,22 @@ import * as enzyme from 'enzyme'
...
+import toJson from 'enzyme-to-json'
 
 expect.addSnapshotSerializer(createSerializer())
+expect.addSnapshotSerializer(createEnzymeSerializer())
 
-expect.extend({
-  toMatchShallowSnapshot: createEnzymeSnapshotMatcher(),
-  toMatchDeepSnapshot: createEnzymeSnapshotMatcher({ mode: 'deep' })
-})
+const expectToMatchSnapshot = component => {
```
Andarist: for cases like this we use `jest-in-case` helper

Andarist:
I've started looking into this, but unfortunately, I'm too tired right now to focus on this properly. Gonna finish the review tomorrow.

packages/jest-emotion/test/react-enzyme.test.js
ajs139: Cool, will check it out.

packages/jest-emotion/src/index.js
```diff
@@ -56,43 +56,76 @@ type Options = {
...
+function filterEmotionProps(props = {}, shouldKnowStyles = true) {
   const {
     css,
     __EMOTION_TYPE_PLEASE_DO_NOT_USE__,
     __EMOTION_LABEL_PLEASE_DO_NOT_USE__,
     ...rest
   } = props
 
-  rest.css = 'unknown styles'
+  if (shouldKnowStyles) {
```
Andarist: Could you explain how this works now? Looks a little bit quirky - why this gets set here based on the argument and without any condition based on actual `props`?

packages/jest-emotion/src/index.js
```diff
@@ -56,43 +56,76 @@ type Options = {
...
         ...val,
         props: filterEmotionProps(val.props),
         type: val.props.__EMOTION_TYPE_PLEASE_DO_NOT_USE__
       })
     }
+    if (isEmotionCssPropEnzymeElement(val)) {
+      let expectedClassNames = (val.props.css.name || '')
+        .split(' ')
+        .map(className => keys.map(key => `${key}-${className}`))
+        .reduce((flat, values) => flat.concat(values), [])
```
Andarist: there is already a `flatMap` in `jest-emotion`, could be reused here
https://github.com/emotion-js/emotion/blob/e20cf6ad6133db82d9991ef354e01811e20aedc5/packages/jest-emotion/src/utils.js#L3

packages/jest-emotion/src/utils.js
```diff
@@ -26,12 +26,20 @@ function isTagWithClassName(node) {
...
-  const actualComponent = shouldDive(node) ? node.dive() : node
-  // Find the first node with a className prop
-  const components = actualComponent.findWhere(isTagWithClassName)
-  const classes = components.length && components.first().prop('className')
-
+  // We need to get the css prop if we have selected a styled child from a shallow render
+  const isShallow = shouldDive(node)
+  let classes
+  if (isShallow) {
+    // Find the first node with a className prop
```
Andarist: i would extract both "finding" algorithms to separate functions and use them here, then you could rewrite it to smth like this:
```js
function getClassNamesFromEnzyme(selectors, node) {
  const isShallow = shouldDive(node)
  const componentWithClassName = isShallow
    ? findComponentWithClassNameShallow(node)
    : findComponentWithClassName(node)

  if (!componentWithClassName) {
      return /* not sure right not if correct is null, empty string, or empty array, need to check */
  }

  return getClassNames(selectors, componentWithClassName.prop('className'))
}
```

packages/jest-emotion/src/index.js
```diff
@@ -56,43 +56,76 @@ type Options = {
...
+    let elements = getStyleElements()
+    let keys = getKeys(elements)
     if (isEmotionCssPropElementType(val)) {
       return printer({
         ...val,
         props: filterEmotionProps(val.props),
         type: val.props.__EMOTION_TYPE_PLEASE_DO_NOT_USE__
       })
     }
+    if (isEmotionCssPropEnzymeElement(val)) {
```
Andarist: could you explain why things weren't working before and how they got fixed now?

packages/jest-emotion/src/index.js
ajs139: This was really an effort to keep existing snapshots the same, but omit "unknown styles" in the shallow rendering. On reflection, it's probably best to simply remove this condition.

packages/jest-emotion/src/index.js
ajs139: When the emotion element is shallow rendered, the actual element that would have the className prop is not rendered. The idea here is to swap the emotion element for an element of the underlying type (from __EMOTION_TYPE_PLEASE_DO_NOT_USE__) with the className prop added.

packages/jest-emotion/src/index.js
ajs139: thanks!

packages/jest-emotion/test/react-enzyme.test.js
ajs139: Looks similar to [`test.each`](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout). I've refactored to use `jest-in-case`, let me know what you think.

packages/jest-emotion/src/index.js
```diff
@@ -69,14 +70,47 @@ function filterEmotionProps(props = {}) {
   return rest
 }
 
+function isShallowEnzymeElement(element, classNames) {
+  let renderedClassNames = (element.children || [])
+    .map(({ props = {} }) => props.className)
+    .filter(Boolean)
+    .join(' ')
+  return !classNames.some(className =>
+    new RegExp(className).test(renderedClassNames)
```
Andarist: `classNames` are just strings, so I think it would be better to just use a simple:
```js
renderedClassNames.includes(className)
```

Or create a regexp out of all class names with smth like `new RegExp(`(${classNames.join('|')})`)` and run this on `renderedClassNames`

Or even create an `areIntersecting` helper that could be used as follows:
```js
return !areIntersecting(classNames, renderedClassNames)
```

packages/jest-emotion/test/matchers.test.js
Andarist: could we maybe split that into a separate test then? Seems like pre condition and post condition assertions are testing slightly different things here

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,60 +1,88 @@
...
-}
+import toJson from 'enzyme-to-json'
 
 expect.addSnapshotSerializer(createSerializer())
 
-expect.extend({
-  toMatchShallowSnapshot: createEnzymeSnapshotMatcher(),
-  toMatchDeepSnapshot: createEnzymeSnapshotMatcher({ mode: 'deep' })
-})
+const testWithMethods = (description, testCase) =>
```
Andarist: would prefer gathering all those test cases in a single object or something to call `jest-in-case` only once (not that it's heavy or this here is bad, I'd only want to have this written in a similar fashion as other, existing, tests are written already). Look for an example here - https://github.com/emotion-js/emotion/blob/8767a69251c8de86e8350ac8788b4129847e15bb/packages/core/__tests__/keyframes.js#L11

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,60 +1,88 @@
...
+  cases(description, testCase, [
+    { name: 'render component with shallow', method: 'shallow' },
+    { name: 'render component with mount', method: 'mount' }
+  ])
 
-test('enzyme mount test', () => {
+const expectToMatchSnapshot = (component, method, mode) => {
```
Andarist: I would move handling `method` into `testWithMethods`, and keep that helper focused on calling appropriate `mode`, that way the last test which is using this helper directly could just use method directly (on its own) and only delegate to this helper for handling mode - that alone test looks a little bit weird when it calls this method parametrized with both

Andarist:
> The snapshot support is less than ideal, but it is much improved over the existing behavior.

Could you describe the limitations of this approach?

> For example, it ensures that the map props aren't included in the snapshot.

Don't quite see if this tested somewhere, could you point me to a test for this?

ajs139:
@Andarist - thanks for the review! Let me know if there's anything else. In terms of the two questions:
> Could you describe the limitations of this approach?
The main problem is that if you shallow render a styled component itself, say:
```
enzmye.shallow(<p css={{ backgroundColor: 'blue' }}>Hello</p>)
```
then we get this:
```
<ContextConsumer>
   <Component />
</ContextConsumer>
```
This unfortunately appears to be a result of way shallow works.

> Don't quite see if this tested somewhere, could you point me to a test for this?
I was referring to what happens when not using the plugin, you get something like this:

```
<EmotionCssPropInternal
       __EMOTION_LABEL_PLEASE_DO_NOT_USE__="FragmentComponent"
       __EMOTION_TYPE_PLEASE_DO_NOT_USE__="div"
       css={
         Object {
           "map": "/*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzI ...
```

Andarist:
Couldn't you recognize that you have rendered a consumer and if yes then grab its first children and dive into it? I'm not too familiar with enzyme, so I'm not sure if this is possible.

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,130 +1,165 @@
...
+)
 
-  const tree = enzyme.mount(
-    <div>
-      <FragmentComponent />
-    </div>
-  )
-  expect(tree).toMatchDeepSnapshot()
-})
+test('', () => {})
```
Andarist: seems like a leftover 😉

ajs139:
As far as I can tell, we're not serializing an enzyme element at this point, so dive won't work. Won't be able to investigate further for a few days.

ajs139:
The serializer as it stands works downstream of the enzyme toJson method, meaning we can’t dive at that point. The only real option to solve this would be to have some sort of “prepare” method to call before passing the component to the expect method.

Andarist:
Thank you! I've just released it and going to also work on pulling this into `next` branch.

chyzwar:
This should be an breaking change for emotion-jest.

Andarist:
@chyzwar how comes? I was expecting this to be what shallow users have expected. Was easier for me to postpone this until upcoming v11 hits stable but thought this is an expected improvement

chyzwar:
Because all of snapshot test are not broken. emotion-jest is now producing different snapshots with the same components.

pleunv:
I have to admit I kind of feel the same way. Releasing things like this and breaking type changes as patches tends to be pretty frustrating. Is there any reason why no minor versions are being released?

We end up with different patch versions for all emotion packages that are all stuck because of varying issues introduced by later ones and it gets hard to keep track.

ajs139:
Seems like this does indeed change existing snapshots, had thought this was going to leave the existing snapshots intact. @Andarist - should I roll back the snapshot changes and/or put them behind a flag? Let me know how I can help.

pleunv:
I also noticed that some of the updated snapshots go back to using the normal emotion classes that contain hashes instead of the ones with numbers that we typically get from the jest serializer, i.e. `emotion-ab12345` instead of `emotion-0`, `emotion-1`, etc.

Andarist:
> Because all of snapshot test are now broken. emotion-jest is now producing different snapshots with the same components.

This is a development-only package. I fail to see how this could be considered a breaking change as it:
- doesn't affect runtime code
- is an improvement which was requested for a long time
- you should be using a lock file, so updating your jest-emotion should be a manual process, accompanied with snapshot update

Semver is hard and sometimes we just need to use common sense to decide what type of change a particular thing is, especially given that literally **every** change can be considered by somebody a breaking change. There is a saying that "your bug fix is my breaking change".

>  Is there any reason why no minor versions are being released?

I'm actually not sure - was following a pattern used up to this date here when deciding about this. I think releasing a minor would be indeed more appropriate here, but I also don't see how this would improve a frustration about this? What type of semver ranges are you using in your package.json? Are you using lock files?

> We end up with emotion packages that are all locked at different patch versions because of varying issues introduced by later ones and it gets hard to keep track. I'm also starting to think that some of the issues are possibly introduced by other packages being locked at lower versions and... I'm starting to avoid updating at all.

We really try to keep backward compatibility seriously - we never introduce runtime breaking changes intentionally. I would encourage you to skip pinning our dependencies (just use lock file) and report any issues encountered - I will gladly take a look at fixing them.

> I also noticed that some of the updated snapshots go back to using the normal emotion classes that contain hashes instead of the ones with numbers that we typically get from the jest serializer, i.e. emotion-ab12345 instead of emotion-0, emotion-1, etc.

This doesn't sound right, could you prepare a repro case at which I could take a look?

chyzwar:
> This is a development-only package. I fail to see how this could be considered a breaking change as it:

This is not how semver work. Any change to public API is breaking change. If this package does not follow semver you should explicitly warn users. Additionally, if you plan to release breaking change on path release you should at least head up people and version this like 0.10.25 where version before 1.0.0 allows for breaking changes. As someone suggested bumping minor versions would also send a stronger signals

It is not the first time emotion broke our builds in path releases.

Andarist:
This change was an unforeseen breaking change and it just got reverted - sorry for the inconvenience.

> It is not the first time emotion broke our builds in path releases.

Sorry to hear that - it's never our intention and as mentioned before we care about semver. I wasn't expecting this to be a breaking change.

That being said I believe it's impractical in general to treat such a change as a breaking one. If this is breaking then basically it would mean that NO change can ever be introduced to jest-emotion (which would change the snapshot output). I just don't see the snapshot output as a public API, it's just a string representation of a particular thing. By using snapshots for testing you are putting yourself at such risk (not from emotion, but in general) - this only IMHO showcase that a snapshot test is brittle and often not a good test. A good test would be a one not failing upon an implementation detail change - after all for your users nothing has changed, so why would your tests suddenly fail without a change in your public API (your application)? I'm not saying that snapshots test should be avoided at all costs - they are easy to write, maintain and have their place in the world BUT you should be ready to just press that `"u"` on your keyboard when necessary.

Also as mentioned before - there is no such thing as "how semver work". It's a guideline, at best, and it's execution is put purely on humans, so by definition, it's error-prone. If we'd like to have ideal semver then EVERY change would bump a major version:
> absolutely correct semver - making semver versioning decisions to ensure nothing less than major is capable of breaking a consumer's code. Because literally any change is technically capable of breaking a user's code, absolutely correct semver requires that all changes are major changes.

but that just wouldn't work well in the real-life, so we need to fallback to:
> pragmatically correct semver - Making semver decisions that you believe to be correct, but may be in error. A pragmatic assessment is likely to change with the number of users of a project, and the API surface area of the project. 

Given the nature of this change, I still believe it was not a breaking change deserving a major bump, however indeed it should have been a minor bump and not a patch.

PR #1650, opened 2019-11-24
Accept custom context in ThemeProvider

Andarist:
It's a followup to https://github.com/emotion-js/emotion/pull/1577 - it probably indeed didn't make much sense, but I believe there is some benefit in accepting a custom context prop in `ThemeProvider`

If we are going to decide to merge this in then I'd like to wait for https://github.com/emotion-js/emotion/pull/1609 to land first, so I can add final TS types in here - based on the work from that other PR.

### DIFF
--- a/packages/core/__tests__/__snapshots__/theme-provider.js.snap
+++ b/packages/core/__tests__/__snapshots__/theme-provider.js.snap
@@ -4,11 +4,23 @@ exports[`ThemeProvider throws the correct errors array 1`] = `"[ThemeProvider] P
 
 exports[`ThemeProvider throws the correct errors boolean 1`] = `"[ThemeProvider] Please make your theme prop a plain object"`;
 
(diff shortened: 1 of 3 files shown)


Andarist:
@mitchellhamilton could u take a look at this one?

mitchellhamilton:
I'm not really big on this given how easy it is to implement yourself and it makes explaining the API harder(and I imagine it would be harder to type correctly)

Andarist:
My main motivation for this is to make it easier to use custom themes, the existing merging behavior of nested providers is good & quite intuitive - I won't insist at getting this merged right now as it can always be added later but is something that I thought was worth presenting as a PR. While this is easy to implement if one wants to have the same logic as builtin ThemeProvider for a custom theme they need to replicate it closely, won't even be able to copy-paste the builtin implementation as it requires subtle tweaks to support a custom theme.

Explaining this would be a separate section in the docs so I don't feel like this would make it particularly harder to understand and as to the types - I think it's possible to support this with TS conditional types.

Feel free to close this PR or comment back to get this a green light, both work for me.

Andarist:
@mitchellhamilton should I close this?

PR #1651, opened 2019-11-24
Allow `labelFormat` option to be a function

### LINKED ISSUE #1559: More options/values for labelFormat
**The problem**

For the moment I use `"labelFormat": "[dirname]-[filename]--[local]",` but that's not enough for me.

**Proposed solution**

I would like that `dirname` takes all the folders.

**Context**

I'm here:  `/src/pages/Home/Main/Content.js` with: 

```javascript
const Heading = styled.div``

class HomePageMainContent extends React.Component {}
```

I would like that the className generated for `Heading` is:

- either `xxx-HomePageMainContent--Heading` (`random-NameOfTheReactClass--NameOfTheLocal`)
- or `xxx-pages-Home-Main-Content--Heading` (`random-The-Whole-Path--NameOfTheLocal`) 

Thank you :)

Andarist:
closes #1559

### DIFF
--- a/packages/babel-plugin-emotion/src/utils/label.js
+++ b/packages/babel-plugin-emotion/src/utils/label.js
@@ -1,6 +1,11 @@
 // @flow
 import nodePath from 'path'
 
+type LabelFormatOptions = {
+  name: string,
+  filename: string
+}
+
 const invalidClassNameCharacters = /[!"#$%&'()*+,./:;<=>?@[\]^`|}~{]/g
 
 const sanitizeLabelPart = (labelPart: string) =>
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-emotion/src/utils/label.js
```diff
@@ -9,11 +14,23 @@ const sanitizeLabelPart = (labelPart: string) =>
...
+  const sanitizedName = sanitizeLabelPart(identifierName)
+
+  if (!labelFormat) {
+    return sanitizedName
+  }
+
+  if (typeof labelFormat === 'function') {
+    return labelFormat({
+      name: sanitizedName,
+      filename
```
mitchellhamilton: Could we call this `path` to distinguish it from the filename used in the string form of labelFormat?

Andarist: Sure, changed.

packages/babel-plugin-emotion/src/utils/label.js
```diff
@@ -1,6 +1,11 @@
 // @flow
 import nodePath from 'path'
 
+type LabelFormatOptions = {
+  name: string,
+  filename: string
```
mitchellhamilton:
```suggestion
  path: string
```

Andarist: 👍  just fixed it

PR #1656, opened 2019-11-27
Optimize Babel output a little bit

Andarist:


packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
```diff
@@ -86,7 +79,7 @@ export let transformExpressionWithStyles = ({
       path.node.arguments.length === 1 &&
       t.isStringLiteral(path.node.arguments[0])
     ) {
-      let cssString = path.node.arguments[0].value
+      let cssString = path.node.arguments[0].value.replace(/;$/, '')
```
Andarist: trailing semi is not needed as we put a semi after `opaqueCss.styles` at runtime anyway:
https://github.com/emotion-js/emotion/blob/0b817dade11dee342a820d6c59e86f8a7ef23493/packages/serialize/src/index.js#L171

packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
```diff
@@ -53,13 +53,6 @@ export let transformExpressionWithStyles = ({
       arg => arg.type !== 'SpreadElement'
     )
 
-    if (canAppendStrings && shouldLabel) {
```
Andarist: Moving this further down allows us to remove labels from production styles when we are able to compute "opaqueCss" object with Babel. A followup PR can also wrap appending label with NODE_ENV check to avoid the same thing for the runtime serialization case.

packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js
```diff
@@ -123,6 +124,14 @@ export let transformExpressionWithStyles = ({
 
       return { node, isPure: true }
     }
+
+    if (canAppendStrings && shouldLabel) {
+      const label = getLabelFromPath(path, state, t)
+      if (label) {
+        appendStringToArguments(path, `;label:${label};`, t)
```
Andarist: Wasn't 100% sure yet if avoiding trailing semi in here would also be OK. I plan this to be dev-only though so it doesn't quite matter.

Andarist:
> I'd like to keep labels even in production because it's useful for some people to see them in production and if people really want to disable them in production, they can do autoLabel: false

The same thing can be said about the reverse - if you want to see them in production, you can do `autoLabel: true` (not sure actually if it's true right now, but we could make it so).

My main motivation here is to optimize the default output - I always thought emotion is a huge proponent of such an approach with various hacks in the codebase just to make the output smaller.

I could live with this not being the default (as long as I can configure things how I like it - which is no labels in production). A problem in this particular case though is that we output opaque css objects with env check and what I put in this PR is my desired output. You assume that the party which transpiles files is the very same one that consumes it - but there are other use cases as well, such as building a library in which I'd like this dual output with labels being visible only during development.

So what I'm saying - maybe we need 3 values for autoLabel? true (always), false (never), 'auto'/undefined (with env checks). An additional question is - which one should be the default?

Andarist:
@mitchellhamilton any more thoughts on this one? How I should proceed with this?

mitchellhamilton:
always, never and auto sounds good to me

Andarist:
So`autoLabel` would become such enum rather than a simple boolean, right? What should be the default? I would vote on `"auto"`

mitchellhamilton:
Agreed

Andarist:
@mitchellhamilton could you review the code & existing tests? Once approved I'm going to adjust docs, add few simple explicit tests for possible `autoLabel` values and add a changeset

packages/babel-plugin-emotion/__tests__/__snapshots__/css-requires-options.js.snap
```diff
@@ -176,11 +176,11 @@ import { css } from 'emotion';
...
 /*#__PURE__*/
 css(process.env.NODE_ENV === \\"production\\" ? {
-  name: \\"ryhvpw-my-css-cls\\",
-  styles: \\"color:hotpink;;label:my-css-cls;\\"
+  name: \\"3sn2xs\\",
+  styles: \\"color:hotpink\\"
 } : {
-  name: \\"ryhvpw-my-css-cls\\",
-  styles: \\"color:hotpink;;label:my-css-cls;\\",
+  name: \\"1pmemk4\\",
```
mitchellhamilton: Why was the label removed here?

Andarist: Good catch, fixed 👍

packages/babel-plugin-emotion/src/utils/arguments.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import { isTaggedTemplateExpressionTranspiledByTypeScript } from './checks'
+
+export const appendExpressionToArguments = (t: *, path: *, expression: *) => {
```
mitchellhamilton: I think this replaces `appendStringToArguments` from `src/utils/strings.js` so we could we delete `appendStringToArguments`?

packages/babel-plugin-emotion/src/utils/arguments.js
```diff
@@ -0,0 +1,25 @@
...
+  let lastIndex = path.node.arguments.length - 1
+  let last = path.node.arguments[lastIndex]
+  if (t.isStringLiteral(last)) {
+    path.node.arguments[lastIndex] = t.binaryExpression('+', last, expression)
```
mitchellhamilton: Is this function meant to accept any expression or only string literals(or I think any expression that we know for sure evaluates to a string)? If only things that evaluate to strings, we might want to rename this function

Andarist:
> Had an idea: Rename auto to dev because auto could really mean anything

Maybe `"dev-only"`? It would be a little bit more descriptive

packages/babel-plugin-emotion/src/utils/arguments.js
Andarist: `appendStringToArguments` is still used here:
https://github.com/emotion-js/emotion/blob/3459ad700b719b2c39c69e315a6fce1808443931/packages/babel-plugin-emotion/src/utils/transform-expression-with-styles.js#L149

both are very similar though so probably this could be unified somehow

packages/babel-plugin-emotion/src/utils/arguments.js
Andarist: > or I think any expression that we know for sure evaluates to a string

This. Any ideas for a better name? `appendStringReturningExpressionToArguments`? 😉

mitchellhamilton:
> Maybe "dev-only"? It would be a little bit more descriptive

SGTM

packages/babel-plugin-emotion/src/utils/arguments.js
mitchellhamilton: `appendStringReturningExpressionToArguments` SGTM

packages/babel-plugin-emotion/src/utils/arguments.js
mitchellhamilton: Unifying them would probably be good. The main reason I brought this up though is that I saw that appendStringToArguments wasn't being run in the tests(https://codecov.io/gh/emotion-js/emotion/pull/1656/src/packages/babel-plugin-emotion/src/utils/strings.js) which probably isn't good

Andarist: I've unified them, please take a look.

Andarist:
I believe this is now ready to be merged, there are new tests, changesets, and updated docs.

PR #1659, opened 2019-11-28
Add handling of the css prop in syntax-preference rule

### LINKED ISSUE #1253: [eslint-plugin-emotion] the syntax-preference rule does not check css props
- `emotion` version: 10.0.7
- `react` version: 16.6.0

**Relevant code:**

```javascript
// eslintrc.js
rules: {
  ...
  'emotion/syntax-preference': ['error', 'string'],
  ...
}

// MyComponent.js
// Despite using the object syntax, the following line doesn't raise an ESLint error
const MyComponent = () => <div css={{background: 'red'}} />;

```

**Problem description:**

The [syntax-preference](https://github.com/emotion-js/emotion/blob/master/packages/eslint-plugin-emotion/src/rules/syntax-preference.js) ESLint rule currently works for `styled` but not with the CSS prop.

**Suggested solution:**

It would be nice if this rule could also enforce a syntax choice for projects using CSS props.

Andarist:
closes #1253 

Our ESLint could certainly get some more love, but for now, this PR focuses on a single thing - validating which kind of syntax gets used in combination with css prop.


Andarist:
@mitchellhamilton friendly 🏓

PR #1664, opened 2019-12-02
Created additional overload to prevent AdditionalProps getting inferr…

JakeGinnivan:
**What**:

Previously this issue was prevented with a mapped type on the interpolation union, instead of restoring the old fix I have added an additional overload to the styled interpolation functions which use used when the user doesn't specify a generic type of AdditionalProps. 

**Why**:

The composed component type gets used for the AdditionalProps generic argument when it's inferred rather than {}.

**How**:

**Checklist**:
- [] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/lucky-swans-kneel.md
+++ b/.changeset/lucky-swans-kneel.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fixed issue when composing components using interpolations the wrong type would be inferred
--- a/.changeset/lucky-swans-kneel.md
+++ b/.changeset/lucky-swans-kneel.md
@@ -2,4 +2,4 @@
 '@emotion/styled': patch
 ---
 
-Fixed issue when composing components using interpolations the wrong type would be inferred
+Fixed issue when using "component as selector" in styled interpolations which caused the wrong type to be inferred.
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -77,10 +77,18 @@ export interface CreateStyledComponent<
       >
     >
   ): StyledComponent<ComponentProps & AdditionalProps, SpecificComponentProps>
+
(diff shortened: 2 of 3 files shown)


packages/styled/types/base.d.ts
```diff
@@ -77,10 +77,18 @@ export interface CreateStyledComponent<
       >
     >
   ): StyledComponent<ComponentProps & AdditionalProps, SpecificComponentProps>
+
+  (
```
Andarist: Could you explain how this has fixed the issue? I don't quite follow this. It's also an overload "in the middle" of listed overloads - is it intentionally at this particular position?

JakeGinnivan: Yeah sure

Overloads are selected from top to bottom. The issue was that the props of the second interpolated component was being inferred as AdditionalProps.

By introducing an overload without AdditionalProps it will be selected if it matches, which happens when it's not specified. Basically the overload removes the ability for AdditionalProps to be inferred.

So for example, this will work:

```
const Input4 = styled.input<AdditionalTest>`
  & + ${Label}: ${props => ({
    marginLeft: props.left
  })}
`
```

but this will not:

```
interface AdditionalTest {
  left: string
}
const Input4 = styled.input`
  & + ${Label}: ${(props: AdditionalTest) => ({
    marginLeft: props.left
  })}
`
```

.changeset/lucky-swans-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fixed issue when composing components using interpolations the wrong type would be inferred
```
Andarist:
```suggestion
Fixed issue when using "component as selector" in styled interpolations which caused the wrong type to be inferred.
```

PR #1665, opened 2019-12-02
Update enzyme snapshots

ajs139:
**What**:
Support for shallow rendering in #1648 introduced an issue where the name of the component could be set to the actual code for that component. This resolves that issue.

**Why**:
The snapshot becomes too large to be useful, as well as brittle.

**How**:
We use the component's `name` property as its name.

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/great-bears-sing.md
+++ b/.changeset/great-bears-sing.md
@@ -2,4 +2,4 @@
 'jest-emotion': patch
 ---
 
-Fix naming of shallow rendered components
+Fix printing names of nested shallow-rendered components.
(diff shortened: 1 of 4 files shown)


.changeset/great-bears-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+'jest-emotion': patch
+---
+
+Fix naming of shallow rendered components
```
Andarist:
```suggestion
Fix printing names of nested shallow-rendered components.
```

PR #1668, opened 2019-12-03
Forward shouldForwardProp through withComponent

### LINKED ISSUE #1394: Ability to set default shouldForwardProp or expose to extended components
_**UPDATED TO REFLECT MORE ACCURATE ISSUE**_

**The problem**

I would love the ability to set a default shouldForwardProp value, project wide by either a simplistic factory extension or via a theme. Currently when I set the shouldForwardProp value on a styled div it works great. However if I extend that styled component to form another one leveraging the `withComponent` helper, that shouldForwardProp value is not exposed to that new styled component and I need to set it again.

ie:
```
import { css } from '@emotion/core';
import * as styledSystem from 'styled-system';
import { COLORS, fontAR, shouldForwardProp } from '../../constants/styles';

const baseStyle = css`
  ${fontAR};
  font-size: 14px;
  line-height: 1.43;
`;

const smallStyle = css`
  font-size: 12px;
  line-height: 1.67;
`;

const largeStyle = css`
  font-size: 16px;
  line-height: 1.91;
`;

const boldStyle = css`
  font-family: Avenir;
  font-weight: 900;
`;

const styleMap = {
  small: smallStyle,
  large: largeStyle,
  bold: boldStyle,
};

interface ParagraphProps
  extends styledSystem.ColorProps,
    styledSystem.FontFamilyProps,
    styledSystem.FontSizeProps,
    styledSystem.FontWeightProps,
    styledSystem.HeightProps,
    styledSystem.LetterSpacingProps,
    styledSystem.LineHeightProps,
    styledSystem.SpaceProps,
    styledSystem.TextAlignProps,
    styledSystem.WidthProps {
  color?: COLORS;
  size?: string;
}

export const Paragraph = styled('p', { shouldForwardProp })<ParagraphProps>`
  ${baseStyle}
  ${styledSystem.color}
  ${styledSystem.fontFamily}
  ${styledSystem.fontSize};
  ${styledSystem.fontWeight}
  ${styledSystem.height}
  ${styledSystem.letterSpacing}
  ${styledSystem.lineHeight}
  ${styledSystem.space}
  ${styledSystem.textAlign}
  ${styledSystem.width}
  ${props => (props.size ? styleMap[props.size] : null)}
`;

const Label = styled(Paragraph)`
  color: ${COLORS.slate};
  display: flex;
  align-items: center;
  margin-bottom: 4px;
`.withComponent('label');
```

Now Label does not know to exclude any props that are a result of the imported function which works as intended in the Paragraph component.

**Proposed solution**

- Add the ability in the core code to consume the underlying shouldForwardProp value if a styled component is extending another styled component combined with `withComponent`.

It works if you use the `as` prop but my intention is to not need to remember to add that additional prop anytime I want to use the new Label component

animecyc:
**What**:
- Fixes a bug that prevented `shouldForwardProp` from forwarding to components created via `withComponent`; Fixes #1394
- Compacts `createStyled`'s options argument

**Why**:
- Ensures similar composition semantics as the `as` prop. This sort of begs the question of whether or not the `withComponent` functionality is needed given we have the `as` prop; Regardless, they should operate the same if the API exists.

**How**:
- Coalesces `shouldForwardProp` together with `options` and `nextOptions`.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [ ] Code complete
- [ ] Changeset


packages/styled-base/src/index.js
```diff
@@ -211,12 +211,11 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
-        nextTag,
-        nextOptions !== undefined
-          ? { ...(options || {}), ...nextOptions }
-          : options
-      )(...styles)
+      return createStyled(nextTag, {
+        shouldForwardProp,
+        ...options,
+        ...nextOptions
+      })(...styles)
```
animecyc: This was done just to simplify the object spreads.

packages/styled/__tests__/styled.js
```diff
@@ -641,6 +641,58 @@ describe('styled', () => {
...
+    })``.withComponent('div', {
+      label: 'cmp-b',
+      shouldForwardProp: prop => prop === 'propB'
+    })
+    const tree = renderer.create(<SomeComponent propA propB />).toJSON()
+    expect(tree).toMatchSnapshot()
+  })
+
+  test('withComponent correctly inherits parent shouldForwardProp through chained withComponent calls', () => {
+    const SomeComponent = styled('div', {
```
Andarist: how is this test different from the previous one? maybe you had in mind a version without `shouldForwardProp` option in the "middle" (in the first `withComponent` call)?

Andarist:
Could you branch your change out from the `next` branch?

packages/styled-base/src/index.js
Andarist: There are 2 potential problems I can see here:
1. `shouldForwardProp` contains composed function, but you can override it in here with the one in `options` (so with the one pre-composition)
2. we allow overriding `shouldForwardProp` completely by spreading `nextOptions` in here - so this isn't composing it

I definitely feel like 1. should get fixed. Still unsure what the expectation for the 2. is though. When using `as` we keep the configured `shouldForwardProp` (if it got configured), but use what would be the default for a new component (the one in `as`) if `shouldForwardProp` was not configured. This API is quirky 🤷‍♂ 

Thoughts?

animecyc:
> Could you branch your change out from the `next` branch?

Sure, reconciling things now.

Andarist:
Not sure exactly why, but the diff includes stuff that it shouldn't. Maybe it would be best to just close this PR and reopen it after branching away from `next` and cherry-picking your single commit?

animecyc:
> Not sure exactly why, but the diff includes stuff that it shouldn't. Maybe it would be best to just close this PR and reopen it after branching away from `next` and cherry-picking your single commit?

I thought I pushed, sorry -- its been reconciled now.

packages/styled-base/src/index.js
animecyc: I am going to mark this comment as resolved as the base branch changed and messed up this audit -- In my recent push i've addressed both points that hopefully satisfy things.

packages/styled/src/base.js
```diff
@@ -196,12 +195,16 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
-      )(...styles)
+      return createStyled(nextTag, {
+        ...options,
+        ...nextOptions,
+        shouldForwardProp:
+          nextOptions && typeof nextOptions.shouldForwardProp === 'function'
+            ? propName =>
+                nextOptions.shouldForwardProp(propName) &&
+                defaultShouldForwardProp(propName)
+            : defaultShouldForwardProp
```
animecyc: My understanding here is that we want to pipe the should forward props like a funnel (C -> B -> A) using either the supplied `shouldForwardProp` falling back to the composed `defaultShouldForwardProp` of the parent.

packages/styled/__tests__/styled.js
```diff
@@ -640,6 +640,55 @@ describe('styled', () => {
...
+    expect(shouldForwardPropB).toHaveBeenCalledTimes(3)
+    expect(shouldForwardPropB).toHaveBeenNthCalledWith(1, 'as')
+    expect(shouldForwardPropB).toHaveBeenNthCalledWith(2, 'disabled')
+    expect(shouldForwardPropB).toHaveBeenNthCalledWith(3, 'readOnly')
+
+    expect(shouldForwardPropA).toHaveBeenCalledTimes(2)
+    expect(shouldForwardPropA).toHaveBeenNthCalledWith(1, 'as')
+    expect(shouldForwardPropA).toHaveBeenNthCalledWith(2, 'disabled')
+
+    expect(tree).toMatchSnapshot()
```
animecyc: I figured it would be good to be a bit more explicit with this test and the calls made to `shouldForwardProp` on each `withComponent`

packages/styled/src/base.js
```diff
@@ -196,12 +197,36 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
+        nextStyles = styles.slice(1)
+
+        if (
+          nextTag.__emotion_real === nextTag &&
+          nextTag.__emotion_identifier !== undefined
+        ) {
+          nextTag = { ...nextTag }
+          nextTag.__emotion_real = nextTag
+          nextTag.__emotion_styles = nextTag.__emotion_styles.slice(1)
+        }
```
animecyc: If we don't remove the prepended label for `nextTag` we end up with duplicate labels.

Andarist:
@mitchellhamilton could you take this over? I'm afraid I won't have time to finish reviewing this right now.

mitchellhamilton:
> @mitchellhamilton could you take this over? I'm afraid I won't have time to finish reviewing this right now.

Absolutely, no problem!

packages/styled/test/prop-filtering.test.js
```diff
@@ -6,17 +6,16 @@ import styled from '@emotion/styled'
 
 test('composes shouldForwardProp on composed styled components', () => {
   const StyledDiv = styled('div', {
-    shouldForwardProp: prop => prop === 'forwardMe'
+    shouldForwardProp: prop => prop !== 'foo'
   })()
 
   const ComposedDiv = styled(StyledDiv, {
-    shouldForwardProp: () => true
+    shouldForwardProp: prop => prop !== 'bar'
```
mitchellhamilton, line 13: Why these changes? Was this test failing with the changes to styled here? If the original test here was failing, I think that's a problem.

Andarist: I don't think the previous version of this test was foo - it was supposed to test `shouldForwardProp` composition but `() => true` doesn't make any difference in regards to the final behavior, because this test would pass with it or without this `shouldForwardProp` as the composition uses `&&` so if we fold this `shouldForwardProp1(prop) && shouldForwardProp2(prop)` to `shouldForwardProp1(prop) && true` then all we really have is `shouldForwardProp1(prop)`

mitchellhamilton (approved):

PR #1672, opened 2019-12-04
Improve support for Enzyme's shallow rendering

### LINKED ISSUE #949: toHaveStyleRule does not work with Emotion 10

ajs139:
**What**:
NOTE: this was originally merged to master, but was backed out as it introduced a breaking change to existing snapshots.

Fix #949, also support better snapshot serialization when using shallow rendering

**Why**:
While using shallow rendering is generally considered unwise, there are projects that wish to convert from CSS to CSS-in-JS and that already use shallow rendering in their tests.

**How**:
The `jest-emotion` package was updated to dive into shallow rendered components to trigger the addition of the styles to the DOM to support `toHaveStyleRule`. Snapshots are supported by adding classNames to the shallow-rendered object by leveraging the `css` prop.

**Checklist**:
- [ ] Documentation (TODO)
- [x] Tests
- [ ] Code complete
- [ ] Changeset N/A


packages/jest-emotion/src/utils.js
```diff
@@ -52,6 +60,7 @@ function getClassNamesFromCheerio(selectors, node) {
 }
 
 function getClassNamesFromDOMElement(selectors, node: any) {
+  console.log(node)
```
Andarist: leftover console.log 😉

packages/jest-emotion/src/utils.js
```diff
@@ -189,7 +198,9 @@ export function getStylesFromClassNames(
 }
 
 export function getStyleElements(): Array<HTMLStyleElement> {
-  let elements = Array.from(document.querySelectorAll('style[data-emotion]'))
+  let elements = global.document
```
Andarist: I think `global` might not be always available - so maybe better to use a simple check of `typeof document !== 'undefined'`

packages/jest-emotion/src/utils.js
```diff
@@ -241,3 +252,11 @@ export function getMediaRules(rules: Array<Object>, media: string): Array<any> {
     })
     .reduce((mediaRules, mediaRule) => mediaRules.concat(mediaRule.rules), [])
 }
+
+export function isPrimitive(test: any) {
```
Andarist: TIL

packages/jest-emotion/src/index.js
```diff
@@ -35,11 +37,50 @@ function getNodes(node, nodes = []) {
   return nodes
 }
 
+function copyProps(src, target) {
```
Andarist: you are defining properties on `src` - which means that those are slightly wrongly named, should be other way around. The first parameter is a target (we copy props onto it) and the second one is a source (we copy from it)

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,130 +1,216 @@
...
+afterEach(() => {
+  document
+    .querySelectorAll('style[data-emotion]')
+    .forEach(element => element.parentNode.removeChild(element))
+})
 
-  return function(val) {
-    return toMatchSnapshot.call(this, serializer.print(val, identityPrinter))
-  }
+const tickle = (wrapper: *) => {
```
Andarist: Is our consumer supposed to do that on their own?

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,130 +1,216 @@
...
-import { toMatchSnapshot } from 'jest-snapshot'
+import { jsx, ThemeProvider } from '@emotion/core'
+import styled from '@emotion/styled'
 import React from 'react'
+import toJson from 'enzyme-to-json'
 
-const createEnzymeSnapshotMatcher = serializerOptions => {
-  const serializer = createEnzymeSerializer(serializerOptions)
-  const identityPrinter = v => v
+afterEach(() => {
```
Andarist: maybe just smth like `document.head.innerHTML = ''`?

packages/jest-emotion/test/react-enzyme.test.js
ajs139: There's two basic options here if we want to publicize exposing the styles in the snapshot for nested components - either we just document this approach, or we expose a helper and document how to use it.

A more ambitious option would be to wrap the underlying `jest-enzyme` implementation. If we take that approach, then users would use `jest-emotion` in place of `jest-enzyme`, instead of using them both together.

packages/jest-emotion/test/react-enzyme.test.js
ajs139: That didn't seem to work, but can take a look after the caching issue is solved. As it stands, clearing the DOM breaks subsequent tests if they use the same styles, since emotion believes the styles have already been added to the DOM as they exist in emotion's cache.

packages/jest-emotion/test/react-enzyme.test.js
ajs139: or maybe a sub-package, like `jest-emotion/enzyme`

Andarist: sub-package sounds good, to create one you can define it like this:
https://github.com/emotion-js/emotion/blob/2fa7a213222fc2bb99ca0a01078148f1a1c6458d/packages/styled/package.json#L55-L58

packages/jest-emotion/test/react-enzyme.test.js
Andarist: you don't clear the cache in this afterEach though, it's only about clearing the emotion-generated style tags, so a simple document.head overwrite should work, right? it's just nitpicking, your version actually covers more cases

Andarist: And not sure if I've commented about this anywhere - but "clearing" the cache could be accomplished by wrapping each test in a separate CacheProvider

docs/testing.mdx
```diff
@@ -18,11 +18,20 @@ npm install --save-dev jest-emotion
 
 Add the snapshot serializer in your [`setupTestFrameworkScriptFile`](http://facebook.github.io/jest/docs/en/configuration.html#setuptestframeworkscriptfile-string) _or_ at the top of your test file.
 
+When using `enzyme`, it's recommended to use the combined enyzme and emotion serializer to support shallow rendering:
```
Andarist:
```suggestion
When using Enzyme, it's recommended to use the combined enzyme and emotion serializer to support shallow rendering:
```

packages/jest-emotion/README.md
```diff
@@ -16,9 +16,18 @@ The easiest way to test React components with emotion is with the snapshot seria
...
   // ... other config
-  snapshotSerializers: [
-    'jest-emotion' /* if needed other snapshotSerializers should go here */
-  ]
+  snapshotSerializers: ['jest-emotion']
+}
+```
+
+To assist with shallow rendering, there's a custon enzyme snapshot serializer, that includes the `enzyme-to-json`
+serializer, which is available by inporting `jest-emotion/enzyme`:
```
Andarist:
```suggestion
serializer, which is available by importing `jest-emotion/enzyme`:
```

packages/jest-emotion/package.json
```diff
@@ -20,6 +22,9 @@
     "css": "^2.2.1",
     "specificity": "^0.4.1"
   },
+  "peerDependencies": {
```
Andarist: This should be an optional peer dependency, to mark is as such you need to add `peerDependenciesMeta` similar to this one - https://github.com/emotion-js/emotion/blob/456be9a602d7d0bac291617f69f59f5ed30d1b84/packages/core/package.json#L38-L42

packages/jest-emotion/src/utils.js
Andarist: also - when this check matters? for the enzyme's case when tests can be run without jsdom available globally?

packages/jest-emotion/types/enzyme.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="jest" />
```
Andarist: If you need help with adding types for this - let me know.

scripts/test-utils/next-env.js
```diff
@@ -1,5 +1,5 @@
 // @flow
 /* eslint-env jest */
-import serializer from 'jest-emotion'
+import serializer from 'jest-emotion/enzyme'
```
Andarist: should this include enzyme configuration? `next-env` seems like something that shouldn't

packages/jest-emotion/types/enzyme.d.ts
ajs139: Let me take a stab, and you can review.

Andarist:
@mitchellhamilton would you like to also take a look at this?

packages/jest-emotion/src/utils.js
ajs139: that's right, this code should always be run in node, but the code shouldn't simply fail if JSDOM is not installed.

packages/jest-emotion/src/enzyme.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+import type { Options } from './serializer'
+import { createSerializer as createEmotionSerializer } from './serializer'
+import { createSerializer as createEnzymeSerializer } from 'enzyme-to-json'
+
+const enzymeSerializer = createEnzymeSerializer({})
+
+const tickle = (wrapper: *) => {
+  if (typeof wrapper.dive === 'function') {
+    wrapper.find('EmotionCssPropInternal').forEach(el => el.dive())
```
mitchellhamilton: Couldn't we have a serializer that calls this and then calls out to the printer so we don't have to call enzyme-to-json ourselves?(We'd also want a WeakSet to store which wrappers we've done the printing on so we don't do it infinitely)

ajs139: The issue is order, we want to execute this on the enzyme objects, but our serializer works on the JSON objects. The other options would either to be to have two serializers, such as an `init-emotion-enzyme` serializer as well as the standard `jest-emotion` one, so the user could setup the correct order of `['init-emotion-enzyme', 'enzyme-to-json', 'jest-emotion']`, or two export an initializer so that the get snapshot call would look something like `expect(initializer(component)).toMatchSnapshot()`. Both of these seemed awkward DX, this alongside the fact that we really do rely on `enzyme-to-json` led to the idea of simply tying them together here. It turns out that this makes it easier to setup, too, as neglecting to include both serializers would cause an error the way it was setup before. Open to other ideas, but this seemed to offer the right balance of DX and functionality.

mitchellhamilton: Okay, SGTM

packages/jest-emotion/README.md
```diff
@@ -16,9 +16,18 @@ The easiest way to test React components with emotion is with the snapshot seria
...
   // ... other config
-  snapshotSerializers: [
-    'jest-emotion' /* if needed other snapshotSerializers should go here */
-  ]
+  snapshotSerializers: ['jest-emotion']
+}
+```
+
+To assist with shallow rendering, there's a custon enzyme snapshot serializer, that includes the `enzyme-to-json`
+serializer, which is available by importing `jest-emotion/enzyme`:
```
mitchellhamilton: Could you add a note mentioning that people should remove the enzyme-to-json if they have it?

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,130 +1,201 @@
...
+      </React.Fragment>
+    )
+
+    const wrapper = enzyme.mount(
+      <div>
+        <FragmentComponent />
+      </div>
+    )
+
+    expect(toJson(wrapper, { mode: 'deep' })).toMatchSnapshot()
```
mitchellhamilton: Is this toJson call meant to be here, I thought that's happening at the snapshot serializer level?

packages/jest-emotion/test/react-enzyme.test.js
```diff
@@ -1,130 +1,201 @@
...
 
-const createEnzymeSnapshotMatcher = serializerOptions => {
-  const serializer = createEnzymeSerializer(serializerOptions)
-  const identityPrinter = v => v
-
-  return function(val) {
-    return toMatchSnapshot.call(this, serializer.print(val, identityPrinter))
-  }
-}
+import serializer from '../src/enzyme'
```
mitchellhamilton: We want to have imports like this so we can run tests against the built bundles
```suggestion
import serializer from 'jest-emotion/enzyme'
```

packages/jest-emotion/src/utils.js
```diff
@@ -189,15 +203,17 @@ export function getStylesFromClassNames(
 }
 
 export function getStyleElements(): Array<HTMLStyleElement> {
-  let elements = Array.from(document.querySelectorAll('style[data-emotion]'))
+  const elements = isBrowser
```
mitchellhamilton: I'm not sure about this, I feel like if we're not in a browser environment when serializing an element we should throw.

packages/jest-emotion/test/react-enzyme.test.js
ajs139: this particular test requires the `deep` option, which enforces all the nodes to have their render methods called.

packages/jest-emotion/README.md
ajs139: sounds good

packages/jest-emotion/test/react-enzyme.test.js
ajs139: thought i was having problems when it was that way, but this works, thanks for the fix

packages/jest-emotion/src/utils.js
ajs139: while i agree with you, i feel we should do it at a higher level if we are to do so, with an informative error message. should i add such to this PR?

mitchellhamilton: If by higher level you mean check in the module scope then I don't think that would be ideal because the module that the serializer is in should be runnable outside of a browser environment so that a user can add the serializer to the `snapshotSerializers` option but still have some tests that use the node jest environment. An informative message would be great though.

ajs139: OK, added a warning.

Issue #1470, opened 2019-08-13
Warn when multiple instances of Emotion are used


FezVrasta:
**The problem**

When multiple instances of Emotion are used (specifically, `@emotion/core`), theming and other parts of the library stop behaving correctly.

**Proposed solution**

Following the React way to solve this problem, we should put a check in Emotion to warn the user if multiple versions are being used together.

**Alternative solutions**

Alternatively we could store the whole Emotion context into a global variable and use it instead of defining a new one for each instance, but I'm not sure about possible downsides of this approach.

**Additional context**

For reference, here's the React PR where the check was first introduced: https://github.com/facebook/react/pull/3579

willgriffiths:
Hi @FezVrasta, did you manage find a workaround? We're hitting what seems like a similar issue.

Andarist:
@willgriffiths what's your exact issue? what kind of a problem are you facing?

willgriffiths:
Thanks for your reply @Andarist We're moving from emotion 9 to emotion 10. And theming isn't working like it used to.

My colleague was explaining the issue to me: 
> There seems to be some kind of problem with emotion-theming and its context.
> If we use ThemeProvider and the useTheme hook from emotion-theming everything works
> ThemeProvider however doesn't interacting properly with the css prop from @emotion/core or styled from @emotion/styled
> The themes in them are coming through as an empty object, which seems to be the default
> I think this is due to emotion 10 switching to the new way of using context, where instead of tying it to an id it ties it to the Context object, and I guess @emotion/styled and emotion-theming aren't using the same Context?
>As far as i can tell the context is made in @emotion/core. So maybe theres multiple copies of it running at once? Im unsure how to find that out

```
@emotion/core: 10.0.17
@emotion/styled: 10.0.17
emotion-theming: 10.0.19
```

We can setup a repro. And also also make a seperate issue once we have more info. :)

FezVrasta:
In any case, this issue aims to provide a way to easily debug the issue, not to provide a solution to the multiple theme contexts. I'm saying this just in case you guys find an issue for this specific problem and may decide this issue can be closed.

Andarist:
> We can setup a repro. And also also make a seperate issue once we have more info. :)

That would be really helpful.

@FezVrasta sure - a warning seems reasonable, shouldn't be a hard error though. Would you like to work on this or should I get to it sometime later?

FezVrasta:
I can try to put a PR together 👍

[...]

PR #1677, opened 2019-12-06
Add warning if multiple versions are running (addresses #1470)

ajs139:
**What**:
Add a register of all the `@emotion` packages, so that a warning can be displayed if the same package is registered with multiple versions.

**Why**:
Running multiple versions of the packages can cause hard to trace issues, see #1470.

**How**:
Each module "registers" itself providing its name and version into a global object. If that module name is already registered for a different version, a warning is displayed.

This is just a draft, for discussion on the approach.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/utils/src/index.js
+++ b/packages/utils/src/index.js
@@ -1,7 +1,7 @@
 // @flow
 import type { RegisteredCache, EmotionCache, SerializedStyles } from './types'
 
-let isBrowser = typeof document !== 'undefined'
+const isBrowser = typeof document !== 'undefined'
 
 export function getRegisteredStyles(
   registered: RegisteredCache,
--- a/.changeset/sour-poets-move.md
+++ b/.changeset/sour-poets-move.md
@@ -1,0 +1,5 @@
+---
+'@emotion/core': patch
+---
+
+Warn if @emotion/core is initialized more than once in the same development environment.
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -6,3 +6,18 @@ export { keyframes } from './keyframes'
 export { ClassNames } from './class-names'
 export { ThemeContext, useTheme, ThemeProvider, withTheme } from './theming'
 export { default as css } from './css'
+
+if (process.env.NODE_ENV !== 'production') {
+  const isBrowser = typeof document !== 'undefined'
+  const globalKey = '__EMOTION_CORE__'
+  const globalContext = isBrowser ? window : global
+  if (globalContext[globalKey]) {
+    console.warn(
+      'You are loading @emotion/core when it is already loaded. Running ' +
+        'multiple instances may cause problems. This can happen if multiple ' +
+        'versions are used, or if multiple builds of the same version are ' +
+        'used.'
+    )
+  }
+  globalContext[globalKey] = true
+}


packages/utils/src/index.js
```diff
@@ -1,7 +1,35 @@
...
+      } is already registered. Running ${version} may cause problems.`
+    )
+  }
+  versions[name] = version
+}
+
+registerEmotionModule(
+  require('../package.json').name,
+  require('../package.json').version
+)
```
ajs139: This is more here for usage, it may be decided that it's OK for this module to have multiple concurrent versions.

packages/jest-emotion/src/index.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import { registerEmotionModule } from '@emotion/utils'
```
Andarist: i think this only matters for `@emotion/core` package - don't think there is a much need for this in other packages

packages/utils/src/index.js
```diff
@@ -1,7 +1,35 @@
...
+  return context[globalVersionsKey]
+}
+
+type ModuleVersions = { [key: string]: string }
+
+const versions: ModuleVersions = getGlobalVersions(isBrowser ? window : global)
+
+export function registerEmotionModule(name: string, version: string) {
+  if (versions[name] && versions[name] !== version) {
+    console.warn(
```
Andarist: It's not only about different versions (but that too) - the problem might also be caused by requiring the same version twice (from different locations at disk).

I think react-router prepared this pretty much in a way that I'd like to have this included here, take a look:
https://github.com/ReactTraining/react-router/blob/e81dfa2d01937969ee3f9b1f33c9ddd319f9e091/packages/react-router/modules/index.js#L1

For that to work exactly like in their case we'd have to have `process.env.ROLLUP_MODULE_FORMAT` available - this could potentially be added in our build tool ([preconstruct](https://github.com/preconstruct/preconstruct)). I expect supporting this would only require changes around those lines:
https://github.com/preconstruct/preconstruct/blob/120f9b6d747e985ed917a898f13d7177fd74ef73/packages/cli/src/build/rollup.ts#L164-L168

ajs139: If we want to warn based on module type as well as version, we would want to split this up:
https://github.com/preconstruct/preconstruct/blob/c1cfc0503f73a197487d6cba6ada607fe957f527/packages/cli/src/build/config.ts#L132-L153
If we allowed the messaging to be more vague, then we wouldn't need to do an explicit check of the module type. For example, we could change the code to this:
```
if (__DEV__) {
  if (versions[name]) {
    if (versions[name] === version) {
      console.warn(
        `Package ${name}@${versions[name]} is already registered, so things won't work right. This would only happen if you are loading two different builds of ${name}.`
      );
    } else {
      console.warn(
        `Package ${name}@${versions[name]} is already registered. Running ${version} may cause problems.`
      );
    }
  }
  versions[name] = version;
}
```

Andarist: The goal of this is to warn about second copy of `@emotion/core` being loaded. It doesn't really matter what was the first version being loaded and it doesn't even matter what is the second one - we only need types of both to provide a helpful error message. I would limit this to a build type error - not sure if we need to complicate this with providing also an exact version.

packages/utils/src/index.js
```diff
@@ -1,7 +1,32 @@
...
+}
+
+type ModuleRegister = { [key: string]: boolean }
+
+const globalContext = isBrowser ? window : global
+
+const instances: ModuleRegister = getGlobalInstances(globalContext)
+
+export function registerEmotionModule(name: string) {
+  if (process.env.NODE_ENV !== 'production') {
```
ajs139: tried using [`babel-plugin-dev-expression`](https://github.com/4Catalyzer/babel-plugin-dev-expression), but it was causing failures when running the test suite.

packages/utils/src/index.js
```diff
@@ -1,7 +1,32 @@
...
+  return context[globalVersionsKey]
+}
+
+type ModuleRegister = { [key: string]: boolean }
+
+const globalContext = isBrowser ? window : global
+
+const instances: ModuleRegister = getGlobalInstances(globalContext)
+
+export function registerEmotionModule(name: string) {
```
Andarist: I don't think this abstraction is much needed, we only need this for a single module (`@emotion/core`), so I would prefer avoiding extra abstractions and just inlining this stuff to `@emotion/core`'s index

I also don't think we really need tests for this - it's rather hard to break it and writing a proper test (without testing implementation details) for this thing would be rather brittle and complicated

packages/core/src/utils.js
```diff
@@ -1,2 +1,19 @@
 // @flow
 export let isBrowser = typeof document !== 'undefined'
+
+const globalKey = '__EMOTION_CORE__'
+const globalContext = isBrowser ? window : global
+
+export function checkForMultipleVersions() {
```
Andarist: this would have to be called to actually make it work :p however, calling this as a function could lead to this not being properly removed in production, so I would just move this function's content to `core/src/index.js` and this should be good to go 👍

Andarist:
@ajs139 the only thing missing is an appropriate changeset, could you add it?

PR #1682, opened 2019-12-09
Added some docs

JakeGinnivan:
Just improving some docs around the types to make it easier for users and also others contributing to the types.

### DIFF
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -172,6 +172,8 @@ const App = () => (
 )
 ```
 
+**NOTE** Specify your props *after* the styled function call to allow emotion to correctly infer the types for you. ie `styled(Component)<Additional>` not `styled<Additional>(Component)`. v11 of emotion fixed a number of issues with type inference, if you are having trouble upgrading, remove any manually specified types.
+
 ### Forwarding props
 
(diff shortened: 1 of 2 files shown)


docs/typescript.mdx
```diff
@@ -199,6 +201,34 @@ const StyledOriginal = styled(Original, {
 ;<StyledOriginal prop1="1" prop2={2} />
 ```
 
+#### Filter props helper
+
+You can use this helper in your project if you want a more concise forward prop syntax
+
+``` ts
+/**
+ * Use `createPropFilter` with shouldForwardProp to filter out props. Will automatically filter out `theme`
```
Andarist: I think you have forgotten to actually filter out `theme` in the implementation of this helper

docs/typescript.mdx
```diff
@@ -199,6 +201,34 @@ const StyledOriginal = styled(Original, {
...
+ * since this prop is passed via emotion's context.
+ */
+
+export function createPropFilter<ComponentProps extends {}>() {
```
Andarist: this is sort of a factory-factory now, it doesnt immediately create prop filter, not a big deal but a little bit misleading - especially because of that it would be great to give a more concrete example below it - one involving a styled call

JakeGinnivan:
@Andarist based on the feedback above, closing and going to open another PR which extends emotion itself to make this nicer.

Issue #1422, opened 2019-06-28
styled types do not support object notation


thomasthiebaud:
**Current behavior:**

`styled` types do not support object notation.

According to the documentation (https://emotion.sh/docs/typescript#passing-props), it should be possible to write 

```
interface ImageProps {
  src: string;
  width: number;
};

const Image1 = styled('div')(
  {
    backgroundSize: 'contain',
  },
  (props: ImageProps) => ({
    width: props.width,
    background: `url(${props.src}) center center`,
  })
);
```

but it give me that error locally

```
Argument of type '{ backgroundSize: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
Object literal may only specify known properties, and 'backgroundSize' does not exist in type 'TemplateStringsArray'.ts(2345)
```

Looking at the types test (https://github.com/emotion-js/emotion/blob/master/packages/styled/types/tests.tsx) it only seems that string literals are supported. 

Where types for object notation supported? Is it a bug?

**Environment information:**

```
"@emotion/core": "10.0.14"
"@emotion/is-prop-valid": "0.8.2"
"@emotion/styled": "10.0.14"
"react": "^16.8.6"
```

Andarist:
Can't reproduce here - https://codesandbox.io/s/stupefied-https-uw7bg

Also we have more TS tests here - https://github.com/emotion-js/emotion/blob/master/packages/styled-base/types/tests.tsx

thomasthiebaud:
Which `tsconfig.json` are you using? I can always reproduce it

eriktoyra:
We have noticed a similar issue since updating to `"@emotion/core": "10.0.14"` and `"@emotion/styled": "10.0.14"`.

We are passing around the `theme` props to our styled components. Up until `10.0.14` this has worked fine with components as the one below. But since `10.0.14`, we started to get errors like 

```bash
error TS2741: Property 'theme' is missing in type '{ children: Element[]; }' but required in type '{ theme: typeof import("/Users/[removed-for-privacy-reasons]/src/themes/[removed-for-privacy-reasons]"); }'.

190                 <Range>
                     ~~~~~

  node_modules/@emotion/styled-base/types/index.d.ts:33:11
    33   : P & { theme: T }
                 ~~~~~
    'theme' is declared here.
```

#### Examples

```tsx
// Worked pre 10.0.14, but blows up in 10.0.14
const Range = styled.div(({ theme }) => ({
    display: 'flex',
[...]

PR #1683, opened 2019-12-09
Update typescript.mdx

s-h-a-d-o-w:
Both types for @emotion/styled and TS have improved (and it has been quite a while since TS 2.9) - adjust docs accordingly.
See also: https://github.com/emotion-js/emotion/issues/1422#issuecomment-506982528

**What**: Outdated docs

**Why**: The best workflows are currently listed last. (And include some excessive type specifications.)

**Checklist**:

- [X] Documentation
- [ ] Tests    // doesn't apply
- [ ] Code complete    // doesn't apply
- [ ] Changeset    // doesn't apply

### DIFF
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -103,44 +103,23 @@ type ImageProps = {
   width: number
 }
 
-const Image0 = styled('div')`
-  width: ${(props: ImageProps) => props.width};
-  background: url(${(props: ImageProps) => props.src})
+const Image0 = styled.div<ImageProps>`
+  width: ${(props) => props.width};
+  background: url(${(props) => props.src})
     center center;
   background-size: contain;
 `
 
 // Or with object styles
-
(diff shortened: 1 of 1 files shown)


Andarist:
Thank you for the PR - but I'd like to close it to avoid potential git conflicts. We are currently preparing v11 (most likely will be released this month) and it already contains a complete TS types makeover. We would love your feedback on it though - so if you are feeling like it please test it (it's already released to npm under `next` tag and also some production sites are already using it). The v11 doesn't come with many breaking changes (besides TS types) - it's mostly just a cleanup, but, of course, you should read changelogs first to get a feeling about how it might affect you.

s-h-a-d-o-w:
I see. I obviously didn't look at the docs change in the `next` branch. 🤦‍♂ 
Good to see that you've updated it already and thanks for the quick feedback!

PR #1685, opened 2019-12-10
Remove no longer needed StyleProps type parameter

Andarist:
cc @JakeGinnivan


Andarist:
There were no real tests for this - so I'll assume the change is good and gonna merge this without waiting for a CR approval.

PR #1690, opened 2019-12-13
fix(types): Fix Matchers type for @types/jest^24.0.20

### LINKED ISSUE #1689: Update type definitions to latest @types/jest
**Current behavior:**

If a project uses Jest types at version 24 or greater, this package will conflict the type definition of `Matchers`. The type error is:

```
node_modules/jest-emotion/types/index.d.ts:33:15 - error TS2428: All declarations of 'Matchers' must have identical type parameters.

33     interface Matchers<R> {
```

The change to the `Matchers` signature happened here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39243/files#diff-013c5ccd6d4d4b59b744d1b1aa39c80fL54

The version this change occurred in is `@types/jest@24.0.20`. This package requires `@types/jest@^23.0.2`.

https://github.com/emotion-js/emotion/blob/master/package.json#L178

Since `jest` is declared globally, these interfaces must have mergeable signatures. Fixing might be a breaking change?

Fixing would require updating the `@types/jest` to the latest version and changing the interface of `Matchers` to:

```ts
interface Matchers<R, T> {
```

https://github.com/NicholasBoll/jest-emotion-issue
Repo:
1. `yarn install`
2. `yarn tsc index.ts`

**Expected behavior:**

There should be no error

**Environment information:**

It only matters that the project depends on `test@24.0.30` or higher or another dependency does. The PR I linked shows many packages that rely on `@types/jest` that were updated together.

NicholasBoll:
**What**:
Fixes #1689

**Why**:
Any project using `@types/jest` >= `24.0.20` (either directly or indirectly through transitive dependencies) will have Typescript errors. I'm unsure if this should be a breaking change

**How**:
I updated the package versions for `@types/jest` and updated the `Matchers` interface to match the interface defined in `@types/jest` for proper interface merging.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/wise-moles-dress.md
+++ b/.changeset/wise-moles-dress.md
@@ -1,0 +1,5 @@
+---
+'jest-emotion': minor
+---
+
+Update Jest types which includes a change to the Matchers type. This may be a breaking change for Typescript users who are using @types/jest < 24.0.20
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "@mdx-js/mdx": "^1.1.0",
     "@mdx-js/react": "^1.0.27",
     "@testing-library/react": "^8.0.2",
-    "@types/jest": "^23.0.2",
+    "@types/jest": "^24.0.20",
     "@types/node": "^10.11.4",
     "@types/react": "^16.8.20",
     "babel-check-duplicated-nodes": "^1.0.0",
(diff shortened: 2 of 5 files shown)


Andarist:
As discussed - this is already a part of the upcoming v11 and we'd rather like to avoid merging this into v10 because it's sort of a breaking change. Gonna close this and I hope we manage to release v11 in a couple of weeks.

PR #1692, opened 2019-12-15
fix(core / styled): fix wrap the display name in styled and withEmotionCache

zthxxx:
**What**: display correct styled component's name in React DevTools

**Why**: 

`@emotion/styled` use `withEmotionCache` function which is a HOC but not wrap a [displayName](https://reactjs.org/docs/higher-order-components.html\#convention-wrap-the-display-name-for-easy-debugging).

So in [React DevTools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en), it will be always show only `render` string as component name.

Before wrap displayName:

image

**How**:

Follow React [displayName docs](https://reactjs.org/docs/higher-order-components.html\#convention-wrap-the-display-name-for-easy-debugging), inject `displayName` before and into `withEmotionCache`.

After wrap displayName:

image

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/packages/core/src/context.js
+++ b/packages/core/src/context.js
@@ -18,10 +18,16 @@ export let ThemeContext = React.createContext<Object>({})
 export let CacheProvider = EmotionCacheContext.Provider
 
 /**
- * https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging 
+ * https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging
  */
-function getDisplayName(WrappedComponent) {
-  return WrappedComponent.displayName || WrappedComponent.name || 'EmotionCacheRender';
+function getDisplayName<Props>(
+  WrappedComponent: React.ComponentType<Props>
+): string {
+  return (
+    WrappedComponent.displayName ||
+    WrappedComponent.name ||
+    'EmotionCacheRender'
+  )
 }
 
 let withEmotionCache = function withEmotionCache<Props>(
@@ -40,7 +46,7 @@ let withEmotionCache = function withEmotionCache<Props>(
     )
   }
 
-  render.displayName = getDisplayName(func)
+  render.displayName = getDisplayName<Props>(func)
 
   // $FlowFixMe
   return React.forwardRef(render)
--- a/packages/core/src/context.js
+++ b/packages/core/src/context.js
@@ -17,6 +17,13 @@ let EmotionCacheContext: React.Context<EmotionCache | null> = React.createContex
 export let ThemeContext = React.createContext<Object>({})
 export let CacheProvider = EmotionCacheContext.Provider
 
+/**
+ * https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging 
+ */
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'EmotionCacheRender';
+}
+
 let withEmotionCache = function withEmotionCache<Props>(
   func: (props: Props, cache: EmotionCache, ref: React.Ref<*>) => React.Node
 ): React.StatelessFunctionalComponent<Props> {
@@ -32,6 +39,9 @@ let withEmotionCache = function withEmotionCache<Props>(
       </EmotionCacheContext.Consumer>
     )
   }
+
+  render.displayName = getDisplayName(func)
+
   // $FlowFixMe
   return React.forwardRef(render)
 }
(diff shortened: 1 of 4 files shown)


zthxxx:
cc @Andarist

.changeset/nervous-otters-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+'@emotion/core': patch
```
Andarist: I would split this into 2 different changesets, one for `@emotion/styled`+`@emotion/styled-base` and the other one for `@emotion/core`. That way we could write specific changelog entries for each one.

packages/styled-base/src/index.js
```diff
@@ -43,6 +43,20 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
+    identifierName !== undefined
+      ? identifierName
+      : `Styled(${
+          typeof baseTag === 'string'
+            ? baseTag
+            : baseTag.displayName || baseTag.name || 'Component'
+        })`
+
+  const injectDisplayName = Component => {
+    Component.displayName = Component.displayName || displayName
```
Andarist: there is no way for the `Component` to have a `.displayName` as it's always the anonymous function

packages/styled-base/src/index.js
```diff
@@ -78,8 +92,8 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
     }
 
     // $FlowFixMe: we need to cast StatelessFunctionalComponent to our PrivateStyledComponent class
-    const Styled: PrivateStyledComponent<P> = withEmotionCache(
-      (props, context, ref) => {
+    const Styled: PrivateStyledComponent<P> = withEmotionCache<P>(
```
Andarist: instead of doing this `injectDisplayName` I think it could be more readable if we could just split this into separate statements, like this:
```js
const render = (props, context, ref) => {}
render.displayName = displayName
const Styled = withEmotionCache(render)
```

That being said - shouldn't we also make this displayName in here a little bit more distinct? Right now 2 component will receive the very same displayName from what I can see (the one being this `render` here and the other one being `Styled`)

packages/core/src/context.js
```diff
@@ -32,6 +45,9 @@ let withEmotionCache = function withEmotionCache<Props>(
       </EmotionCacheContext.Consumer>
     )
   }
+
+  render.displayName = getDisplayName<Props>(func)
```
Andarist: does `func` have a name under any circumstances? I think we always pass anonymous functions to it

.changeset/nervous-otters-sell.md
zthxxx: done, new changesets appended

packages/styled-base/src/index.js
zthxxx: > Right now 2 component will receive the very same displayName from what I can see

In my practice, I found that `Styled.displayName` has no effect.

packages/core/src/context.js
zthxxx: Not before, so I made `injectDisplayName` in `styled-base` to inject displayName before `withEmotionCache`, 
it still has no `name` property, but maybe post in the future

packages/styled-base/src/index.js
zthxxx: But first, i wrote codes like yours comment, but i found some problems that 

- its less semantic than `injectDisplayNameToComponent(Component, displayName)` func
- if you want use the same code to solve same problems with other codes where `withEmotionCache` used, it will make lots of code changed; 

  On the other hand `injectDisplayName` func is reusable logic if export and necessary

zthxxx: > I found that Styled.displayName has no effect.

i guess, its due to `forwardRef` in HOC

zthxxx: > i guess, its due to forwardRef in HOC

yes, source code at [forwardRef.js#L46](https://github.com/facebook/react/blob/master/packages/react/src/forwardRef.js#L46)

seems do not accept `Component.displayName` after `forwardRef` return;

some POC codes and results here:

屏幕快照 2019-12-15 下午8 24 30

屏幕快照 2019-12-15 下午8 25 08

屏幕快照 2019-12-15 下午8 29 17

zthxxx:
I try to fix it in react devtools in facebook/react#17613

Andarist:
Hm, this seems like some sort of regression in devtools. I've found a PR supposedly fixing this: https://github.com/facebook/react-devtools/pull/1154/files

zthxxx:
yep, I think it's actually a problem in React DevTools.

zthxxx:
Thanks for review, I have fixed the real reason in facebook/react#17613

Andarist:
No problem, I'm glad that you have fixed this in devtools ❤️

PR #1696, opened 2019-12-18
Fix issue with rehydrated style tags not taking into account prepend option

Andarist:
This fixes an issue with rehydrated tags always being appended - regardless of `prepend` option.

It also makes sheet own "its" rehydrated styles - which seems like a good change, because without that `flush` didn't affect those.

### DIFF
--- a/packages/sheet/src/index.js
+++ b/packages/sheet/src/index.js
@@ -84,20 +84,27 @@ export class StyleSheet {
     this.before = null
   }
 
+  _insertTag = (tag: HTMLStyleElement) => {
+    let before
+    if (this.tags.length === 0) {
+      before = this.prepend ? this.container.firstChild : this.before
+    } else {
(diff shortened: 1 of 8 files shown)


packages/sheet/src/index.js
```diff
@@ -84,20 +84,27 @@ export class StyleSheet {
     this.before = null
   }
 
+  _insertTag = (tag = createStyleElement(this)) => {
```
mitchellhamilton: This will probably be slightly smaller if we don't do a default arg and instead call it with that where we need to

Andarist: Agreed - proposed change is much better.

Andarist:
@mitchellhamilton fixed a nit-pick about `_insertTag` and added some tests, hard to test actual SSRed styles when used in combination with `@emotion/cache` so I've just skipped this. Should be ready to review & merge.

PR #1698, opened 2019-12-22
Add LICENSE files

### LINKED ISSUE #1694: LICENSE file should be included in npm bundle
The license text is currently not provided with the npm bundle.
In order to allow automation tools like `license-webpack-plugin` to pick it up, it should be included there.
Also the license states: The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
This includes the npm registry.

Andarist:
fixes #1694


LICENSE
```diff
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2016 Kye Hohenberger
+Copyright (c) Emotion team and other contributors
```
Andarist: React doesn't include the year - so I thought it's good to just drop it. If they can do it - we can also 😅 

I was not entirely sure who to put as copyright holders, chosen "Emotion team and other contributors". If you want to change this, let me know.

PR #1703, opened 2019-12-30
Should forward props options

JakeGinnivan:
**What**: Add options object to shouldForwardProp which takes include and exclude.

**Why**: Most of the time we just want to either include a subset of props or exclude a few props and have to write a custom filter function. This allows an options object to be specified instead making it much easier.

Also writing type guards to prevent prop collisions is a little painful in TypeScript, I started putting a helper function in the docs but then thought why not just add support into the library.

**How**: In both the react native and normal create styled functions, if a function is not specified assume it's an options object and inspect the includes/excludes keys to generate the appropriate forwarding function.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

This replaces #1682, I changed the approach after the initial feedback, so there are a few docs changes in here too.

### DIFF
--- a/packages/primitives/test/emotion-primitives.test.js
+++ b/packages/primitives/test/emotion-primitives.test.js
@@ -188,18 +188,86 @@ describe('Emotion primitives', () => {
   })
 
   test('custom shouldForwardProp works', () => {
-    const Text = styled.Text``
-    const Title = props => <Text {...props} />
-    // $FlowFixMe
+    let forwardedProps: any
+    const Title = props => {
+      forwardedProps = props
+      return null
+    }
     const StyledTitle = styled(Title, {
       shouldForwardProp: prop => prop !== 'color' && prop !== 'theme'
     })`
       color: ${props => props.color};
     `
 
-    const tree = renderer
-      .create(<StyledTitle color="hotpink">{'Emotion'}</StyledTitle>)
-      .toJSON()
-    expect(tree).toMatchSnapshot()
+    renderer.create(
+      // $FlowFixMe
+      <StyledTitle color="hotpink" disabled>
+        {'Emotion'}
+      </StyledTitle>
+    )
+
+    expect(forwardedProps).toEqual({
+      children: 'Emotion',
+      disabled: true,
+      style: {
+        color: 'hotpink'
+      }
+    })
+  })
+
+  test('shouldForwardProp exclude option works', () => {
+    let forwardedProps: any
+    const Title = props => {
+      forwardedProps = props
+      return null
+    }
+    const StyledTitle = styled(Title, {
+      shouldForwardProp: { exclude: ['color', 'theme'] }
+    })`
+      color: ${props => props.color};
+    `
+
+    renderer.create(
+      // $FlowFixMe
+      <StyledTitle color="hotpink" disabled>
+        {'Emotion'}
+      </StyledTitle>
+    )
+
+    expect(forwardedProps).toEqual({
+      children: 'Emotion',
+      disabled: true,
+      style: {
+        color: 'hotpink'
+      }
+    })
+  })
+
+  test('shouldForwardProp include option works', () => {
+    let forwardedProps: any
+    const Title = props => {
+      forwardedProps = props
+      return null
+    }
+    const StyledTitle = styled(Title, {
+      shouldForwardProp: { include: ['disabled'] }
+    })`
+      color: ${props => props.color};
+    `
+
+    renderer.create(
+      // $FlowFixMe
+      <StyledTitle color="hotpink" disabled>
+        {'Emotion'}
+      </StyledTitle>
+    )
+
+    expect(forwardedProps).toEqual({
+      children: 'Emotion',
+      disabled: true,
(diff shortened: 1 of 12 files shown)


packages/primitives/test/emotion-primitives.test.js
```diff
@@ -188,18 +188,86 @@ describe('Emotion primitives', () => {
...
       color: ${props => props.color};
     `
 
-    const tree = renderer
-      .create(<StyledTitle color="hotpink">{'Emotion'}</StyledTitle>)
-      .toJSON()
-    expect(tree).toMatchSnapshot()
+    renderer.create(
+      // $FlowFixMe
+      <StyledTitle color="hotpink" disabled>
```
JakeGinnivan: I changed these tests because the Text component has it's own filters, so the test would always pass even if you added more props

packages/styled/types/base.d.ts
```diff
@@ -104,28 +177,55 @@ export interface CreateStyled {
   >(
     component: C,
     options: FilteringStyledOptions<PropsOf<C>, ForwardedProps>
-  ): CreateStyledComponent<Pick<PropsOf<C>, ForwardedProps> & { theme?: Theme }>
```
JakeGinnivan: I moved the theme prop down a level to remove a generic argument

packages/styled/types/base.d.ts
```diff
@@ -96,6 +146,29 @@ export interface CreateStyledComponent<
  * @example styled('div')<Props>(props => ({ width: props.width })
  */
 export interface CreateStyled {
+  // Component + shouldForwardProp: { include: [] }
```
JakeGinnivan: I opted to use overloads here rather than mapped types because i thought it would be simpler

JakeGinnivan:
@mitchellhamilton that was the approach I started with, the problem is that for typing you lose all context if you create a separate helper.

See the comment in https://github.com/emotion-js/emotion/pull/1682#discussion_r356331578

I can't see any way this can be implemented in userland without making the user double up typings and removing all type inference.

JakeGinnivan:
@mitchellhamilton any further thoughts about this, suggestions welcome on how to write something external which doesn't suffer the problems I outlined.

Andarist:
The base form of a `shouldForwardProp` - function, doesn't suffer from inference problems, right? As you can just write it inline and have types to be filled in in the very same way your exclude/include proposal does this?

Btw - when using helper it should actually be possible (although it might have some quirks associated with it) to leverage inference if only helper is used "inline". [XState](https://xstate.js.org/) is relaying on this behavior of TS heavily. Not sure how it would play out in this particular case though.

JakeGinnivan:
> The base form of a shouldForwardProp - function, doesn't suffer from inference problems, right? As you can just write it inline and have types to be filled in in the very same way your exclude/include proposal does this?

Yes.

## Current
``` ts
shouldForwardProp(propName): propName is 'width' | 'height' {
    return propName === 'width' || propName === 'height'
}
// or
shouldForwardProp(propName): propName is 'width' | 'height' {
    return ['width', 'height'].indexOf(propName) !== -1
}
```

## Proposed
``` ts
shouldForwardProp: { include: ['width', 'height'] }
```

While only a little bit of code it is just painful having to specify the props in the function and the type guard. Also this sort of logic, especially when dealing with more is super easy to get the wrong way. Cognitively is it hard to read, especially if you are not polyfilling `.includes`

## Using helper

``` ts
shouldForwardProp: createFilterForProps<typeof JSX.IntrinsicElements['div']>()(['width', 'height'])
// or maybe

shouldForwardProp: createFilterForProps<typeof JSX.IntrinsicElements['div']>().include(['width', 'height'])
// and
shouldForwardProp: createFilterForProps<typeof JSX.IntrinsicElements['div']>().exclude(['width', 'height'])
```

There are 2 things which make this annoying, first you have to specify the types of the component props again, this *will* get out of sync with the actual component over time. As far as I know you can't get TypeScript to infer a generic type based on an assignment (the helper is being assigned to shouldForwardProp). Taking a look at xstate it uses function chaining which doesn't suffer from this issue because the chained function would be a closed generic, with this helper we have to start from scratch.

Second you need to do it in 2 function calls due to the limitation of having to infer all or none of the generic params at the moment in TypeScript.

The actual code changes to the library are really small and IMO greatly improves the readability of this API for the majority use case, especially for TypeScript users.

https://github.com/emotion-js/emotion/pull/1703/files#diff-b27635bde7e4523923453024f4a15dfb

I understand not wanting to add more to the core library, in this case I can't see a good alternative.

Andarist:
> As far as I know you can't get TypeScript to infer a generic type based on an assignment (the helper is being assigned to shouldForwardProp).

This is what I was talking about before. This very thing is possible, it works by default - it just works under TS inference capabilities and it's hard for me to assess if this particular helper would play ok with TS inference or not. Take a look at this example - https://codesandbox.io/s/small-fast-h23l3 . Notice how `ctx` and `ev` are inferred inside `assign` based on its generic parameters being inferred from Machine.

Issue #1707, opened 2020-01-08
Types: `props.theme` should not be optional in Emotion 11


fabien0102:
**Current behavior:**

The `props.theme` is `Theme | undefined` instead of `Theme`, so it's very hard to migrate to emotion 11 since we need to make sure that the theme is there on every usage.

This was previously discussed and fixed here -> https://github.com/emotion-js/emotion/pull/1501#issuecomment-553489241

I also follow the project to migrate to the new way, as describe here -> https://github.com/emotion-js/emotion/blob/next/docs/typescript.mdx

Note: Since the theme is a runtime variable inject by the `ThemeProvider`, I totally get that `Theme` is actually in reality optional and the current implementation is more typesafe, but in another hand, this will be very painful to use if we need to check everytime if the theme is there before consuming it… 

**To reproduce:**

https://codesandbox.io/s/brave-tereshkova-lx8sn

![image](https://user-images.githubusercontent.com/1761469/71978846-9811d500-321c-11ea-9df5-6229562949d4.png)

=> `props.theme` is `Theme | undefined` instead of `Theme`

**Expected behavior:**

I was expecting `props.theme` to be `Theme`, so the migration to emotion 11 is way less painful 😬 

**Environment information:**

- `react` version: 16.8.4
- `emotion` version: 11.0.0-next.10

Andarist:
https://github.com/emotion-js/emotion/blob/31e610f2385d5a3dfd532b31f743e5f6b9fee43b/docs/typescript.mdx#define-a-theme
> If you were previously relying on theme being an any type, you have to restore compatibility with: [...]

fabien0102:
Thanks, but I like my typesafety, this was not `any` and putting `any` there will be a HUGE lost…

So I guess you suggest that we need to check if theme is there everytime? 😕

fabien0102:
For the record, this was considered like a bug two months ago and fixed -> https://github.com/emotion-js/emotion/pull/1501#issuecomment-553840838

fabien0102:
Also, this is not consistant with the object notation, `Theme` is well defined in this case.

```tsx
import styled from "@emotion/styled";

export default styled.h2(({ theme }) => ({
  color: theme.color // `theme` is `Theme` here
}));
```

Andarist:
[...]

PR #1708, opened 2020-01-08
Fix the type of `Theme` for StyledComponent

fabien0102:
**What**:

Fix the type of `Theme` for styledComponent

**Why**:

https://github.com/emotion-js/emotion/issues/1707

**How**:

Trying to follow the previous resolution from https://github.com/emotion-js/emotion/pull/1632

**Checklist**:

- [ ] Documentation N/A
- [x] Tests (I would love too but not sure how to add them since `Theme` must be globally override)
- [x] Code complete
- [ ] Changeset  N/A

### DIFF
--- a/.changeset/tasty-goats-crash.md
+++ b/.changeset/tasty-goats-crash.md
@@ -2,4 +2,4 @@
 '@emotion/styled': patch
 ---
 
-Fix `props.theme` type in styled component interpolation, this was `Theme | undefined` instead of `Theme`.
+Fix `props.theme` type in styled component interpolation being optional (`Theme | undefined`) for components without `AdditionalProps` being specified.
--- a/.changeset/tasty-goats-crash.md
+++ b/.changeset/tasty-goats-crash.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fix `props.theme` type in styled component interpolation, this was `Theme | undefined` instead of `Theme`.
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -68,7 +68,9 @@ export interface CreateStyledComponent<
 
   (
     template: TemplateStringsArray,
-    ...styles: Array<Interpolation<ComponentProps & SpecificComponentProps>>
+    ...styles: Array<
+      Interpolation<ComponentProps & SpecificComponentProps & { theme: Theme }>
+    >
   ): StyledComponent<ComponentProps, SpecificComponentProps>
 
   /**
--- a/packages/styled/types/tests.tsx
+++ b/packages/styled/types/tests.tsx
@@ -86,7 +86,6 @@ const Button = styled.button``
 const Input = styled.input`
   & + ${Label}: {
     margin-left: 3px;
-    color: ${props => props.theme.primary};
   }
 `
 
@@ -112,6 +111,9 @@ const Input4 = styled.input<AdditionalTest>`
   & + ${Label}: ${props => ({
   marginLeft: props.left
 })}
+`
+const Input5 = styled.input`
+  color: ${props => props.theme.primary};
 `
 ;<Input
   onChange={(evt: React.ChangeEvent<HTMLInputElement>) =>
--- a/packages/styled/types/tests.tsx
+++ b/packages/styled/types/tests.tsx
@@ -86,6 +86,7 @@ const Button = styled.button``
 const Input = styled.input`
   & + ${Label}: {
     margin-left: 3px;
+    color: ${props => props.theme.primary};
   }
 `
 


packages/styled/types/base.d.ts
```diff
@@ -68,7 +68,9 @@ export interface CreateStyledComponent<
 
   (
     template: TemplateStringsArray,
-    ...styles: Array<Interpolation<ComponentProps & SpecificComponentProps>>
+    ...styles: Array<
```
Andarist: could you add a type test for this?

fabien0102: I will try, still need to figure how ^^ I was trying with local project to see if everything was following 😉

fabien0102: I was hoping to have some tests failing in the CI to help me 😅 but everything is green 😕

Andarist: Should be fairly simple to add stuff here:
https://github.com/emotion-js/emotion/blob/31e610f2385d5a3dfd532b31f743e5f6b9fee43b/packages/styled/types/tests-base.tsx

fabien0102: For sure, I just need to see how dtslint is working (my projects are usually directly in typescript 😅 ) Do you have a wait to execute `yarn test:typescript` on one file only or at least one project (and even better with a watch mode? 😇)

fabien0102:
@Andarist Another quick question since I'm trying to make my project working with the emotion@next.

`@emotion/serialized` seams to not be up to date in npm (still consume `Theme` from `@emotion/core` instead of `@emotion/react`) can you publish a new version of `@emotion/serialized` to have the `next` version?

image

packages/styled/types/tests.tsx
```diff
@@ -86,6 +86,7 @@ const Button = styled.button``
 const Input = styled.input`
   & + ${Label}: {
     margin-left: 3px;
+    color: ${props => props.theme.primary};
```
Andarist: Have you confirmed that this would be broken without your fix?

packages/styled/types/tests.tsx
```diff
@@ -86,6 +86,7 @@ const Button = styled.button``
 const Input = styled.input`
```
Andarist: could you also add one a little bit simpler test that would test this thing alone? this one is testing that `Label` can be interpolated as a part of the selector

packages/styled/types/tests.tsx
fabien0102: Yep! definetely broken before and fix after 😉

Andarist:
> @emotion/serialized seams to not be up to date in npm (still consume Theme from @emotion/core instead of @emotion/react) can you publish a new version of @emotion/serialized to have the next version?

Interesting, this might be a slight problem with our defined dependencies. Gonna discuss with @mitchellhamilton and release a fix for this later.

Andarist:
I have prepared a fix for `@emotion/serialize`: https://github.com/emotion-js/emotion/pull/1710

Could you add a changeset to this PR?

fabien0102:
> I have prepared a fix for `@emotion/serialize`: #1710
> 
> Could you add a changeset to this PR?

image

Something like this? (sorry, first time, so I prefer check ^^)

Andarist:
Yes - exactly. You should only list `@emotion/styled` in it though (`@emotion/serialize` problem is going to be handled through other PR) and write a little bit more descriptive message about the fix: what was broken and how it got fixed (not necessarily with code examples, just please provide more context about the fix). Not sure if you have used it but you can run `yarn changeset` to get a nice CLI experience for creating a changeset (you don't have to do that now, as you already have a file created and you can just adjust it manually).

fabien0102:
I used `yarn changeset`, was very nice 👌 I may stole this tool for some of my projects 😁

Andarist:
> I used yarn changeset, was very nice 👌 I may stole this tool for some of my projects 😁

👌 highly recommend it, much better than semantic commits in my opinion. Great especially for monorepos (it manages version bumps between workspace packages), but can be also used for single-package repositories as well.

.changeset/tasty-goats-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fix `props.theme` type in styled component interpolation, this was `Theme | undefined` instead of `Theme`.
```
Andarist:
```suggestion
Fix `props.theme` type in styled component interpolation being optional (`Theme | undefined`) for components without `AdditionalProps` being specified.
```

Andarist: I've wanted to commit this suggestion through the GitHub interface - but it seems like I don't have sufficient permissions to push to your fork.

fabien0102: image

This was already set, you are maybe not considered as maintainer 🙄

Andarist: That's weird then, I'm an organization member with admin rights here - also I've been successfully committing such suggestions in the past in other PRs. Maybe GitHub has some hiccup 🤷‍♂

Andarist:
Thanks ❤️

PR #1710, opened 2020-01-09
Specify `@emotion/react` as optional peer dependency of `@emotion/serialize`

### LINKED ISSUE #1718: @emotion/serialize is not published with the renamed packages.
**Current behavior:**

@emotion/serialize is not published and its types still refer to @emotion/core.

**To reproduce:**

1. Install @emotion/react@11.0.0-next.10
2. Open node_modules/@emotion/serialize/types/index.d.ts
3. Third line reads: import { Theme } from '@emotion/core'

Or

See: https://cdn.jsdelivr.net/npm/@emotion/serialize@0.12.0-next.3/types/index.d.ts

**Expected behavior:**

The already updated version is published. The renamed packages can be used with TypeScript again.

**Environment information:**

- `react` version: 16.11.0
- `emotion` version: 11.0.0-next.10 `@emotion/serialize` version: 0.12.0-next.3

Andarist:
fixes #1718

Lack of this has caused `@emotion/serialize` not being released along with the pkg-renaming stuff.

This is causing a `many-pkg` error:
> @emotion/css the package @emotion/css depends on @emotion/serialize which has a peerDependency on @emotion/react but @emotion/react is not specified in the dependencies or peerDependencies of @emotion/css. please add @emotion/react to the dependencies or peerDependencies of @emotion/css

And it's right about the error, but it would also be weird to specify `@emotion/react` as a peer of `@emotion/css`. On the other hand, due to the current dependency graph structure, this is needed, unless we want to restructure stuff so this could be avoided - I'm not sure how this could be done though (but I also don't have much time right now to think this through).

### DIFF
--- a/.changeset/six-moons-deliver.md
+++ b/.changeset/six-moons-deliver.md
@@ -1,0 +1,5 @@
+---
+'@emotion/serialize': patch
+---
+
+Specify `@emotion/react` as optional peer dependency as it's used by TS types.
--- a/packages/serialize/package.json
+++ b/packages/serialize/package.json
@@ -20,6 +20,14 @@
     "@emotion/utils": "0.11.2",
     "csstype": "^2.6.7"
   },
+  "peerDependencies": {
+    "@emotion/react": "^11.0.0-next.10"
+  },
+  "peerDependenciesMeta": {
+    "@emotion/react": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "dtslint": "^0.3.0"
   },


mitchellhamilton:
To solve the manypkg error we could make manypkg not error when a dependent of something which has an optional peer dependency but the dependent does not depend on the optional peer dependency right?

Andarist:
Maybe, it's a little blurry for me how this spreads. But yeah, when I'm thinking about it - it should be ok to skip this check when considering a generic situation. I'm unsure how this plays out with strict package managers exactly.

However, in this situation here, it doesn't quite seem OK? Because current typings for `@emotion/serialize` depending on `@emotion/react`, which is not really optional for TS users

mitchellhamilton:
> To solve this we could make serializeStyled generic and just inject the correct typing from the outside, removing this dependency from here. WDYT?

This sounds good to me. I would definitely prefer this

Andarist:
Lol, it seems that this is already done and the problematic import in types is some kind of leftover, it's not used anywhere 😅

PR #1712, opened 2020-01-10
Warn about css prop/babel plugin compatibility earlier

Slapbox:
Since there's only the one warning I think it makes more sense to put it at the top where it shows implementing the plugin in `.babelrc`. I also noticed that the full page for the docs on the `@emotion/babel-preset-css-prop` doesn't make any mention of this incompatibility.

Inspired by a comment by @GuillaumeCisco - https://github.com/emotion-js/emotion/pull/348#issuecomment-458461505


docs/css-prop.mdx
```diff
@@ -31,7 +31,7 @@ Use the [JSX Pragma](#jsx-pragma) method instead.
   "presets": ["@emotion/babel-preset-css-prop"]
 }
 ```
-
+Compatibility Note: If you include the plugin `babel-plugin-transform-react-inline-elements` in your `.babelrc` your styles will not be applied. The plugin is not compatible with the `css` prop.
```
Andarist: I'm not so sure about the change - it obviously doesn't hurt anybody, but I'm wondering if it's the best place for such information. @mitchellhamilton wdyt?

Andarist: I've decided not to include it here as it's already mentioned on the bottom - **but** I've added a new warning about this to the `@emotion/babel-preset-css-prop`'s README

PR #1721, opened 2020-01-14
`theme` not forwarded to custom components

blicksky:
**What**: Updated the documentation for `shouldForwardProp` to indicate that the `theme` prop is not forwarded by default for custom components.

**Why**: The code at https://github.com/emotion-js/emotion/blob/97673098945a75b716d4cac100c1af46a5ae18f2/packages/styled-base/src/utils.js#L32 omits `theme` and `innerRef` from the props forwarded by default for custom components. The documentation currently states that all props are forwarded for custom components, which is misleading. I decided not to mention `innerRef`, because my understanding is that this is a deprecated feature.

**Checklist**:
- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset  N/A


docs/styled.mdx
```diff
@@ -159,7 +159,7 @@ This API was inspired by [glamorous](https://github.com/paypal/glamorous). ❤
 
 ### Customizing prop forwarding
 
-By default, Emotion passes all props to custom components and only props that are valid html attributes for string tags. You can customise this by passing a custom `shouldForwardProp` function. You can also use `@emotion/is-prop-valid` (which is used by emotion internally) to filter out props that are not valid as html attributes.
+By default, Emotion passes all props except for `theme` to custom components and only props that are valid html attributes for string tags. You can customise this by passing a custom `shouldForwardProp` function. You can also use `@emotion/is-prop-valid` (which is used by emotion internally) to filter out props that are not valid as html attributes.
```
Andarist: I'm not sure if this is really that misleading. "all props" are mentioned as all given to the component, `theme` is usually injected to interpolations by emotion and not given explicitly by a user.

blicksky: Our use case is for a component library that exports a set of theme objects and components that expose a `theme` prop. These components use emotion internally, and it wasn't clear at first that we needed to set `{ shouldForwardProp: () => true }` so that the `theme` prop would be passed down into internal components.

Andarist: What do you think about such wording?
> By default, Emotion passes all externally received props to custom components and only props that are valid html attributes for string tags.

blicksky: In our case the `theme` prop is externally received. It doesn't seem to me that our case is that unusual.

Andarist: Hm, I believe that most people use the `theme` from the context and not from the external props, so it seems to me that your use case is a little bit unusual. I get what you are saying though, my version wouldn't be correct - because you actually pass `theme` externally.

blicksky: Couldn't it also be possible (though maybe unlikely) that someone could have a component that has a prop that happens to be named `theme`, but is unrelated to emotion's concept of a theme?

Andarist: It's possible - we are in JavaScript world after all 😅 This is completely not supported though and can break badly. It's also not allowed by provided TS types.

Issue #1598, opened 2019-11-01
Link to nonce 404s


bolatovumar:
**Description:**

See "nonce" link here: https://emotion.sh/docs/create-emotion#primary-use-cases
It leads [here](https://emotion.sh/docs/cache#nonce-string) which is a 404 page.

I would expect it to take you the page explaining what the nonce does.

Andarist:
Thank you for the report! I'll try to fix it as soon as I can - in the meantime the section you were looking for is available here: https://emotion.sh/docs/@emotion/cache#nonce

Andarist:
Got fixed by https://github.com/emotion-js/emotion/pull/1903 for the `next` branch. I hope to ship v11 soon and when this happens the main site will benefit from this fix.

PR #1724, opened 2020-01-15
Updating Link to Nonce

lauchness:
Documentation fix for #1598. Link had been broken to the documentation for nonce.

**What**: Documentation fix related to #1598 

**Why**: Link to the nonce documentation is currently broken

**How**: Updating the docs to include the correct link

**Checklist**:
- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset N/A

Hopefully this is a good first PR, just a fix to the documentation to get used to the contribution process.

### DIFF
--- a/packages/create-emotion/README.md
+++ b/packages/create-emotion/README.md
@@ -39,7 +39,7 @@ export const {
 
 - Using emotion in embedded contexts such as an `<iframe/>`
 
-- Setting a [nonce](/packages/cache#nonce-string) on any `<style/>` tag emotion creates for security purposes
+- Setting a [nonce](/packages/@emotion/cache#nonce) on any `<style/>` tag emotion creates for security purposes
 
 - Use emotion with a developer defined `<style/>` tag
 


packages/create-emotion/README.md
```diff
@@ -39,7 +39,7 @@ export const {
 
 - Using emotion in embedded contexts such as an `<iframe/>`
 
-- Setting a [nonce](/packages/cache#nonce-string) on any `<style/>` tag emotion creates for security purposes
```
Andarist: problem is that we'd like those URLs to work on GitHub as well. Those URLs are rewritten when building site which is I think is happening here - https://github.com/emotion-js/emotion/blob/d0b2a94ab9d5648667447dbd78e7a2e3e93de42a/site/plugins/gatsby-remark-fix-links/index.js . This script has to have some bug in it. I think I've tried to fix this in the past, just very briefly though - couldn't find more time to finish the work. My attempt can be found here https://github.com/emotion-js/emotion/pull/1432/files . Maybe you would like to give it a go?

lauchness: Will do. 

Do you think the fact that the hash-link in the docs says `nonce-string`, and the correct hash is `nonce` could be the culprit?

Andarist: It should be corrected as well, but its not the main culprit i think. The main one being this broken rewrite script.

Andarist:
Got fixed by https://github.com/emotion-js/emotion/pull/1903 for the `next` branch. I hope to ship v11 soon and when this happens the main site will benefit from this fix.

PR #1729, opened 2020-01-19
ThemeOverride to make define custom theme type

morlay:
**What**:

We need to define project theme type to make theme parameters clear.

**Why**:

we could not do this by 
```ts
declare module 'react' {
  interface DOMAttributes<T> {
    css?: InterpolationWithTheme<ThemeObject>
  }
}

declare global {
  namespace JSX {
    /**
     * Do we need to modify `LibraryManagedAttributes` too,
     * to make `className` props optional when `css` props is specified?
     */

    interface IntrinsicAttributes {
      css?: InterpolationWithTheme<ThemeObject>
    }
  }
}
```

**How**:

after new changes. we could do like this.

```ts
declare module "@emotion/core" {
  export interface ThemeOverride {
    theme(): { 
       color: { primary: "#000" } 
    }; 
  }
}
```

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/core/types/index.d.ts
+++ b/packages/core/types/index.d.ts
@@ -26,7 +26,22 @@ export {
 
 export { EmotionCache, Interpolation, SerializedStyles, css }
 
-export const ThemeContext: Context<object>
+/**
+ *
+ * ```
+ * declare module "@emotion/core" {
+ *   export interface ThemeOverride {
+ *     theme(): TTheme; // my custom theme
+ *   }
+ * }
+ * ```
+ */
+export interface ThemeOverride {
+  theme(): any;
+}
+type ThemeObject = ReturnType<ThemeOverride["theme"]>
+
+export const ThemeContext: Context<ThemeObject>
 export const CacheProvider: Provider<EmotionCache>
 export function withEmotionCache<Props, RefType = any>(
   func: (props: Props, context: EmotionCache, ref: Ref<RefType>) => ReactNode
@@ -45,7 +60,7 @@ export interface GlobalProps<Theme> {
  * @desc
  * JSX generic are supported only after TS@2.9
  */
-export function Global<Theme = any>(props: GlobalProps<Theme>): ReactElement
+export function Global<Theme = ThemeObject>(props: GlobalProps<Theme>): ReactElement<any>
 
 export function keyframes(
   template: TemplateStringsArray,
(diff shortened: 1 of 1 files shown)


Andarist:
This has already be done for the upcoming v11. Im afraid that we cannot backport this to v10. V11 is available on npm though so i would encourage you to migrate to it

morlay:
cool. i will migrate to the new version tomorrow 😀.

Andarist:
Give me a ping if you encounter any problems.

morlay:
@Andarist the way for defining type of v11 is a little dirty.

for theming, we usually define a theme object first. 
```ts
const defaultTheme = {
   fontFamily: { 
     mono: "",
     system: "",
   }, 
   ...
}
```

then got the theme type through `typeof`
```ts
type TTheme = typeof defaultTheme
```

and now we have to use this way to overwrite 
```
declare module "@emotion/react" {
  export interface Theme {
    fontFamily: TTheme["fontFamily"];
    weight: TTheme["weight"];
    size: TTheme["size"];
    color: TTheme["color"];
    zIndex: TTheme["zIndex"];
    radius: TTheme["radius"];
    spacing: TTheme["spacing"];
    shadow: TTheme["shadow"];
}
```

I prefer the way of this pr
```ts
declare module "@emotion/core" {
  export interface ThemeOverride {
    theme(): TTheme
  }
}
```

Andarist:
Doesnt this work for you?
```ts
declare module '@emotion/react' {
  export interface Theme extends TTheme {}
}
```

morlay:
@Andarist 
it will not work.
typeof object is type，not interface. so not be extendable.

sorry for replaying late. busying with Chinese New Year

Andarist:
> sorry for replaying late. busying with Chinese New Year

No problem :)

> typeof object is type，not interface. so not be extendable.

Can't you do something like this?
```ts
type TTheme = typeof defaultTheme

interface ITTheme extends TTheme {}

declare module '@emotion/react' {
  export interface Theme extends ITTheme {}
}

```

morlay:
@Andarist 
Sorry, I hold old context about `TS2312: An interface may only extend a class or another interface.`.
TypeScript team upgrade it since version 2.2. 
Now only `typeof` literal value (`1`, `true`, `"v"`...) has the limits.
object `{}`, array `[]`, even `new String()`,`new Boolean()` could be extendable. 

So the first way will be work too.

```ts
const defaultTheme = {} 

type TTheme = typeof defaultTheme

declare module '@emotion/react' {
  export interface Theme extends TTheme {}
}
```

Thanks

PR #1732, opened 2020-01-22
Fixed some composition issues with styles not containing a final semi

### LINKED ISSUE #1730: [babel-plugin@v11] transform css prop without css wrapper for array value
**Current behavior:**

work well Component in v10
```ts
export const Graph = ({ children, inline, ...otherProps }: IGraphProps) => {
  return (
    <span
      css={[
        {
          alignItems: "center",
          verticalAlign: "middle",
          lineHeight: "inherit",
        },
        {
          "& > svg": {
            width: "100%",
            height: "auto",
          },
        },
        { display: inline ? "inline-flex" : "flex" },
      ]}
      {...otherProps}>
      {children}
    </span>
  );
};
```

in v11 it is transformed to 

```js
var Graph = function Graph(_ref) {
  var children = _ref.children,
      inline = _ref.inline,
      otherProps =
  /*#__PURE__*/
  _babel_runtime_helpers_objectWithoutProperties__WEBPACK_IMPORTED_MODULE_1___default()(_ref, ["children", "inline"]);

  return Object(_emotion_react__WEBPACK_IMPORTED_MODULE_3__["jsx"])("span",
  /*#__PURE__*/
  _babel_runtime_helpers_extends__WEBPACK_IMPORTED_MODULE_0___default()({
    // here 
    //  in v10, there have css of `@emotion/css` wrapper
    css: ["align-items:center;vertical-align:middle;line-height:inherit;& > svg{width:100%;height:auto;}", {
      display: inline ? "inline-flex" : "flex"
    },  false ? undefined : ";label:Graph"]
  }, otherProps), children);
};
```
and styles broken

after added css wrapper of `@emotion/react`, it work well.

**Expected behavior:**

should transformed with `css` wrapper for array value.
or process it by core lib.

**Environment information:**

- `react` version: 16.8.x
- `emotion` version: v11.0.0-next.11

Andarist:
fixes #1730


packages/serialize/src/index.js
```diff
@@ -264,7 +264,12 @@ function createStringFromObject(
 
   if (Array.isArray(obj)) {
     for (let i = 0; i < obj.length; i++) {
-      string += handleInterpolation(mergedProps, registered, obj[i], false)
+      string += `${handleInterpolation(
```
Andarist: @mitchellhamilton any ideas when this could break stuff? I cant find any, so I think this is safe to do, but would be great to give this a little bit more extra thought.

Andarist:
@mitchellhamilton would you be able to take a look at this?

PR #1736, opened 2020-01-23
Allow inert attribute

bezoerb:
**What**: 
Allow inert attribute

**Why**:
When an element has an attribute `inert="true"`, browser may "ignore" it from assistive technologies, page search and text selection. This can be useful when building UIs such as modals where you would want to "trap" the focus inside the modal when it's visible.
See https://html.spec.whatwg.org/multipage/interaction.html#inert and https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

**How**:
Adds `inert: true` to the list of allowed props

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Changeset

### DIFF
--- a/.changeset/six-cows-tie.md
+++ b/.changeset/six-cows-tie.md
@@ -1,0 +1,5 @@
+---
+'@emotion/is-prop-valid': patch
+---
+
+Adds inert to the list of allowed props
--- a/packages/is-prop-valid/src/props.js
+++ b/packages/is-prop-valid/src/props.js
@@ -184,6 +184,8 @@ const props = {
   autoSave: true,
   // color is for Safari mask-icon link
   color: true,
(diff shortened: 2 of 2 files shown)


Andarist:
Is this supported by any browser?

bezoerb:
> Is this supported by any browser?

I think it's behind a flag in chrome: https://www.chromestatus.com/feature/5703266176335872
But there are polyfills for it. Currently we need to use a small hack to pass the attribute to the dom using `styled-components` and it would be nice if this hack wouldn't be necessary

Andarist:
@mitchellhamilton @probablyup WDYT? I'm okay with adding this.

probablyup:
Lgtm

Issue #1749, opened 2020-02-05
[v11] @emotion/native types are based on old react-native types.


Zn4rK:
**Current behavior:**

Both ActivityIndicatorIOS and WebView are included in the Pick from `@types/react-native@0.57.0`.
Both of which seems to be gone in the latest version of `@types/react-native@^0.61.10`

I'm getting:
`Type '"ActivityIndicatorIOS"' is not assignable to type '"ActivityIndicator" | "Button" | "DatePickerIOS" | "DrawerLayoutAndroid" | "FlatList" | "Image" | "ImageBackground" | "KeyboardAvoidingView" | "ListView" | "Modal" | "NavigatorIOS" | ... 147 more ... | "ViewPropTypes"'.`

**To reproduce:**

Compile TypeScript-project that uses the types and a newer version of react-native.

**Expected behavior:**

I would expect to be able to use the latest version of react-native (and @types/react-native).

**Environment information:**

- `react` version: n/a
- `emotion` version: `@emotion/native@11.0.0-next.11`

Andarist:
Feel free to open a PR to fix this.

cc @patsissons , i guess there is really no way to support a wide range of RN versions, right? So we should just stick to supporting latest

patsissons:
I think I just aligned with whatever was being used in the native package at the time.

Andarist:
closed by https://github.com/emotion-js/emotion/pull/1750

PR #1750, opened 2020-02-06
[v11] fixed native typings

Zn4rK:
**What**:
Fixes https://github.com/emotion-js/emotion/issues/1749

**Why**:
Typings for `@emotion/native`  wasn't aligned with the latest `react-native` types.

**How**:
I upgraded the react-native versions in `packages/native/package.json` and removed the missing types from the union.

I also removed the two entries in the components-array thats sets getters on the styled-object.
So the types are aligned with the expected functionality.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [X] Code complete
- [X] Changeset

### DIFF
--- a/.changeset/angry-cheetahs-march.md
+++ b/.changeset/angry-cheetahs-march.md
@@ -2,4 +2,4 @@
 '@emotion/native': patch
 ---
 
-updating @types/react-native and react-native
+Match supported components to what is exported from the latest version of React Native (0.61.5).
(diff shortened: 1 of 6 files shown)


.changeset/angry-cheetahs-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/native': patch
+---
+
+updating @types/react-native and react-native
```
Andarist: Match supported components to what is exported from the latest version of React Native (0.61.5).

Andarist:
```suggestion
Match supported components to what is exported from the latest version of React Native (0.61.5).
```

Andarist:
Thanks for the PR! Sorry that it took so long to handle this.

PR #1761, opened 2020-02-13
Fix how ILLEGAL_ESCAPE_SEQUENCE_ERROR is raised in some corner cases

### LINKED ISSUE #1755: Unexpected "You have illegal escape sequence..." in Chrome DevTools
**Current behavior:**

I have got the error message saying
> You have illegal escape sequence in your template literal, most likely inside content's property value.
Because you write your CSS inside a JavaScript string you actually have to do double escaping, so for example "content: '\00d7';" should become "content: '\\00d7';".
You can read more about this here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#ES2018_revision_of_illegal_escape_sequences

While my one and only `tsx` file never contains a backslash.
```tsx
import React from 'react'
import ReactDOM from 'react-dom'
import { css } from '@emotion/core'

const RED = 'red'

ReactDOM.render(
  <div
    css={css`
      color: ${RED};
    `}
  >
    Hey
  </div>,
  document.getElementById('root')
)
```

**To reproduce:**

1. `git clone https://github.com/ThisIsManta/parcel-ts-emotion`
2. `yarn`
3. `yarn start`
4. Open http://localhost:1234/ in Google Chrome (may possible to reproduce in other browsers as well)

**Expected behavior:**

Expect NOT to see the error message in Google console.

The error message **will be gone** if I put `"resolutions": { "@emotion/serialize": "0.11.4" }` in my [_package.json_](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) as the latest version of `@emotion/core` auto installs **v0.11.5** which prints the error message unexpectedly.

image

**Environment information:**

- I'm using Parcel + TS + Emotion via [_.babelrc_](https://github.com/ThisIsManta/parcel-ts-emotion/blob/master/.babelrc)
- `react` version: 16.12.0
- `@emotion/core` version: **v10.0.27**
- See the rest in [_package.json_](https://github.com/ThisIsManta/parcel-ts-emotion/blob/master/package.json)
- Chrome for Windows: 80.0.3987.87 (Official Build) (64-bit)

mansourkheffache:
Fixes #1755

**What**:
Better handling of how styles are serialized

**Why**:
Source map string is inserted in a confusing way for tagged templates. It creates one extra argument, with no string to append it to - making the string `undefined` and raising an `ILLEGAL_ESCAPE_SEQUENCE_ERROR` mistakenly. 

**How**:
Prevent checking for strings that are out of the bounds of the input strings array.

**Checklist**:
- [ ] Documentation "N/A"
- [ ] Tests "N/A"
- [x] Code complete
- [ ] Changeset  "N/A"


packages/serialize/src/index.js
```diff
@@ -374,7 +374,7 @@ export const serializeStyles = function(
       args[i],
       styles.charCodeAt(styles.length - 1) === 46
     )
-    if (stringMode) {
+    if (stringMode && i < strings.length) {
```
Andarist: This requires a new test to be added. I don't think we need to assert `console.error` being called or not, but it seems that this has also caused `"undefined"` being appended to the styles, so it would be good to test that it no longer happens.

If you need any help with setting up a repository for the local development or if you are unsure how exactly such a test should be added then please reach out to me on our Slack so I could provide the assistance for you.

Andarist:
I've identified this issue as https://github.com/emotion-js/emotion/pull/1538 not covering all cases.

The core issue here was that TS-transpiled output was handed to us and we have failed to recognize it, so we have just fallen back to appending sourcemaps to the end of arguments, causing this error.

By actually inserting them into correct locations we can avoid an extra runtime check for args & strings length mismatch.

Andarist:
@mansourkheffache thanks for looking into this and the initial fix, I've tweaked it after recognizing the underlying issue and going to release a new version of Emotion today.

Issue #1549, opened 2019-10-15
@emotion/core doesn't handle React.Fragment shorthand correctly.


shakib609:
**Current behavior:**
Throws an error "React is not defined".
If React is imported then throws a warning "'React' is defined but never used."

**To reproduce:**
1. Create an app with `create-react-app`
2. Create a component which uses `jsx` instead of `React`.
3. Use `React.Fragment` shorthand as the container of that component.
4. Import the component and render it.

**CodeSandbox**: [https://codesandbox.io/s/pensive-edison-bdkpp](https://codesandbox.io/s/pensive-edison-bdkpp)

**Expected behavior:**
It would be great if `emotion` could handle the shorthand out of the box. If not, it should not at least warn me about React not being used.

**Environment information:**

- `react` version: 16.10.2
- `@emotion/core` version: 10.0.21

Andarist:
> It would be great if emotion could handle the shorthand out of the box.

It can't - because it's handled by Babel and not by Emotion.

> If not, it should not at least warn me about React not being used.

You get runtime `Uncaught ReferenceError: React is not defined`. Emotion can't handle this because it's not capable of doing static analysis. This is a job for ESLint.

shakib609:
Thanks for the explanation. :+1: 
Should I open an issue in `eslint-plugin-react` to solve this?

Andarist:
If they have rules for jsx itself requiring a React to be in scope or custom jsx pragma, then yes - it should account for Fragments as well. Please verify if it ain't handled already before reporting an issue.

shakib609:
Ok. I'll take a look. :)
Closing this for now.

justingrant:
@shakib609 @Andarist - I think that the issue tracking this problem is https://github.com/yannickcr/eslint-plugin-react/issues/2156.  I added [this comment](https://github.com/yannickcr/eslint-plugin-react/issues/2156) to the end of that issue. 

P.S. - Anyone ending up here from Google, the workaround is simple: use `<React.Fragment>` instead of the shorthand `<>` syntax.

PR #1767, opened 2020-02-18
Add `fragment` option to handle fragment shorthand

SidneyNemzer:
**What**: Added a new option, `fragment`. This is a new opt-in feature for `@emotion/babel-plugin-jsx-pragmatic` that will add `import { default as React } from 'react'` when a JSX fragment is detected.

`babel-plugin-jsx-pragmatic`
In:
```js
const F = () => <>Test</>
```

Out:
```js
import { jsx as ___EmotionJSX } from "@emotion/core";
import { default as React } from "react";

const F = () => <>Test</>;
```

**Why**: Without this option, all JSX is transformed to Emotion's JSX function which means React does not need to be in scope (unlike normal JSX). However, fragments `<></>` compile to `React.Fragment` by default, which causes a runtime error if React is not imported.

This new option is opt-in, and can be specified in `@emotion/babel-preset-css-prop`. I made it opt-in because it might be considered a breaking change. Let me know if you think this should be enabled by default.

Also, if you think there is a better way to do this, let me know! I decided to just open a PR first because this was a pretty simple change.

**How**: I added a new state key, `jsxFragmentDetected`, which causes an additional import to be inserted by the plugin. I refactored the plugin a tiny bit to support two declarations instead of one.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

related: https://github.com/emotion-js/emotion/issues/1549

### DIFF
--- a/.changeset/two-coins-grab.md
+++ b/.changeset/two-coins-grab.md
@@ -1,0 +1,6 @@
+---
+'@emotion/babel-plugin-jsx-pragmatic': minor
+'@emotion/babel-preset-css-prop': minor
+---
+
+Add new option, `fragment`, to automatically import React when shorthand fragments are used. Previously, using a shorthand fragment would cause `React.Fragment` to be used in the output, but this could cause a runtime error if React was not manually imported.
--- a/packages/babel-plugin-jsx-pragmatic/README.md
+++ b/packages/babel-plugin-jsx-pragmatic/README.md
@@ -163,6 +163,24 @@ String. The export that you want to import as `import` from `module`. Default va
 // import {something as x} from "whatever"
 ```
 
+### `fragment`
+
+Object. Disabled by default. Adds a React import when fragment shorthand appears in the source. By default, the shorthand is transpiled to `React.Fragment`, and if this is not enabled, React must be manually imported to use fragment shorthand (otherwise, a runtime error will be thrown).
+
+```js
+{
+  fragment: {
+    module: "react"
+    import: "React",
+  },
+  // other options
+  module: "whatever",
+  import: "x",
+}
+// import {default as x} from "whatever"
+// import {default as React} from "react"
+```
+
 # Known Issues
 
 - Doesn't do anything special in the case that the file being transformed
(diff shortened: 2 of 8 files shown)


Andarist:
Thanks for PR - I hope to get to reviewing it in the following days, but I'm really busy lately and don't have much free time for this right now. So please wait patiently and thanks for the understanding 🙏

Andarist:
@mitchellhamilton what do you think about this one? seems fine to me, although we should check if React is not already declared/imported to avoid parsing errors

SidneyNemzer:
I figured it was OK to potentially have duplicate React imports because this plugin might already be creating duplicate `@emotion/core` imports. 

I also briefly tested with webpack to make sure it would not error on duplicate imports. I'm pretty sure you can't have duplicates in your source, but webpack doesn't care if babel plugins create duplicates.

Andarist:
You can have multiple import declarations with the same source ("react") but you cant have 2 bindings to React. So if there is already a React import in the file (for example when using hooks) then this would error.

SidneyNemzer:
Ah that makes sense. I'll make that change when I have a chance.

mitchellhamilton:
The idea behind this sgtm though I think the option should be `fragment` and that should be an object with `export` `module` and `import` and in the preset, we should set it to import fragment from react(and shouldn't let people change it in the preset).

SidneyNemzer:
@Andarist @mitchellhamilton Updated!

Andarist:
I would like [this PR](https://github.com/babel/babel/pull/11154) to get merged into Babel before we proceed any further with this.

Andarist:
@SidneyNemzer it seems that with that mentioned PR landed and being already published we don't need to handle this on our side as it will be handled "soon" by the official React jsx transform - given new React version with new jsx factories gets released. I'd like to avoid adding this right now as the problem should get fixed by itself with new React and new Emotion version in the near future. Let me know what do you think about this.

SidneyNemzer:
It does appear that PR will address the issue I was trying to solve here, so I'll close this PR

PR #1775, opened 2020-02-20
Optimize hashing for size and performance

kripod:
**What**: Optimized the implementation of `@emotion/hash`, with concisely inlined `Math.imul` operations.

**Why**: According to [my benchmarks](https://jsperf.com/math-imul-alternatives/5), `Math.imul` can be implemented more efficiently, and with less code:

image

**How**: I ported `MurmurHash2`, following the [reference implementation written in C](https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash2.cpp#L37-L86). After minification, the refactored code weighs about 25% less than the original:

image

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset


Andarist:
This looks awesome! ❤️ 

Could you also prepare a perf comparison of the previous version of this hash function and the proposed one? Would be great to also measure the overall gains, rather than focusing solely on the `Math.imul` perf.

kripod:
Sure, thank you! A benchmark is now available [here](https://jsperf.com/murmurhash2-performance-comparison). While iOS Safari 13 and Edge 18 performs way better than before, Chrome and Firefox have performance regressions:

image

image

image

image

---

By replacing:

```js
((x * 0x5bd1) << 16) + x * 0xe995
```

With:

```js
(((x >>> 16) * 0x5bd1e995) << 16) + (x & 0xffff) * 0x5bd1e995
```

Results similar to the old implementation can be achieved. I'm not sure why this happens, but browser engineers _(@mathiasbynens?)_ may come up with an explanation.

Andarist:
Not sure if you have expertise in that area but maybe you could compare generated assembly code for both versions?

kripod:
Unfortunately, I’m not sure how to do that in browsers. I’d appreciate any tips on how to get started.

Andarist:
As the regression~ happens in v8 we can just use node to check this, you can use `node --print-code script.js` to see the generated bytecode.

This is of course rather hard to analyze if you don't have a good knowledge of assembly code. Mine is definitely rusty and I wouldn't be able to analyze this quickly myself.

Some starting materials https://medium.com/dailyjs/understanding-v8s-bytecode-317d46c94775

Let me know if you are capable of jumping into this, if not - let's wait to see if Mathias answers.

kripod:
I will try my best during the weekend, thank you for the advice!

Andarist:
I've asked a friend about this and he has suggested to just perform 32bit operations, avoiding those 16bit conversions. I've prepared a [small benchmark](https://esbench.com/bench/5e4fda6b328a52009e8b1b01) to test this out and it seems to work great for the `Math.imul` alone.

Could you try this out in this PR?

packages/hash/src/index.js
```diff
@@ -1,47 +1,57 @@
...
+  // They're not really 'magic', they just happen to work well.
 
-  while (l >= 4) {
+  // const m = 0x5bd1e995;
+  // const r = 24;
+
+  // Initialize the hash to a 'random' value
+
+  var len = str.length,
+    h = len ^ len;
```
Andarist: This is always equal to 0 (`len ^ len === 0`). Can we just use 0 here? "Initialize the hash to a 'random' value" seems to be incorrect here - original algo initializes this using a received `seed` value (making it random~), but we dont have this here.

kripod:
I started investigating the bytecode and noticed the following:

image

(The bytecodes used in V8 are defined [here](https://github.com/v8/v8/blob/503692b7c435a77a841d5dc2ba585dfb3a010bbd/src/interpreter/bytecodes.h#L127-L150)).

Now it's time to convince engines that we need bitwise operations instead of `MulSmi.Wide` and `MulSmi.ExtraWide`.

kripod:
It turns out that `2 ** 30` or `2 ** 31` are [the boundaries](https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/test/mjsunit/math-imul.js#L113-L125) to cross in order to force performing bitwise multiplication. While the implementation below is not mathematically correct (it produces extra bits above the lower 32-bits), it fits the use-case for Murmur2, because bitwise operations trim the overflowing bits anyway:

image

kripod:
I’m not quite sure about why the algorithm performs worse in the case of Firefox. I went through the generated V8 bytecode line by line and could not notice why the regression occurs. The best-performing benchmark is available [here](https://jsperf.com/murmurhash2-performance-comparison/23).

kripod:
After a lot of experimentation, I finally managed to achieve the ~2x performance boost in iOS Safari while keeping the performance nearly intact in Firefox and Chrome, and also improving it in IE and Edge.

@Andarist This PR is now ready for merging from my side. Please see the final benchmarks [over here](https://jsperf.com/murmurhash2-performance-comparison/29) to verify my results below. They may vary between 5–10%, so I would recommend repeating the benchmark multiple times if necessary.

image

image

image

image

image

Andarist:
Would you be willing to write down what has been changed, what have you tried in the process and what you have learned about engines while doing this? It would be both super interesting, helpful for reviewing this and would serve as a reference for future readers.

kripod:
I worked all day to create a visual comparison between the old and new bytecode _(click on the image to view it in large):_

comparison

The diff is shown in the center with red and blue indicators. Swapped variables are denoted with grey lines and are mostly done to improve in-memory caching.

Instructions have been re-structured to support code minification while keeping the performance intact.

Andarist:
Would you be able to just give a few words about this as well? While I understand **some** of this, it's really low level and I might miss a bunch of stuff about this. Have you learned something new about engines while working on this? Would you have any tips that could be re-used for similar code in other scenarios?

In general, this looks good to me 👍 I would also love @mitchellhamilton to take a look at it before merging.

kripod:
Mainly, I’ve embedded the value of variables instead of referring to them, in order to reduce the size of the code. E.g. by replacing `a = b; a = Math.imul(a, c)` with `a = Math.imul(b, c)`.

After that, I observed the inlined `imul` operation and saw an opportunity for optimization by removing unnecessary operations. I’ve benchmarked multiple implementations of 32-bit integer multiplication and found the one in this PR consustently fast amongst different browsers.

iOS Safari caused the biggest surprise, as its native `Math.imul` function is slower than its userspace counterparts by multiple orders of magnitude.

I’m not sure if any of the techniques applied would be useful in other parts of Emotion without sacrificing code readability.

kripod:
@mitchellhamilton Done, thank you for the review!

.changeset/two-kings-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/hash': minor
+---
+
+Optimize hashing for size and performance
```
Andarist:
```suggestion
Optimized hashing for performance while also reducing the size of the function.
```

Andarist:
@kripod thanks for the great work!

kripod:
I’ve just noticed that c2fc3cc28bf9edb923da8ce01ff37b96e63b7e55 has just reverted all the work I put into this 😔

I’m curious how that performs compared to the contents of this PR, both in terms of operations/second on all the major JavaScript VMs, and compressed bundle size.

PR #1788, opened 2020-03-05
Fix required theme with TypeScript 3.8

### LINKED ISSUE #1779: Type Error with TypeScript 3.8.2 and theme 🏷
**Current behavior:**

After updating typescript to 3.8.2 I've got an error with the theme.
If a styled component uses the `theme` from the props the typescripts throw an error where I import and use the styled-component.
The error tells that I don't pass the `theme` in the styled-component.
Error: `'theme' is declared here.`

**To reproduce:**

1. Update TS to 3.8.2
2. Create a styled-component and use the `theme` from props
3. Implement the styled-component
4. Check the type error 

**Expected behavior:**

The styled-component shouldn't expect the `theme` where we plan to use it.

If I make the `theme` [here](https://github.com/emotion-js/emotion/blob/master/packages/styled-base/types/index.d.ts#L33) an optional parameter here the error dissolves but It creates another error because `theme` becomes an optional parameter.
I think the `theme` should be omitted in this file, maybe that would solve the issue.

**Environment information:**

- `react` version: 16.9.0
- `emotion` version: 10.0.27
- `typescript` version: 3.8.2

aaronjensen:
Fixes #1779

**What**: https://github.com/emotion-js/emotion/issues/1779

**Why**: https://github.com/emotion-js/emotion/issues/1779

**How**: ???

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A - the existing tests, when run against 3.8 should catch this 
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/nine-shoes-sniff.md
+++ b/.changeset/nine-shoes-sniff.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled-base': patch
+---
+
+Fix TypeScript 3.8 required theme prop
--- a/.changeset/nine-shoes-sniff.md
+++ b/.changeset/nine-shoes-sniff.md
@@ -2,4 +2,4 @@
 '@emotion/styled-base': patch
 ---
 
-Fix TypeScript 3.8 required theme prop
+Fixed issue with TypeScript 3.8 where `theme` was inferred to be required for a styled component.
--- a/packages/styled-base/types/index.d.ts
+++ b/packages/styled-base/types/index.d.ts
@@ -39,7 +39,7 @@ export interface StyledOptions {
 }
 
 export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+  extends React.SFC<InnerProps & Omit<StyleProps, 'theme'> & { theme?: Theme }>,
     ComponentSelector {
   /**
    * @desc this method is type-unsafe


packages/styled-base/types/index.d.ts
```diff
@@ -39,7 +39,7 @@ export interface StyledOptions {
 }
 
 export interface StyledComponent<InnerProps, StyleProps, Theme extends object>
-  extends React.SFC<InnerProps & StyleProps & { theme?: Theme }>,
+  extends React.SFC<InnerProps & Omit<StyleProps, 'theme'> & { theme?: Theme }>,
```
Andarist: Could you add a failing test for this?

Andarist:
Do you know by any chance **why** this is necessary now? What has changed in TS and how exactly this change helps?

packages/styled-base/types/index.d.ts
aaronjensen: I'm not sure how to ensure this is tested against 3.8. My suspicion is that it isn't and that's why this wasn't caught. But if you could point me to test this against 3.8, I can ensure that a test is added that fails without this change.

aaronjensen:
> Do you know by any chance **why** this is necessary now? What has changed in TS and how exactly this change helps?

No, I don't.

aaronjensen:
> Do you know by any chance **why** this is necessary now? What has changed in TS and how exactly this change helps?

My guess is that something about the type inference changed such that `StyleProps` is now inferring a `theme` property instead of it being inferred that it does not have one because `WithTheme` adds it in. There are no breaking changes in 3.8 that look like it should have impacted this, but that doesn't necessarily mean anything. Perhaps they actually fixed a bug and this is now behaving as desired.

I could see someone reporting this to TypeScript, but I'm not planning on taking the time to do that. You're more than welcome to if you'd like to try and get to the bottom of this. In the meantime, this is preventing emotion users from upgrading to TS 3.8 and it's a change that introduces a slight redundancy at worst.

Andarist:
It seems that TS tests are actually not running right now on the `master` branch (on the CI) - this got fixed on `next` but we've decided against porting this back to the `master` as we didn't expect v10 to get any new releases besides bug fixes.

However, you can run those tests by entering the appropriate directory and running `yarn test:typescript` - this should run tests against all supported TS versions, including their `next`. Would be great if you could run them without your fix and with your fix to see if they would fail right now on the `master`. If not then it would be great to add a failing test and then confirm if it got fixed by your change.

aaronjensen:
@Andarist it installs 3.4 at highest and next. For whatever reason, 3.6, 3.7, 3.8 are not tested against. Several tests fail in `styled` with `next` with and without my changes. You have bigger problems on `master` with `next` than my PR fixes, but my PR does fix something that several people including us have run into. 

I'm sorry, but I don't have the time to add a new test to an already failing test suite. This PR is an improvement, and fixes things for people, but it's obviously your call if you'd like to merge it without a test. 

Thanks!

.changeset/nine-shoes-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled-base': patch
+---
+
+Fix TypeScript 3.8 required theme prop
```
Andarist:
```suggestion
Fixed issue with TypeScript 3.8 where `theme` was inferred to be required for a styled component.
```

aaronjensen:
Thank you for the merge!

PR #1792, opened 2020-03-09
Add disablePictureInPicture to prop-types

egdbear:
**What**: Add video prop-type

**Why**: https://github.com/styled-components/styled-components/issues/2988

### DIFF
--- a/packages/is-prop-valid/src/props.js
+++ b/packages/is-prop-valid/src/props.js
@@ -63,6 +63,7 @@ const props = {
   defer: true,
   dir: true,
   disabled: true,
+  disablePictureInPicture: true,
   download: true,
   draggable: true,
   encType: true,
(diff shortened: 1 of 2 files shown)


packages/is-prop-valid/src/props.js
```diff
@@ -471,7 +471,8 @@ const props = {
   // preact
   for: true,
   class: true,
-  autofocus: true
+  autofocus: true,
+  video: true
```
Andarist: I believe that you have wanted to add `disablePictureInPicture` and not the `video`. Also - you should move it to the "main" list of props so it doesn look like it's preact-specific prop.

egdbear: Sure! :)

Andarist:
Thanks!

Issue #1794, opened 2020-03-10
[babel-preset-css-prop] only css attr in jsx enabled css prop.


morlay:
**The problem**

babel-preset-css-prop will transform all files which has react component in whole project.
but not all components with css props，especially monorepo，some sub modules need to import `emotion jsx` but some are not.

**Proposed solution**

so we could enhance this preset，to import `emotion jsx` only css attr exists. 

```js
const r = (pkg) => {
  // eslint-disable-next-line @typescript-eslint/no-var-requires,no-undef
  const m = require(pkg);
  return m.default || m;
};

const autoImportEmotionJSXIFNeed = (api) => {
  const t = api.types;

  const pragmaName = "___EmotionJSX";
  const pragmaFragName = "___Fragment";

  const findLast = (arr, predicate) => {
    for (let i = arr.length - 1; i >= 0; i--) {
      if (predicate(arr[i])) {
        return arr[i];
      }
    }
  };

  const addPragmaImport = (path, opts) => {
    const importDeclar = t.importDeclaration(
      [t.importSpecifier(t.identifier(opts.import), t.identifier(opts.export || "default"))],
      t.stringLiteral(opts.module),
    );

    const targetPath = findLast(path.get("body"), (p) => p.isImportDeclaration());

    if (targetPath) {
      targetPath.insertAfter([importDeclar]);
    } else {
      // Apparently it's now safe to do this even if Program begins with directives.
      path.unshiftContainer("body", importDeclar);
    }
  };

  const createComponentBlock = (value) => ({
    type: "CommentBlock",
    value,
  });

  return {
    name: "auto-import-emotion-jsx-if-need",
    inherits: r("@babel/plugin-syntax-jsx"),
    visitor: {
      Program: {
        enter: function(path, state) {
          let needEmotion = false;

          // scan css prop to check if need to enable emotion jsx
          path.traverse({
            JSXAttribute(path) {
              if (path.get("name").node.name === "css") {
                needEmotion = true;
              }
            },
          });

          if (needEmotion) {
            const { file } = state;

            addPragmaImport(path, {
              module: "@emotion/core",
              export: "jsx",
              import: pragmaName,
            });

            addPragmaImport(path, {
              module: "react",
              export: "Fragment",
              import: pragmaFragName,
            });

            file.ast.comments.push(
              createComponentBlock(`@jsx ${pragmaName}`),
              createComponentBlock(`@jsxFrag ${pragmaFragName}`),
            );
          }
        },
      },
[...]

PR #1795, opened 2020-03-11
feat(babel-preset-css-prop): option cssPropOnly #1794

morlay:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/babel-preset-css-prop/__tests__/__fixtures__/index.without.css.js
+++ b/packages/babel-preset-css-prop/__tests__/__fixtures__/index.without.css.js
@@ -1,0 +1,9 @@
+import * as React from 'react'
+
+export let Button = props => {
+  return (
+    <>
+      <button {...props} />
+    </>
+  )
+}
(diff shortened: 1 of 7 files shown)


morlay:
@Andarist 

case like below may just notice in document. "when cssPropOnly enabled, don't use like this."
```
const props = { css: { color: 'hotpink' } }
;<div {...props}/>
```

Andarist:
I would like [this PR](https://github.com/babel/babel/pull/11154) to get merged into Babel before we proceed any further with this.

Andarist:
This is not something that we'd like to get in right now but is **maybe** something we'd consider in the future. For now, you have been the only user asking for it, so we don't consider this a high priority feature. We might revisit this when we upgrade to using those automatic JSX transforms introduced by Babel and React team. We are postponing working on this though as, for example, TypeScript has not yet implemented a similar feature.

morlay:
@Andarist 
it's ok. i have switch to my custom [babel plugin](https://github.com/querycap/webappkit/tree/master/%40querycap-ui/babel-preset-css-prop) to do this.

btw. how about the v11?

PR #1802, opened 2020-03-14
Fix issue with cx not being able to compose opaque styles

### LINKED ISSUE #1769: ClassNames does not appear consume classes passed to cx properly
**Current behavior:**
I've noticed that using the ClassNames component does not properly utilize the cx function the way it behaves in the vanilla emotion package. The documentation states that "[If you have used versions of Emotion prior to Emotion 10 or used vanilla Emotion, the css and cx functions work exactly like they do in those versions.](https://emotion.sh/docs/class-names)". However, this does not appear to work this way. Calling `cx(someEmotionClass, 'some-external-class')` returns a malformed class list like `class="name styles toString App"`.

**To reproduce:**

Here is a link to a simple codesandbox that uses css to create a style, then passes two classes: the emotion class and a string name. In vanilla emotion, this would output the classnames as expected

https://codesandbox.io/s/wonderful-wind-gr8nv

**Expected behavior:**

I was discussing with with @Andarist on slack and there's a little bit of confusion here as to how this behaves. In vanilla emotion, you can create a css class and call cx like `cx(someEmotionClass, 'some-external-class')` and the output would behave as expected. Since the documentation says it behaves the same here, I would expect to be able to call it here like this as well.

**Environment information:**

React version: 16.12.0
Emotion: @emotion/core 10.0.27

Where we left this is that this might be by design. If so, the documentation should provide a better example of how to call this here.

Andarist:
closes #1769

### DIFF
--- a/.changeset/early-boats-draw.md
+++ b/.changeset/early-boats-draw.md
@@ -1,0 +1,5 @@
+---
+'@emotion/utils': minor
+---
+
+`getRegisteredStyles` accept now an extra boolean argument telling it if it should retrieve registered styles or if it should just extract registered class name.
--- a/.changeset/pink-insects-yell.md
+++ b/.changeset/pink-insects-yell.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fixed issue with `cx` received inside `<ClassNames/>` not being able to compose styles created by `css` from the `@emotion/react` package.
--- a/packages/css/src/create-instance.js
+++ b/packages/css/src/create-instance.js
@@ -19,7 +19,8 @@ function merge(registered: Object, css: (*) => string, className: string) {
   const rawClassName = getRegisteredStyles(
     registered,
     registeredStyles,
-    className
+    className,
+    true
   )
 
   if (registeredStyles.length < 2) {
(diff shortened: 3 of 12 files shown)


Andarist:
In such a case the documentation should be tweaked because right now it states:
> If you have used versions of Emotion prior to Emotion 10 or used vanilla Emotion, the css and cx functions work exactly like they do in those versions.

And while it is true to some extent this is surprising that it was not able to compose regular React styles (so not exactly the same like in previous versions, people dont tend to think about those things as "string styles" and "opaque styles", conceptually they are just "styles").

The use case like this seems valid:
```js
import { css } from `@emotion/react`

;<ClassNames>
  {({ cx }) => <div className={cx(css`color:hotpink;`, 'other-cls')} />}
</ClassNames>
```
This can be "fixed" by calling received `css` on your styles, but both identifiers are the same, so it become quite cumbersome as you either have to hoist the styles or rename one of those identifiers.

At the very least we could introduce a dev warning telling people what to do if they misuse the API.

mitchellhamilton:
> At the very least we could introduce a dev warning telling people what to do if they misuse the API.

This sounds good to me.

Andarist:
Closed in favor of https://github.com/emotion-js/emotion/pull/1810

Issue #1744, opened 2020-01-29
[v11] @emotion/react types implicitly depend on @emotion/styled


billyjanitsch:
**Current behavior:**

The typings for `@emotion/react` [import from `@emotion/styled`](https://unpkg.com/browse/@emotion/react@11.0.0-next.11/types/theming.d.ts), which results in an error when typechecking if the latter is not installed:

```
node_modules/@emotion/react/types/theming.d.ts:12:8 - error TS2307: Cannot find module '@emotion/styled'.

12 } from '@emotion/styled'
          ~~~~~~~~~~~~~~~~~


Found 1 error.
```

**To reproduce:**

Install `@emotion/react` and run TS against the following snippet:

```ts
import {css} from '@emotion/react'
```

**Expected behavior:**

I would expect to be able to use `@emotion/react` without having `@emotion/styled` installed.

**Environment information:**

- `react` version: n/a
- `emotion` version: @emotion/react@11.0.0-next.11

Andarist:
Thanks for bringing this up - this definitely should get fixed before the final release.

JakeGinnivan:
There are dependencies the other direction too, from `@emotion/styled` to `@emotion/react` too.

What is the responsibility for each package, I can try and clean up the circular dependencies in the types once I have a clear idea of what should rely on what.

Andarist:
It seems that those were just some leftovers, not actually used in that file. I've removed them and tests pass - https://github.com/emotion-js/emotion/pull/1804 

@JakeGinnivan `@emotion/styled` can depend on `@emotion/react` as it's its peer dep. The other situation should not occur (`@emotion/react` should not depend on `@emotion/styled`). I've double-checked now and there is no such situation anymore, so everything seems goo.

PR #1804, opened 2020-03-14
Remove unnecessary @emotion/styled import from theming declaration file

Andarist:
closes https://github.com/emotion-js/emotion/issues/1744


JakeGinnivan:
Awesome, thanks @Andarist

PR #1805, opened 2020-03-14
Relax requirement for label extraction - containing functions dont have to be PascalCased anymore

Andarist:
It IMHO makes sense to relax this, otherwise styles from functions calling `css` dont receive labels.

### DIFF
--- a/.changeset/neat-geese-knock.md
+++ b/.changeset/neat-geese-knock.md
@@ -1,0 +1,5 @@
+---
+'@emotion/babel-plugin': minor
+---
+
+Requirements for a label extraction have been relaxed. In certain situations it was previously required for a containing function to have a PascalCased name.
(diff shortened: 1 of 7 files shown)


Andarist:
This only gives a label based on a function if there is no intermediate declaration, etc, in the ancestors' chain. Right now this gets a label:
```js
const foo = css`color: green;`
```
but this doesnt
```js
const foo = () => css`color: green;`
```
This is surprising for some (had a couple of questions regarding this on Slack and here on GitHub).

I don't see much difference between a "partial" styles (exhibit one) and a helper function creating such (exhibit two). Could you maybe prepare a case where you think this would become problematic?

Andarist:
@mitchellhamilton all "locals" are being sanitized when actually generating labels: https://github.com/emotion-js/emotion/blob/a267a9a578869219f9b967e1f5dd7de0d94c096e/packages/babel-plugin-emotion/src/utils/label.js#L16 , so it's already covered

Issue #1727, opened 2020-01-16
[jest.mock + Typescript] You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems.


bmvantunes:
**Current behaviour:**

Inside a unit test, after using jest.mock('./MyComponent') we get the error message added at #1677 

```sh
console.warn node_modules/@emotion/react/dist/react.cjs.dev.js:490
    You are loading @emotion/react when it is already loaded. Running multiple instances may cause problems. This can happen if multiple versions are used, or if multiple builds of the same version are used.
```

**To reproduce:**
Option 1 - Cloning:
```sh
git clone https://github.com/bmvantunes/emotion-v11-bug-jest.git
cd emotion-v11-bug-jest
npm test
```

Option 2 - Step by step after "npx create-react-app my-app --typescript":

1. Install @emotion/react and @emotion/styled@next
```sh
npm install --save @emotion/react @emotion/styled@next
```

2. Create a component "ComponentChild.tsx"
```typescript
import styled from '@emotion/styled';

export const ComponentChild = styled.div`
  background: red;
  height: 300px;
`;
```

3. Create a component "ComponentParent.tsx"
```typescript
import { ComponentChild } from "./ComponentChild";
import styled from "@emotion/styled";

export const ComponentParent = styled(ComponentChild)`
  background: blue;
`;
```
4. Change your "App.tsx" to be:
```typescript
import React from 'react';
import { ComponentParent } from './ComponentParent';
import { ComponentChild } from './ComponentChild';

const App: React.FC = () => {
  return (
    <>
      <ComponentParent />
      <ComponentChild />
      <p>Test</p>
    </>
  );
}

export default App;
```
5. Then in your "App.test.tsx":
```typescript
import React from 'react';
import { render } from '@testing-library/react';
import App from './App';

jest.mock('./ComponentParent');

test('renders learn react link', () => {
  const { getByText } = render(<App />);
  const test = getByText(/Test/i);
  expect(test).toBeInTheDocument();
});
```

6. Run test command
```sh
npm test
```

**Expected behaviour:**

No warning should be seen in the console

**Versions**
React - latest
typescript - 3.7.4
react: 16.12.0
@emotion/react  - 11.0.0-next.11
@emotion/styled - 11.0.0-next.11

Andarist:
Thanks for the report - I'm going to take a look at this later.

bmvantunes:
Thanks @Andarist

Andarist:
[...]

PR #1806, opened 2020-03-14
Do not warn about @emotion/react being loaded twice in Jest

Andarist:
closes https://github.com/emotion-js/emotion/issues/1727

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -9,15 +9,20 @@ export { default as css } from './css'
 
 if (process.env.NODE_ENV !== 'production') {
   const isBrowser = typeof document !== 'undefined'
-  const globalKey = '__EMOTION_REACT__'
-  const globalContext = isBrowser ? window : global
-  if (globalContext[globalKey]) {
-    console.warn(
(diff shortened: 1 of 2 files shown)


packages/react/src/index.js
```diff
@@ -9,15 +9,20 @@ export { default as css } from './css'
...
-  const globalContext = isBrowser ? window : global
-  if (globalContext[globalKey]) {
-    console.warn(
-      'You are loading @emotion/react when it is already loaded. Running ' +
-        'multiple instances may cause problems. This can happen if multiple ' +
-        'versions are used, or if multiple builds of the same version are ' +
-        'used.'
-    )
+  // #1727 for some reason Jest evaluates modules twice if some consuming module gets mocked with jest.mock
+  const isJest = typeof process !== 'undefined' && process.env.JEST_WORKER_ID
```
mitchellhamilton: Why not `typeof jest !== 'undefined'`? I'm afraid that using `process` in a webpack build would pull in a bunch of code

Andarist: This was suggested in the issue and I forgot that this can be simplified like this 😅 changed it right now to what you have suggested

PR #1810, opened 2020-03-16
Add a dev-only warning about opaque styles being passed to `cx`

Andarist:
closes https://github.com/emotion-js/emotion/pull/1802

### DIFF
--- a/packages/react/src/class-names.js
+++ b/packages/react/src/class-names.js
@@ -30,6 +30,16 @@ let classnames = (args: Array<ClassNameArg>): string => {
         if (Array.isArray(arg)) {
           toAdd = classnames(arg)
         } else {
+          if (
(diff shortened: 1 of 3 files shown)


packages/react/src/class-names.js
```diff
@@ -30,6 +30,15 @@ let classnames = (args: Array<ClassNameArg>): string => {
         if (Array.isArray(arg)) {
           toAdd = classnames(arg)
         } else {
+          if (
+            process.env.NODE_ENV !== 'production' &&
+            arg.styles !== undefined
```
mitchellhamilton: Could we check `name` as well just in case someone has a class name called `styles`?

Andarist: added

PR #1812, opened 2020-03-18
(styled-base) Changed default props.theme type to any

hasparus:
### What:

- **Changed default `props.theme` type to _any_**
- Updated dtslint and added `--expectOnly` flag
- Corrected stylis errors from ts-styled-plugin

### Why:

The [docs][theme-docs] were wrong.

#### Updated dtslint and added `--expectOnly` flag

Not sure if anybody cares about the string coercion warnings which were already here.

Then, only following error remains:
```
background-color: ${props => props.theme.backColor};
```
```
ERROR: 122:44  expect  TypeScript@next compile error: 
Property 'backColor' does not exist on type 'object'.
```

#### Changed default `props.theme` type to _any_ 

It stems from *CreateStyledComponentBaseThemeless* defaulting Theme to `*object*.

https://github.com/emotion-js/emotion/blob/ef7794f50a1ef9790ea6ebe530b6fd8e0b7b0942/packages/styled-base/types/index.d.ts#L57-L77

*CreateStyledComponentBaseThemeless* is used when the theme isn't given to *CreateStyled*.

https://github.com/emotion-js/emotion/blob/ef7794f50a1ef9790ea6ebe530b6fd8e0b7b0942/packages/styled-base/types/index.d.ts#L103-L115

The [docs][theme-docs] claim that props.theme is _any_ and the type definition also suggests so.
```
interface CreateStyled<Theme extends object = any>
```

[theme-docs]: https://emotion.sh/docs/typescript#define-a-theme

However, I encountered *object*, both in **styled-base** and **styled** packages.

ThemeContext is [defined][theme-context] as _React.Context\<Object\>_ in **@emotion/core** flow source, but in TypeScript _any_ or _Record<string, any>_ would be more suitable. 

I'm changing it to _any_ to match the docs.

[theme-context]: https://github.com/emotion-js/emotion/blob/10514a8635dcaa55b85c7bff90e2a9e14d1ba61f/packages/core/src/context.js#L17

#### Corrected stylis errors from ts-styled-plugin

Fixed problems:
  - camelCase properties instead of kebab-case
  - missing semis

### Checklist ###:

- [ ] Documentation N/A
- [x] Tests (I fixed an error in existing test)
- [x] Code complete 
- [x] Changeset

### DIFF
--- a/.changeset/eight-feet-hide.md
+++ b/.changeset/eight-feet-hide.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled-base': patch
+---
+
+Default props.theme to any to match the docs
--- a/packages/styled-base/package.json
+++ b/packages/styled-base/package.json
@@ -12,7 +12,7 @@
   "license": "MIT",
   "repository": "https://github.com/emotion-js/emotion/tree/master/packages/styled-base",
   "scripts": {
-    "test:typescript": "dtslint types"
+    "test:typescript": "dtslint --expectOnly types"
   },
   "dependencies": {
     "@babel/runtime": "^7.5.5",
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@emotion/core": "^10.0.28",
     "@types/react": "^16.8.20",
-    "dtslint": "^0.3.0",
+    "dtslint": "^3.3.0",
     "react": "^16.5.2"
   },
   "peerDependencies": {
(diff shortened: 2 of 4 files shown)


Andarist:
Sorry for the lack of an answer so far. We didn't get many reports regarding this, so hopefully, people have managed to work with this, unfortunate, discrepancy. I'm deeply sorry for the confusion and problems this has introduced. 

At this point in time, I don't think it's worth fixing though as we are getting close to releasing v11 which comes with the complete types overhaul. I hope you understand.

PR #1817, opened 2020-03-20
Stylis v4

Andarist:
This is **not** ready by any means - just an initial attempt at using new Stylis.

It looks really promising, when preparing this I've found some minor problems with Stylis v4, for most of them I've already opened PRs against Stylis. The biggest unsolved problem right now is prefixing by adding a completely new rule (adding webkit keyframes) **in combination with** rulesheet plugin. Truth to be told - I'm not even sure if we really need this, but from Stylis perspective it should be possible to do this nevertheless.

cc @mitchellhamilton


packages/cache/src/index.js
```diff
@@ -115,46 +115,59 @@ let createCache = (options: Options): EmotionCache => {
   ) => string | void
 
   if (isBrowser) {
-    stylis.use(options.stylisPlugins)(ruleSheet)
+    let currentSheet
+    const finalizer = middleware(
```
Andarist: seems to me that we only need to accept a list of plugins and if one wants to use custom ones they should provide our default plugins (compat and prefixer) back to us as well - this solves the ordering problem, we don't need to decide anything regarding this

OTOH removeLabel + stringify + rulesheet are essential and they should always be given as last and in that order, so I think it's safe to "hardcode" them here

packages/cache/src/index.js
```diff
@@ -201,62 +214,33 @@ let createCache = (options: Options): EmotionCache => {
...
-                )
-              }
-            })
-          }
-
-          break
-        }
-      }
-    })
+    // stylis.use((context, content, selectors) => {
```
Andarist: this should be easy to bring back, I just hadn't yet time to do so

packages/cache/src/index.js
```diff
@@ -201,62 +214,33 @@ let createCache = (options: Options): EmotionCache => {
...
-        case -1: {
-          while (commentStart.test(content)) {
-            commentEnd.lastIndex = commentStart.lastIndex
-
-            if (commentEnd.test(content)) {
-              commentStart.lastIndex = commentEnd.lastIndex
-              continue
-            }
-
-            throw new Error(
```
Andarist: this is not needed as this doesnt cause an infinite loop in Stylis anymore

packages/css/test/__snapshots__/css.test.js.snap
```diff
@@ -83,8 +84,8 @@ exports[`css boolean as value 1`] = `
 exports[`css composition stuff 1`] = `
 .emotion-0 {
   -webkit-box-pack: center;
-  -webkit-justify-content: center;
   -ms-flex-pack: center;
+  -webkit-justify-content: center;
```
Andarist: the order of those properties has changed (its how they are generated now), I dont think it matters though

packages/css/test/__snapshots__/css.test.js.snap
```diff
@@ -129,7 +130,7 @@ exports[`css composition with objects 1`] = `
   color: red;
 }
 
-@media (min-width:420px) {
+@media (min-width: 420px) {
```
Andarist: the output regarding some whitespaces has changed, this might be visible in a bunch of updated snapshots

packages/utils/src/index.js
```diff
@@ -47,7 +47,7 @@ export const insertStyles = (
     let current = serialized
     do {
       let maybeStyles = cache.insert(
-        `.${className}`,
+        serialized === current ? `.${className}` : '',
```
Andarist: the alternate is for keyframes (need to insert them without scope

packages/styled/test/__snapshots__/prop-filtering.test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`custom shouldForwardProp works 1`] = `
 .emotion-0,
-.emotion-0 * {
+.emotion-0* {
```
Andarist: a bug in stylis, I've a pending PR with a fix for this

packages/css/test/instance/__snapshots__/css.test.js.snap
```diff
@@ -35,8 +35,6 @@ exports[`css array with explicit true 1`] = `
 exports[`css auto px 1`] = `
 .emotion-0 {
   display: flex;
-  -webkit-flex: 1;
```
Andarist: those disappeared because this file is testing a custom instance and I have not added a prefixer to it

packages/jest/test/printer.test.js
```diff
@@ -149,18 +149,6 @@ header .emotion-0 {
   })
 })
 
-test('throws nice error for invalid css', () => {
-  const tree = renderer.create(<div css={css`jnnjvh@'jevhevhb`} />).toJSON()
```
Andarist: stylis produces an empty rule for this, so this invalid stuff gets lost and we do not have an easy way to throw a nice error for this

packages/react/__tests__/__snapshots__/global.js.snap
```diff
@@ -18,25 +18,19 @@ exports[`basic 1`] = `
     data-emotion="css-global"
   >
     
-    @font-face{font-family:some-name;}
-  </style>
-  <style
-    data-emotion="css-global"
-  >
-    
-    @import url('something.com/file.css');
```
Andarist: have a pending fix in stylis for `@import` rules not being passed to the rulesheet plugin

packages/cache/src/stylis-plugins.js
```diff
@@ -1,89 +1,47 @@
...
-  if (
-    context === 1 &&
-    // charcode for l
-    content.charCodeAt(0) === 108 &&
-    // charcode for b
-    content.charCodeAt(2) === 98
-    // this ignores label
-  ) {
-    return ''
+export let createUnsafeSelectorsAlarm = cache => (element, index, children) => {
```
mitchellhamilton: Could you check that this is dead code eliminated in prod?

packages/jest/test/printer.test.js
mitchellhamilton: This error existed because the css parser we use in @emotion/jest threw a useless error on invalid rules so as long as there's no useless errors, I think it's fine

packages/react/__tests__/__snapshots__/server.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`ssr @import 1`] = `
 <div class="css-1lrxbo5">
 </div>
 <style data-emotion="css-global 1sh9nmh">
-  @import url('https://some-url');h1{color:hotpink;}
+  h1{color:hotpink;}@import url('https://some-url');
```
mitchellhamilton: `@import` has to be the first rule(or at least before non `@import` rules) when using insertRule, could you check if this not being first here is fine?

packages/stylis/src/index.js
```diff
@@ -1,2 +1,1215 @@
...
+        var pattern = ''
+        while (current < next - 1) {
+          pattern += tokens[current++]
+        }
+        props[i] = replace(prop, ' ' + pattern, pattern)
+      }
+  }
+}
+
+export {
```
mitchellhamilton: Do we really want to export all of this?

packages/stylis/src/index.js
```diff
@@ -1,2 +1,1215 @@
-// @flow
-export { default } from './stylis.min'
+/* eslint-disable */
```
mitchellhamilton: Could you try running this through closure compiler and see if it has any size wins?

packages/cache/src/index.js
```diff
@@ -42,11 +47,11 @@ let getServerStylisCache = isBrowser
       }
     })
 
+const defaultStylisPlugins = [compat, prefixer]
```
mitchellhamilton: Why is compat not an omnipresent plugin?

packages/cache/src/index.js
```diff
@@ -1,30 +1,35 @@
...
 export type PrefixOption =
   | boolean
   | ((key: string, value: string, context: 1 | 2 | 3) => boolean)
-type StylisPlugins = StylisPlugin[] | StylisPlugin
 
 export type Options = {
   nonce?: string,
-  stylisPlugins?: StylisPlugins,
+  stylisPlugins?: StylisPlugin[],
   prefix?: PrefixOption,
```
mitchellhamilton: Am I correct in thinking that the prefix option is gone? And people will be able to pass in their own prefixer into plugins?

packages/cache/src/stylis-plugins.js
Andarist: sure, I'll recheck this later, but i don't see any reason why it wouldnt:
1. it's a declaration (side-effect freeeee)
2. it's called conditionally

packages/react/__tests__/__snapshots__/server.js.snap
Andarist: It probably is not - I've changed tests to always put `@import` first: https://github.com/emotion-js/emotion/pull/1817/commits/2d95e2c3c349e0d2c1d352f0c0fb2fd8fd02c2ee

Stylis v3 was special-casing this and it actually has moved import rules to the first position when compiling, v4 doesn't do that anymore. It seems to me that it's a valid approach to just require from the user to put those first "manually". Special-casing this would require from us writing a special stylis plugin for the server path - IMHO it's not worth it. 

If we decide this being a requirement on the consumer side then this special case could be dropped I think: https://github.com/emotion-js/emotion/blob/ff9746b8d6f5e8dca89bc9a7340a90b91240800b/packages/sheet/src/index.js#L118
`@import` rules should not occur in the scoped styles and for global ones they should just be put first. Keep also in mind that `@charset` should have even higher priority than `@import` and we do not special case it anyway - it's way less common though.

packages/stylis/src/index.js
Andarist: Nah, it's just a copy paste from the stylis itself right now as v4 is not yet published to npm. I think it should be fine to just use the stylis itself and make `@emotion/stylis` obsolete.

packages/stylis/src/index.js
Andarist: I've actually already tried this, the output for ESM build was slightly higher, UMD had like 100b in favour of closure.

I've also compared output and the main difference (for ESM) was:
- closure have inlined "enum" strings
- closure have inlined 2 small helpers responsible for reading string and array lengths

Other than that there was just some changes to the minified code, slightly different structure of loops etc but nothing substantial really.

packages/cache/src/index.js
Andarist: it's up for discussion, main reason is that omnipresent ones come last currently and this probably should come first. I see omnipresent as really essentials and compat is not that - using a custom stylis plugins is already a really advanced use case and it seems that greater control over plugin ordering is good, by adding `compat` manually we have to decide about its order.

packages/cache/src/index.js
Andarist: > Am I correct in thinking that the prefix option is gone? And people will be able to pass in their own prefixer into plugins?

Yes, that's my current thinking and the way to customize prefixer - this includes removing it or changing what gets prefixed. You don't pass any special options, you just customize plugins however you want.

Andarist: This is also a good time to reevaluate what browsers should we support in the builtin prefixer. We could save some bytes both from the bundlesize (less prefixing logic) and from SSRed styles

Andarist:
I've rewritten [the compat plugin](https://github.com/emotion-js/emotion/blob/174395ac3a59b9bc7cd0423bd2412b29d858cf9e/packages/cache/src/stylis-plugins.js#L5-L54) and moved it to this repo - should cover all cases now (our docs are built correctly now).

I'm going to include tests later - have actually written them but have them on the branch in the stylis repo on the branch so i need to port them here.

thysultan:
@Andarist Is there any reason you aren't using `element.props` and instead re-tokenizing `element.value` in https://github.com/emotion-js/emotion/blob/174395ac3a59b9bc7cd0423bd2412b29d858cf9e/packages/cache/src/stylis-plugins.js#L5-L54 as `element.props` already has comma separated selectors tokenized. `element.props`?

packages/stylis/src/index.js
Andarist: cc @thysultan - this might slightly interest you, not sure if you have seen this comment and my "findings"

also - an additional note as I've just remembered one more thing observed, the closure compiler was able to minify 2 properties of the node structure (dont recall which ones), which is not correct from our perspective as node's shape is stylis public API, this probably could be prevented somehow, but im not too familiar with the closure compiler, so this would require some research on my part

but as mentioned - i haven't really observed any substantial changes between terser and closure compiler outputs, so not sure if it's worth looking into using it. Keep in mind that I've really looked quickly through those outputs and haven't done extensive analysis

Andarist:
@thysultan `element.props` are not tokenized, they are computed selectors. For an input like this:
```js
compile('.user{.a,.b,:focus{color:hotpink;}}')
```
we get such `element.props`
```js
[".user .a", ".user .b", ".user :focus"]
```
and based on this alone it's hard to determine what was the original input as it could have been this:
```js
compile('.user{.a,.b,& :focus{color:hotpink;}}')
```
which produces the same `element.props` but we shouldn't adjust the output in this case.

thysultan:
@Andarist What is the compat plugin meant to achieve? Is it just meant to bring back the old orphaned `:focus` behaviour?

packages/css/test/__snapshots__/selectivity.test.js.snap
```diff
@@ -123,3 +123,45 @@ exports[`css media query with nested selector without declarations on root 1`] =
...
+.css-b4zt77 .qwe span,
+.css-b4zt77 .foo:last-child,
+.css-b4zt77 .bar div:last-child,
+.css-b4zt77 .qwe:last-child {
+  color: hotpink;
+}"
+`;
+
+exports[`orphanated pseudos overlapping 1`] = `
+".css-1pl3kvr:first-child:first-child {
```
Andarist: this is actually **not** a correct output, it has surfaced a bug in:
https://github.com/emotion-js/emotion/blob/4ed12cb22bbc08fd8c131446d37eb46bac798b9a/packages/babel-plugin/src/utils/minify-utils.js#L45

the correct output is:
```css
.css-h4sh :first-child:first-child {
  color: hotpink;
}
```

Andarist:
@thysultan yes, I realize that the current implementation looks overly complex but I was not able so far to figure out a simpler solution. Take a look at tests I've pushed out: https://github.com/emotion-js/emotion/pull/1817/commits/87bd0bacd59f98a9ac90be261819323a2b6877cd

I've also pushed them out in the stylis repo if you want to take a look there:
https://github.com/thysultan/stylis.js/tree/compat-plugin

packages/cache/src/stylis-plugins.js
```diff
@@ -1,89 +1,100 @@
...
+      var value = element.value
+      var propsLength = props.length
 
-function toSheet(block) {
-  if (block) {
-    Sheet.current.insert(block + '}')
-  }
-}
+      // short-circuit for the simplest case
+      if (propsLength === 1 && value.charCodeAt(0) !== 58 /* colon */) {
```
thysultan: This will not handle this variant `h1,:hover` assuming you want to support that?

packages/cache/src/stylis-plugins.js
```diff
@@ -1,89 +1,100 @@
...
-        }
-        // charcode for l
-        case 108: {
-          // charcode for b
-          // this ignores label
-          if (content.charCodeAt(2) === 98) {
-            return ''
-          }
+      for (; i < valueTokens.length; i++) {
+        if (valueTokens[i].charCodeAt(0) === 44 /* , */) {
```
thysultan: Commas in css functions i.e `:not(h1,h2)`, aren't handled here, you might be better of just re-passing `element.value` to `compile` and extract  it i.e `var props = compile(element.value+'{}')[0].props` if you are re-parsing it all over again here.

```js
compile(element.value+'{}')[0].props.map(value => {
if (value.charAt(0) === ':')
return // handle your psuedo selector
})
```

I've come up with a short function for this

```js
var compat = function ({type, value, props} => {
if (type === 'rule')
	if (~value.indexOf(':'))
		for (var i = 0; i < props.length; i++);
		    props[i] = props[i].replace(' :', ':')
})
```

But it doesn't yet handle the `.xyz{& first-child{}}` case.

packages/cache/src/stylis-plugins.js
Andarist: yes, this is only intended as a short-circuit for the most common case, this line is not relevant for the algorithm to actually work

packages/cache/src/stylis-plugins.js
Andarist: > Commas in css functions i.e :not(h1,h2)

Are u sure? parens are actually grouped by `tokenize`, so we get this:
```js
tokenize('.foo :not(h1,h2)') // [".foo", " ", ":", "not", "(h1,h2)"]
```
so I don't encounter a comma token at all for this case - a comma in a css function won't be accounted as a "point" (rightfully so).

> you might be better of just re-passing element.value to compile and extract it i.e var props = compile(element.value+'{}')[0].props 

as far as I can tell this cannot be done so easily because:
- i can't just override props like this as original props have already full computed selectors and value holds only the "current level" stuff, so doing it like this i would remove selector parts from the "parent levels" (those from lower depths of the input)
- i still have no notion of `&` is at the "current level" and need to extract it from the props anyhow

> I've come up with a short function for this compat plugin

This doesn't handle `div,:focus` case - a colon doesn't have to be first in the value, it has to be first in the "group" (`:focus` being a second group here)

It would also fail on `:first-child,div :last-child{}` as it would remove the space from the second group.

> But it doesn't yet handle the .xyz{& first-child{}} case.

Well, that's exactly that pesky case I've tried to handle through all of this overcomplicated stuff 😅

Andarist:
I was thinking about the compat plugin a lot in the past days. I think it should be fine to just change the behavior regarding this. I know it's a breaking change, **but**:
- most of the docs (ours and styled-components) already use explicit `&`
- I've asked several people which form they use and the answer was **always** that they use explicit `&`
- it matches the behavior of popular CSS preprocessors like Sass, Less, etc
- we wouldn't have to do hoops to bring back the old behavior and currently it's hard to implement such compact plugin to handle all the possible cases. It also means that the bundlesize would not be increased by the potential compact plugin and perf would be slightly better as there would be no extra work to be done for those cases
- we can implement ESLint rule that would warn people about this changed behavior for most cases **and** we can implement a dev-only plugin that would warn about this for **all** cases and which would suggest using explicit `&:focus` or `& :focus`

Thoughts?

mitchellhamilton:
I'm quite strongly of the opinion that we shouldn't change this. I personally always use the `&`less form. Changing it feels like churn that doesn't really give users a benefit tbh.

> It also means that the bundlesize would not be increased by the potential compact plugin and perf would be slightly better as there would be no extra work to be done for those cases

How big of a size and perf difference would this be?

---

Aside: Does this also affect cases like this or no?

```jsx
{
  '.class': styles
}
```

Andarist:
> How big of a size and perf difference would this be?

Probably not **that** much. It's a secondary concern though, right now I'm unable to cover all cases - not without forking Stylis, which I'd rather like to avoid because this single change doesn't seem worth it in my opinion. 

> Aside: Does this also affect cases like this or no? `{'.class': styles}`

No, it's exclusively about pseudo selectors - see the prior discussion [here](https://github.com/thysultan/stylis.js/issues/64) where people mention that they have actually thought this is a bug (and imho rightfully so as `&` convention was established long before css-in-js and how it behaves has been set in people minds by its predecessors).

packages/react/__tests__/__snapshots__/server.js.snap
mitchellhamilton: > Keep also in mind that `@charset` should have even higher priority than `@import` and we do not special case it anyway - it's way less common though.

Yeah, I'm not super worried about `@charset` given I've never seen anyone use it.

Okay - I think I'm fine with requiring `@import` be first and only be in `Global`. We definitely need a better error than what we would have if we just removed the isImportRule stuff though.

packages/cache/src/index.js
mitchellhamilton: Happy to re-evaluate what prefixes we support by default. (and ideally we could offer a prefixer that's equivalent for anyone who cares about those browsers)

packages/css/test/__snapshots__/selectivity.test.js.snap
Andarist: this has been fixed by https://github.com/emotion-js/emotion/pull/1836

packages/cache/src/index.js
Andarist: > Happy to re-evaluate what prefixes we support by default.

Do you have any possible criteria in mind? I would be in favor of supporting flex prefixing for IE11 and some reasonable market share criteria (1%?) for everything else.

> (and ideally we could offer a prefixer that's equivalent for anyone who cares about those browsers)

This is basically what the currently included prefixer does - it's supporting quite a bit of properties. I think it actually supports a little bit more than what was included in Stylis v3.

mitchellhamilton: > Do you have any possible criteria in mind? I would be in favor of supporting flex prefixing for IE11 and some reasonable market share criteria (1%?) for everything else.

That sounds good to me, though maybe create an issue and tweet about it to get some more perspectives?

packages/cache/src/index.js
Andarist: Ok, I've pushed around some bits and here you go: https://github.com/emotion-js/emotion/pull/1817/commits/48629c723c59d6203694875c3760c8cfab2cc65f

packages/react/__tests__/__snapshots__/server.js.snap
Andarist: Ok, I've pushed out changes around this - together with new dev-only warnings for incorrect inputs.

packages/sheet/__tests__/index.js
```diff
@@ -100,18 +102,6 @@ describe('StyleSheet', () => {
     document.body.removeChild(container)
   })
 
-  it('should not throw an error when inserting a @import rule in speedy when a rule has already been inserted', () => {
```
Andarist, line 103: as this is not special-cased now this test became obsolete, I also couldn't convert it to a test which would check if we have warned correctly on non-leading imports as cssom package doesn't implement throwing for this case:
https://github.com/NV/CSSOM/blob/d600816ea9442c5e33f27ff59de536d7b7ccd239/lib/CSSStyleSheet.js#L39-L47

Andarist:
@mitchellhamilton I've made some minor changes, removed outdated docs and added changesets. Please check out the latest changes - I'd like to make sure this is ready to be merged (besides waiting for the stylis v4 being actually released).

packages/sheet/src/index.js
```diff
@@ -111,30 +111,22 @@ export class StyleSheet {
...
       } catch (e) {
         if (process.env.NODE_ENV !== 'production') {
           console.error(
             `There was a problem inserting the following rule: "${rule}"`,
             e
           )
+          let isImportRule =
+            rule.charCodeAt(0) === 64 && rule.charCodeAt(1) === 105
+          if (isImportRule) {
+            console.error(
```
mitchellhamilton, line 126: This should throw

Andarist:
> Just realized a problem with the @import thing. It will mean that this will be broken:

Good catch, I totally have missed this breaking the vanilla emotion.

I like the idea of making each `injectGlobal` having its own style tag (it would nicely match how `<Global/>` behaves) but when I think about it then it seems to complicate the APIs around flushing and `.sheet` access (although I'm not completely sure why `.sheet` is available directly). As to flushing - we could maintain a list of sheets inside `@emotion/css` to flush them all? I have also thought about how this would play with rehydration etc but at the first glance this doesn't seem to be problematic. From what I understand calling `hydrate` is required anyway and it inserts ids into a cache, not sheets, so everything is fine.

mitchellhamilton:
> calling hydrate is required anyway and it inserts ids into a cache, not sheets, so everything is fine.

This isn't true, the same auto-hydration that happens for `@emotion/react` happens here.

---

So I just ran this code

```jsx
const { injectGlobal } = require("emotion");
const { renderStylesToString } = require("emotion-server");

injectGlobal`.thing {display:flex;}`;

injectGlobal`@import 'custom.css';`;

console.log(renderStylesToString(""));
```

It outputs

```html
<style data-emotion-css="njnz1q mndbu7">.thing{display:-webkit-box;display:-webkit-flex;display:-ms-flexbox;display:flex;}@import 'custom.css';</style>
```


So I'm kind of just inclined to say "@imports must only happen in the first injectGlobal call in your app" and then all of our problems are solved right? As long as we error with a nice message when this happens, I'm okay with it.

Andarist:
> This isn't true, the same auto-hydration that happens for @emotion/react happens here.

You are right, it seems like `hydrate` is not needed at all then?

> So I'm kind of just inclined to say "@imports must only happen in the first injectGlobal call in your app" and then all of our problems are solved right? 

If we assume that `@import` cant be lazily inserted during a page's lifetime - then yeah, there is no problem. 

> As long as we error with a nice message when this happens, I'm okay with it.

Okay - gonna improve this error a little bit tomorrow:
https://github.com/emotion-js/emotion/blob/5e39d81bf9177e68e78d23bd869c1592341d9768/packages/sheet/src/index.js#L127

mitchellhamilton:
> You are right, it seems like hydrate is not needed at all then?

https://github.com/emotion-js/emotion/pull/1876#discussion_r428350808

Andarist:
@mitchellhamilton I've improved the errors logged on incorrect `@import` usage in the latest commit: https://github.com/emotion-js/emotion/pull/1817/commits/a1d71730b1b6195730d242379d16ea4715197024 . Please take a look if this looks good to you. We also have other, somewhat lighter/more-specific warnings concerned with a single stylis call which you can check out in action here: https://github.com/emotion-js/emotion/blob/a1d71730b1b6195730d242379d16ea4715197024/packages/react/__tests__/warnings.js#L256-L311 , this new commit improves a situation when `@import` gets inserted incorrectly between stylis calls. Keep also in mind that this mostly affects vanilla Emotion users as `@emotion/react` scopes `<Global/>` styles to separate `<style/>` elements.

I believe this was the last remaining item here so if you approve this then we should finally be ready to merge this 🎉

packages/sheet/src/index.js
```diff
@@ -108,27 +108,31 @@ export class StyleSheet {
...
+        // but we don't want enabling speedy to affect the observable behavior
+        // so we report this error at all times
+        console.error(
+          `You have attempted inserting such rule:\n` +
+            rule +
+            '\n\n`@import` rules must precede all other types of rules in a stylesheet. Please ensure that you insert them first.'
+        )
+      }
+
+      ;(this: any)._alreadyInsertedOrderInsensitiveRule =
```
mitchellhamilton, line 126: Not really a blocker but why not make this an optional property on the class?

packages/sheet/src/index.js
```diff
@@ -108,27 +108,31 @@ export class StyleSheet {
...
+        rule.charCodeAt(0) === 64 && rule.charCodeAt(1) === 105
+
+      if (isImportRule && (this: any)._alreadyInsertedOrderInsensitiveRule) {
+        // this would only cause problem in speedy mode
+        // but we don't want enabling speedy to affect the observable behavior
+        // so we report this error at all times
+        console.error(
+          `You have attempted inserting such rule:\n` +
+            rule +
+            '\n\n`@import` rules must precede all other types of rules in a stylesheet. Please ensure that you insert them first.'
```
mitchellhamilton, line 122:
```suggestion
          `You're attempting to insert the following rule:\n` +
            rule +
            '\n\n`@import` rules must be before all other types of rules in a stylesheet but other rules have already been inserted. Please ensure that `@import` rules are before all other rules.'
```

.changeset/warm-ties-drop.md
```diff
@@ -0,0 +1,14 @@
...
+
+- plugins written for the former Stylis v3 are not compatible with the new version. To learn more on how to write a plugin for Stylis v4 you can check out its [README](https://github.com/thysultan/stylis.js#middleware) and the source code of core plugins.
+- vendor-prefixing was previously customizable using `prefix` option. This was always limited to turning off all of some of the prefixes as all available prefixes were on by default. The `prefix` option is gone and to customize which prefixes are applied you need to fork (copy-paste) the prefixer plugin and adjust it to your needs. While this being somewhat more problematic to setup at first we believe that the vast majority of users were not customizing this anyway. By not including the possibility to customize this through an extra option the final solution is more performant because there is no extra overhead of checking if a particular property should be prefixed or not.
+- Prefixer is now just a plugin which happens to be put in default `stylisPlugins`. If you plan to use custom `stylisPlugins` and you want to have your styles prefixed automatically you must include prefixer in your custom `stylisPlugins`.
```
mitchellhamilton, line 13: You should tell users where to get the default prefixer from here.

.changeset/warm-ties-drop.md
```diff
@@ -0,0 +1,14 @@
...
+- `@import` rules are no longer special-cased. The responsibility to put them first has been moved to styles author. They also can't be nested within other rules now - this excludes them being usable within `css`. It's only possible to write them at the top level of global styles.
```
mitchellhamilton, line 14:
```suggestion
- `@import` rules are no longer special-cased. The responsibility to put them first has been moved to the author of the styles. They also can't be nested within other rules now - this means they are not usable within `css`. It's only possible to write them at the top level of global styles.
```

> this means they are not usable within `css`.

This is wrong, no?

```jsx
<Global styles={css`@import whatever it is`} />
```

packages/cache/src/stylis-plugins.js
```diff
@@ -1,89 +1,170 @@
...
+}
+
+export let incorrectImportAlarm = (element, index, children) => {
+  if (!isImportRule(element)) {
+    return
+  }
+
+  if (element.parent) {
+    console.error(
+      "`@import` rules can't be nested inside other rules. Please move it to the top and put it before regular rules. Keep in mind that they can only be used within global styles."
```
mitchellhamilton, line 161:
```suggestion
      "`@import` rules can't be nested inside other rules. Please move it to the top level and put it before regular rules. Keep in mind that they can only be used within global styles."
```

packages/cache/src/stylis-plugins.js
```diff
@@ -1,89 +1,170 @@
...
+    )
+    nullifyElement(element)
+  } else if (isPrependedWithRegularRules(index, children)) {
+    console.error(
+      "`@import` rules can't be preceded by regular rules. Please put it before them."
```
mitchellhamilton, line 166:
```suggestion
      "`@import` rules can't be after other rules. Please put your `@import` rules before your other rules."
```

packages/sheet/src/index.js
Andarist: Just a minor difference - an optional property would be visible as part of the public type and we don't need it to be there.

.changeset/warm-ties-drop.md
Andarist: Right, was more thinking about vanilla Emotion where this would not be possible, but you are right - `css` from `@emorion/react` can be fed to `<Global/>` so this wasn't true. I've removed this part of a changeset.

.changeset/warm-ties-drop.md
Andarist: Added such mention just now.

Andarist:
> I assume you're gonna update docs in another PR or something?

Do you have anything in mind that needs to be documented besides vendor-prefixing stuff covered by those 2nd and 3rd points of this changeset:
https://github.com/emotion-js/emotion/blob/703b1ddb2b754002ef1c5c98ad522edfe60191d4/.changeset/warm-ties-drop.md#L1

mitchellhamilton:
Stuff like https://github.com/emotion-js/emotion/blob/master/docs/cache-provider.mdx, https://github.com/emotion-js/emotion/tree/master/packages/cache#options and potentially anything else that exists.

zenininja:
Stylis upgrade seems to have broken the React Native compatibility :(. Please check here https://github.com/emotion-js/emotion/issues/1986

PR #1825, opened 2020-03-27
Add toHaveStyleRule support for React Native

### LINKED ISSUE #1790: document is not defined when using jest-enzyme with React Native
**Current behavior:**

Asserting `toHaveStyleRule` on a `shallow` rendered React Native component will fail with error `ReferenceError: document is not defined` 

**To reproduce:**

1. Create component:
```jsx
import React from 'react';

  height: 100px;
  width: 100px;
`;

const MyComponent = () => <Container />;

export default MyComponent;
export { Container };
```
2. create test:
```jsx
import React from 'react';
import { matchers } from 'jest-emotion';
import { shallow } from 'enzyme'
import MyComponent, { Container } from './MyComponent';

expect.extend(matchers);

describe('<MyComponent />', () => {
  it('should have height 100', () => {
    const wrapper = shallow(<MyComponent />);
    expect(wrapper.find(Container)).toHaveStyleRule('height', '100px');
  });
});
```

**Expected behavior:**

`toHaveStyleRule` should pass.

**Environment information:**

- `react` version: `16.9.0`
- `react-native` version: `0.61.5`
- `@emotion/core` version: `^10.0.28`
- `@emotion/native` version: `^10.0.27`
- `jest-emotion` version: `^10.0.27`

MrCreeper1008:
Closes #1790

**What**:  
Currently, `toHaveStyleRule` is having issue with React native.

**Why**:
React Native components do not have `className` prop which this rule relies on to perform checks. Instead, they have a `style` prop which takes a style object. However, when `@emotion/native` is used with React Native components, the original components are nested 3 levels deep, making the `style` prop inaccessible unless we `dive`.

**How**:
By `dive`ing into the received component three times, the `style` prop is exposed, which allows us to perform checks. However, style checking for React Native components is implemented differently, so I added a `native` option. To use,
```javascript
expect(ReactNativeComponent).toHaveStyleRule('flexDirection, 'row', { native: true });
```
Alternatively, we can create a native version of `jest-emotion`:
```javascript
import { matchers } from 'jest-emotion/native';
```

**Checklist**:
- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [x] Changeset 

**Caveats**
- This fix is mainly geared towards components shallowly rendered with `enzyme`, so I have no idea if this would work with other testing libraries like `native-testing-library`.
- Also, I have no idea how to test this. (Do we install `react-native` just to get a couple components for us to test with?)  

I would appreciate any feedback given to this PR!

Edit: Add usage details.  
Edit 2: Link to issue


Andarist:
Sorry for the late reply, are you still interested in working on this? If yes then I think we'd like:
- rebase this on top of `next` branch
- refactor to a separate entry point like `/native`. A problem is that there is already `/enzyme` entry point there, so maybe this should actually be a nested one, like `/enzyme/native`?

packages/jest-emotion/src/matchers.js
```diff
@@ -42,8 +42,64 @@ function toHaveStyleRule(
...
+    if (typeof received.props === 'function') {
+      // the received component is rendered by enzyme shallow()
+      style = received
+        .dive()
+        .dive()
+        .dive()
+        .prop('style')
+    } else {
+      // the received component is rendered by native-testing-library
+      style = received.props.style
```
Andarist, line 60: is this guaranteed to be an array or an object? I recall RN styles being sometimes referenced by integers (produced by RN's `StyleSheet`), I'm not sure about this as I have only played a little bit with RN and I don't have any actual experience with it.

MrCreeper1008: I will double check. Thanks for the review!

MrCreeper1008:
Thanks for the review. I’ll work on this.

MrCreeper1008:
Moving this to #1906

PR #1836, opened 2020-04-05
Fix tagged templates minifier

Andarist:
Just how [HTML should not be parsed with RegExp](https://stackoverflow.com/a/1732454), CSS should not be either 🤪

By using a proper CSS parser we can write context-based minifying rules to avoid some pitfalls, the code also becomes much more readable.

I'm using a forked Stylis here to achieve the goal because I don't want to expand the input string to proper~ CSS nodes, but rather maintain hierarchy so I can serialize it back to a string which is just a minified input and nothing more than that. It's just a *single* change in Stylis source code, but it can't be achieved without forking. cc @thysultan - it might be interesting to you to see how I've bent and used Stylis here. Would you maybe also like some license/author header put into the fork file? If yes - how it should look like?

Keep in mind that this probably should land after Stylis v4 gets published to npm. It would be great to avoid checking in its source code.


Andarist:
@mitchellhamilton this is ready for review now - I'm targeting `stylis-v4` branch with this now as ideally, we are going to wait for stylis v4 to be published (this should happen soon-ish) before merging both of those PRs into next. Both can be reviewed beforehand though.

packages/babel-plugin/__tests__/__snapshots__/css.js.snap
```diff
@@ -710,7 +710,7 @@ import { css } from '@emotion/react';
 function media(...args) {
   return (
     /*#__PURE__*/
-    css(\\"@media (min-width:100px){\\",
+    css(\\"@media (min-width: 100px){\\",
```
mitchellhamilton, line 713: Why is a space added here?

Andarist: Because stylis is crazy-ass parser 🤣and it doesn't try to minify the input too hard, just some whitespace collapsing, nothing more.

It also tries to minimize information that it needs to track along the way, especially given the fact that it often doesn't yet know what it is currently parsing - is it a rule? is it a declaration? don't know - until it reaches either `{` or `;`

So what happens here is basically that whitespace needs to be preserved in css functions like `calc` (probably not all of it, but at least collapsed to a single space - although it doesn't try to collapse it here). So it just consumes anything between `(` and `)` as is - or rather between opening and closing tokens. The same `delimit` function is used for single quotes, double quotes, parenthesis, square brackets and obviously for strings we need to preserve all whitespace whatsoever.

I will probably revisit this later because we can minify this on our side, but didn't want to overcomplicate this PR.

PR #1841, opened 2020-04-10
Add commonjs option to babel-plugin-jsx-pragmatic

ckknight:
**What**:
The commonjs option, when `true`, will allow output to be a CommonJS-style require statement rather than an ECMAScript import statement.

**Why**:
The intent behind this is to route around problems of babel configuration within Jest, primarily.

**How**:
Within babel-plugin-jsx-pragmatic, a check is done for `opts.commonjs` which, if truthy, will insert a `VariableDeclaration` instead of an `ImportDeclaration`

**Checklist**:

- [x] Documentation
- [ x Tests
- [x] Code complete
- [x] Changeset 

Alternatively, if anyone knows how to ensure that Babel consistently processes import statements after babel-plugin-jsx-pragmatic when running within Jest, this effort might be unnecessary.

### DIFF
--- a/.changeset/fifty-rabbits-switch.md
+++ b/.changeset/fifty-rabbits-switch.md
@@ -1,0 +1,5 @@
+---
+'@emotion/babel-plugin-jsx-pragmatic': minor
+---
+
+Add commonjs option to babel-plugin-jsx-pragmatic
--- a/packages/babel-plugin-jsx-pragmatic/README.md
+++ b/packages/babel-plugin-jsx-pragmatic/README.md
@@ -163,6 +163,20 @@ String. The export that you want to import as `import` from `module`. Default va
 // import {something as x} from "whatever"
 ```
 
+### `commonjs`
+
+Boolean. If provided and set to `true`, then CommonJS-style `require` statements will be produced instead of `import` statements. This should not be necessary in the typical case.
+
+```js
+{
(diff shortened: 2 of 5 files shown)


Andarist:
Could you check why [this fix](https://github.com/emotion-js/emotion/pull/1576) is not enough? I'm rather hesitant of including the proposed change and I would like to understand the full context before making a final decision about this.

ckknight:
@Andarist in the way that I had jest configured, the conversion of `import` to `require` seemed to occur before this plugin ever executed, so the fix you mentioned was not adequate.

All this being said, I ended up fully redoing my jest configuration and now things seem to work.

This PR might be theoretically valuable for someone else, but I no longer personally need it.

If you decide to close the PR, my feelings won't be hurt.

Andarist:
I prefer to close it for now until I get a compelling repro case to reconsider this. I'm not saying that this would be a bad thing in general, but I think a slightly alternative approach should be taken instead - we should check if a particular file is a module, like Babel does it:
https://github.com/babel/babel/blob/0aa5a4719618caf4f2118f82409854ba01079dcc/packages/babel-helper-module-imports/src/is-module.js

I'm not 100% sure if we don't rely on things being modules elsewhere, so until somebody comes with this problem again I would like to put this on hold until then.

PR #1848, opened 2020-04-21
Add speedy method to the TS type declaration of the sheet object available on Emotion instances

osdiab:
**What**:
Adds `speedy()` type to `StyleSheet` type

**Why**:
- [Docs say specifically to use `speedy()` when rendering with Puppeteer](https://emotion.sh/docs/ssr#puppeteer)
- [Tests show use of `sheet.speedy()`](https://github.com/emotion-js/emotion/blob/6cb8d15438b01b8623af42f304d5ea3032332187/packages/emotion/test/sheet.dom.test.js)

**How**: Added a function to the `StyleSheet` definitions

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset Not sure?

- Idk if anything needs to be done for Flow


Andarist:
This method is only available when access `emotion.sheet`: https://github.com/emotion-js/emotion/blob/d24177672912095895fda09d584431261f745bbf/packages/create-emotion/src/index.js#L70 . It's not part of the core sheet API.

osdiab:
@andarist Not sure I follow - are you saying it works when importing from `emotion` but not from some other place where a `StyleSheet` object would be present?

Andarist:
Yes, the sheet object is being mutated (receives a new method) when emotion instance is being created

osdiab:
Then does this update work for you @Andarist ?

Andarist:
@osdiab sorry for the lack of response for quite some time. I've adjusted your PR - so you can take a look at what I have meant as this method only being available on Emotion's sheets and how this should be typed.

@mitchellhamilton could you review this? I'm mostly interested in the approval of the chosen type name (`CSSStyleSheet`). The rest is quite a no-brainer 😉

Issue #1847, opened 2020-04-19
[jest-emotion] snapshots show every style changed when any html attribute or style changes


Jimmydalecleveland:
**Current behavior:**

Snapshot errors do not show only the changed styles or attributes of a styled component when using the snapshotSerializer.

**To reproduce:**
Here is a minimal repo I created with the snapshot failure committed. Simply install (I'm using yarn) and run `yarn test` to see it.
[emotion-snapshot-issue-reproduction](https://github.com/Jimmydalecleveland/emotion-snapshot-issue-reproduction)

1. Update snapshots
2. Change any style or html attribute (such as the `id` value)
3. Snapshot will fail as expected but show that every style in the component has changed

**Expected behavior:**

Snapshots should only show diff lines for the things that changed in a component. This used to be the behavior for all our tests but at some point it changed. I'm sorry I can't figure out when, despite trying.

**Environment information:**

- `react` version: ^16.12.0
- `@emotion/core` version: ^10.0.27
- `@emotion/styled` version: ^10.0.2
- `jest-emotion` version: ^10.0.27

Andarist:
This is actually a bug in [`jest-snapshot`](https://github.com/facebook/jest/tree/master/packages/jest-snapshot) package. I've located the problem and gonna try to prepare a fix.

Jimmydalecleveland:
Oh really? Well good luck, I am shivering with anticipation.

Jimmydalecleveland:
I've confirmed locally that `jest@25.0.0` introduced the bug when they changed to the new diff highlighting, so I'm going to close this issue. Thanks for narrowing it down to that @Andarist .

Andarist:
Ok, so I actually was mistaken - although how things are handled in jest around this are somewhat confusing. The actual snapshot is being dedented (the indentation is completely removed) but the received snapshot is not - it relies on the printer indenting it correctly based on the received arguments.

And that's why I have thought it's a bug in `jest-snapshot` because I would expect it to both actual and expected snapshots being handled the same way. It's still might be a bug there - depends on how one looks at it. The printed leading whitespace doesn't have to be strictly related to classic code-indentation purposes which most commonly only improves the readability of nested logic and structures, whereas indentation can be used for other reasons. I think I'm going to propose a fix for this to the jest team regardless of handling this issue here.

[...]

PR #1850, opened 2020-04-23
Jest serializer new api

Andarist:
It [has been reported](https://github.com/emotion-js/emotion/issues/1847) that snapshots show **every** style to be changed when snapshot & received don't match. After investigating this:
- it's somewhat a regression in Jest 25 - I've reported it to the jest team
- we have been ignoring `config.indent` option which somewhat comes helpful here because we were using the old serializers API

So I've decided to refactor those to use "new" serializers API (it got introduced in Jest 21) to eventually handle `config.indent` regardless of what happens to the reported ticket in Jest repo. Also, it was impossible to handle it (at least not in a straightforward way) using the old API

In the process, I've removed `css` dependency (a parser and stringifier) as we already depend on Stylis and we don't need two separate parsers. This has also actually "fixed"/improved printing of nested at rules.


packages/css-prettifier/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@emotion/css-prettifier",
```
Andarist, line 2: I'm not super fond of this name - it seems super close to `@emotion/css` but yet it doesn't have to do anything directly with it.

packages/css-prettifier/src/index.js
```diff
@@ -0,0 +1,53 @@
+import memoize from '@emotion/memoize'
+import { compile, serialize, combine, tokenize } from '@emotion/stylis'
+
+// adjusted https://github.com/thysultan/stylis.js/blob/68b9043427c177b95a0fd6a2a13f5b636bf80236/src/Serializer.js#L26-L34
```
Andarist, line 4: I've started by forking the original `stringify` because I thought I would only have to change a thing or two... but I had to change a little bit more than that, somewhat to the point that it might not be worth it and maybe it would be better to just implement a specific to this use case `serialize`, rather than acting like its a Stylis plugin. OTOH - this works now and Im not sure if I want to refactor it.

packages/css/test/__snapshots__/css.test.js.snap
```diff
@@ -427,11 +427,11 @@ exports[`css nested at rule 1`] = `
     background-color: blue;
   }
 
-@supports (width: 100vw) {
+  @supports (width: 100vw) {
```
Andarist, line 430: this is an example of improved nested atrules printing

packages/jest/src/serializer.js
```diff
@@ -77,18 +77,10 @@ function deepTransform(node, transform) {
...
+  indentation: string
 ) {
-  const styles = getStylesFromClassNames(classNames, elements)
-
-  let prettyStyles
-  try {
-    prettyStyles = css.stringify(css.parse(styles))
-  } catch (e) {
-    console.error(e)
-    throw new Error(`There was an error parsing the following css: "${styles}"`)
```
Andarist, line 89: `stylis` doesnt throw on invalid input, it should just swallow it

.changeset/breezy-carpets-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': major
+---
+
+`test` & `print` are no longer exported as named exports. If you want to access the default serializer just access the default export.
```
mitchellhamilton, line 5: `test` & `print` were exported so you could add `@emotion/jest` to `snapshotSerializers`, I think I'd like to keep `@emotion/jest` working and throw an error when `@emotion/jest/serializer` is imported that tells people to use `@emotion/jest`

Andarist: Hm, isn't it packing too much into a single module? Having a separate entry point for this seems conceptually easier as with `test` & `print` being exported we kinda trick Jest to think that it's a serializer - structurally ok-ish but nominally not so much.

mitchellhamilton: Yeah, okay, sure. Let's use the seperate module. Could you make the changeset clearer that using the raw package name in the `snapshotSerializers` option will no longer work and they'll have to use `@emotion/jest/serializer` instead?

Andarist: added this info, could you re-check?

.changeset/breezy-carpets-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': major
+---
+
+`test` & `print` are no longer exported as named exports. If you want to access the default serializer just access the default export. This means that `@emotion/jest` can be used directly as the value for `snapshotSerializers`, but rather you should use `@emotion/jest/serializer` for this.
```
mitchellhamilton, line 5:
```suggestion
`test` & `print` are no longer exported as named exports. If you want to access the default serializer just access the default export. This means that `@emotion/jest`(previously `jest-emotion`) can't be used directly in the `snapshotSerializers` option, you should use `@emotion/jest/serializer` instead for this.
```

Andarist:
I've merged this after applying your suggested changeset changes as this was the only comment. Feel free to comment further in https://github.com/emotion-js/emotion/pull/1817 if needed.

packages/css-prettifier/package.json
mitchellhamilton: I think it's fine. For mostly internal packages, I don't think it matters much

PR #1859, opened 2020-05-04
Make memoize, is-prop-valid ES modules

wereHamster:
Also mark those packages as side-effect free.

**What**: Node.js 13+ supports ES modules out of the box. This PR makes the code in the two packages `@emotion/memoize` and `@emotion/is-prop-valid` available as ES modules.

**Why**: To allow modern Node.js versions to consume the ES modules directly instead of going through CJS. Using "exports" also prevents users from importing internal files (on purpose or by accident, `import "@emotion/memoize/dist/memoize.esm.js"` will throw an error).

**How**: Only changes in the two respective package.json files are needed. There should not be any regressions for older Node.js versions since they ignore the two new fields "type" and "exports".

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [ ] Changeset 

I'm not certain whether this should be a major bump or not. It should not cause any regressions, so I'm tempted to give it a patch bump.

### DIFF
--- a/packages/is-prop-valid/package.json
+++ b/packages/is-prop-valid/package.json
@@ -1,10 +1,13 @@
 {
+  "type": "module",
   "name": "@emotion/is-prop-valid",
   "version": "0.8.8",
   "description": "A function to check whether a prop is valid for HTML and SVG elements",
   "main": "dist/is-prop-valid.cjs.js",
+  "exports": "./dist/is-prop-valid.esm.js",
   "module": "dist/is-prop-valid.esm.js",
   "types": "types/index.d.ts",
+  "sideEffects": false,
   "license": "MIT",
   "repository": "https://github.com/emotion-js/emotion/tree/master/packages/is-prop-valid",
   "scripts": {
--- a/packages/is-prop-valid/package.json
+++ b/packages/is-prop-valid/package.json
@@ -1,11 +1,15 @@
 {
-  "type": "module",
   "name": "@emotion/is-prop-valid",
   "version": "0.8.8",
   "description": "A function to check whether a prop is valid for HTML and SVG elements",
   "main": "dist/is-prop-valid.cjs.js",
-  "exports": "./dist/is-prop-valid.esm.js",
-  "module": "dist/is-prop-valid.esm.js",
+  "exports": {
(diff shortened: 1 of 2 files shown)


Andarist:
We use http://preconstruct.tools/ to keep our dist files and package.json metadata in check. The support for conditional exports there is being discussed: https://github.com/preconstruct/preconstruct/issues/251 . I wouldn't like to introduce this change manually but rather wait for preconstruct to add support for this. We are mostly waiting for things to clear up a little bit before we implement it there - for instance we'd like to know how exactly bundlers will utilize conditional exports.

> I'm not certain whether this should be a major bump or not. It should not cause any regressions, so I'm tempted to give it a patch bump.

Technically it's a breaking change for ESM consumers in node 13+ because it changes what is being imported, currently when CJS gets imported (which works) you import an object with `.default` property but after this change you would import a function. We, of course, actually want a consumer to import a function and not a compat-wrapper object, but... technically it's a breaking change 😉

wereHamster:
@Andarist I just pushed a change. Like this it should not be a breaking change except for disallowing imports of everything but the entry point (ie. no morn `require("@emotion/memoize/dist/…")` or `require("@emotion/memoize/package.json")`). Though technically a breaking change for node 13+, both packages still have major version zero, so… maybe a chance to bump to 1.0?

It does not work however because [preconstruct does not allow .mjs file extensions](https://github.com/preconstruct/preconstruct/issues/284).

Andarist:
@wereHamster sorry for not getting to you back sooner. We really want to make all of our packages proper modules for Node, but we don't feel that it's worth it **right now**. There are still some unanswered questions for us in regard to interop with bundlers etc. For example, in webpack there is an ongoing discussion on how to support this exactly and if maybe there is also a need for yet-another-exports-key to accommodate some semantics preferred by web/bundler community:
- https://github.com/webpack/webpack/issues/11014
- https://gist.github.com/sokra/e032a0f17c1721c71cfced6f14516c62

We want to wait until the dust settles to move with this forward - once it does we'll just implement this within Preconstruct and benefit from it here and in any other repositories using that tool.

PR #1870, opened 2020-05-14
Fix typo

karur4n:
customise => customize


Andarist:
Well, "customise" is also correct but given that the header is already using "Customizing" then both should use the same spelling. Thanks!

PR #1876, opened 2020-05-20
Remove hydrate method as there is auto-hydration anyway

Andarist:
Seems like `hydrate` is not required as auto-hydration is happening at all times. 

Additionally, I've renamed `rehydrate` methods to `hydrate` as both we were using `hydrate` and ReactDOM is using the same verb.

This is just a quick PR - opening a discussion about this, if this is something we want to do then I'm going to add changeset and fix tests.

### DIFF
--- a/docs/ssr.mdx
+++ b/docs/ssr.mdx
@@ -118,24 +118,14 @@ const stream = renderToNodeStream(<App />).pipe(renderStylesToNodeStream())
 
 ### extractCritical
 
-This returns an object with the properties `html`, `ids` and `css`. It removes unused rules that were created with emotion(it still includes rules that were inserted with `injectGlobal`).
+This returns an object with the properties `html` and `css`. It removes unused rules that were created with emotion(it still includes rules that were inserted with `injectGlobal`).
 
 ```jsx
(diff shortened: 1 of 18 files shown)


docs/ssr.mdx
```diff
@@ -118,24 +118,14 @@ const stream = renderToNodeStream(<App />).pipe(renderStylesToNodeStream())
...
 import { renderToString } from 'react-dom/server'
 import { extractCritical } from '@emotion/server'
 import App from './App'
 
-const { html, ids, css } = extractCritical(renderToString(<App />))
-```
-
-#### hydrate
-
-`hydrate` should be called on the client with the `ids` that `extractCritical` returns. If you don't call it then emotion will reinsert all the rules. `hydrate` is **only** required for `extractCritical`, **not** for `renderStylesToString` or `renderStylesToNodeStream`, hydration occurs automatically with `renderStylesToString` and `renderStylesToNodeStream`.
```
mitchellhamilton, line 133: ☝️  what this says, `hydrate` is still required for people who use `extractCritical` and pass the ids to the client themselves

Issue #1878, opened 2020-05-25
[jest-emotion] snapshots reordering css based on test order, causing failures.


Jimmydalecleveland:
**Current behavior:**
Rendering styled components in `jest` using `@testing-library/react` and `@testing-library/jest-dom` is causing some sort of mutation that changes the order css is written in the snapshot. This causes snapshots to fail when a test is skipped, removed, or re-ordered, among other situations.

It took me many hours to track this down in a large project and it makes our tests very unreliable and causes a ton of noise when the code in many of our styled components changes.

**To reproduce:**
I've created a reproduction repo here:
[https://github.com/Jimmydalecleveland/emotion-jest-reordering-bug](https://github.com/Jimmydalecleveland/emotion-jest-reordering-bug)

The instructions are in the readme. This is not the only way to reproduce this bug, but it's the most simple and clear that I could come up with. Thank you, in advance, for looking at this and any help you can give.

Here is a quick example screenshot from the repo I linked which shows the css being re-ordered:
![image](https://user-images.githubusercontent.com/3894120/82841617-fff74e00-9e93-11ea-8897-018ffb5b4826.png)

**Expected behavior:**
The order of the tests should not mutate other test snapshots. Removing, skipping, or moving a test should not fail other tests snapshots due to css being moved around.

**Environment information:**

- `react` version: 16.12.0
- `emotion` version: 10.0.27
- `jest` version: 24.9.0 (25's new diffing tool still has conflicts with `jest-emotion`)
- `jest-emotion` version: 10.0.27

Andarist:
`getStylesFromClassNames` currently works by iterating through extracted `classNames`:
https://github.com/emotion-js/emotion/blob/fc119766b539f73546dc8e2863cca1439af3cb1c/packages/jest/src/utils.js#L168
and appending matches to the output string, so yes - this currently is prone to style elements being reordered.

A possible solution to this would be to change how this is being iterated in a way that would preserve the order of input `classNames`. Would you like to take a stab at this?

Jimmydalecleveland:
That makes sense, but what I'm quite confused on is why rendering a component in separate tests (like in my repo example) would cause other tests snapshots to re-order. It does not seem like they should affect each other at all, and in most cases they do not.

Thanks for linking to the code, I'll try looking around there, but I don't know if it is exactly the same issue.

Andarist:
[...]

PR #1880, opened 2020-05-27
Fix snapshot styles order

Jimmydalecleveland:
**What**: Fix testing snapshot style reordering when the order that tests are run changes. I have an issue open discussing the details of the bug here: #1878 

Due to jest writing styles in the order it encounters rendered styled components across tests, an issue can occur where skipping/removing/reordering your tests will invalidate test snapshots.

**Why**: Also described in the issue linked above, but the gist is that reordering/skipping/removing tests causes other test snapshots to fail in certain situations where that components styles were already being rendered by emotion. This causes the style order to be changed, but the same styles to be present, causing a lot of noise in test and spooking my coworkers on our code reliability. It's also really hard to track down that this is actually what is going on (many hours for me to figure that out)

**How**: This fix sorts the style elements when serializing emotion styles.

**Checklist**:

- [ ] Documentation: Not unless you count the changeset explanation.
- [x] Tests: New tests were not written for the sorting, and I am not sure how I would write them, but many (32) snapshots needed updating from this change.
- [x] Code complete
- [x] Changeset 

I made this a patch with changeset, because it doesn't require users to change any code to use it. But I was pretty iffy about that, because it does break many snapshots by reordering their css, and they would have to run update for all their snapshots when installing this patch. I'm open to critique there and changing the semver for that if need be.

I realize this is a pretty drastic change outcome, even if the code change is pretty small. In the reproduction repo ([https://github.com/Jimmydalecleveland/emotion-jest-reordering-bug](https://github.com/Jimmydalecleveland/emotion-jest-reordering-bug)) I linked in the previous issue I have examples of how to produce this bug, and I used this repo to test that my proposed change works to fix that issue.

I don't really know if this is the right way to go about it, and many people probably don't even know they have an issue waiting to happen when they move their tests around or rerun them in different orders. We thought this was general snapshot noise in our project, which is quite large, for many months. Again, I'm open to a complete rejection of this PR or any feedback on a better way to handle it.


Andarist:
Sorry for the late reply, I was really busy in the past days.

I'm afraid that the proposed solution is too simplistic - for example, it prints rules for pseudo-classes and pseudo-elements before the actual classes. It also can print `emotion-2` before `emotion-0` which might be confusing.

I think a better solution would be to figure out how to sort them based on the order of the `classNames` argument. This would ensure stable & more readable output.

Please let me know if you would like to work on this one, if not I might take it over but it also might take me some time before I do that. Any help is highly appreciated 🙏

Jimmydalecleveland:
Ok, thanks for looking into this. I've also been super busy lately so I can totally sympathize, no worries. I'd like to take a look into it but I don't know how long it will take me to get back to. Feel free to take it on at any point, but I will make an effort to find time for it.

Andarist:
@Jimmydalecleveland I've finally found some time to fix this properly - u can take a look at how I have done it if you are interested.

PR #1900, opened 2020-06-13
Rename rehydrate

Andarist:
I've extracted this from https://github.com/emotion-js/emotion/pull/1876 as this part of the change is way less controversial and shouldn't be blocked by the discussion about the rest of the stuff in that PR.

### DIFF
--- a/.changeset/few-forks-cheer.md
+++ b/.changeset/few-forks-cheer.md
@@ -1,0 +1,5 @@
+---
+'@emotion/sheet': patch
+---
+
+Renamed `rehydrate` method to `hydrate` to align naming to what is already used in Emotion and React itself.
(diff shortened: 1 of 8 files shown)


Andarist:
I think this is not controversial at all, so I'm just going ahead and merging this. Feel free to leave the review comments at any time though.

.changeset/few-forks-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/sheet': patch
```
mitchellhamilton, line 2: Should be a major

Andarist: Technically in a regular situation - yes, this kind of a change should be a major. Ive deliberetely made it patch because this method was only added within current prerelease cycle so technically this is not a breaking change for v10 users. 

I will be removing this changeset and changing the original one before landing v11 so no mention of rehydrate will be included in the final v11 changelogs.

Also - as I plan to remove this particular changeset i didnt want to trigger too high bump in the current prerelease cycle as removing it later would cause a final version to be downgraded (in comparison to one of released prereleases) and this would only cause confusion and potentially be a problem later.

PR #1906, opened 2020-06-16
Add toHaveStyleRule support for React Native

### LINKED ISSUE #1790: document is not defined when using jest-enzyme with React Native

MrCreeper1008:
Closes #1790

**What**:  
Currently, `toHaveStyleRule` is having issue with React native.

**Why**:
React Native components do not have `className` prop which this rule relies on to perform checks. Instead, they have a `style` prop which takes a style object. However, when `@emotion/native` is used with React Native components, the original components are nested 3 levels deep, making the `style` prop inaccessible unless we `dive`.

**How**:
By `dive`ing into the received component three times, the `style` prop is exposed, which allows us to perform checks. However, style checking for React Native components is implemented differently, so I added a `native` option. To use,
```javascript
expect(ReactNativeComponent).toHaveStyleRule('flexDirection, 'row', { native: true });
```
Alternatively, we can create a native version of `jest-emotion`:
```javascript
import { matchers } from 'jest-emotion/native';
```

**Checklist**:
- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [x] Changeset 

**Caveats**
- This fix is mainly geared towards components shallowly rendered with `enzyme`, so I have no idea if this would work with other testing libraries like `native-testing-library`.
- Also, I have no idea how to test this. (Do we install `react-native` just to get a couple components for us to test with?)  

I would appreciate any feedback given to this PR!

Edit: Add usage details.  
Edit 2: Link to issue


MrCreeper1008:
Sorry for screwing up the commit history a bit.

MrCreeper1008:
@Andarist I don't think having a `/enzyme/native` entry point is a good idea, since this should also work with `native-testing-library`, so doesn't make sense to have it under `/enzyme`.

Andarist:
@MrCreeper1008 do you think this should be a part of a new `@emotion/jest/native` entrypoint? or should it be just a part of the "core" `toHaveStyleRule` (like it is now)?

MrCreeper1008:
I think creating a new entry point makes more sense. React Native projects will be only using the native version of `toHaveStyleRule` anyway, so passing `{ native: true }` to it creates unnecessary noise. Developers can choose to alias `@emotion/jest/native` to `@emotion/jest` if they find the extra entry point annoying.

Andarist:
1. Would you be willing to implement the required changes?
2. Have you tested this in any RN project? Are you confident that this is working for the most common cases?

MrCreeper1008:
1. I'd be more than happy to implement the change. But just to be clear, can you list out all the changes you want?  
2. I created a new react native project to test this out and everything seems to be working fine.

Andarist:
Please add `/native` entry point by adding it [here](https://github.com/emotion-js/emotion/blob/4d3b60d0d448a61d762ee150e6cb7a2c995ccc2f/packages/jest/package.json#L62) and creating a file like [this](https://github.com/emotion-js/emotion/blob/4d3b60d0d448a61d762ee150e6cb7a2c995ccc2f/packages/jest/serializer/package.json). The source code for this entry should look similar to this one: https://github.com/emotion-js/emotion/blob/4d3b60d0d448a61d762ee150e6cb7a2c995ccc2f/packages/jest/src/index.js . Technically there might also be a need for native serializer, right? But not sure if you want to work on it as well. If not - then this should only export `matchers`.

MrCreeper1008:
Can you elaborate more on the need for serializer? Is it used for snapshot testing?

Andarist:
Yes, exactly.

MrCreeper1008:
Shouldn't be hard to implement the serializer. I think I can handle it.

MrCreeper1008:
@Andarist is this structure you're aiming for?

Andarist:
Yes, pretty much - need to review more closely later, especially around extending global `jest` types etc, but looks good at the first glance.

MrCreeper1008:
Thank you. I'll start implementing the serializer.

PR #1920, opened 2020-06-27
Cleanup @emotion/jest entrypoints

Andarist:
1. make `@emotion/jest/serializer` & `@emotion/jest/enzyme-serializer` primarily useful for https://jestjs.io/docs/en/configuration#snapshotserializers-arraystring , they now contain only 2 exports needed for that: `test` & `serialize`
2. use preconstruct to create `@emotion/jest/serializer`
3. removed the default export from `@emotion/jest`, it only exports serializer factories now (and matches) - i believe that in presence of `@emotion/jest/serializer` this makes more sense and ultimately is a little bit less confusing as the same thing is not exported from different entrypoints using different export specifiers


Andarist:
@mitchellhamilton I've [added changesets](https://github.com/emotion-js/emotion/pull/1920/commits/cb94c9825d20747742a382a62349101fd4853a3d), could you take a final look?

.changeset/wet-cycles-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': major
+---
+
+`@emotion/jest/serializer`'s main purpose is compatibility with Jest's `snapshotSerializers` option, so it no longer has a default export. You can import `createSerializer` from the root entry (`@emotion/jest`) and create your own serializer if needed.
```
mitchellhamilton, line 5: I assume `import * as serializer from '@emotion/jest/serializer'` would work?

mitchellhamilton: And if so, in this changeset and the one for enzyme stuff, explain that they're individual exports now.

Andarist: > I assume import * as serializer from '@emotion/jest/serializer' would work?

Yes, it is used like that somewhere in our tests.

> And if so, in this changeset and the one for enzyme stuff, explain that they're individual exports now.

Done - gonna go ahead and merge this PR, but if you have any other comments then please leave them here and I will make sure to address them.

PR #1931, opened 2020-07-09
Add missing unitless CSS properties

kripod:
**What**: Similarly to the pull requests I made towards [React](https://github.com/facebook/react/pull/19286) and [Preact](https://github.com/preactjs/preact/pull/2608), I'm adding all the missing standard and experimental unitless CSS properties, [collected from MDN data](https://github.com/kripod/otion/blob/9b0a30d51bb9bcd13778d6ae913f6e1ca948cac7/packages/otion/src/propertyMatchers.test.ts#L87-L91).

**Why**: To postfix _all_ the numeric values passed to unitless CSS properties with 'px'.

**How**: With thorough testing, as [described](https://github.com/preactjs/preact/issues/2607) in the related Preact issue.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [ ] Changeset


Andarist:
Thank you for this! I'd like to see what happens with the React PR before we merge this here - although I don't expect any objections to this coming from the React team.

kripod:
After the discussion on the React issue, I decided to close that PR, due to the reasons outlined over there. I even opened a new issue [for otion](https://github.com/kripod/otion/issues/25), so that once we may all migrate to a more sustainable approach.

Andarist:
What do you think should happen with this PR here then?

kripod:
I think this should be closed for now, until the React team decides upon a better solution.

kripod:
My [PR towards Preact](https://github.com/preactjs/preact/pull/2608) has been accepted, so we may want to re-evaluate the value of this fix.

Andarist:
I must say that in the end, I feel very similar as React team. Maintaining a whitelist is just troublesome - there was also no actual user request about those being missing. If that's OK with you - I would like to decline this for now.

kripod:
That’s totally reasonable, thank you for answering!

PR #1941, opened 2020-07-27
Localized jsx namespace

### LINKED ISSUE #1257: Augmenting the DOMAttributes interface with 'css' causes conflicts
- `emotion` version: 10.0.7
- `react` version: 16.8.3

**Relevant code:**

```tsx
// NOTE: the code doesn't actually use emotion at all,
// it's an indirect devDependency; but if emotion's types are loaded at all,
// it causes conflicts.
import {} from '@emotion/core'
import styled from 'styled-components'
/// <reference types="styled-components/cssprop"/>

// If I use React.ComponentPropsWithoutRef<'a'> instead there is no conflict,
// but the component I am trying to use does use AnchorHTMLAttributes
interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
  // the foo doesn't matter, it's just to demonstrate object rest
  foo: boolean
}

export default function WrappedLink({ foo, ...props }: Props) {
  return <A {...props} />
}

// the style doesn't matter
const A = styled.a``
```

**What you did:**

In a project that already used `styled-components`, `@emotion/core` started being loaded as an indirect devDependency (`@storybook/theming`, to be more specific).

**What happened:**

The combination of an augmented `React.DOMAttributes` with a double-augmented `JSX.Attributes` is breaking pass-through of the `props` because the `React.DOMAttributes` makes the component think it can receive `css` as a prop.

```
Type '{ download?: any; href?: string | undefined; hrefLang?: string | undefined; media?: string | undefined; rel?: string | undefined; target?: string | undefined; type?: string | undefined; ... 251 more ...; css?: InterpolationWithTheme<...>; }' is not assignable to type '{ className?: string | undefined; download?: any; href?: string | undefined; hrefLang?: string | undefined; media?: string | undefined; rel?: string | undefined; target?: string | undefined; ... 253 more ...; key?: string | ... 1 more ... | undefined; }'.
  Types of property 'css' are incompatible.
    Type 'InterpolationWithTheme<any>' is not assignable to type 'string | (string & ComponentSelector) | (string & { name: string; styles: string; anim: number; toString: () => string; }) | (string & SerializedStyles) | (string & ArrayInterpolation<undefined>) | ... 12 more ... | undefined'.
      Type 'null' is not assignable to type 'string | (string & ComponentSelector) | (string & { name: string; styles: string; anim: number; toString: () => string; }) | (string & SerializedStyles) | (string & ArrayInterpolation<undefined>) | ... 12 more ... | undefined'.
```

**Reproduction:**

CodeSandbox can't typecheck, even in typescript sandboxes :(

**Problem description:**

`@emotion/core` unconditionally augments both `React.DOMAttributes` and `JSX.Attribute`, but ideally such augmentation should be optional to avoid this conflict.

**Suggested solution:**

`@types/styled-components` made it optional by splitting just the augmentation itself into a separate file, and you can request it by adding `"styled-components/cssprop"` to your `tsconfig.json`'s `compilerOptions.types`, using a triple-slash reference (only needed once, on any file that's in the project) as in the above sample, or using an empty import (also only needed once).

The augmentation of `React.DOMAttributes` is technically unnecessary, augmenting `React.Attributes` is sufficient for all JSX use cases (`JSX.Attributes` inherits from `React.Attributes`). The only thing it makes possible is _receiving_ `css` as a prop, which you can't, even with / especially with the `jsx` pragma.

The added advantage of making the augmentation optional is that it makes it possible for the user to augment it with a more specific type -- perhaps `InterpolationWithTheme<MyTheme>`, for example.

---

EDIT: This also, of course, affects actual uses of `css` with the `styled-components` `css` function, as it becomes impossible to fulfill the type. The only workarounds I can do until this is fixed is to locally augment the `css` prop in both interfaces with `any` (!).

### LINKED ISSUE #1800: Emotion types are polluting the global scope
**Current behavior:**

The `@emotion/core/types/index.d.ts` file from the package includes the following declarations:
```ts
declare module 'react' {
  interface DOMAttributes<T> {
    css?: InterpolationWithTheme<any>
  }
}

declare global {
  namespace JSX {
    /**
     * Do we need to modify `LibraryManagedAttributes` too,
     * to make `className` props optional when `css` props is specified?
     */

    interface IntrinsicAttributes {
      css?: InterpolationWithTheme<any>
    }
  }
}
```

This causes the global scope to become polluted with `@emotion` types, even though the user might not be asking for it - eg. it came as a surprise to me, because `@emotion` is a dependency of `storybook`, and I didn't know about it until I tried using the `css` prop for `styled-components` and there was no error.

Further, after including `styled-components` type declaration for the `css` prop, some strange errors emerged because of the way TypeScript handles declaration merging:

```
  Types of property 'css' are incompatible.
    Type 'string | CSSObject | FlattenInterpolation<ThemeProps<any>> | undefined' is not
assignable to type 'InterpolationWithTheme<any>'.
      Type 'FlattenInterpolation<ThemeProps<any>>' is not assignable to type 'InterpolationWithTheme<any>'.
        Type 'FlattenInterpolation<ThemeProps<any>>' is not assignable to type 'ObjectInterpolation<undefined>'.
```

**Expected behavior:**

I'd expect a mechanism similar to `styled-components/cssprop`: the user has to manually create a file which imports global declarations instead of pullution by default.

**Environment information:**

@emotion/core@10.0.27

Andarist:
fixes #1800 fixes #1257

I've discovered this magic recently and it solves, quite ideally, the problem with our types augmenting global types and thus causing conflicts with other libraries that define types for the `css` prop. 🎉 

More info can be found here:
- https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions
- https://github.com/microsoft/TypeScript/pull/22207

TODO: 
- [x] docs
- [x] changeset


packages/react/types/css-prop.d.ts
```diff
@@ -0,0 +1,9 @@
+import {} from 'react'
```
Andarist, line 1: This file can be included manually by people not wanting to use `@jsx` pragma. Those types here are slightly worse than the ones provided for `@jsx` pragma but AFAIK there is no way to override a type alias already defined in a namespace so I can't redefine `LibraryManagedAttributes` to only allow `css` prop for components already accepting `className` prop.

packages/react/types/index.d.ts
```diff
@@ -94,8 +92,42 @@ export interface ClassNamesProps {
  */
 export function ClassNames(props: ClassNamesProps): ReactElement
 
-declare module 'react' {
-  interface Attributes {
-    css?: Interpolation<Theme>
+type WithConditionalCssProp<P> = P extends { className?: string }
```
mitchellhamilton, line 95: I think this should be:
```tsx
type WithConditionalCssProp<P> = "className" extends keyof P
  ? P extends { className?: string }
    ? P & { css?: Interpolation<Theme> }
    : P
  : P;
```

So that it doesn't add className for `{}`

Andarist: Eh, I always forget about this surprising, to me, behavior of the `{}`. Good catch.

Andarist:
> This is awesome 😍 Love that we can conditionally add css rather than always

Right! I was quite pleasantly surprised by this as well. There is one downside though - errors about `css` prop being missed in the props types are more cryptic. One could think that they should add it to their props instead of adding `className`. 

> It's probably worth trying this out on some codebases to see how the type checking performance is, I've got a reasonably big one so I can try it out but others trying it would probably be good.

That would be sweet! I will add the appropriate changeset and docs today/tomorrow so we'll should be able to ship new prerelease this week.

Andarist:
@mitchellhamilton I've added a changeset and short docs mentioning how it behaves now. Please take a look.

JakeGinnivan:
This is cool, I will upgrade soon and see how perf is going on our project

PR #1948, opened 2020-07-30
Minor TS types tweaks in @emotion/styled

Andarist:
As far as I can tell those are equivalent but written in a less awkward way.

### DIFF
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -37,7 +37,7 @@ export interface StyledComponent<
   ComponentProps extends {},
   SpecificComponentProps extends {} = {}
 > extends React.FC<ComponentProps & SpecificComponentProps>, ComponentSelector {
-  withComponent<C extends React.ComponentType<React.ComponentProps<C>>>(
+  withComponent<C extends React.ComponentType<any>>(
     component: C
   ): StyledComponent<ComponentProps & PropsOf<C>>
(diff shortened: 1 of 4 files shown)


packages/styled/types/base.d.ts
```diff
@@ -99,23 +99,23 @@ export interface CreateStyledComponent<
...
   <
-    C extends React.ComponentType<React.ComponentProps<C>>,
+    C extends React.ComponentType<any>,
     ForwardedProps extends keyof React.ComponentProps<
       C
     > = keyof React.ComponentProps<C>
   >(
     component: C,
-    options: FilteringStyledOptions<PropsOf<C>, ForwardedProps>
+    options: FilteringStyledOptions<React.ComponentProps<C>, ForwardedProps>
```
Andarist, line 108: this is not an equivalent, but actually I think the new version is more correct - we don't need to resolve defaultProps here (I don't care as much about resolving propTypes)

Andarist:
It fails on this component:
https://github.com/emotion-js/emotion/blob/f57a72299cd4025a725bd5bd1b966a8f9df16cd8/packages/styled/types/tests-base.tsx#L365-L373
which is wrongly typed (`getDerivedStateFromProps` should be declared using `Readonly<Props>`) but also this is for some reason allowed by TS and probably popular in the wild. The current version of our typings allows this so this change here is probably not worth it.

PR #1951, opened 2020-07-30
Minor TS types tweak in @emotion/styled

Andarist:
Extracted from the closed https://github.com/emotion-js/emotion/pull/1948

### DIFF
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -105,7 +105,7 @@ export interface CreateStyled {
     > = keyof React.ComponentProps<C>
   >(
     component: C,
-    options: FilteringStyledOptions<PropsOf<C>, ForwardedProps>
+    options: FilteringStyledOptions<React.ComponentProps<C>, ForwardedProps>
   ): CreateStyledComponent<
     Pick<PropsOf<C>, ForwardedProps> & {
       theme?: Theme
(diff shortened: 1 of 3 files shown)


packages/styled/types/base.d.ts
```diff
@@ -105,7 +105,7 @@ export interface CreateStyled {
     > = keyof React.ComponentProps<C>
   >(
     component: C,
-    options: FilteringStyledOptions<PropsOf<C>, ForwardedProps>
+    options: FilteringStyledOptions<React.ComponentProps<C>, ForwardedProps>
```
Andarist, line 108: this is not an equivalent, but actually I think the new version is more correct - we don't need to resolve defaultProps here (I don't care as much about resolving propTypes)

the even more correct version would use `React.ComponentPropsWithoutRef` but that is using conditional type + Exclude and Pick combo, so probably better to avoid it if we can, and the gain from using this over a simpler `React.ComponentProps` wouldnt be that big

packages/styled/types/tests-base.tsx
```diff
@@ -432,3 +432,13 @@ const Section = styled('section')`
 `
 ;<StyledWithAs as="section" />
 ;<StyledWithAs as={Section} />
+
+{
+  const MemoedComponent = React.memo((_props: { foo: string }) => {
```
Andarist, line 437: this is not relevant for this change, just something I've wanted to check - there were no other types tests involving `.memo` so I've decided to add this here.

PR #1952, opened 2020-07-30
Fix v11 jsx factory type not allowing children when implictly in component type

mitchellhamilton:
I tried out the definitions and found this problem(it caught some bugs in the repo I tested it on because of the conditional adding of the css prop + the css prop not being global 🎉)

Another thing related to this: We should probably omit the `css` prop from where styled-components and emotion v10 add it so this is compatible with styled-components and emotion v10(wouldn't work with both emotion v10 and s-c but having one of them + emotion v11 working together would be nice), I haven't looked into it but just wanted to raise it. (For the repo I was testing the types on, I deleted @emotion/core's global declarations to get around it)

I'm not 100% sure about the fix but I _think_ it's correct?

### DIFF
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -116,9 +116,8 @@ export namespace jsx {
     interface ElementChildrenAttribute
       extends ReactJSXElementChildrenAttribute {}
 
-    type LibraryManagedAttributes<C, P> = C extends React.ComponentType<infer T>
-      ? WithConditionalCSSProp<T>
-      : WithConditionalCSSProp<ReactJSXLibraryManagedAttributes<C, P>>
+    type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+      ReactJSXLibraryManagedAttributes<C, P>
 
     interface IntrinsicAttributes extends ReactJSXIntrinsicAttributes {}
     interface IntrinsicClassAttributes<T>
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -177,4 +177,11 @@ const anim1 = keyframes`
   ;<CompWithoutProps css={{ backgroundColor: 'hotpink' }} />
 }
 
-;<React.Fragment>content</React.Fragment>
+{
+  // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40993
+  // this is really problematic behaviour by @types/react IMO
+  // but it's what @types/react does so let's not break it.
+  const CompWithImplicitChildren: React.FC = () => null;
+  ;<CompWithImplicitChildren>content<div/></CompWithImplicitChildren>
+
+}
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -183,5 +183,4 @@ const anim1 = keyframes`
   // but it's what @types/react does so let's not break it.
   const CompWithImplicitChildren: React.FC = () => null;
   ;<CompWithImplicitChildren>content<div/></CompWithImplicitChildren>
-
 }
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -176,3 +176,5 @@ const anim1 = keyframes`
   // $ExpectError
   ;<CompWithoutProps css={{ backgroundColor: 'hotpink' }} />
 }
+
+;<React.Fragment>content</React.Fragment>


packages/react/types/index.d.ts
```diff
@@ -116,9 +116,8 @@ export namespace jsx {
     interface ElementChildrenAttribute
       extends ReactJSXElementChildrenAttribute {}
 
-    type LibraryManagedAttributes<C, P> = C extends React.ComponentType<infer T>
-      ? WithConditionalCSSProp<T>
-      : WithConditionalCSSProp<ReactJSXLibraryManagedAttributes<C, P>>
+    type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+      ReactJSXLibraryManagedAttributes<C, P>
```
Andarist, line 120: Huh, I was under impression that `P` stands for props passed to the JSX element, and not for all props of `C`. It seems that the compiler resolves those on its own so my `React.ComponentType<infer T>` was completely redundant - and it even has introduced this problem with `children`. 

I was curious to see how this `LibraryManagedAttributes` works and I've prepared a TS playground to test various scenarios, might be interesting for somebody: [TS playground](https://www.typescriptlang.org/play/index.html?jsx=2&ssl=32&ssc=30&pln=32&pc=47#code/PQKhAIAECsGcA9x0SYAoAlgWwA4HsAnAF3AgENZwAlAUzIGMSAzAvLcAcgLsY7TSYBXAHaMMeYUgQAKAJTgA3uAC+aYWSw1YOBjSmIFacOBrx8xcOs3bd4AFIBlABqKjxk2cIkiATxx6AGQwAIwIyAh8AWTJ1AHMaABMAQSIiAhDBIi0AHgBhABpwAAUAPnAAXmLwADJFcHoAfgAucALwHGaq1WNVVQERMQlWthxpHFYcWBalJjw8FuFBLGCaAkLg8M7YNIxhWJV5Q2NuIkECSUWAG0u0PoBucDyR8Fm8coUAFgAmZXr3q8uvxw-0E11+wBK-HoEm24EiNCweESuWelVoDCIADpNIjpCjcLJ+A9svDEcjnq93t9fvQQWD2nTAeAIVCYSR8TgAOoYIgACwAkrhLhh6DzcryMJcEtxhC10YxMQAxXLZGZzBZLFZrcAbAhbHZ7FRlSpCUREcSSDljCawQ5uE5nC6gm59YzEjncvmCnDC0VEcWS6U0SSUz4-P4KAFAxngyFoaHCWEenm8gAiNCYZFBREolWteEm0xe6ssmtW63CLW26UNynk5TKR3ADvOpeutzQyb56cz2dgmISGazlyIRRtFTqupaHHUs7IHBURMeXbTQ77xbeYZpMYZkedsaAA)

I was also curious why there is a difference between both of those and I think that is because the TS compiler gets `P` from a callable signature, which is defined here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0f095853ae6233ad5482a7204dab9326b128ad5c/types/react/index.d.ts#L547
but `React.ComponentType` infers a generic given to `FunctionComponent` which is defined here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0f095853ae6233ad5482a7204dab9326b128ad5c/types/react/index.d.ts#L546
and thus it doesn't have those implicit `children`.

As to the proposed change - there is probably no real difference between what u got and this, but maybe this would look slightly more naturally?
```suggestion
    type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<ReactJSXLibraryManagedAttributes<C, P>>
```

mitchellhamilton: I originally did that, though that doesn't seem to add the css prop for components: https://www.typescriptlang.org/play/index.html?jsx=1&ssl=1&ssc=1&pln=50&pc=25#code/PQKhAIAECsGcA9x0SYAoAlgWwA4HsAnAF3AG9wBjAgUwEMjqBRAG2q2oDsSBfcAMwJ4s4AOQ1aFIiLRoiATxzVwAdQxEAFgGE8HACZqMO2s00BlUwAVBOADwWAfOAC8ois1qxYAOVrsR4angGPVhwAGtqOTw+cAs0cHAAfnAACgsAoM5dUPI3D29fakSALnBYIgIMDgBzcF5k9IAyMkpPEvA8ACNoakk68FKLAEp4gdiZYGBwAFcOHAkw8GNmcHVqGnAiPCWANzwMXXAqviq1JVhqZj4AWho+dc4KKtqAdzWOcF1qE45njumCB0Xh8AFKmAAa4A4hVg8wo1FkCiUACU6JIweCWGxOCQXBiAHRY9hcRGKcCoiREDFEnGadyeZzgAk0rh0-KklFoqkQllEACCRAqGE60wYsCseEUxDkjOZrGJ-MFlRFYolUvkHPJXOp8tp6gwzF0NA4AqFKqUeIhhN1rP1huNpuVooR8jJFPREIAMsKCLQCHIALK0aHVai6R3C52wGyaAA0sUclvB+O9nV9-qDIbDEfN0bjCc17u54IAklxKhxYBgKDmo7KrWWhZXq7WxYXtRDGxWqxQ2Z5W9RowAVRNMhvlqo9vuwAfD+ztykYruT6u80JJ-HL5sUNcyL55DYUHTlJAIUqu6jRSjiBi8tD79wbaHsWESJTIMijKEwuFKDGfhIEiqBgCD4N9wF5DJgmyLVFx5G0SFIbgvyOLh1jA+EIIQ6coKyUIix1bFWXpHJkMA1CQIwpReVnNV1g1ciEkCaD8I7TEENo6x6JlJCUOA9DwN5TQ7SNTgBxQpjMhCWCPXYoiiGEg1RJNJVIwYMhkJQi9wFTdNA2DWhQ3DVTcxjeMHEZVQNG0PQDCMExzDVGwCK9H0-X0rNjLNKMzITedGPAGRyP40DwK3HtZ1w6SXNLCdt0i3jgrQ0LMPC6tp1nGwRwkqKYJitLexIzKRw0oLAO0gq10ZUgcoAbQAaVQ8JIivfK4p7NcAF1SjapsOoQ2AGs68BmhqgKEgoNpSi6HpJBysjAIW5DNIAbhsfQdlaWAnFIJDeGAewVpkNA+FmSRDA+AM5G0XAdBxFIcGsWBilIPJPB8dgSnKCtqm4IYxvAGgiABD4OGmZhmDQbgjrQGwrpu-AOBxLadr28ADqAA

mitchellhamilton: Oh wait - If you distribute over props, it works([playground](https://www.typescriptlang.org/play/index.html?jsx=1&ssl=50&ssc=25&pln=49&pc=1#code/PQKhAIAECsGcA9x0SYAoAlgWwA4HsAnAF3AG9wBjAgUwEMjqBRAG2q2oDsSBfcAMwJ4s4AOQ1aFIiLRoiATxzVwAdQxEAFgGE8HACZqMO2s00BlUwAVBOADwWAfOAC8ois1qxYAOVrsR4angGPVhwAGtqOTw+cAs0cHAAfnAACgsAoM5dUPI3D29fakSALnBYIgIMDgBzcF5k9IAyMkpPEvA8ACNoakk68FKLAEp4gdiZYGBwAFcOHAkw8GNmcHVqGnAiPCWANzwMXXAqviq1JVhqZj4AWho+dc4KKtqAdzWOcF1qE45njumCB0Xh8AFKmAAa4A4hVg8wo1FkCiUACU6JIweCWGxOCQXBiAHRY9hcRGKcCoiREDFEnGadyeZzgAk0rh0-KklFoqkQllEACCRAqGE60wYsCseEUxDkjOZrGJ-MFlRFYolUvkHPJXOp8tp6gwzF0NA4AqFKqUeIhhN1rP1huNpuVooR8jJFPREIAMsKCLQCHIALK0aHVai6R3C52wGyaAA0sUclvB+O9nV9-qDIbDEfN0bjCc17u54IAklxKhxYBgKDmo7KrWWhZXq7WxYXtRDGxWqxQ2Z5W9RowAVRNMhvlqo9vuwAfD+ztykYruT6u80JJ-HL5sUNcyL55DYUHTlJAIUqu6jRSjiBi8tD79wbaHsWESJTIMijKEwuFKDGfhIEiqBgCD4N9wF5DJgmyLVFx5G0SFIbgvyOLh1jA+EIIQ6coKyUIix1bFWXpHJkMA1CQIwpReVnNV1g1ciEkCaD8I7TEENo6x6JlJCUOA9DwN5TQ7SNTgBxQpjMhCWCPXYoiiGEg1RJNJVIwYMhkJQi9wFTdNA2DWhQ3DVTcxjeMHEZdJmLwpYOBlZICK9H0-X0rNjLNKMzJUNQtB0fQiEMaETHMNU7Hsexig4agdnWCTwBkcj+NA8Ctx7WdcOkxzSwnbd0t4xK0OSzDUuradZxsEc4uszK2JK3sSPKkcNISwDtLqtdGVIOKAG0AGlUPCSIryy9qENgABdUoRpyns1z68bwGaLrGMAig2lKLoekkOKyMA3bkM0gBuGx9B2VpYCcUgkN4YB7EOmQ0D4WZJEC8AAzkbRcB0HEUhwaxYGKUg8k8Hx2BKcoK2qbghmW8AaCIAEPg4aZmGYNBuHutAbHez78Cirhzsu67wFuoA)) - I feel like the intersection is kinda better, I would assume an intersection is cheaper than a conditional type?

Andarist: It also works with a simpler:
```ts
```

Andarist: I don't understand why the composition of the types doesn't work here, but if it doesn't then the intersection is fine. 👍

Andarist:
> Another thing related to this: We should probably omit the css prop from where styled-components and emotion v10 add it so this is compatible with styled-components and emotion v10(wouldn't work with both emotion v10 and s-c but having one of them + emotion v11 working together would be nice), I haven't looked into it but just wanted to raise it. 

Not sure if I understand this concern. Both emotion 10 and s-c add support for the css prop through extending global/react interfaces - not sure how we could be compatible with that. I don't feel it's much of a concern for s-c because their css prop types are opt-in and only added per project so they shouldn't really clash anyhow with Emotion 11. Emotion 10 might only be somewhat problematic in a similar way s-c has been clashing with Emotion 10 (types being pulled into the compilation without consumer being even aware of it) - is it a common enough scenario that we should be worried about it though?

Andarist:
Merging this as it's a fix but we can discuss [this](https://github.com/emotion-js/emotion/pull/1952#issuecomment-667077116) more here.

Issue #1942, opened 2020-07-27
`ref` prop type missing with @emotion/native in Emotion 11


efoken:
**Current behavior:**

If I have this code:

```tsx
import React from 'react';
import styled from '@emotion/native';

const Container = styled(View)({ margin: 8 });

const Example: React.FC = () => {
  const containerRef = useRef<View>(null);

  return <Container ref={containerRef} />;
};
```

I get the following TypeScript error, because the `ref` property does not exist:

```
Type '{ children: Element[]; allowFontScaling?: boolean | undefined; autoCapitalize?: "none" | "sentences" | "words" | "characters" | undefined; autoCorrect?: boolean | undefined; ... 116 more ...; theme: Theme; }' is not assignable to type 'IntrinsicAttributes & ViewProps & { theme?: Theme | undefined; } & { children?: ReactNode; }'.
  Property 'ref' does not exist on type 'IntrinsicAttributes & ViewProps & { theme?: Theme | undefined; } & { children?: ReactNode; }'.
```

But it works as expected.

**To reproduce:**

1. Just create a component in React Native with the code from above
2. See TypeScript error

**Expected behavior:**

Using `ref` should be working without TypeScript errors.

**Environment information:**

- `react` version: 16.11.0
- `emotion` version: 11.0.0-next.13

Andarist:
Thank you for reporting this - I will take a look at this soon.

Andarist:
Hm, `ref` is not part of `ViewProps` - do you know why that may be?

efoken:
@Andarist I think `ref` is defined somewhere else – but I have really no idea, it's defined somewhere in the React types – not React Native

Andarist:
Implemented this in https://github.com/emotion-js/emotion/pull/1954

jongbelegen:
I'm currently on the latest next:
```
"@emotion/core": "^11.0.0-next.10",
"@emotion/native": "^11.0.0-next.19",
"@emotion/react": "^11.0.0-next.10",
```

And I still have this issue:
```
Property 'ref' does not exist on type 'IntrinsicAttributes & PressableProps & { children?: ReactNode; } & { theme?: Theme | undefined; } & CustomProps & { ...; }'.

ref={ref}
```

efoken:
@jongbelegen Can you update all 3 Emotion packages to `^11.0.0-rc.0`? Then try and post your component code here – I'm not having issues with the `ref` anymore...

jongbelegen:
[...]

PR #1954, opened 2020-08-01
Add TS types support for `ref` on class components (including host components) in @emotion/native

Andarist:
closes https://github.com/emotion-js/emotion/issues/1942

cc @efoken @patsissons


Andarist:
@JakeGinnivan I would appreciate you taking a look at those improvements - just at the conceptual level, so you could confirm that introducing a new type param to solve the `ref` prop problem is a good idea. However, in the meantime - I'm merging this so other work can be rebased on top of this.

JakeGinnivan:
It appears createStyled will have a ref even if the underlying component does not support refs? Is that right, I haven't pulled it down to actually check.

I think unblocking is the better option, when we get time we can look at the new ref approach, I just have been short of time

Andarist:
> It appears createStyled will have a ref even if the underlying component does not support refs? Is that right, I haven't pulled it down to actually check.

All class components support refs - you only get a component's instance as a value, rather than the underlying host element. I believe that React Native components were all implemented using classes and "native methods" were just hoisted onto those classes, so you could use - let's say - `focus` method on that instance and it would just proxy to a native/inner handler, but technically this was equivalent to something like:
```
class Input extends React.Component {
  _ref = React.createRef()

  focus() {
    this._ref.current.focus()
  }

  render() {
    return <input ref={this._ref} {...this.props} />
  }
}
```

Nowadays those host components are implemented using `React.forwardRef`, but in the past they were classes:
https://github.com/facebook/react-native/blob/2e384184bc7b3173e3bb41e2128eb75ba8aee8db/Libraries/Components/View/View.js#L35
`@types/react-native` still type most of them as classes though - as ultimately it doesn't matter much, the implementation stays rather opaque to the user.

**Note**: Technically typings for `@emotion/styled` are not 100% correct right now as they should support `ref` in a similar manner - one should be able to get a ref to the underlying class' instance (if one wraps a class component with `styled`), but it's rather not worth fixing as no one expects this to be typed correctly right now and class components are legacy-ish anyway.

PR #1956, opened 2020-08-03
feat(serialize): Bump csstype

eps1lon:
**What**:

Bump `csstype` in `@emotion/serialize`. 

**Why**:

Align type ecosystem to avoid potential type conflicts. All `@types/*` packages already switched https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46435

**How**:

- make sure `test:typescript` is ran in CI
- fix `test:typescript` first 
- bump `csstype` and see if `test:typescript` still passes (it does)

**Checklist**:

- ~[ ] Documentation~
- [x] Tests
- [x] Code complete
- [x] Changeset


.circleci/config.yml
```diff
@@ -77,6 +77,7 @@ jobs:
             - v4-dependencies-
       - run: yarn install --pure-lockfile
       - run: yarn lint:check
+      - run: yarn workspace @emotion/serialize test:typescript
```
Andarist, line 80: not running `test:typescript` has already been fixed on the `next` branch, I would keep this out from here to keep merging simpler

packages/serialize/types/tests.ts
```diff
@@ -16,41 +16,41 @@ const testObjectInterpolation1: ObjectInterpolation<undefined> = {
 }
 
 // $ExpectType SerializedStyles
-serializeStyles({}, [])
+serializeStyles([], {})
```
Andarist, line 19: any particular reason why those tests had to be updated?

Andarist:
Do you know what kind of breaking changes were included in v3? It's not super apparent from this: https://www.npmjs.com/package/csstype#version-30

packages/serialize/types/tests.ts
eps1lon: They did not pass on master

eps1lon:
> Do you know what kind of breaking changes were included in v3? It's not super apparent from this: [npmjs.com/package/csstype#version-30](https://www.npmjs.com/package/csstype#version-30)

This seems sufficient to me. With types breaking changes reveal themselves. Anything that's not necessarily findable with a structural type system is listed e.g. `string & {}` for autocomplete.

.circleci/config.yml
eps1lon:
```suggestion
```

packages/serialize/types/tests.ts
eps1lon: More specifically: Most of the packages do not pass their own `test:typescript` on `master` and types aren't tested in CI. I don't know about the state of `next` but it's not apparent from CONTRIBUTING.md on which branch development should happen.

Andarist:
> This seems sufficient to me. With types breaking changes reveal themselves.

Our consumers might often not be aware that they are using an external library like `csstype` - and when we publish a new version of our package it would be ideal to document breaking changes from our PoV.

> Anything that's not necessarily findable with a structural type system is listed e.g. string & {} for autocomplete.

Isn't it a new feature, sort-of, rather than a breaking change? As is - it seems to allow more than before?

eps1lon:
> Isn't it a new feature, sort-of, rather than a breaking change? As is - it seems to allow more than before?

Can't fault by making a bigger bump than necessary.

> Our consumers might often not be aware that they are using an external library like csstype - and when we publish a new version of our package it would be ideal to document breaking changes from our PoV.

Quickly scanned the project and could not see any exposure of these namespaces. I could not find any breakage in dependent packages when csstype was bumped in the `@types` packages (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46435). Only direct consumers had to change code. Seems safe to me.

Andarist:
Ok - thanks for checking this. I will handle this soon-ish - need to recheck versioning between master&next and other organizational stuff in the repo. I assume you are interested in including this change in the current latest version (v10) and waiting for it to be released in the upcoming major version (v11) is not good enough, right?

eps1lon:
> I assume you are interested in including this change in the current latest version (v10) and waiting for it to be released in the upcoming major version (v11) is not good enough, right?

It's not blocking me in any way. Whatever timeline works best for you should be preferred.

PR #1958, opened 2020-08-03
Add pressable native component

sanpoChew:
**What**: Add support for the new Pressable component introducted in React Native 0.63 https://reactnative.dev/blog/2020/07/06/version-0.63#pressable

**Why**: Without these changes users are unable to use `styled.Pressable` to create Pressable components.

**How**: Added Pressable to the components array in the code and added it to the ReactNativeComponentNames union type.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset 

Not sure if I should run `yarn changeset` seeing as this would go into `next`?


sanpoChew:
guess this will require a bump of the `@types/react-native` dependency?

Andarist:
> guess this will require a bump of the @types/react-native dependency?

I have done this recently on another branch - so don't worry about it.

> Not sure if I should run yarn changeset seeing as this would go into next?

Yes, as we are publishing prereleases from this `branch` quite often. I can do that later myself though.

Issue #1960, opened 2020-08-05
The jest snapshot serializer is failing to properly replace hashed classnames.


vorillaz:
**Current behavior:**

The `jest-emotion` snapshot serializer is failing to properly replace the generated class names when styling gets generated from the passed props.

**To reproduce:**
You can find a [CodeSandbox](https://codesandbox.io/s/emotion-issue-template-ju60p?file=/src/index.test.js) replicating the bug. Since snapshots can't get generated properly though CodeSandbox you have to download the `zip` file

The bug can also get easily reproduced as follows:
1. Create a dummy component that it's styling is generated through prop passing.
```jsx
const Box = styled.div`
  ${({ colored }) => {
    if (colored) {
      return {
        color: "red",
      };
    }
  }}
`;
```
2. Setup `jest-emotion` snapshot serializer as described in [docs](https://www.npmjs.com/package/jest-emotion)
3. Create two new tests with Jest snapshot checking:
```jsx
import React from "react";
import { render } from "@testing-library/react";
import styled from "@emotion/styled/macro";
import Box from "./box";

const Failing = styled.div``;

test("Failing is failing ¯\_(ツ)_/¯", () => {
	const { container } = render(<Failing />);
	expect(container).toMatchSnapshot();
});

test("Box without props is failing badly", () => {
	const { container } = render(<Box />);
	expect(container).toMatchSnapshot();
});

test("Box with props is passing", () => {
	const { container } = render(<Box truncated />);
	expect(container).toMatchSnapshot(``);
});
```

**Expected behavior:**

The snapshots are failing as the generated classnames are not properly replaced from the serializer, eg:
```html
<div>
  <div
    class="css-1efsrtw-Box emotion-0"
  />
</div>
```

**Environment information:**

- `react` version: `16.8.6`
- `emotion` version: `10.0.10` and `11.0.0-next.12`

Andarist:
It's not that this is not being replaced correctly when styling is generated based on props or not but rather it fails if you want to snapshot a component without styles, for example
```js
styles.div``
```
**without** rendering a component with styles first.

So given your repro case (thanks for that!), this doesn't work:
```js
test("Box without props is failing", () => {
  const { container } = render(<Box />);
  expect(container).toMatchSnapshot();
});

test("Box with props is passing", () => {
  const { container } = render(<Box truncated />);
  expect(container).toMatchSnapshot();
});
```

but this does work:
```js
test("Box with props is passing", () => {
[...]

PR #1962, opened 2020-08-07
Insert empty style elements eagerly in dev

Andarist:
It would remove confusion around this, like in here: https://github.com/emotion-js/emotion/issues/1960 . This was the second report about this - I recall debugging a similar scenario in the past after another user's report.


packages/jest/test/__snapshots__/react-enzyme.test.js.snap
```diff
@@ -264,7 +264,7 @@ exports[`enzyme shallow basic 1`] = `
 
 exports[`enzyme shallow empty styled 1`] = `
 <div
-  className="css-1k6nb1k-Greeting emotion-0"
+  className="emotion-0 emotion-1"
```
Andarist, line 267: in here we can see how this has "improved" the situation

Issue #1969, opened 2020-08-12
babel-preset: Support plugin-transform-react-jsx's automatic runtime


FLGMwt:
## The problem

[`@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx) powers the transformation of JSX to standard JavaScript method invocations: `<div />` => `React.createElement('div')`.

[@emotion/babel-preset-css-prop](https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop#emotionbabel-preset-css-prop) is a babel preset that wraps the JSX transform configuration to change the default JSX pragma (`React.createElement`) to a custom pragma (`___EmotionJSX`) defined in `@emotion/core`.

React 17 introduces an element creation paradigm which is an alternative to `React.createElement`. It is described in the [Create Element Changes RFC](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) and is implemented for React itself in [this PR](https://github.com/facebook/react/pull/18299). The new element creation paradigm can be used with a new option in the [@babel/plugin-transform-react-jsx](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154).

The current `@emotion/babel-preset-css-props` configuration and `@emotion/core`'s `jsx` transform are built for the `React.createElement` paradigm, now referred to as the "classic JSX runtime". This configuration and transform implementation are incompatible with the new "automatic JSX runtime" option.

I would like to use the automatic JSX runtime option with the emotion CSS prop.

## Proposed solution

I've done some digging and I think I have a handle on what's required:

### `@emotion/babel-preset-css-props` should check for `options.runtime === 'automatic`

Currently, `babel-preset-css-props` forwards all jsx plugin options but *also* configures `pragma` and `pragmaFrag` https://github.com/emotion-js/emotion/blob/master/packages/babel-preset-css-prop/src/index.js#L26. These options are incompatible with the automatic runtime. We shouldn't send them when runtime is automatic. Instead, we should set the [`importSource`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#importsource) option to `@emotion/core`.

We also don't need pragmatic because its role is now handled by the jsx transform in automatic mode.

### `@emotion/core` should implement and expose jsx-runtime

[...]

PR #1970, opened 2020-08-12
Added new JSX entry points for the automatic runtime

FLGMwt:
**What**: Draft / PoC implementation of the automatic jsx runtime #1969 

**Why**: Examine feasibility of supporting the automatic jsx runtime

**How**: 

1. I added handling for `runtime === 'automatic` in the babel css preset to configure for the automatic runtime
2. I implemented `jsx-runtime` in `@emotion/core` mirroring the existing `jsx` function. (this was copied wholesale for ease of modification, it should be factored once the solution is finalized)
3. I added a `jsx-runtime` preconstruct entrypoint configuration to `@emotion/core` to allow babel to import `@emotion/core/jsx-runtime` (need to figure out UMD)
4. I added a new from-scratch webpack / babel / react playground app to test babel configuration

**Needs discussion / solutioning**

We need react@^17.0.0 for forwarding our modified jsx args to `react/jsx-runtime`'s jsx. How do we conditionally depend on react^17 without forcing a client upgrade? can we?

I couldn't get UMDs to work with `@emotion/core/jsx-transform`. It errored looking for an external for `react/jsx-runtime` I expected this to be fixed by adding `{ globals: { 'react/jsx-runtime': 'JSXRuntime' } }` [based on this](https://github.com/facebook/react/pull/18299/files#diff-6c6554c881731a0c8ac33c8ddba3876eR91), but it didn't work. I just disabled the UMD build to get things working.

The JSX implementation is incomplete and needs factoring. I was able to get it working with almost the only changes being `React.createElement` => React's `jsx`, but I know there are issues with keys. I think I'll need some guidance here on how `withEmotionCache` works. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset


package.json
```diff
@@ -121,7 +121,8 @@
       "packages/*"
     ],
     "globals": {
-      "react": "React"
+      "react": "React",
+      "@emotion/core": "emotionCore"
```
FLGMwt, line 125: I don't recall if I added this to whack-a-mole an error or if preconstruct added this automatically, but if this is needed, it probably shouldn't be in root, right?

Probably has something to do with me disabling `@emotion/core`'s UMD.

packages/babel-preset-css-prop/src/index.js
```diff
@@ -13,17 +13,22 @@ export default (
   api,
   { pragma, sourceMap, autoLabel, labelFormat, instances, ...options } = {}
 ) => {
+  const isAutomaticRuntime = options.runtime === 'automatic'
+  console.log({ isAutomaticRuntime })
+  const jsxOptions = isAutomaticRuntime
+    ? { importSource: '@emotion/core' }
+    : { pragma: pragmaName, pragmaFrag: 'React.Fragment' }
```
FLGMwt, line 20: `importSource` and `pragma`/`pragmaFrag` are mutually exclusive and based on the runtime https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#react-automatic-runtime-1

packages/babel-preset-css-prop/src/index.js
```diff
@@ -13,17 +13,22 @@ export default (
...
   return {
     plugins: [
-      [
+      !isAutomaticRuntime && [
```
FLGMwt, line 23: We don't need this since the module the automatic runtime uses to resolve `jsx` is based on `importSource`.

packages/core/jsx-runtime/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "main": "dist/core.cjs.js",
+  "module": "dist/core.esm.js",
+  "browser": {
+    "./dist/core.cjs.js": "./dist/core.browser.cjs.js",
+    "./dist/core.esm.js": "./dist/core.browser.esm.js"
```
FLGMwt, line 6: it's weird that these are named `core`, yeah? I tried to make it `jsx-runtime` but preconstruct overrode it. In any case, it works. We have:

`@emotion/core/dist/core.esm.js`
`@emotion/core/jsx-runtime/dist/core.esm.js`

and so on. The imports seem to work in the `minimal-react` so 🤷‍♀️ maybe it doesn't matter.

packages/core/package.json
```diff
@@ -37,15 +37,18 @@
...
   },
   "repository": "https://github.com/emotion-js/emotion/tree/master/packages/core",
   "publishConfig": {
     "access": "public"
   },
-  "umd:main": "dist/core.umd.min.js",
   "preconstruct": {
-    "umdName": "emotionCore"
+    "source": "src/index.js",
+    "entrypoints": [
```
FLGMwt, line 49: Just followed the instructions in https://preconstruct.tools/guides/adding-a-second-entrypoint

The UMD build for `jsx-runtime` failed on a "missing external" for `react/jsx-runtime`. This makes sense, but nothing I tried worked to fix it. My expectation was that adding the following to preconstruct would get things working

```js
"globals": {
  "react/jsx-runtime": "JSXRuntime"
}
```
would fix it, but it didn't. Since umd has to be on for all entrypoints or none, I disabled it for `@emotion/core` to get things working.

packages/core/src/jsx-runtime.js
```diff
@@ -0,0 +1,174 @@
+/* eslint-disable no-unused-vars */
```
FLGMwt, line 1: This file is copied almost exactly from `jsx.js` but I did what I *think* was necessary to replace calls to `React.createElement` with `jsx` from `react/jsx-runtime`.

Since most of the code doesn't care about the factory method, it should definitely factored to support both.

packages/core/src/jsx-runtime.js
```diff
@@ -0,0 +1,174 @@
...
+    )
+  }
+  return render(cache, props, null, ref)
+})
+
+if (process.env.NODE_ENV !== 'production') {
+  Emotion.displayName = 'EmotionCssPropInternal'
+}
+
+const jsx = (type, props, maybeKey) => {
```
FLGMwt, line 113: change between this and old `jsx` is that it's called w/ a props object that includes children *without` `key` and passes key as the third arg. 

Gigantic caveat: I don't really grok this code very well yet, but I was able to get it sort of working. There's definitely something broken with keys.

packages/core/src/jsx-runtime.js
```diff
@@ -0,0 +1,174 @@
...
+  createElementArgArray[1] = newProps
+
+  for (let i = 2; i < argsLength; i++) {
+    createElementArgArray[i] = args[i]
+  }
+  // $FlowFixMe
+  return reactJsx.apply(null, createElementArgArray)
+}
+
+const jsxs = jsx
```
FLGMwt, line 172: https://github.com/facebook/react/pull/18299/files#diff-97fbb0b56908c06793243318ab1cb29aR19

image

playgrounds/minimal-react/package.json
```diff
@@ -0,0 +1,47 @@
...
+      "react/react-in-jsx-scope": "off"
+    }
+  },
+  "babel": {
+    "presets": [
+      "@babel/preset-env",
+      [
+        "@babel/preset-react",
+        {
+          "runtime": "automatic"
```
FLGMwt, line 36: oh: I think this is unnecessary to configure here if babel goes after, right? 

In any case, I'd guess users should probably have a shared `reactJsxOptions` config object they use for both in babel.config.js that they use for both. Something about consistency.

playgrounds/minimal-react/src/App.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+      }}
+    />
+    <h1
+      css={{
+        color: 'hotpink',
+        animation: `${animation} 1s infinite`
+      }}
+    >
+      wow, some hotpink text!!
```
FLGMwt, line 29: all identical to the CRA playground

Andarist:
Please note that I have seen your PR - I just haven't yet time to look at it properly. We are also aware of this new feature - but it has not been released yet to React. We are also watching closely how TypeScript will handle this - there are still some unknowns for us.

packages/core/src/jsx-runtime.js
hasparus: I don't understand this part.

Looking at the initial RFC for create element changes 

> My proposal is that we compile <div>{a}{b}</div> to jsxs('div', {children: [a, b]}) and <div>{a}</div> to jsx('div', {children:a}). The jsxs function indicates that the top array was created by React. The nice property of this strategy is that even if you don't have separate build steps for PROD and DEV, we can still issue the key warning and still pay no cost in PROD.

and the line of code you linked

https://github.com/facebook/react/pull/18299/files#diff-97fbb0b56908c06793243318ab1cb29aR19

```jsx
const jsx = __DEV__ ? jsxWithValidationDynamic : jsxProd;
// we may want to special case jsxs internally to take advantage of static children.
// for now we can ship identical prod functions
const jsxs = __DEV__ ? jsxWithValidationStatic : jsxProd;
const jsxDEV = __DEV__ ? jsxWithValidation : undefined;
```

it seems to me that the difference between _jsxWithValidationStatic_ and _jsxWithValidationDynamic_ is signifcant during development, and Emotion shouldn't discard jsxs.

Andarist: If anything we should just forward to whatever React exports - but I'm quite hesitant about moving forward with this PR at this point in time as React 17 is not released yet. We'll have to wait for it and then think how to support both React 16 and React 17 at the same time or if we should cut a new major for React 17.

hasparus: > how to support both React 16 and React 17 at the same time

This would be preferable. I was undder impression that React 17 still supports React.createElement, and the only addition is `react/jsx-runtime` entrypoint.

Emotion would have to export three things to match React's exports.

1. current __jsx__ function to support old runtime, using __React.createElement__ underneath
   ```(type, props, ...children) => ReactElement```
2. a new __jsx__ function, using _jsx_ from `react/jsx-runtime` (*)
   ```(type, config, maybeKey, source?, self?) => ReactElement```
3. __jsxs__ function, using _jsxs__ from `react/jsx-runtime`

This PR already creates the the `@emotion/core/jsx-runtime` entrypoint.

(*) There's a warning about explicitly using new `jsx` functions in React docs.
It's probably meant for projects that don't implement their own `jsx` function, though.

> The functions inside react/jsx-runtime and react/jsx-dev-runtime must only be used by the compiler transform. If you need to manually create elements in your code, you should keep using React.createElement. It will continue to work and is not going away.
>
> _https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html_

I'm not sure yet how `react/jsx-dev-runtime` fits the puzzle.

Andarist: > This would be preferable. I was under impression that React 17 still supports React.createElement, and the only addition is react/jsx-runtime entrypoint.

Yes, that's correct.

> Emotion would have to export three things to match React's exports.

That's correct as well - the only problematic part is that TypeScript doesn't import fully automatic runtime with custom modules 😞 but that's kinda their limitation and not ours, even though it affects us greatly.

> (*) There's a warning about explicitly using new jsx functions in React docs.
> It's probably meant for projects that don't implement their own jsx function, though.

Yes, I believe we don't have to care about this warning - API of those functions have to be stable after all, otherwise, compilers would break, so we should be free to use them directly.

> I'm not sure yet how react/jsx-dev-runtime fits the puzzle.

I think it's the same - we should just forward stuff to the appropriate function and that's all.

morlay:
could we just added a hack `jsx-runtime` until next `react/jsx-runtime` support old version (now seems only 17, if changes a lot, may broken @emotion for old version).  

```js
const _jsx = require("@emotion/core").jsx;

exports.Fragment = require("react").Fragment;
exports.jsx = (Component, { children, ...otherProps }, maybeKey) => _jsx(Component, maybeKey ? { key:maybeKey, ...otherProps }: otherProps, children);
exports.jsxs = (Component, { children, ...otherProps }, maybeKey) => _jsx(Component, maybeKey ? { key:maybeKey, ...otherProps }: otherProps, ...children);
```

Andarist:
Yes, we probably could add smth like this to bring automatic runtime for Emotion users stuck with React 16 but it seems to be counter-productive. I would much more prefer to push the ecosystem forward and use the actual new factories available in the React 17.

SimenB:
>users stuck with React 16

React just had releases backports of the new transform to v0.14, v15 and v16

https://github.com/facebook/react/releases/tag/0.14.10
https://github.com/facebook/react/releases/tag/v15.7.0
https://github.com/facebook/react/releases/tag/v16.14.0

Andarist:
Wow, this is unexpected. In such a case I guess we should aim at shipping this for Emotion 10 as well.

@FLGMwt are you still eager to work on this PR? or should I take over? I would appreciate any help so you working on it and me giving feedback, reviews etc is definitely what I would prefer. Please let me know.

packages/core/src/jsx-runtime.js
```diff
@@ -0,0 +1,174 @@
...
+
+export { jsx, jsxs, Fragment, createElement }
```
morlay, line 174: `createElement` not need to export in `@emotion/core/jsx-runtime`

if have to, need to export in `@emotion/core`. but i think this may some order issue of babel.

result by my custom `jsx-runtime`

image

morlay: got it, codes like `<span {...v} key={i}>{i}</span>` will be use `createElement` of `@jsxImportSource`

@babel/plugin-transform-react-jsx: ^7.10.4

```jsx
const SomeComponent = () => {
  return (
    <>
      <div>123</div>
      <span>{undefined}</span>
      <>
        {[{}, {}].map((v, i) => <span {...v} key={i}>{i}</span>)}
      </>
    </>
  );
};

// ↓ ↓ ↓ ↓ ↓ ↓

const SomeComponent = () => {
  return /*#__PURE__*/_jsxs(_Fragment, {
    children: [/*#__PURE__*/_jsx("div", {
      children: "123",
    }), /*#__PURE__*/_jsx("span", {
      children: undefined,
    }), /*#__PURE__*/_jsx(_Fragment, {
      children: [{}, {}].map((v, i) => /*#__PURE__*/_createElement("span", {
        ...v,
        key: i,
      }, i)),
    })],
  });
}; 
```

morlay: https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx-experimental/src/index.js#L270
```
  // We want to use React.createElement, even in the case of
  // jsx, for <div {...props} key={key} /> to distinguish it
  // from <div key={key} {...props} />. This is an intermediary
  // step while we deprecate key spread from props. Afterwards,
  // we will stop using createElement in the transform.
```

may no need to export the `createElement`

morlay:
@Andarist 

if the fallback logic still needed. https://github.com/facebook/react/issues/20031

Have to keep old `jsx` of `@emotion/core`, and alias to `createElement`.

Andarist:
I've pushed some minor commits, cleaned up repo-related stuff etc. Gonna continue work on this later today/tomorrow.

TODO:
- [ ] review & try to dedupe the classic/automatic runtimes in our code
- [ ] tests
- [ ] add dev runtime

packages/core/jsx-runtime/package.json
Andarist: yes, it doesnt matter - i agree the naming convention used for those files is confusing, we have already agreed in the past on making that change but I'm not sure when it will land

Andarist:
I believe this is ready to review and all - the only leftovers to do are:
- docs
- changeset

Gonna prepare them 🔜

package.json
```diff
@@ -121,7 +122,9 @@
       "packages/*"
     ],
     "globals": {
-      "react": "React"
+      "react": "React",
+      "react/jsx-runtime": "ReactJSX",
+      "react/jsx-dev-runtime": "ReactJSXDev"
```
mitchellhamilton, line 127: These don't actually exist but we still want UMD builds for Emotion so these are just to make it so Preconstruct can make a build

packages/core/src/jsx-dev-runtime.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+import * as ReactJSXRuntimeDev from 'react/jsx-dev-runtime'
```
billyjanitsch, line 2: OOC do you have a particular reason to prefer the namespace import over named ones? Does it work around webpack's `(0, r.jsx)(...)` compile output, or something?

billyjanitsch:
@Andarist you mentioned wanting some help with docs in https://github.com/emotion-js/emotion/issues/2052#issuecomment-713343719. I'm happy to help; can you answer a few questions?
- Since this hasn't been merged, would you like me to submit a PR to @FLGMwt's fork?
- Is there anything specific that you'd like documented other than the new `runtime` option?
- If there's anything high-level, can you give me a general pointer to the area of the docs where you'd like it mentioned?

Andarist:
> Since this hasn't been merged, would you like me to submit a PR to @FLGMwt's fork?

I don't think it matters much but actually, it's probably best to prepare a separate PR against master because we auto-deploy docs from the master and there will be a gap between merging this PR and doing a final release having those changes (even if just a few hours or whatever). So it would be best to merge a docs update **after** we release this to avoid potential confusion among users looking at the docs that got updated too quickly 😉 

> Is there anything specific that you'd like documented other than the new runtime option?

We should add:
- specific docs for things affected directly here - so probably just the docs for the preset.
- explain different options on how one can integrate Emotion nowadays in their app, mentioning both new runtimes, older `@jsx` etc, each should explain briefly when it should be used (mainly env/tool requirements)

At all places where we are going to mention new runtimes, we should mention that they require the latest React 16 or React 17 and the latest version of `@emotion/core`.

> If there's anything high-level, can you give me a general pointer to the area of the docs where you'd like it mentioned?

The most detailed "guide" explaining the old and new approaches should probably be mentioned here: https://emotion.sh/docs/css-prop

billyjanitsch:
Thanks, this is really useful. I'll try to get a PR up tonight.

packages/core/src/jsx-dev-runtime.js
billyjanitsch: FYI while exploring this, I discovered that the namespace import is actually preferable for the reason I suspected (it bundles more efficiently), but only if the runtime is CJS. So it makes no difference for Emotion but it does make a difference for React. I opened https://github.com/facebook/react/issues/20113 to suggest changing the React Babel plugin output to use a namespace import. 🙂

billyjanitsch:
Sorry @Andarist, the docs are taking me longer than expected, mainly because I only use the CSS prop Babel transform and hadn't initially considered the pragma option. The product of those two options and the two runtime options all need to be documented, and it's tricky to explain four possible options with slightly different constraints/outcomes in a way that isn't confusing. I'll try to have a PR up soon.

.changeset/nine-waves-play.md
```diff
@@ -0,0 +1,7 @@
+---
+'@emotion/core': minor
+---
+
+Support for [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) has been added. They require compatible React versions and shouldn't be manually used.
+
+To use them them you can use the new `@jsxImportSource @emotion/core` pragma instead of the old `@jsx jsx` or you can use `@emotion/babel-preset-css-prop` with `{ runtime: 'automatic' }` option to have it handled automatically for you for the whole project.
```
mitchellhamilton, line 7:
```suggestion
To use them you can use the new `@jsxImportSource @emotion/core` pragma instead of the old `@jsx jsx` or you can use `@emotion/babel-preset-css-prop` with `{ runtime: 'automatic' }` option to have it handled automatically for you for the whole project.
```

mitchellhamilton (approved):

cybercitizen7:
Hey guys,

Was this fixed? I just created a new React App and installed emotions library. When I try to use it like this:
```
/** @jsx jsx */
import {jsx} from '@emotion/core'
```

But I still get the error message:
`pragma and pragmaFrag cannot be set when runtime is automatic.`

Here is my `package.json `dependencies:
```
  "dependencies": {
    "@testing-library/jest-dom": "^5.11.9",
    "@testing-library/react": "^11.2.5",
    "@testing-library/user-event": "^12.6.3",
    "react": "^17.0.1",
    "react-dom": "^17.0.1",
    "react-scripts": "4.0.2",
    "web-vitals": "^1.1.0"
  },
```

I haven't done any other changes, just a pure newly created React App with command `create-react-app`

karlhorky:
You're probably looking for:

```js
/** @jsxImportSource @emotion/react */
```

As in the docs:

> If you are using a zero-config tool with automatic detection of which runtime (classic vs. automatic) should be used and you are already using a React version that has the new JSX runtimes (hence runtime: 'automatic' being configured automatically for you) such as Create React App 4 then /** @jsx jsx */ pragma might not work and you should use /** @jsxImportSource @emotion/react */ instead.

Source: https://emotion.sh/docs/css-prop#jsx-pragma:~:text=Create%20React%20App%204%20then%20%2F**,use%20%2F**%20%40jsxImportSource%20%40emotion%2Freact%20*%2F%20instead.

PR #1971, opened 2020-08-13
adds translate to the list of valid props

mjcampagna:
**What**:

Adds `translate` to the list of valid props.

**Why**:

Because it's a valid HTML attribute:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1759

... and we'd like to be able to use it.

**How**:

Just added it to the list. Pretty straight-forward.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [x] Code complete
- [ ] Changeset N/A

### DIFF
--- a/.changeset/zany-cougars-shop.md
+++ b/.changeset/zany-cougars-shop.md
@@ -1,0 +1,5 @@
+---
+"@emotion/is-prop-valid": minor
+---
+
+`translate` got added to the list of the valid props.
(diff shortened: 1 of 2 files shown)


Andarist:
Whitelisting properties strikes again. It seems that it's a recommendation to support this as per HTML 5.1 which has been released in 2016. What's your opinion about this - is it OK to add this to a whitelist of the current major version of Emotion & Styled-Components? cc @mitchellhamilton @probablyup

Andarist:
@mitchellhamilton @probablyup 🏓

probablyup (approved):

mitchellhamilton (approved):

Andarist:
@mjcampagna thanks!

PR #1973, opened 2020-08-14
Improve regex for getting labels on Node.js

santialbo:
**What**:
I'm trying to fix some rehydration missmatch warnings using Chakra-UI (https://github.com/chakra-ui/chakra-ui/issues/1446 , subsequent comments are discussing some other issue).
I found that the root of the issue was that on the node side the stack trace looked like this:
```
Error
    at jsx (...../node_modules/@emotion/core/dist/core.cjs.dev.js:246:17)
    at jsx (...../node_modules/@chakra-ui/system/dist/cjs/jsx.js:57:20)
    at Object.Collapse [as render] (...../node_modules/@chakra-ui/collapse/dist/cjs/collapse.js:119:26)
    ...
```
and the regex for getting the label was missing the name.

I'm not sure why the stack trace looks like this on the server side but this case can be easily handled by tweaking the regex a bit. Please let me know if I'm missing something or if it's there a workaround to this.

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/.changeset/six-geckos-battle.md
+++ b/.changeset/six-geckos-battle.md
@@ -1,0 +1,5 @@
+---
+"@emotion/core": patch
+---
+
+Fixed label extraction from the stack traces in node for components wrapped in `React.forwardRef`. This has affected only development builds.
--- a/packages/core/src/jsx.js
+++ b/packages/core/src/jsx.js
@@ -154,7 +154,7 @@ export const jsx: typeof React.createElement = function(
     if (error.stack) {
       // chrome
       let match = error.stack.match(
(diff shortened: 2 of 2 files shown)


Andarist:
Thanks for the report and the fix. I've looked into the repro in the linked issue and my guess is that webpack-processed bundles result in a slightly different stack traces than when the code gets executed "as is". The stack trace produced in node (the one that has failed to match against our RegExp) is more "correct" - I was able to get the same one in Chrome when I have executed code in the console. My first intuition was that this is caused by enabled source maps but I couldn't make it to produce "correct" stack traces even after disabling source maps.

Anyhow - there is no point in investigating this further. Going to merge your fix soon.

PR #1994, opened 2020-09-04
Remove all sourcemaps

danielmoore:
**What**: This change removes _all_ sourcemaps from rendered CSS during serialization, (as opposed to just the first one). This seems to happen when composing serialized styles together. 

**Why**:  Here's the (very contrived) test case:

```jsx
test('Emotion sourcemaps composition', () => {
  const bgColor = fill => css({ backgroundColor: fill });

  const Box = styled.div(
    props => ({
      '@media (max-width: 500px)': bgColor(props.fill[0]),
      '@media (min-width: 501px)': bgColor(props.fill[1]),
    }),
    props => ({ width: props.size[0], height: props.size[1] }),
  );

  const { container } = render(<Box fill={['blue', 'green']} size={[20, 30]} />);

  expect(container.firstChild).toMatchSnapshot();
});
```

Without sourcemaps:
```css
.emotion-0 {
  width: 20px;
  height: 30px;
}

@media (max-width:500px) {
  .emotion-0 {
    background-color: blue;
  }
}

@media (min-width:501px) {
  .emotion-0 {
    background-color: green;
  }
}
```

With sourcemaps:

```css
@media (max-width:500px) {
  .emotion-0 {
    background-color: blue;
  }

@media (min-width:501px) {
    .emotion-0 {
      background-color: green;
      width: 20px;
      height: 30px;
    }
}
}
```

**How**: This is a backport of a fix available in `next` by simply cherry-picking 5c55fd1. 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [x] Code complete 
- [x] Changeset 

I'm not sure how to add this test case to the test suite. I'd be happy to add it if someone could suggest a strategy.

### DIFF
--- a/.changeset/lazy-elephants-stare.md
+++ b/.changeset/lazy-elephants-stare.md
@@ -1,0 +1,6 @@
+---
+'@emotion/core': patch
+'@emotion/serialize': patch
+---
+
+Fix to what location generated source maps are pointing in case of composed styles.
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -332,7 +332,7 @@ let labelPattern = /label:\s*([^\s;\n{]+)\s*;/g
 
 let sourceMapPattern
 if (process.env.NODE_ENV !== 'production') {
-  sourceMapPattern = /\/\*#\ssourceMappingURL=data:application\/json;\S+\s+\*\//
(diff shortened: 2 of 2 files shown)


Andarist:
We’d really like to avoid releasing v10 without a serious reason to do so - I hope to release v11 soon so everyone can benefit from all things included in it. Its already very stable and used at scale in production - I encourage you to just try it out.

Issue #1987, opened 2020-09-01
`ref` prop missing in `@emotion/styled` in Emotion 11


RoystonS:
**Current behavior:**

The `ref` property doesn't appear in the typing for styled components with Emotion 11.

That causes this error when compiling with TypeScript and attempting to provide a `ref` prop:
```
 Property 'ref' does not exist on type 'IntrinsicAttributes & Pick<FixedSizeGridProps, "columnCount" | "columnWidth" | "height" | "width" | "children" | "itemKey" | "overscanCount" | ... 16 more ... | "rowCount"> & Partial<...> & Partial<...> & { ...; } & { ...; }'.
```

(This may be related to issue #1942 which was for `@emotion/native`?)

**To reproduce:**

Basically, use `styled` to style a component:
```
const MyStyledComponent = styled(MyComponent)({
  background: "pink"
});
```
then use it with a `ref` prop:
```
  <MyStyledComponent ref={ref2} message="MyStyledComponent" />
```

Here's a CodeSandbox illustrating the issue:
https://codesandbox.io/s/emotion-11-ref-issue-wi5dt?file=/src/index.tsx

**Expected behavior:**

Wrapping a component with `styled` should not remove the typing for the `ref` prop.

**Environment information:**

- `react` version: 16.13.1
- `emotion` version: 11.0.0-next.15

Andarist:
Yes, this looks very similar to the issue regarding `@emotion/native` - my bet is the issue is the same. 

Question: why do u want to access a ref of a class component?

RoystonS:
In the actual use-case in our full codebase we're using the third-party `react-window` library and that has a `FixedSizeList` class component which exposes a `scrollTo` method which we need to be able to call, sadly.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-window/index.d.ts#L375

Andarist:
Thank you for sharing - the problem is that `ref` is not part of the props of your component. It's part of the global `JSX.IntrinsicClassAttributes` which is an extension used when type-checking JSX-created elements (`<MyClass ref={() => {}}/>`). We can't test if a received component is a class though and we could add support for `ref` that way - gonna discuss this with @JakeGinnivan some time this week (hopefully).

RoystonS:
Thanks. So this is only an issue with class components?

Andarist:
Yes, i believe so. We have type tests for other component types and it works fine.

PR #1996, opened 2020-09-04
Fixed an issue with wrapped class components not having a type for the `ref` prop

Andarist:
fixes https://github.com/emotion-js/emotion/issues/1987

### DIFF
--- a/.changeset/cyan-apes-join.md
+++ b/.changeset/cyan-apes-join.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fixed an issue with wrapped class components not having a type for the `ref` prop.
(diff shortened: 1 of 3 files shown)


Andarist:
Going to merge this, but feel free to review at any time. I've noticed that this still has an issue:
```ts
  class FixedSizeList extends React.Component {
    scrollTo = () => {}

    render() {
      return null
    }
  }

  const StyledFixedList = styled(FixedSizeList)`
    color: hotpink;
  `

  // this has wrong type of the `ref`
  const WrappedStyledFixedList = styled(StyledFixedList)``.withComponent('li')
```

but the issue is not really a regression from this PR because this doesn't work correctly for a very similar reason:
```ts
  const StyledLabel = styled('label')`
    color: hotpink;
  `

  // this has wrong type of the `ref`
  const WrappedStyledLabel = styled(StyledLabel)``.withComponent('span')
```

PR #2000, opened 2020-09-06
Concept of native inferred typing improvement

louisgv:
**What**:

I would like to propose a solution for better typing that consume existing type instead of explicitly stating each export for component for emotion native styled

**Why**:

Since react-native export component will change in new update, it is likely better to get the type from source, also keep typing DRY, also since react-native is not a focus of emotion.

**How**:

Import ReactNative type and instead of explicitly stating the keys, put the object in wholesale. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset


packages/native/src/index.js
```diff
@@ -5,46 +5,7 @@ import { styled } from './base'
...
-  'TextInput',
-  'ToolbarAndroid',
-  'TouchableHighlight',
-  'TouchableNativeFeedback',
-  'TouchableOpacity',
-  'TouchableWithoutFeedback',
-  'View',
-  'ViewPagerAndroid'
-]
+const components = Object.keys(reactNative)
```
Andarist, line 8: Expecting all of those to be components seems to be rather inaccurate/not future-proof, isnt it?

packages/native/types/index.d.ts
```diff
@@ -49,60 +47,9 @@ type HostClassComponent<
...
-  FlatList: HostClassComponent<ReactNative['FlatList']>
-  Image: HostClassComponent<ReactNative['Image']>
-  ImageBackground: HostClassComponent<ReactNative['ImageBackground']>
-  KeyboardAvoidingView: HostClassComponent<ReactNative['KeyboardAvoidingView']>
-  ListView: HostClassComponent<ReactNative['ListView']>
-  Modal: HostClassComponent<ReactNative['Modal']>
-  NavigatorIOS: HostClassComponent<ReactNative['NavigatorIOS']>
-  Picker: HostClassComponent<ReactNative['Picker']>
-  PickerIOS: HostClassComponent<ReactNative['PickerIOS']>
-  Pressable: HostFunctionComponent<ReactNative['Pressable']>
```
Andarist, line 75: This was using the HostFunctionComponent - u assume now that all components are typed as classes, this probably wont work correctly

packages/native/src/index.js
louisgv: My initial thought was maybe it could be done that way just for the sake of typing since majority of user who use styled.* would probably only do it to those component they know - but after thinking a bit, it'd be confusing for new developer. The new iteration should address this

packages/native/types/index.d.ts
```diff
@@ -13,6 +13,9 @@ import {
   ReactNativeStyleType
 } from './base'
 
+import * as classComponent from '../src/classComponent'
```
louisgv, line 16: Since this module is shipping its source and the definition are stripped after compile, I think this should be ok? The key idea here is to keep the new work on this dry by just altering the js export of `react-native`.

PR #2003, opened 2020-09-07
Fixed an issue with orphaned pseudos having their selector doubled for descendant at-rules

### LINKED ISSUE #2002: Emotion 11 mangles selectors when mixing pseudo-elements and media queries
**Current behavior:**

If we have a pseudo-element such as `::before` (or `dialog`'s `::backdrop`) and attempt to style it using media queries, the generated CSS selector is incorrect.  It worked fine with Emotion 10.

A little code:
```
const MyControl = styled.div({
  "::before": {
    content: '"::before content"',
    background: "pink",
    "@media screen and (max-width: 800px)": {
      background: "cyan"
    }
  }
});
```

This is the CSS that Emotion 10 generates with that input:
```
.css-z60a1g::before{content:"::before content";background:pink;}
@media screen and (max-width:800px){.css-z60a1g::before{background:cyan;}}
```

This is the CSS that Emotion 11 generates with that input:
```
.css-z60a1g::before{content:"::before content";background:pink;}
@media screen and (max-width: 800px){.css-z60a1g::before::before{background:cyan;}}
```
(Note the doubled-up `::before` selector.)

This then fails to apply the style correctly because it's attempting to target the 'before' pseudo element of the before pseudo 
element.

**To reproduce:**

Here's the above code running in CodeSandbox:

- Emotion 10: https://codesandbox.io/s/emotion-10-pseudo-media-query-busww?file=/src/index.tsx
- Emotion 11: https://codesandbox.io/s/emotion-11-pseudo-media-query-issue-1e80z?file=/src/index.tsx

If you make the screen narrower it should trigger the media query into making the background cyan rather than pink.  That only works on Emotion 10.

This is a cut-down example and not the same use case we found:  we first noticed this when our Visual Regression tests spotted an issue with our `dialog` `::backdrop` styling not working with print media queries.  But that's a little trickier to reproduce and diagnose.  Using `::before` demonstrates the issue more easily.

**Expected behavior:**

It should generate a correct CSS selector as Emotion 10 did.

**Environment information:**

- `react` version: 16.13.1
- `emotion` version: 11.0.0-next.15

Andarist:
fixes #2002

### DIFF
--- a/.changeset/gentle-colts-think.md
+++ b/.changeset/gentle-colts-think.md
@@ -1,0 +1,5 @@
+---
+'@emotion/cache': patch
+---
+
+Fixed an issue with orphaned pseudo selectors (e.g. `:hover` - where `&:hover`, `div:hover`, etc are not considered orphaned) having the context selector (the one computed based on all ancestor levels selectors) doubled in a descendant at-rule.
--- a/packages/cache/src/stylis-plugins.js
+++ b/packages/cache/src/stylis-plugins.js
@@ -63,11 +63,19 @@ export let compat = element => {
     return
   }
 
-  let parent = element
+  let { parent } = element
+  let isImplicitRule =
+    element.column === parent.column && element.line === parent.line
 
-  do {
+  // if this is an implicitly inserted rule (the one eagerly inserted at the each new nested level)
+  // then the props has already been manipulated beforehand as they that array is shared between it and its "rule parent"
+  if (isImplicitRule) {
+    return
+  }
+
+  while (parent.type !== 'rule') {
     parent = parent.parent
-  } while (parent.type !== 'rule')
+  }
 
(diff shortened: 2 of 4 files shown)


.changeset/gentle-colts-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/cache': patch
+---
+
+Fixed an issue with orphaned pseudo selectors (e.g. `:hover` - where `&:hover`, `div:hover`, etc are not considered orphaned) having the context selector (the one computed based on all ancestor levels selectors) doubled in a descendant at-rule.
```
Andarist, line 5: quite a mouthful, but not sure how to describe it better and more precise

mitchellhamilton (approved):

Issue #1955, opened 2020-08-03
Theming in react-native


chochihim:
**The problem**
In web, I can do following with css props:
```js
import React from "react";
import { css } from "@emotion/core";

const myStyle = theme => css`
  background-color: ${theme.backgroundColor};
`;

const MyComponent = () => <div css={myStyle}>Foo</div>;

export default MyComponent;
```
The theme object is abstracted from the component.

However, on react-native I need to explicitly provide the theme object using `useTheme`:
```js
import React from "react";
import { View, Text } from "react-native";
import { css } from "@emotion/native";
import { useTheme } from "emotion-theming";

const myStyle = (theme) => css`
  background-color: ${theme.backgroundColor};
`;

const MyComponent = () => {
  const theme = useTheme();
  return <View style={myStyle(theme)}><Text>Foo</Text></View>;
};

export default MyComponent;
```

I think it is more consistent if we also abstract the theme object on react-native.

Andarist:
In the `@emotion/native` `props.style` are being concatenated with styles passed to a `styled` factory so this is treated as any interpolation and thus you can access theme by destructuring it from your `props`:
```diff
-const myStyle = (theme) => css`
+const myStyle = ({ theme }) => css`
  background-color: ${theme.backgroundColor};
`;
```

Whereas in the web version `css` prop is handled separately.

I agree that this doesn't look consistent though, so I will evaluate if we should change this in `@emotion/native`.

chochihim:
> In the `@emotion/native` `props.style` are being concatenated with styles passed to a `styled` factory so this is treated as any interpolation and thus you can access theme by destructuring it from your `props`:
> 
> ```diff
> -const myStyle = (theme) => css`
> +const myStyle = ({ theme }) => css`
>   background-color: ${theme.backgroundColor};
> `;
> ```

Does this mean that I need to use `styled` API like this?
```js
import styled, { css } from "@emotion/native";

const myStyle = ({theme}) => css`
  background-color: ${theme.backgroundColor};
`;

const MyVIew = styled.View`
  ${myStyle}
`;

const MyComponent = () => {
  return <MyVIew><Text>Foo</Text></MyVIew>;
};
```
But I want to stick with the "css prop" API. I wonder if we can have this in `@emotion/native`

Andarist:
> Does this mean that I need to use styled API like this?

No, you can do this right now:
```js
const myStyle = ({theme}) => css`
  background-color: ${theme.backgroundColor};
`;

[...]

PR #2014, opened 2020-09-14
Disallow functions as value of style prop in libs related to React Native

Andarist:
Partially addresses https://github.com/emotion-js/emotion/issues/1955


packages/primitives-core/src/styled.js
```diff
@@ -54,6 +54,11 @@ export function createStyled(
 
         let stylesWithStyleProp = styles
         if (props.style) {
+          if (typeof props.style === 'function') {
```
Andarist, line 57: alternatively, we could resolve those with `theme` - the main problem was that those were resolved with `props` which didn't even match css prop signatures

I'm mostly interested in avoiding any other solutions from being breaking changes in v11 - we don't have to have a final solution for this problem right now, we only need to make sure that we can implement a final solution within v11 line. But also - I'm not super interested in working on any improvements here 🤷‍♂️ The community must step up if they want anything from the libs related to React Native.

mitchellhamilton: This still allows functions here doesn't it because I assume you can pass in an array with a function in it and it'll be called?

Why don't we make `newProps.style = [emotionStyles, props.style]`?

Andarist: so that way RN should just resolve on its own the array and we omit the `props.style` from being resolved by Emotion? that would be sweet!

mitchellhamilton: Yeah

Andarist: I've used your advice and changes the code accordingly. Had to update snapshots and I've decided to use actual web implementation of `react-primitives` in tests so things get tested in more e2e manner - with the previous mock array got forwarded to the native (JS)DOM's style prop which didn't look OK.

packages/primitives-core/package.json
```diff
@@ -13,7 +13,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "css-to-react-native": "^2.2.1"
+    "css-to-react-native": "^3.0.0"
```
mitchellhamilton, line 16: There should be a changeset for this, no?

Andarist: added

mitchellhamilton (approved):

PR #2023, opened 2020-09-24
RFC: Expose interface for custom CSSObject declaration

tpict:
**What**:

Modifications to `@emotion/serialize/types` so that the user may specify a custom `CSSObject` declaration in preference to the default.

**Why**:
- To allow users to constrain what styles may be applied at any leaf in their React tree
- To allow users to opt-out of arbitrary CSS selectors etc

**How**:
The existing `CSSObject` is renamed to `DefaultCSSObject`. A new, empty interface named `CustomCSSObject` is added. Finally, a new type named `CSSObject` is exported that takes the value of `CustomCSSObject` is it's provided, otherwise `DefaultCSSObject`. Users can augment `CustomCSSObject` in their project to override the value of `CSSObject` globally. This shouldn't have any impact on users who don't provide their own declaration.

**Checklist**:
- [x] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset

I'm not sure if/how this can be tested without splitting `dtslint` into two stages, since declaring a `CustomCSSObject` anywhere changes the value globally. Feedback appreciated!

### DIFF
--- a/.changeset/early-jokes-applaud.md
+++ b/.changeset/early-jokes-applaud.md
@@ -1,0 +1,5 @@
+---
+'@emotion/serialize': minor
+---
+
+Add interface for custom CSSObject declaration
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -310,6 +310,28 @@ declare module '@emotion/react' {
 }
 ```
 
+### Define a CSS Interface
+
+By default, the interface of style objects is based on
+[CSSType](https://github.com/frenic/csstype). This library provides type safety where possible without
+restricting valid CSS values.
+
+If you want to constrain which styles are globally allowed in your project, you can provide your own
+`CustomCSSObject` interface:
+
+_emotion.d.ts_
+
+```typescript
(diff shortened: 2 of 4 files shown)


Andarist:
Please describe what issue this tries to solve.

tpict:
The issue is the difficulty of enforcing design constraints in a large-scale React application where styles can be freely applied to any component. This change would allow design constraints to be enforced at compile time. For example, if my designer prescribes a palette of 12 colors only, any violations of that constraint could be reported in CI.

It would be useful even for ad-hoc styling. CSSType has certain limitations. The [README](https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors) describes how to work around missing properties, but missing values are a possibility too:
```ts
// this results in a TS2322 error
<div css={{ fontWeight: "var(--verybold)" }}>
  Hello world
</div>
```

Of course you can work around this on a case-by-case basis, but that gets pretty tiresome, and there's no autocomplete hint for values that are implicitly valid in your environment.

Issue #2025, opened 2020-09-28
Error from Enzyme serializer when creating snapshot for component accepting another element as prop


ctumolosus:
**Current behavior:**

The following error is displayed when creating a snapshot for a React component accepting another element as prop.

```
PrettyFormatPluginError: Cannot redefine property: propsTypeError: Cannot redefine property: props
```

**To reproduce:**

You should be able to see the error on this sandbox URL:
https://codesandbox.io/s/emotion-issue-template-forked-pvxld?file=/index.test.js

Basically, setup your environment to work with Emotion 11, Enzyme, and Jest. A simple test for a component accepting another element as prop should reveal the error afterwards.

```jsx
import { configure, mount } from "enzyme";
import { createEnzymeSerializer } from "@emotion/jest";
import React from "react";
import Adapter from "enzyme-adapter-react-16";

configure({ adapter: new Adapter() });
expect.addSnapshotSerializer(createEnzymeSerializer());

const FixMe = ({ element }) => element;

test("works", () => {
  const wrapper = mount(<FixMe element={<button>Foo</button>} />);
  // PrettyFormatPluginError: Cannot redefine property: propsTypeError: Cannot redefine property: props
  expect(wrapper).toMatchSnapshot();
});

```

**Expected behavior:**

Should create snapshot without errors.

**Environment information:**

- `react` version: 16.13.1
- `emotion` version: 11.0.0-next.17

Andarist:
Thank you for reporting your issue with the appropriate repro case! I've prepared a fix for it here: https://github.com/emotion-js/emotion/pull/2027

A sideways question - I've noticed that our serializer serializes "mounted" components by serializing both the full, mounted, tree but also by serializing all props and keeping intermediate components in the output. Would you consider this correct when dealing with `mount`? Or would you rather like to see a "flattened" output - basically what you would actually see rendered in the browser?

Currently, things like [this](https://github.com/emotion-js/emotion/blob/d55556ec2f3bbaf8ed8b3b3574069d93a3c83334/packages/jest/test/__snapshots__/react-enzyme.test.js.snap#L122-L140) happen and such input:
```js
const Test1 = () => <span>BAZINGA</span>
const Test2 = () => <Test1 test="foo" />
mount(<Test2 />)
```
yields such output:
```
<Test2>
  <Test1
    test="foo"
  >
    <span>
      BAZINGA
    </span>
  </Test1>
</Test2>
```

Andarist:
[...]

PR #2027, opened 2020-10-01
Fixed an issue with serializing Enzyme's `ReactWrapper` (what is returned from `mount`) with props containing elements without the css prop

Andarist:
fixes https://github.com/emotion-js/emotion/issues/2025


package.json
```diff
@@ -228,13 +228,13 @@
...
-    "jest-environment-jsdom": "^25.4.0",
+    "jest": "^26.4.2",
+    "jest-environment-jsdom": "^26.3.0",
     "jest-in-case": "^1.0.2",
-    "jest-junit": "^10.0.0",
+    "jest-junit": "^11.1.0",
     "jest-serializer-html": "^7.0.0",
-    "jest-watch-typeahead": "^0.5.0",
-    "jsdom": "^16.2.2",
+    "jest-watch-typeahead": "^0.6.1",
```
Andarist, line 236: I've wanted to finally leverage [improved exact typeahead matching](https://github.com/jest-community/jest-watch-typeahead/pull/39) hence this update here

packages/jest/src/create-serializer.js
```diff
@@ -36,10 +36,11 @@ function getNodes(node, nodes = []) {
   return nodes
 }
 
-function copyProps(src, target) {
-  return Object.defineProperties(src, {
-    ...Object.getOwnPropertyDescriptors(target)
-  })
+function copyProps(target, source) {
```
Andarist, line 39: the previous names were confusing for me, usually when copying props from one object to another they are called the other way around (how I've adjusted them here)

packages/jest/src/create-serializer.js
```diff
@@ -49,27 +50,14 @@ function deepTransform(node, transform) {
 
   const transformed: any = transform(node)
 
-  if (transformed !== node) {
-    if (transformed.props) {
```
Andarist, line 53: I can't say that I understand the original issue in full, but I've found out that just removing this fixes the problem and no other tests fail, so... let's roll with that.

mitchellhamilton (approved):

Issue #1945, opened 2020-07-29
Loading a second emotion bundle shouldn't break the first


craigsketchley:
**Current behavior:**

I'm trying to run emotion 11 in a browser extension content script. The trouble is if that content script then runs in a page that is already running an emotion 11 app, then the app styles break. After a chat with @Andarist on the emotion-slack, he suggested I open an issue here and that this is probably the culprit: https://github.com/emotion-js/emotion/blob/1e10d8c9aa91ce443d29661a7f830190d42aae92/packages/cache/src/index.js#L55-L60

**To reproduce:**

1. Build and install a browser extension that uses emotion 11 in a content script (I've made a test extension here: https://github.com/craigsketchley/emotion-11-extension).
2. Visit a page that uses emotion 11 in production mode (I've made a test app here: https://github.com/craigsketchley/emotion-11-app).
3. Watch styles break as soon as the content script loads.

**Expected behavior:**

Loading a second bundle which includes emotion 11 shouldn't break the styles of the existing emotion 11 app already running in the page.

**Environment information:**

In the app and extension:

- `react` version: 16.13.1
- `@emotion/cache` version: 11.0.0-next.13
- `@emotion/react` version: 11.0.0-next.13
- `@emotion/styled` version: 11.0.0-next.13

Andarist:
I finally got to it and prepared a PR for this, you can track it here: https://github.com/emotion-js/emotion/pull/2028

criles25:
Hey @Andarist , i am running into a similar issue as ^.

I have a page with multiple bundles. Each of these bundles is using emotion v11.1.1 and each bundle has its own emotion cache with a unique `key`. 

Additionally, my team has a design system built with emotion and styled-system https://github.com/styled-system/styled-system. This design system gets imported as an npm package and bundled into each of the bundles. I noticed the styles from the design system are trying to write to the `css` key for each of the bundles, which ultimately breaks the styles.

You can see what I mean in this image:

![image (41)](https://user-images.githubusercontent.com/6254912/101099102-1b08f000-3592-11eb-95d7-4afe99e7f0a5.png)

criles25:
i think my issue is a duplicate of this issue https://github.com/emotion-js/emotion/issues/1386. i am going to try making `@emotion/react` external to achieve a single context.

Andarist:
Ye, thats preferable - same way that its expected that u only have a single React used on your page.

PR #2028, opened 2020-10-01
Fixed an issue with multiple Emotion caches moving SSRed styles to head at their initialization times

Andarist:
fixes https://github.com/emotion-js/emotion/issues/1945

If this gets accepted then I'll add changeset and some basic tests. Personally, I'm not super fond of this - I would much rather like to remove this initialization side-effect altogether, but 🤷‍♂️ 

This ensures that no styles are moved twice by adding a marker data attribute to:
- moved styles
- all styles inserted at runtime

### DIFF
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -41,7 +41,6 @@ let getServerStylisCache = isBrowser
     )
 
 const defaultStylisPlugins = [prefixer]
-let movedStyles = false
 
 let createCache = (options: Options): EmotionCache => {
   let key = options.key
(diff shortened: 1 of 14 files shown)


packages/cache/src/index.js
```diff
@@ -56,11 +56,14 @@ let createCache = (options: Options): EmotionCache => {
...
 
-    const ssrStyles = document.querySelectorAll(`style[data-emotion]`)
+    const ssrStyles = document.querySelectorAll(
+      `style[data-emotion]:not([data-s])`
+    )
     // get SSRed styles out of the way of React's hydration
     // document.head is a safe place to move them to
     Array.prototype.forEach.call(ssrStyles, (node: HTMLStyleElement) => {
       ;((document.head: any): HTMLHeadElement).appendChild(node)
+      node.dataset.s = 1
```
mitchellhamilton, line 66: Let's use setAttribute here instead of dataset so that this doesn't need to be polyfilled if people need to support ie10(iirc, the only things people would need to polyfill rn would be WeakMap and Map to be support ie10 and given there's basically no cost to using setAttribute, we should just do it IMO)

mitchellhamilton (commented):
✅ on the approach

mitchellhamilton (approved):

PR #2034, opened 2020-10-03
Allow ESLint 7 as peer dep of @emotion/eslint-plugin

Andarist:


mitchellhamilton (approved):
Needs a changeset, otherwise, LGTM

PR #2057, opened 2020-10-26
[TypeScript] Fix that some functions unnecessarily require their  arguments to inc…

lazytype:
…lude mutable arrays

**What**: Type definitions

**Why**: A TypeScript error is emitted when trying to pass in readonly arrays to some functions. Ex.

```tsx
const classnames = [classname1, classname2] as const;
cx(styles);
```

**How**: Updated type definitions to use `ReadonlyArray` instead of `Array`. 

Since this only relaxes types that only used in function arguments, it's not a backwards-incompatible change.

I also fixed a couple pre-existing errors coming from `yarn test:typescript`.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/serious-melons-sip.md
+++ b/.changeset/serious-melons-sip.md
@@ -1,0 +1,6 @@
+---
+'@emotion/native': patch
+'@emotion/styled': patch
+---
+
+Fixed TypeScript false positive with `styled` components
--- a/.changeset/serious-melons-sip.md
+++ b/.changeset/serious-melons-sip.md
@@ -1,6 +1,0 @@
----
-'@emotion/native': patch
-'@emotion/styled': patch
----
-
-Fixed TypeScript false positive with `styled` components
--- a/packages/native/types/base.d.ts
+++ b/packages/native/types/base.d.ts
@@ -30,7 +30,7 @@ export type ObjectInterpolation<
 
 export interface ArrayCSSInterpolation<
   StyleType extends ReactNativeStyle = ReactNativeStyle
-> extends Array<CSSInterpolation<StyleType>> {}
+> extends ReadonlyArray<CSSInterpolation<StyleType>> {}
 
 export type CSSInterpolation<
   StyleType extends ReactNativeStyle = ReactNativeStyle
@@ -39,7 +39,7 @@ export type CSSInterpolation<
 export interface ArrayInterpolation<
   MergedProps,
   StyleType extends ReactNativeStyle = ReactNativeStyle
-> extends Array<Interpolation<MergedProps, StyleType>> {}
+> extends ReadonlyArray<Interpolation<MergedProps, StyleType>> {}
 
 export interface FunctionInterpolation<
   MergedProps,
@@ -101,8 +101,12 @@ export interface CreateStyledComponent<
 > {
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
+   * @typeparam InferredAdditionalProps  Should not be explicitly specified
    */
-  <AdditionalProps extends {} = {}>(
+  <
+    AdditionalProps extends {} = {},
+    InferredAdditionalProps extends AdditionalProps = AdditionalProps
+  >(
     ...styles: ArrayInterpolation<
       ComponentProps &
         SpecificComponentProps &
@@ -110,19 +114,23 @@ export interface CreateStyledComponent<
       StyleType
     >
   ): StyledComponent<
-    ComponentProps & AdditionalProps,
+    ComponentProps & InferredAdditionalProps,
     SpecificComponentProps,
     JSXProps
   >
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
+   * @typeparam InferredAdditionalProps  Should not be explicitly specified
    */
-  <AdditionalProps extends {} = {}>(
+  <
+    AdditionalProps extends {} = {},
+    InferredAdditionalProps extends AdditionalProps = AdditionalProps
+  >(
     template: TemplateStringsArray,
     ...styles: ArrayInterpolation<
       ComponentProps &
         SpecificComponentProps &
-        AdditionalProps & { theme: Theme },
+        InferredAdditionalProps & { theme: Theme },
       StyleType
     >
   ): StyledComponent<
--- a/packages/react/types/tests-css.tsx
+++ b/packages/react/types/tests-css.tsx
@@ -32,8 +32,8 @@ css`
 // $ExpectError
 css(() => 'height: 300px;')
 
-// $ExpectError
 css`
   position: relative;
-  flexgrow: ${() => 20};
+  flexgrow: ${// $ExpectError
+  () => 20};
 `
--- a/packages/react/types/tests-css.tsx
+++ b/packages/react/types/tests-css.tsx
@@ -32,8 +32,11 @@ css`
 // $ExpectError
 css(() => 'height: 300px;')
 
+// prettier-ignore
 css`
   position: relative;
-  flexgrow: ${// $ExpectError
-  () => 20};
+  flexgrow: ${
+    // $ExpectError
+    () => 20
+  };
 `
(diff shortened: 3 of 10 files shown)


Andarist:
I'm afraid that we won't be changing this in Emotion 10, but I'm open to discussing this change for the upcoming Emotion 11. Could you rebase your work (and resolve conflicts) on top of the `next` branch?

lazytype:
@Andarist sure thing, updated

packages/native/types/base.d.ts
```diff
@@ -101,28 +101,36 @@ export interface CreateStyledComponent<
 > {
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
+   * @typeparam InferredAdditionalProps  Should not be explicitly specified
```
Andarist, line 104: Could u explain what's happening here?

.changeset/serious-melons-sip.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/native': patch
+'@emotion/styled': patch
+---
+
+Fixed TypeScript false positive with `styled` components
```
Andarist, line 6: I don't think this changeset is particularly useful - in general, we try to express user-facing changes, and AFAIK what has been published from `next` to npm did not suffer from many problems and this doesn't quite even explain what it has fixed/changed/added. 

If this is related to the introduced changes in this PR then this can be safely removed because the core of the change is already explained in the other changeset. But maybe I've misunderstood intention here.

Andarist:
Are you aware of any downsides of using readonly arrays instead of regular ones? I would like to understand the potential blowback a little bit better.

packages/react/types/tests-css.tsx
```diff
@@ -32,8 +32,8 @@ css`
 // $ExpectError
 css(() => 'height: 300px;')
 
-// $ExpectError
 css`
   position: relative;
-  flexgrow: ${() => 20};
+  flexgrow: ${// $ExpectError
```
Andarist, line 37: I think this might have broken `yarn test:typescript` task.

packages/native/types/base.d.ts
lazytype: there's an interesting bug in the following scenario:
```tsx
const Label = styled.label``
const Input = styled.input`
  & + ${Label}: {
    margin-left: 3px;
  }
`
```
`Label` is a `StyledComponent` but it since it's a function component it also matches the `FunctionInterpolation<Props>` interface. TypeScript sees this and thinks that the props accepted by the styled input component must match the props of any `FunctionInterpolation<Props>` provided to it, so the props accepted by `Input` end up being the intersection of those accepted by `input` *and* `label` elements. 

What my fix here does is make sure that the the props of a `FunctionInterpolation<Props>` is constrained by the props of `AdditionalProps` via an intermediate type `InferredAdditionalProps`, but the inferred type of `AdditionalProps` will not be affected by the props of the function passed in.

.changeset/serious-melons-sip.md
lazytype: This changeset was specifically for the `InferredAdditionalProps` change I added. Since it shouldn't have a user-visible facing effect I'll remove it.

packages/react/types/tests-css.tsx
lazytype: ah, the comment was supposed to go above the arrow function, but prettier moved it during the pre-commit hook. I've added a comment to tell prettier to ignore this line.

lazytype:
> Are you aware of any downsides of using readonly arrays instead of regular ones? I would like to understand the potential blowback a little bit better.

@Andarist when they're used within function arguments it shouldn't have any negative effect as all previous function calls should still work. There's the tiniest risk that if one of these types such as `Interpolation` is *returned* by a function, and there's code that depends on mutating the arrays within that function, they'll now get a type error. I think that's highly unlikely, and even if it does occur, easily remedied (e.g. cast the readonly array to be writeable or create a shallow copy).

Andarist:
This tradeoff seems reasonable - I can't see any obvious use case that would break because of that. Users rarely interact with those types directly.

A little bit offtopic but related questions:
1. why a mutable array is accepted by a function that is expecting a readonly one?
2. why readonly arrays are not accepted by functions expecting mutable arrays while the similar for objects is not true (see [this playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBFCm0YF4YAoQCMBWAuGA3jAA4BOIx+0pAlmAOYwC+AlCgHyFMBQ3oksAGYgQKQmQr4A5ADo5UpjACGEGP2jcE0NMJAte62FqgAmMWiWlSVKLQYBtALptknAj0MxMl-ACV4SgAm4AA2AJ4AglZKYQA81HT0nKhOmohQaN6kLEA))?

packages/native/types/base.d.ts
Andarist: > so the props accepted by Input end up being the intersection of those accepted by input and label elements.

I've just tested this on `next` branch and this doesn't happen there. How your changes have affected the inference algorithm for this? 🤔

lazytype:
Both are related to wonderful nature of TypeScript's structure-based sub-typing.

> 1. why a mutable array is accepted by a function that is expecting a readonly one?
> 2. why readonly arrays are not accepted by functions expecting mutable arrays

The reason actually doesn't have anything to do with whether the indices of the array are assignable. It's determined by the properties that are explicitly defined on these types. Mutable arrays have all the properties that readonly arrays have. Whatever properties the function would use on a readonly array, e.g. `map`, `slice`, `filter`, can be safely used on a mutable array. The reverse is not true since readonly arrays are not defined to have methods like `push`, `pop`, `shift`, `unshift`, or `splice`. 

> while the similar for objects is not true

Though the keys of `{prop: '...'} as const` are `readonly`, TypeScript considers this to be "structurally" the same as a non-const object, so they're mutually assignable.

In short, the `Readonly` aspect of `ReadonlyArray` is kind of a red herring. It might be better to look at them as "ArrayWithReadMethods" and `Array`s as a sub-type "ArrayWithReadAndWriteMethods".

Andarist:
Thank you! This makes sense.

This leaves us only with this issue: https://github.com/emotion-js/emotion/pull/2057#discussion_r519230689

@mitchellhamilton do u have any concerns regarding the `readonly` modifier?

mitchellhamilton:
i've got no concerns with that

packages/native/types/base.d.ts
Andarist: Could this be solved by using a conditional type within `Interpolation` type that would match our components as "strings" and thus not allowing them to be matched as `FunctionInterpolation<Props>`? Alternatively, we could type our components slightly differently - they are not truly `React.FC` so typing them as such is a stretch. However, from what I remember this just made things easier for us to type and a similar technique is used within `@types/react` to allow ExoticComponents to be handled by TS.

I've also thought that this should work:
```diff
diff --git a/packages/serialize/types/index.d.ts b/packages/serialize/types/index.d.ts
index 5fc08511..017731c9 100644
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -41,7 +41,7 @@ export interface CSSObject
     CSSOthersObject {}
 
 export interface ComponentSelector {
-  __emotion_styles: any
+  __emotion_styles: string
 }
 
 export type Keyframes = {
@@ -56,6 +56,7 @@ export interface ArrayInterpolation<Props>
 
 export interface FunctionInterpolation<Props> {
   (props: Props): Interpolation<Props>
+  __emotion_styles?: never
 }
 
 export type Interpolation<Props> =
```
but apparently it doesnt.

.changeset/serious-melons-sip.md
Andarist: I have concerns regarding this additional type param - would prefer solving this without introducing it. I plan to release v11 today but as this is type-only change I think it will be OK to just release it in a minor version later when we settle on a final solution.

Andarist:
@lazytype This was closed as the PR was targeting the `next` branch which got deleted. I've missed that and forgot about this PR - sorry for that! I will try to restore your commits and get this through the finish line.

PR #2063, opened 2020-10-31
Add docs about new JSX factories

Andarist:
This is super hard to document because there are many subtle details that one should account for, especially around how Babel doesn't (currently) allow `@jsxImportSource` to be used when the runtime is configured to be classic and how using `@jsx` throws when the runtime is configured to be automatic. At the same time, it's easy to overload users with those subtle details and it doesn't seem right. 

I propose rolling with smth like this and observe reported issues to assess which parts require more details to be put in the docs. At the same time, I'm still discussing the possibility to lift some of those restrictions in Babel so using this might become more straight-forward in the future

### DIFF
--- a/docs/css-prop.mdx
+++ b/docs/css-prop.mdx
@@ -55,6 +55,8 @@ This option works best for testing out the `css` prop feature or in projects whe
 
 Similar to a comment containing linter configuration, this configures the [jsx babel plugin](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx) to use the `jsx` function instead of `React.createElement`.
 
+If you are using a zero-config tool with automatic detection of which runtime (classic vs. automatic) should be used and you are already using a React version that has [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) (hence `runtime: 'automatic'` being configured automatically for your) such as Create React App 4 then `/** @jsx jsx */` pragma might not work and you should use `/** @jsxImportSource @emotion/core */` instead. Keep in mind that this also requires a compatible version of `@emotion/core` which is `^10.1.0`.
+
 > [JSX Pragma Babel Documentation](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragma)
 
 #### Import the `jsx` function from `@emotion/core`
--- a/docs/css-prop.mdx
+++ b/docs/css-prop.mdx
@@ -13,7 +13,7 @@ There are 2 ways to get started with the `css` prop.
 
 Both methods result in the same compiled code.
 After adding the preset or setting the pragma as a comment, compiled jsx code will use emotion's `jsx` function instead of `React.createElement`.
-  
+
 | | Input | Output |
 | ------ | -------------------------- | --------------------------------------------------- |
 | Before | `<img src="avatar.png" />` | `React.createElement('img', { src: 'avatar.png' })` |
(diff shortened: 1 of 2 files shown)


packages/babel-preset-css-prop/README.md
```diff
@@ -139,6 +139,8 @@ Options for both `babel-plugin-emotion` and `@babel/plugin-transform-react-jsx`
 >
 > - [`@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/en/next/babel-plugin-transform-react-jsx)
 
+You can opt into [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) by configuring this preset with `{ runtime: 'automatic' }`. Keep in mind that you have to use compatible React version (`>=16.14.0`) to use this option and a compatible version of `@emotion/core` (`>=10.1.0`).
```
billyjanitsch, line 142: As I was reading through these, I noticed that it's ambiguous whether you're supposed to also include `@babel/preset-react`. (The only mention says "If you use `@babel/preset-react` [...]".) If I understand correctly, the story is a bit more complicated with the new runtime.

If you're using the classic runtime, then you "need" the preset because it includes the source/self transforms in dev and the pure annotations in prod. Whereas in the modern runtime, that functionality has all been absorbed into `@babel/plugin-transform-react-jsx{,-dev}`, so additionally applying the preset has no effect. Do you think it's worth adding a note to explain this?

Btw, I noticed that you had a related comment:

https://github.com/emotion-js/emotion/blob/71514b06fe172517168f98321499f05e74388de6/packages/babel-preset-css-prop/src/index.js#L32-L34

Not sure if you considered this but I think another advantage you'd get from switching to the preset would be that users _never_ need to apply the React preset themselves, regardless of the runtime. Maybe something to consider for Emotion 11 if you plan to deprecate Babel 6 support there.

billyjanitsch (commented):
Sorry I didn't end up contributing these but this is a lot better than the draft I was working on anyway. 🙂 Thanks @Andarist!

docs/css-prop.mdx
```diff
@@ -45,6 +55,8 @@ This option works best for testing out the `css` prop feature or in projects whe
 
 Similar to a comment containing linter configuration, this configures the [jsx babel plugin](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx) to use the `jsx` function instead of `React.createElement`.
 
+If you are using a zero-config tool with automatic detection of which runtime (classic vs. automatic) should be used and you are already using a React version that has [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) (hence `runtime: 'automatic'` being configured automatically for your) such as Create React App 4 then `/** @jsx jsx */` pragma might not work and you should use `/** @jsxImportSource @emotion/core */` instead. Keep in mind that this also requires a compatible version of `@emotion/core` which is `^10.1.0`.
```
mitchellhamilton, line 58:
```suggestion
If you are using a zero-config tool with automatic detection of which runtime (classic vs. automatic) should be used and you are already using a React version that has [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) (hence `runtime: 'automatic'` being configured automatically for you) such as Create React App 4 then `/** @jsx jsx */` pragma might not work and you should use `/** @jsxImportSource @emotion/core */` instead. Keep in mind that this also requires a compatible version of `@emotion/core` which is `^10.1.0`.
```

mitchellhamilton (approved):

packages/babel-preset-css-prop/README.md
Andarist: > Do you think it's worth adding a note to explain this?

Might be worth adding. Could you prepare a follow-up PR for this?

> Not sure if you considered this but I think another advantage you'd get from switching to the preset would be that users never need to apply the React preset themselves, regardless of the runtime.

Yeah, I'd like to do that - just need to verify how Babel merges presets etc. Ideally even if one has React preset configured but our comes later then it should just turn off the React preset. It works like this for plugins but I'm not sure if this happens when all presets are flattened into a plugins list or does it already happen at each step of resolution or smth.

> Maybe something to consider for Emotion 11 if you plan to deprecate Babel 6 support there.

Support for Babel 6 has been removed from Emotion 11.

Issue #2065, opened 2020-11-02
10.1.0 jsx implementation causes key errors


billyjanitsch:
**Current behavior:**

This component:

```js
import * as React from 'react'

export default function MyComponent() {
  return (
    <div css={{color: 'red'}}>
      <div />
      <div />
    </div>
  )
}
```

Transpiled with `@emotion/babel-preset-css-prop@10.1.0` (`{runtime: 'classic'}`), using `@emotion/core@10.1.0`, produces the following key error:

```
Warning: Each child in a list should have a unique "key" prop. See https://reactjs.org/link/warning-keys for more information.
    at div
    at EmotionCssPropInternal
    at MyComponent
```

In general, this happens for any element that uses the `css` prop and has two or more children. If you remove the `css` prop, or if you remove either or both children, there is no key error.

**To reproduce:**

See [this CodeSandbox](https://codesandbox.io/s/emotion-1010-key-error-zeghw?file=/src/App.js). Note that the code is the transpiled output from the Babel preset. (You can't reproduce it on CodeSandbox otherwise because it only manifests with the preset.)

**Expected behavior:**

There should be no key error.

**Environment information:**

- `react` version: `^17.0.1` (but reproduces in `16.x` as well)
- `emotion` version: `10.1.0`

billyjanitsch:
The issue is with this code path:

https://github.com/emotion-js/emotion/blob/0abe324862b733d27a4b47833527fed6bc4e209d/packages/core/src/jsx.js#L20-L33

In the dev build, React runs key validation in two places: once [in the dev implementation of `createElement`](https://github.com/facebook/react/blob/56e9feead0f91075ba0a4f725c9e4e343bca1c67/packages/react/src/ReactElementValidator.js#L476-L478) and once [in the reconciler](https://github.com/facebook/react/blob/56e9feead0f91075ba0a4f725c9e4e343bca1c67/packages/react-reconciler/src/ReactChildFiber.new.js#L751). (Note: the Emotion comment links to the prod implementation of `createElement`, but the dev build actually takes [an entirely different code path](https://github.com/facebook/react/blob/56e9feead0f91075ba0a4f725c9e4e343bca1c67/packages/react/src/React.js#L62).)

[...]

PR #2066, opened 2020-11-02
Fixed an issue with React giving warning about static children not having unique keys when using classic `jsx` factory

Andarist:
fixes https://github.com/emotion-js/emotion/issues/2065

### DIFF
--- a/packages/core/src/jsx.js
+++ b/packages/core/src/jsx.js
@@ -15,20 +15,15 @@ export const jsx: typeof React.createElement = function(
     return React.createElement.apply(undefined, args)
   }
 
-  const emotionProps = createEmotionProps(type, props)
+  let argsLength = args.length
+  let createElementArgArray = new Array(argsLength)
+  createElementArgArray[0] = Emotion
+  createElementArgArray[1] = createEmotionProps(type, props)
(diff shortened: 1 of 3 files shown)


packages/core/src/jsx.js
```diff
@@ -15,20 +15,15 @@ export const jsx: typeof React.createElement = function(
     return React.createElement.apply(undefined, args)
   }
 
-  const emotionProps = createEmotionProps(type, props)
+  let argsLength = args.length
+  let createElementArgArray = new Array(argsLength)
+  createElementArgArray[0] = Emotion
+  createElementArgArray[1] = createEmotionProps(type, props)
 
-  // https://github.com/facebook/react/blob/fd61f7ea53989a59bc427603798bb111c852816a/packages/react/src/ReactElement.js#L386-L400
```
Andarist, line 20: I'm kinda not sure why I have changed this logic to replicate what React is doing under the hood. I've implemented support for those new runtimes and refactor to the classic `jsx` a few months back, so 🤷‍♂️

Andarist:
@mitchellhamilton merging this as its sort of a hotfix

PR #2078, opened 2020-11-09
Export JSX namespace from automatic runtime entries

Andarist:


packages/react/types/index.d.ts
```diff
@@ -92,41 +92,21 @@ export interface ClassNamesProps {
...
-    interface IntrinsicClassAttributes<T>
-      extends ReactJSXIntrinsicClassAttributes<T> {}
-
-    type IntrinsicElements = {
-      [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
-        css?: Interpolation<Theme>
-      }
-    }
+    export type Element = EmotionJSX.Element
+    export type ElementClass = EmotionJSX.ElementClass
```
mitchellhamilton, line 99: These should probably be interfaces no?

PR #2081, opened 2020-11-09
Fixed an issue with dev-only labels not being added when source maps were disabled

Andarist:


mitchellhamilton (approved):
I'm assuming the end of the title is "when source maps are disabled"?

Assuming that, sgtm (besides needing a changeset)

PR #2087, opened 2020-11-10
DOCS: Typescript typeof theme for less typing

dsod:
**What**: I made a small change to the typescript documentation page and the contribution guildelines.

**Why**: The current typescript docs suggest to create a type of the Theme object. I changed this to let Typescript generate the type for us, with `typeof theme`. The update to CONTRIBUTION.md was made because the installation of the project failes with a node version above 12. More specifically the compilation of the sharp dependency (https://github.com/lovell/sharp/issues/1909).

**How**: Just some text changes to the docs.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 ## Prerequisites
 
-- [Node.js](http://nodejs.org/) >= v7 must be installed.
+- [Node.js](http://nodejs.org/) >= v7 <= 12 must be installed.
 - [Yarn](https://yarnpkg.com/en/docs/install)
 
 ## Installation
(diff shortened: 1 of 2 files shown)


Andarist:
Thank you for your contribution but I'm planning to release v11 tomorrow - or a day after that, so I would like to avoid updating v10 docs right now. This is also subjective and could be documented as an alternative way for defining theme type, I'm not 100% sold that this should replace what has been documented before.

dsod:
Sure, i will wait until the v11 is released and update the PR per your feedback. Thank you and good luck with the release! :)

Andarist:
The Emotion 11 got released a few hours ago so if you want to improve this area of the docs please take a look at the current docs and adjust your PR in a way that makes sense to you.

PR #2101, opened 2020-11-13
Improved tree-shakeability of the `@emotion/react` package

### LINKED ISSUE #2100: Webpack is unable to tree-shake unused theming code
Related to https://github.com/emotion-js/emotion/issues/2097.

**Current behavior:**

`withTheme` is now exported from `@emotion/react`. It uses `hoist-non-react-statics`:

https://github.com/emotion-js/emotion/blob/f4b9dc6ea6863e499586a61b9fd64f2c2cdf4092/packages/react/src/theming.js#L75

Unfortunately, this package doesn't provide an ESM entry point, which interferes with webpack's ability to tree-shake it if it's unused. `hoist-non-react-statics` also [depends on `react-is@16`](https://github.com/mridgway/hoist-non-react-statics/blob/f3de655212b47f20b9c6a27e89a60d23f358925a/package.json#L25), which is especially unfortunate for React 17 projects.

**To reproduce:**

See [this repo](https://github.com/billyjanitsch/emotion-theming). Specifically, note that [this entrypoint](https://github.com/billyjanitsch/emotion-theming/blob/01f2886939a889729dcf8b78edbecbeaa67667b6/src/index.js) results in [this bundle](https://github.com/billyjanitsch/emotion-theming/blob/01f2886939a889729dcf8b78edbecbeaa67667b6/dist/main.js), which includes the entirety of `hoist-non-react-statics` and `react-is`. (n.b. I ran Prettier on the bundle so that it's intelligible, but it was generated in prod mode.)

**Expected behavior:**

I'd expect `withTheme`, as well as all of `hoist-non-react-statics` and `react-is`, to be pruned from the bundle. I don't personally use Emotion's theming features so this would be really nice. 🙂 

There's an [open PR to add ESM support](https://github.com/mridgway/hoist-non-react-statics/pull/86) but it's 1.5 years old so I wouldn't expect it to be merged soon. Perhaps it would be worth forking the package to add the ESM entry and update the `react-is` dependency to `^16.7.0 || ^17.0.0`?

I'm happy to work on whatever approach you prefer.

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version: 11.0.0

Andarist:
fixes #2100

### DIFF
--- a/packages/react/src/isolated-hoist-non-react-statics-do-not-use-this-in-your-code.js
+++ b/packages/react/src/isolated-hoist-non-react-statics-do-not-use-this-in-your-code.js
@@ -1,0 +1,10 @@
+// this file isolates this package that is not tree-shakeable
+// and allows it to be dropped - if it stays unused
+// it happens thanks to sideEffects: false in our package.json
+import hoistNonReactStatics from 'hoist-non-react-statics'
+
+// have to wrap it in a proxy function because Rollup is too damn smart
+// and if this module doesn't actually contain any logic of its own
+// then Rollup just use 'hoist-non-react-statics' directly in other chunks
+export default (targetComponent, sourceComponent) =>
+  hoistNonReactStatics(targetComponent, sourceComponent)
--- a/.changeset/hungry-suns-accept.md
+++ b/.changeset/hungry-suns-accept.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Improved tree-shakeability of the package. The main benefit is that bundlers should be able now to drop [`hoist-non-react-statics`](https://github.com/mridgway/hoist-non-react-statics) if you don't actually use our `withTheme` export.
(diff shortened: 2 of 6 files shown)


packages/react/src/isolated-hoist-non-react-statics-do-not-use-this-in-your-code.js
```diff
@@ -0,0 +1,10 @@
+// this file isolates this package that is not tree-shakeable
+// and allows it to be dropped - if it stays unused
+// it happens thanks to sideEffects: false in our package.json
+import hoistNonReactStatics from 'hoist-non-react-statics'
+
+// have to wrap it in a proxy function because Rollup is too damn smart
+// and if this module doesn't actually contain any logic of its own
+// then Rollup just use 'hoist-non-react-statics' directly in other chunks
```
mitchellhamilton, line 8: 😂

mitchellhamilton (approved):

PR #2109, opened 2020-11-16
`createEnzymeSerializer` export has been removed from the root entrypoint and moved to dedicated `@emotion/jest/enzyme`

Andarist:


TrySound (approved):

packages/jest/src/index.js
```diff
@@ -1,4 +1,3 @@
 // @flow
 export { createSerializer } from './create-serializer'
-export { createEnzymeSerializer } from './create-enzyme-serializer'
```
mitchellhamilton, line 3: I feel pretty uncomfortable removing this in a minor. How about we have a `createEnzymeSerializer` implementation that's exported from `@emotion/jest` that looks like this?
```js
export function createEnzymeSerializer(options) {
  return require('@emotion/jest/enzyme').createEnzymeSerializer(options)
}
```

Andarist: Usually, this is problematics as well - some tools would treat such a `require` just as a variable from the global scope and this would only work because usually the tools that would consume this file wrap the transpiled file in such a way that `require` is available in the closure. But this is not guaranteed or anything. 

This export is not even properly documented - one could only find it in the source code or in one of the prerelease changelog entries that was not included in the final v11 changelog. Given this and that this was a mistake on my part that was conflicting with the declared optional peer dependency contract I wouldn't be as concerned about it (especially just after the v11 release). This seems like a pragmatic semver decision for me.

mitchellhamilton: ok 👍  (though the require would work fine though since this will only be used in jest where require will always be available)

PR #2120, opened 2020-11-18
Fix links for npm

karlhorky:
**What + Why**:

Currently, the npm page at https://www.npmjs.com/package/@emotion/babel-preset-css-prop has broken links pointing to:

- https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop/blob/HEAD/packages/babel-plugin
- https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop/blob/HEAD/docs/css-prop.mdx##babel-preset

**How**:

Fixing broken links

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset N/A

### DIFF
--- a/packages/babel-preset-css-prop/README.md
+++ b/packages/babel-preset-css-prop/README.md
@@ -1,6 +1,6 @@
 # @emotion/babel-preset-css-prop
 
-> A Babel preset to automatically enable Emotion's css prop when using the classic JSX runtime. If you want to use [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) please do not use this preset but rather just include our [`@emotion/babel-plugin`](/packages/babel-plugin) directly and follow instructions for configuring the new JSX runtimes [here](/docs/css-prop.mdx##babel-preset).
+> A Babel preset to automatically enable Emotion's css prop when using the classic JSX runtime. If you want to use [the new JSX runtimes](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) please do not use this preset but rather just include our [`@emotion/babel-plugin`](https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin) directly and follow instructions for configuring the new JSX runtimes [here](https://github.com/emotion-js/emotion/blob/master/docs/css-prop.mdx#babel-preset).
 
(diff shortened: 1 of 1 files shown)


Andarist:
This introduces a regression for GitHub and our docs site which I believe are more important from the npm page. It would be great to fix this for the npm page as well but I'm not sure how to accommodate all of those use cases at the same time - this would probably require a custom script that would alter the packed content of a package? 🤔

karlhorky:
> This introduces a regression for GitHub and our docs site

Ah, I was afraid of that :( Yeah, that sounds like a bit of a bind... Maybe a custom publish script that makes such a change to all relative links temporarily, does the `npm publish` and then reverts the changes?

Andarist:
The ideal solution (although it's a tad complicated for the issue at hand) would be to:
- add support to Changesets that would handle a special `pack` script for a package - it would just default to `npm pack`
- use the resulting archive when calling `npm publish`
- write that custom pack script in Emotion's repo that would rewrite those URLs
- integrate both

Having this implemented in Changesets would be quite awesome but not sure if you want to go that extra mile to make it work right now.

karlhorky:
Oh, that makes sense. Yeah, I'm not sure when I would get around to this... not familiar with Changesets and would have to dig in a bit.

Closing this PR for now.

Andarist:
Not a problem, I've created a feature idea here: https://github.com/atlassian/changesets/issues/492

Andarist:
It seems that such a feature in Changesets is not actually needed: https://github.com/atlassian/changesets/issues/492#issuecomment-731078478 . Would you like to work on this given that the scope of the task has been dramatically downsized now?

karlhorky:
Ah, so the workflow would be... create this `rewrite-links.js` file using `unist-util-visit` with `remark` to edit this one file?

Maybe that would be doable... I've worked with some of these things in projects like this: https://github.com/upleveled/mdx-local-link-checker/blob/main/index.js

Andarist:
> Ah, so the workflow would be... create this rewrite-links.js file using unist-util-visit with remark to edit this one file?

Yep, something like that - ideally it would try to patch all `packages/*/README.md`.

karlhorky:
I'll reopen, let's see when I get time for this.

PR #2122, opened 2020-11-19
Fix emotion label not being applied to components with numbers in name

dcastil:
**What**:

Fixes emotion label in development not being applied to components with numbers in component name.

```js
/** @jsx jsx */
import { jsx } from '@emotion/react'

// className: css-2a3l4r-MyComponent
const MyComponent = () => <div css={{}}  />

// className: css-fhsrqh
const MyComponentNo1 = () => <div css={{}}  />
```

**Why**:

Components with numbers in their name deserve the same love and recognition within the Elements tab in Chrome devtools!

**How**:

Adding numbers to the regex which is used to identify the emotion label.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/nasty-starfishes-retire.md
+++ b/.changeset/nasty-starfishes-retire.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fix emotion label not being applied to components with numbers in name
--- a/.changeset/nasty-starfishes-retire.md
+++ b/.changeset/nasty-starfishes-retire.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-Improved label extraction from the stack traces. We should be able to extract labels from components with numbers in their names now.
+Improved label extraction from the stack traces to handle components with numbers in their names.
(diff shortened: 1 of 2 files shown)


Andarist:
Looks good, could u add a changeset?

dcastil:
@Andarist I added the changeset. 😊

.changeset/nasty-starfishes-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fix emotion label not being applied to components with numbers in name
```
Andarist, line 5:
```suggestion
Improved label extraction from the stack traces. We should be able to extract labels from components with numbers in their names now.
```

.changeset/nasty-starfishes-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Improved label extraction from the stack traces. We should be able to extract labels from components with numbers in their names now.
```
Andarist, line 5:
```suggestion
Improved label extraction from the stack traces to handle components with numbers in their names.
```

Andarist:
Thanks!

Issue #2118, opened 2020-11-18
[Documentation] Should "advanced" vanilla emotion ssr be documented (v11)


Jareechang:
**Description:**

#### ⚠️ Note: I'm using version 11 for the packages

I noticed on the [@emotion/css page](https://emotion.sh/docs/@emotion/css) documentation page that [Server Side Rendering](https://emotion.sh/docs/ssr#api) goes to the "Default Approach" and "Advanced Approach". 

Do we intend to document the "Advanced Approach" for vanilla emotion ?

I followed the "advanced approach" but couldn't get it to work. However, I figured it out. 

It should use "cache" from the `@emotion/css` pkg which is different than the docs. Otherwise, no css will be extracted in `extractCritical` call.

#### To get SSR to work on vanilla emotion, I used:

```diff
import createEmotionServer from '@emotion/server/create-instance'
+import { cache } from '@emotion/css'
-import createCache from '@emotion/cache'
import { CacheProvider } from '@emotion/react';

const key = 'custom';
-const cache = createCache({ key })
const { extractCritical } = createEmotionServer(cache);

```

Please let me know what you think. I can PR the change if we are ok with updating. 

**Documentation links:**

[@emotion/css details page](https://emotion.sh/docs/@emotion/css) 
["Server Side Rendering" link out](https://emotion.sh/docs/ssr#api) 
["Default Approach" SSR](https://emotion.sh/docs/ssr#default-approach)
["Advanced Approach" SSR](https://emotion.sh/docs/ssr#advanced-approach)

Andarist:
Sure - more documentation the better.

kireerik:
Here is a full example:
- # [refo](https://github.com/kireerik/refo)

&nbsp;
example:

index/
- [style](https://github.com/kireerik/refo/tree/48072d16f27d007e5c16d628cec1c00a91a67235/example/index/style)/

   - [instance/index.js](https://github.com/kireerik/refo/blob/48072d16f27d007e5c16d628cec1c00a91a67235/example/index/style/instance/index.js)

   - [extract.js](https://github.com/kireerik/refo/blob/48072d16f27d007e5c16d628cec1c00a91a67235/example/index/style/extract.js)
&nbsp;
- index/index/site/

   - [index/index/index.js#L4-L6](https://github.com/kireerik/refo/blob/48072d16f27d007e5c16d628cec1c00a91a67235/example/index/index/site/index/index/index.js#L4-L6)

   - main/

      - header/

[...]

PR #2127, opened 2020-11-21
[Documentation] Update advanced ssr example for emotion 11

Jareechang:
**What**:

Updated documentation for "advanced approach" ssr. The existing documentation example does not work when upgrading to emotion 11.

As noted in #2118, the following example results in no css being extracted when calling `extractCritical`.

**Why**:

- Update the "advanced approach" for SSR - [existing advanced approach doc](https://emotion.sh/docs/ssr#advanced-approach)

**How**:

Updating the documentation in ssr.mdx

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/.changeset/nasty-starfishes-retire.md
+++ b/.changeset/nasty-starfishes-retire.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Improved label extraction from the stack traces to handle components with numbers in their names.
--- a/packages/react/src/emotion-element.js
+++ b/packages/react/src/emotion-element.js
@@ -44,11 +44,11 @@ export const createEmotionProps = (type: React.ElementType, props: Object) => {
     if (error.stack) {
       // chrome
       let match = error.stack.match(
-        /at (?:Object\.|Module\.|)(?:jsx|createEmotionProps).*\n\s+at (?:Object\.|)([A-Z][A-Za-z$]+) /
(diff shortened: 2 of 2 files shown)


docs/ssr.mdx
```diff
@@ -39,10 +39,8 @@ You can also use the advanced integration, it requires more work but does not ha
 import { CacheProvider } from '@emotion/react'
 import { renderToString } from 'react-dom/server'
 import createEmotionServer from '@emotion/server/create-instance'
-import createCache from '@emotion/cache'
+import { cache } from '@emotion/css'
```
Andarist, line 42: This is actually not because what is documented doesn't work but rather because you are using `@emotion/css` rather than `@emotion/react`.

This would actually need a separate recipe below this one to document your solution that is intended to work with `@emotion/css`. In your case most likely there is also no need to use `@emotion/server/create-instance` - `import { extractCritical } from '@emotion/server'` probably works just fine.

Jareechang: ok got it, that makes sense. I got it mixed up. 

Let me add a new section under `@emotion/css`. I got an example running in [codesandbox](https://codesandbox.io/s/festive-water-b374y?file=/src/index.ts), if it looks good, i'll add it.

Andarist: Yes, it looks OK. Please try to add the section where you see fit and we'll continue discussing from there.

PR #2132, opened 2020-11-23
Add displayNames to React components

dcastil:
**What**:

Add displayNames to Global, ClassNames and component returned from `withEmotionCache` to prevent components being labeled as Anonymous. Display names are not added in production.

**Why**:

Easier for devs to orient themselves in React devtools and tests, also easier to find Emotion-related components through search.

**How**:

Manually attaching `displayName` to Global and ClassNames components and the component returned from withEmotionCache.

Note to `withEmotionCache`:

```js
// → Component has displayName "WithEmotionCache(MyComponent)"
const Component = withEmotionCache(function MyComponent(…) {…})
```

In React devtools this component will appear as `MyComponent` with the modifiers `ForwardRef` and `WithEmotionCache`.

Component in React Components tab:
Screenshot 2020-11-23 at 21 00 36

Component in detail view of React Components tab:
Screenshot 2020-11-23 at 21 01 15

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/kind-dots-confess.md
+++ b/.changeset/kind-dots-confess.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-Add displayNames to Global, ClassNames and component returned from withEmotionCache to prevent components being labeled as Anonymous.
+`displayName`s have been added to `Global` and `ClassNames` components so React Developer Tools should display those better now.
--- a/.changeset/kind-dots-confess.md
+++ b/.changeset/kind-dots-confess.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Add displayNames to Global, ClassNames and component returned from withEmotionCache to prevent components being labeled as Anonymous.
--- a/packages/react/src/class-names.js
+++ b/packages/react/src/class-names.js
@@ -142,3 +142,7 @@ export const ClassNames: React.AbstractComponent<
   }
   return ele
 })
+
+if (process.env.NODE_ENV !== 'production') {
+  ClassNames.displayName = 'EmotionClassNames'
+}
--- a/packages/react/src/context.js
+++ b/packages/react/src/context.js
@@ -19,27 +19,16 @@ let EmotionCacheContext: React.Context<EmotionCache | null> = /* #__PURE__ */ Re
 
 export let CacheProvider = EmotionCacheContext.Provider
 
-let withEmotionCache = function withEmotionCache<
-  Props,
-  Ref: React.Ref<*>
->(func: {
-  (props: Props, cache: EmotionCache, ref: Ref): React.Node,
-  displayName?: string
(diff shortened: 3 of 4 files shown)


packages/react/src/class-names.js
```diff
@@ -142,3 +142,7 @@ export const ClassNames: React.AbstractComponent<
   }
   return ele
 })
+
+if (process.env.NODE_ENV !== 'production') {
+  ClassNames.displayName = 'EmotionClassNames'
```
Andarist, line 147: Shouldn't this be just `ClassNames`? It would reflect the actual name and match the inferred name if only this would be inferrable here.

packages/react/src/global.js
```diff
@@ -136,3 +136,7 @@ export let Global: React.AbstractComponent<
 
   return null
 })
+
+if (process.env.NODE_ENV !== 'production') {
+  Global.displayName = 'EmotionGlobal'
```
Andarist, line 141: Same as here: https://github.com/emotion-js/emotion/pull/2132/files#r530285457

Please note that I'm open for discussion about this.

dcastil: I prefixed it with Emotion for following reasons:
* It shows up when I search for `emotion` in React devtools which might help me find emotion-related components.
* It is more specific than `Global`. If someone new to a codebase sees this component and search for `EmotionGlobal` on Google, they find the correct docs immediately.
* It doesn't seem to have any disadvantages. When I search for `Global`, I still find this component.

But I don't have a big opinion on this. I'm happy to change them to their exact name if you think it makes more sense.

packages/react/src/context.js
```diff
@@ -19,16 +19,27 @@ let EmotionCacheContext: React.Context<EmotionCache | null> = /* #__PURE__ */ Re
 
 export let CacheProvider = EmotionCacheContext.Provider
 
-let withEmotionCache = function withEmotionCache<Props, Ref: React.Ref<*>>(
-  func: (props: Props, cache: EmotionCache, ref: Ref) => React.Node
-): React.AbstractComponent<Props> {
+let withEmotionCache = function withEmotionCache<
```
Andarist, line 22: I don't think changes here are required - `withEmotionCache` is purely internal and you have already covered all cases that use it.

dcastil: Fixed in 5227c72

.changeset/kind-dots-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Add displayNames to Global, ClassNames and component returned from withEmotionCache to prevent components being labeled as Anonymous.
```
Andarist, line 5:
```suggestion
`displayName`s have been added to `Global` and `ClassNames` components so React Developer Tools should display those better now.
```

PR #2133, opened 2020-11-23
Add server side rendering documentation for emotion/css

Jareechang:
**What**:

The ssr docs for vanilla emotion (`@emotion/css`) links out to example with react. 

Add example documentation for ssr using vanilla emotion.

**Why**:

Developers may want to use vanilla emotion ssr with other frameworks other than react. So, it'll be good to document ssr with vanilla emotion. 

**How**:

- Added new "Server Side Rendering" section `@emotion/css` README

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/css/README.md
+++ b/packages/css/README.md
@@ -11,7 +11,7 @@ The [@emotion/css](https://www.npmjs.com/package/@emotion/css) package is framew
   - [Animation Keyframes — `keyframes`](#animation-keyframes)
   - [Composing Class Names — `cx`](#cx)
 - [Custom Instances](#custom-instances)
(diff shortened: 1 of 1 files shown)


TimDaub:
@Jareechang thanks for this contribution. I was quite confused not to find a "framework agnostic" way of using emotion in SSR on [this page](https://emotion.sh/docs/ssr#api). But I've just tested your solution and it seems to work 👍🏼

szagi3891:
what about this PR ? maybe it's worth to merge it

PR #2136, opened 2020-11-25
Changed the implementation of label extraction from the stack traces

Andarist:


packages/react/src/emotion-element.js
```diff
@@ -42,16 +42,23 @@ export const createEmotionProps = (type: React.ElementType, props: Object) => {
   if (process.env.NODE_ENV !== 'production') {
     const error = new Error()
     if (error.stack) {
-      // chrome
+      // hooks|classes|ssr
       let match = error.stack.match(
-        /at (?:Object\.|Module\.|)(?:jsx|createEmotionProps).*\n\s+at (?:Object\.|)([A-Z][A-Za-z0-9$]+) /
+        /(^.+)\n.*(renderWithHooks|finishClassComponent|processChild)/m
```
mitchellhamilton, line 47: It seems like only V8 is the only engine to include the class name in the stack trace, the others only include `render` so using finishClassComponent is probably not gonna go well and it looks like processChild is used for both class and function components so maybe we should not match `ComponentName.render` for the V8 case?

Andarist: So just to be sure - are you propose matching lines with `ComponentName` (but without `.render`!) followed by lines with `renderWithHooks|processChild`?

mitchellhamilton: Yes, unless there's another way we could not match class components?

Andarist:
This got superseded by https://github.com/emotion-js/emotion/pull/2534

PR #2137, opened 2020-11-25
Update type import

hedi-far:
What changes are being made? (What feature/bug is being fixed here?) 

When using the css prop in my Next.js project as suggested in the docs, TypeScript gave me an error message (see screenshot).

Why are these changes necessary? 

The suggested solution did not work. It gave me this error message:
error_message

How were these changes implemented?

By adding the word "type" (as described in https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export). 

Have you done all of these things?  

**Checklist**:

- [ ] Documentation "N/A"
- [ ] Tests "N/A"
- [ ] Code complete "N/A"
- [ ] Changeset

### DIFF
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -62,7 +62,7 @@ When using our jsx pragma the support for `css` prop is being added only for com
 However, it's not possible to leverage `css` prop support being added conditionally based on a type of rendered component when one is not using our jsx pragma. For those cases when people use our pragma implicitly (for example when using our `@emotion/babel-preset-css-prop`) we have a special file that can be imported once to add support for the `css` prop globally, for all components. Use it like this:
 
 ```ts
-import {} from '@emotion/react/types/css-prop'
+import type {} from '@emotion/react/types/css-prop'
 ```
 
 ## @emotion/styled


Andarist:
Could you share a runnable repro case of your problem? Also - have you tries using:
```js
/// <reference types="@emotion/react/types/css-prop" />
```
that is mentioned [here](https://emotion.sh/docs/emotion-11#css-prop-types)

hedi-far:
> Could you share a runnable repro case of your problem? 

https://github.com/hedi-far/emotion-types-example

1. Clone the repository and install the dependencies with Yarn
2. Start the server with yarn dev
3. Go in the browser to http://localhost:3000

Please look at components/header to see the error message and comments!

Also - have you tried using:
> 
> ```js
> /// <reference types="@emotion/react/types/css-prop" />
> ```
> 
> that is mentioned [here](https://emotion.sh/docs/emotion-11#css-prop-types)

No, I have not as I am using TypeScript 4.1.2.

Andarist:
Given you are using already the automatic runtime the best option to resolve this is to apply this patch:
```diff
diff --git a/tsconfig.json b/tsconfig.json
index dbef917..c912c9d 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,6 +13,7 @@
     "strict": true,
     "noEmit": true,
     "jsx": "preserve",
+    "jsxImportSource": "@emotion/react",
     "esModuleInterop": true,
     "downlevelIteration": true,
     "resolveJsonModule": true,
```

And information about this might be actually missing from the docs - I would highly appreciate a PR for this 🙏 It can just be in this PR - I wouldn't mind.

> No, I have not as I am using TypeScript 4.1.2.

This solution works for any TS version (or at least for a lot of TS versions) - while the `import type` only works since TS 3.8.

You could try this patch:
```diff
diff --git a/next-env.d.ts b/next-env.d.ts
index 7b7aa2c..6bd2980 100644
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,2 +1,3 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
+/// <reference types="@emotion/react/types/css-prop" />
```

That being said - this is only needed if you are stuck with the classic runtime. The first solution (`"jsxImportSource": "@emotion/react"`) is the preferred one for the automatic runtime.

Andarist:
@hedi-far were you able to take a look at the given solutions?

hedi-far:
> @hedi-far were you able to take a look at the given solutions?

Hi, yes, sorry that it took so long! 

First, I removed 'import {} from '@emotion/react/types/css-prop'/'import type {} from '@emotion/react/types/css-prop' completely. Then I tried out the changes you proposed in the tsconfig.json, and the page built up without any errors. This is with TypeScript version 4.1.2. 
However if I change to version 4.0.3. I get an error message that this configuration requires TypeScrip 4.1. or higher. In this case, it works with /// <reference types="@emotion/react/types/css-prop" />

So, the pull request should look like approximately this...?: 

Eomtion/react 11/Typescript:

When using automatic runtime, for TypeScript versions from 3.8 use import type {} or /// <reference types="@emotion/react/types/css-prop" />
and from 4.1.++ use "jsxImportSource": "@emotion/react".

(I am very new to coding, so excuse me if this is not accurate.)

Andarist:
> (I am very new to coding, so excuse me if this is not accurate.)

Not a problem at all! I'm happy to help.

I've already prepared a PR with an update for the css prop section here: https://github.com/emotion-js/emotion/pull/2151 . I have forgotten to add a note that the `jsxImportSource` requires TS 4.1, so thanks for reminding me about this. I will have to add this as this is an important piece of information. Could you review that PR to see if the proposed changes make it more clear how to configure stuff? If something could still be improved - please comment there and I will try to adjust this further.

hedi-far:
Hello Andarist,

I think I would probably put it more like this:

-------------------------------------------------------------------------

When using our JSX factories, the support for `css` prop is being added only for components that accept the `className` prop as they take the provided `css` prop, resolve it and pass the generated `className` to the rendered component.

When using the automatic runtime, you should just add this to your `tsconfig.json` to let TypeScript know where it should look for the `JSX` namespace:

	```json
	{
	"compilerOptions": {
	"jsxImportSource": "@emotion/react"
	}
	}
	```

Please be aware that this requires TypeScrip 4.1 or higher. 

If you are still using the classic runtime, the ‘jsx’ namespace is resolved through the `jsx` pragma.(But to me, this information is quite confusing here in this place because the same information has been given already higher above on the same page, right? Why is there an extra chapter about the css prop in the first place? - But maybe I only misunderstand this bc of my lack of coding knowledge).

When you use our pragma implicitly (for example when using our `@emotion/babel-preset-css-prop`), we have a special file that can be imported once to add support for the `css` prop globally, for all components. Add it to your next-env.d.ts file (?) like so:

/// <reference types="@emotion/react/types/css-prop" />

------------------------------------------------------------------------------------

Okay, so maybe this helps a little :-)

PR #2140, opened 2020-11-27
Fix `css` prop missing from props types in TypeScript

hasparus:
**What**:

Sup folks. I fixed one small problem with WithConditionalCSSProp.

I was actually fixing a similar thing in another library https://github.com/system-ui/theme-ui/issues/1307.
I accidentaly auto-imported WithConditionalCSSProp instead of WithConditionalSxProps while writing tests, and I noticed you have exactly the same problem :D

**Why**:

**`css` prop is not added to all props types which accept `className` of type `string`, and it's also added to props which accept `className?: undefined`.**

[**Take a look at this TS Playground.**](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgSQHYwKZUgGwIYzASoA0cAKgBYYgZwC+cAZlBCHAOQACNEhxAeigY8AYxgcA3AChpMAJ5g6AdWAxKAYWIATNUVR4cGgMrGACqzAAeMwD44AXk6j8AZ1cA5PLQ5wMAD0xUbVc4AGsMeQgmODNpODgAfjgACjM-QIxg0KQXPHcvWkSALjhXGChgVABzBiTYuAAyRDhRdxKUdCxcAn0rKhoMe0ZSswBKeLhR2TlFOgB9AAl8jXcLCDAAIQBXGGNKCG2cbQ8+RzhVdS1gvWJDE3NLK1y3T28MDu3gjCYqjG0GLZZAolHB5qdVq51ltdvtDsdlgA3DAARnOl00OluBiMpmhz1ar0KH1KXzCqAgAHdUIDZqDwRBIdCdnsDkdtEiMAAmdFqTE3fg4h74l75N5FUrlSo1OAAHzKFSq1QA2gBdWl0hbLVxMywAQVcaCC2n+aKcGOuukF9zxT1FBXeksVMvoQJBWpWa31hq6320PPNfMt2Jtjw2BLyDolCultVdM2kAgEcAAtGnZCa8sI4O6LkGsdbcWGwB4AKLKGz2ByTBL28UYJ2xyR1ALG0JxBKdzvJdLNXLtUpG7oQfCC-rUWjDGtdqaxYFzMEQr0bA1Dv1mvNXAv6UPQssVuvEz7fX6of4a3NLT1Q71rk3aADMvK3Ap3Rb35YjRPeHSlStpl7arqK4+sa-wACzPvyVpvsKlj7l+YrEo2-7xoB17Qquvr3gArFBwaFnBGwIYeP6kqg5JUjSaELleOrLmAWFgdoABs+Hbnc77wZ+pHRn+MryvxKrqq6QA)

**How**:

We can read `extends` as _is assignable to_ or _is subset of_.
We need to add `css` prop to all props objects that can accept `className` prop of type `string`, so
- if `{ className: string }` is assignable to P
- or if we think of sets, if a set of all objects that have className property of type string is a subset of set P.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

<!-- feel free to add additional comments

### DIFF
--- a/.changeset/perfect-spoons-whisper.md
+++ b/.changeset/perfect-spoons-whisper.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-Add `css` prop to all types of props with className: string
+Fixed an issue with `css` prop type not being added to all components that accept a string `className` correctly.
--- a/.changeset/perfect-spoons-whisper.md
+++ b/.changeset/perfect-spoons-whisper.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Add `css` prop to all types of props with className: string
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -2,8 +2,8 @@ import 'react'
 import { Interpolation } from '@emotion/serialize'
 import { Theme } from './index'
 
-type WithConditionalCSSProp<P> = 'className' extends keyof P
-  ? (P extends { className?: string } ? P & { css?: Interpolation<Theme> } : P)
+declare type WithConditionalCSSProp<P> = { className: string } extends P
+  ? P & { css?: Interpolation<Theme> }
   : P
 
 // unpack all here to avoid infinite self-referencing when defining our own JSX namespace
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -2,8 +2,12 @@ import 'react'
 import { Interpolation } from '@emotion/serialize'
 import { Theme } from './index'
 
-type WithConditionalCSSProp<P> = { className: string } extends P
-  ? P & { css?: Interpolation<Theme> }
+// prettier-ignore
+type WithConditionalCSSProp<P> =
+  'className' extends keyof P
+    ? string extends P['className' & keyof P]
+      ? P & { css?: Interpolation<Theme> }
+      : P
   : P
 
 // unpack all here to avoid infinite self-referencing when defining our own JSX namespace
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -2,12 +2,10 @@ import 'react'
 import { Interpolation } from '@emotion/serialize'
 import { Theme } from './index'
 
-// prettier-ignore
-type WithConditionalCSSProp<P> =
-  'className' extends keyof P
-    ? string extends P['className' & keyof P]
-      ? P & { css?: Interpolation<Theme> }
-      : P
+type WithConditionalCSSProp<P> = 'className' extends keyof P
+  ? string extends P['className' & keyof P]
+    ? P & { css?: Interpolation<Theme> }
+    : P
   : P
 
 // unpack all here to avoid infinite self-referencing when defining our own JSX namespace
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -2,7 +2,7 @@ import 'react'
 import { Interpolation } from '@emotion/serialize'
 import { Theme } from './index'
 
-declare type WithConditionalCSSProp<P> = { className: string } extends P
+type WithConditionalCSSProp<P> = { className: string } extends P
   ? P & { css?: Interpolation<Theme> }
   : P
 
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -8,6 +8,7 @@ import {
   keyframes,
   withEmotionCache
 } from '@emotion/react'
+import { WithConditionalCSSProp } from './jsx-namespace'
 
 declare module '@emotion/react' {
   // tslint:disable-next-line: strict-export-declare-modifiers
@@ -159,14 +160,20 @@ const anim1 = keyframes`
 
   // TS@next reports an error on a different line, so this has to be in a single line so `test:typescript` can validate this on all TS versions correctly
   // $ExpectError
-  ;<CompWithoutClassNameSupport prop1="test" css={{ backgroundColor: 'hotpink' }} />
+  ;<CompWithoutClassNameSupport
+    prop1="test"
+    css={{ backgroundColor: 'hotpink' }}
+  />
 
   const MemoedCompWithoutClassNameSupport = React.memo(
     CompWithoutClassNameSupport
   )
   // TS@next reports an error on a different line, so this has to be in a single line so `test:typescript` can validate this on all TS versions correctly
   // $ExpectError
-  ;<MemoedCompWithoutClassNameSupport prop1="test" css={{ backgroundColor: 'hotpink' }} />
+  ;<MemoedCompWithoutClassNameSupport
+    prop1="test"
+    css={{ backgroundColor: 'hotpink' }}
+  />
 }
 
 {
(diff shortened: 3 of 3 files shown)


hasparus:
I see I failed tests on CircleCi. I'll take a look at it in a moment.

packages/react/types/jsx-namespace.d.ts
```diff
@@ -2,8 +2,12 @@ import 'react'
 import { Interpolation } from '@emotion/serialize'
 import { Theme } from './index'
 
-type WithConditionalCSSProp<P> = 'className' extends keyof P
-  ? (P extends { className?: string } ? P & { css?: Interpolation<Theme> } : P)
+// prettier-ignore
```
Andarist, line 5: Any reason why this is ignored here? Its in the source code so formatting shouldnt affect much

packages/react/types/tests.tsx
```diff
@@ -159,14 +160,14 @@ const anim1 = keyframes`
...
-  ;<CompWithoutClassNameSupport prop1="test" css={{ backgroundColor: 'hotpink' }} />
+  ;<CompWithoutClassNameSupport prop1="test" css={{ color: 'hotpink' }} />
 
   const MemoedCompWithoutClassNameSupport = React.memo(
     CompWithoutClassNameSupport
   )
   // TS@next reports an error on a different line, so this has to be in a single line so `test:typescript` can validate this on all TS versions correctly
   // $ExpectError
-  ;<MemoedCompWithoutClassNameSupport prop1="test" css={{ backgroundColor: 'hotpink' }} />
+  ;<MemoedCompWithoutClassNameSupport prop1="test" css={{ color: 'hotpink' }} />
```
Andarist, line 170: Why this has to be changed? 🤔

packages/react/types/tests.tsx
```diff
@@ -181,6 +182,46 @@ const anim1 = keyframes`
...
-  ;<CompWithImplicitChildren>content<div/></CompWithImplicitChildren>
+  const CompWithImplicitChildren: React.FC = () => null
+  ;<CompWithImplicitChildren>
+    content
+    <div />
+  </CompWithImplicitChildren>
+}
+
+// Tests for WithConditionalCSSProp
+{
```
Andarist, line 193: Im reviewing from mobile and everything is squeezed so maybe im seeing this wrong but it seems that those tests only compute types. Couldwe rewrite them in a way that they would act as assertions? So tests would a tually fail if we introduce a regression? Such test could be written by comparing the output to the expected type

packages/react/types/jsx-namespace.d.ts
```diff
@@ -2,8 +2,12 @@ import 'react'
...
+type WithConditionalCSSProp<P> =
+  'className' extends keyof P
+    ? string extends P['className' & keyof P]
```
Andarist, line 8: Could u explain the machinery if this type step by step? Especially the `& keyof` bit - is this just supposed to produce never if there is no className prop?

Ill have to think the whole thing through in terms of possible perf regressions as well.

packages/react/types/tests.tsx
hasparus: This line with `backgroundColor` is too long for Prettier. Even if you don't format it on auto-save in your IDE, you need to remember to do `git commit --no-verify`, because there's a `prettier --write` running on pre-commit.

```json
  "lint-staged": {
    "*.{js,ts,tsx,md}": [
      "prettier --write",
      "git add"
    ]
  }
```


Obviously, I kept forgetting about it and failing tests, because the error occurs in different places (on `css` or over entire element) by different TypeScript version.

I can change it to `backgroundColor` if you want, but that property is irrelevant from the perspective of the test, because the css prop is missing, and with `backgroundColor` it was just more unpleasant to contribute to this file.

Andarist: Right - I don't mind the change, was just curious why it was needed. Your explanation makes sense 👍

packages/react/types/jsx-namespace.d.ts
hasparus: Sorry, this isn't necessary anymore. Thanks!

packages/react/types/jsx-namespace.d.ts
hasparus: sure. I assume you get the gists of it, so I'll focus on that confusing `& keyof`.

Let's just translate it into pseudocode, so we're on the same page.
```
  if className is a key in Props
    then if Props["className"] can be a string
      then return Props merged with { css?: Interpolation<Theme> }
      else return Props
    else return Props
```

Previous code (pasted below for comparison) was checking _can I assign Props to `{ className?: string }`_.
You can assign `{ className?: undefined }` to `{ className?: string }`, and you can't assign `{ className: string[] | string }` or `{ className: unknown }` to it.

```
type WithConditionalCSSProp<P> = 'className' extends keyof P
  ? (P extends { className?: string } ? P & { css?: Interpolation<Theme> } : P)
  : P
```

---

> Especially the & keyof bit - is this just supposed to produce never if there is no className prop?

Yup, that's correct.

image

This is the most "interesting" part. The props will surely be an object, and we just care about the `className` prop.
But we can't read that `className` from `P` until we know that such a key exists in `P`. 

image

This code is sufficient in TS 4.0, but TS 3.6 is a bit dumber.
I had to add that `& keyof P`, because `className` is really a key of `P` at this point, and we'll never get `never` there. TS 3.6 just doesn't remember that we already checked it :)

But hey! Doesn't this mean that we can get rid of that first `extends keyof P` clause and use only one conditional type?

Like so
```tsx
type WithConditionalCSSProp<P> =
  string extends P['className' & keyof P]
    ? P & { css?: Interpolation<Theme> }
    : P
```

We could if we didn't care about TS 3.4 anymore.

image

3.4 is before TypeScript was smart enough to know that you can't be two different strings at once. An intersection of `'any-string' & 'className'` doesn't collapse into `never`, we get an error, and a nice `any`, `css` is added to _all props_ regardless if they have className and our tests fail.

image

That `dtslint` was a smart choice.

> Ill have to think the whole thing through in terms of possible perf regressions as well.

Do you have any benchmarks / performance tests for Emotion types?

packages/react/types/tests.tsx
hasparus: Good point. I wrote hoping they'll crash if there's a regression, but any `any` in LibraryManageAttributes would wreak them. I can certainly improve them with $ExpectType.

packages/react/types/jsx-namespace.d.ts
Andarist: I have done Very Scientific Research™ by checking a fresh CRA with just a few css prop-based components and those are some numbers from the `--diagnostics` (I've ignored mem consumption and times as those are not stable between reruns):

### Current

```
Nodes:           336996
Identifiers:     114711
Symbols:          81060
Types:             4642
Instantiations:   14645
```

### This PR

```
Nodes:           336998
Identifiers:     114711
Symbols:          80811
Types:             4181
Instantiations:    4824
```

### Without `& keyof P`

```
Nodes:           336994
Identifiers:     114710
Symbols:          80811
Types:             4180
Instantiations:    4809
```

--- 

As we may notice there is only a rather negligible difference between your PR and the "without `& keyof P`" variant that aims to satisfy older TS versions **but** there is actually **a big improvement** over the current version in terms of instantiations that are often responsible for slowdowns

.changeset/perfect-spoons-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Add `css` prop to all types of props with className: string
```
Andarist, line 5:
```suggestion
Fixed an issue with `css` prop type not being added to all components that accept a string `className` correctly.
```

PR #2141, opened 2020-11-28
fix: memoize broken on props present in prototype chain

ChALkeR:
**Why**:

Test case (invalid results):
```js
> require('@emotion/memoize').default(x => x)('toString')
[Function: toString]
> require('@emotion/memoize').default(x => x)('constructor')
[Function: Object]
```

Memoize (and is-prop-valid) was broken on a number of arguments.

```js
> require('@emotion/is-prop-valid').default('constructor')
[Function: Object]
> require('@emotion/is-prop-valid').default('toString')
[Function: toString]
> require('@emotion/is-prop-valid').default('__proto__')
{}
```
Note that those are all truthy.

**How**:

Create the underlying storage object with a null prototype.
Could also have used a `Map`.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests — No tests in that dir.
- [ ] Code complete N/A
- [ ] Changeset

### DIFF
--- a/.changeset/unlucky-swans-care.md
+++ b/.changeset/unlucky-swans-care.md
@@ -1,0 +1,5 @@
+---
+"@emotion/memoize": patch
+---
+
+Fixed an issue with prototype properties not being correctly handled.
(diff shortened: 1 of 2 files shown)


Andarist:
How this have affected you?

ChALkeR:
@Andarist It hasn't. This was found during code review of linaria deps.

PR #2147, opened 2020-12-01
Fixes babel plugin crashing when there was no string after the last interpolation

macalinao:
**What**:

Plugin crashed, so I added a presence check to see if the node actually exists.

**Why**:

Bug

**How**:

Tested locally

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset N/A


Andarist:
Thank you! I've adjusted the PR slightly to patch the root cause - `interleave` should always receive one array of N expressions and N+1 strings.

PR #2159, opened 2020-12-07
Replace is-prop regex to O(1) hash table compare

tlgimenes:
**What**:
This PR improves is-prop-valid performance by replacing a huge (4Kb) RegExp with a small RegExp and a O(1) hash table.

**Why**:
I am developing an e-commerce website using Gatsby.js and theme-ui. Theme-ui uses emotion-js. Analyzing my website in Google's Lighthouse, I figured out a huge TBT (total blocking time). Using Google Chrome's flamechart performance analyzer, I've figured out that a huge time was spent running is-prop-valid.

Using some webpack aliases, I've managed to use my own implementation of `@emtion/is-prop-valid`. This change completely removed the blocking time in Chrome's performance flamechart. [This PR](https://github.com/vtex/faststore/pull/370) explains a little bit better the changes I've done in theme-ui.

I'm now backporting my custom is-prop-valid implementation to emotion-js so everybody can benefit from this performance improvement 

**How**:
`@emotion/is-prop-valid` is a small, two files package. One package contains the is-prop-valid function implementation and the other contains all the data for generating a RegExp that computes if a prop is valid or not. 

This final RegExp is something like `/^(children|dangerouslySetInnerHTML|key)$/`. This RegExp could be replaced by the following hash table
```
{
   children: true,
   dangerouslySetInnerHTML: true,
   key: true
}
```

By doing this replacement, we can achieve greater performance. 

Also, serializing this hash table in the final JS is important. This PR uses the old JSON.parse/JSON.serialize technique for speeding up compile/parse times. To understand a little bit more about this technique, take a look at [this video](https://www.youtube.com/watch?v=ff4fgQxPaO0)

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset 

I'd like to thank you for reading my small contribution. If you have any suggestions/comments, please feel free to let me know

Thanks !

### DIFF
--- a/.changeset/ninety-toes-nail.md
+++ b/.changeset/ninety-toes-nail.md
@@ -1,0 +1,5 @@
+---
+'@emotion/is-prop-valid': patch
+---
+
+Replaces is-prop-valid HUGE RegExp to a small RegExp and a hash table increasing the performance
--- a/packages/is-prop-valid/src/index.js
+++ b/packages/is-prop-valid/src/index.js
@@ -1,17 +1,21 @@
 // @flow
 import memoize from '@emotion/memoize'
 
-declare var codegen: { require: string => RegExp }
+declare var codegen: { require: string => string }
 
-const reactPropsRegex = codegen.require('./props')
+const reactProps = JSON.parse(codegen.require('./props'))
+const reactPropsRegex = /^(([Dd][Aa][Tt][Aa]|[Aa][Rr][Ii][Aa]|x)-.*)$/
 
-// https://esbench.com/bench/5bfee68a4cd7e6009ef61d23
-const isPropValid = /* #__PURE__ */ memoize(
-  prop =>
-    reactPropsRegex.test(prop) ||
-    (prop.charCodeAt(0) === 111 /* o */ &&
+const isPropValid = /* #__PURE__ */ memoize(prop => {
+  const chars =
+    prop.charCodeAt(0) === 111 /* o */ &&
     prop.charCodeAt(1) === 110 /* n */ &&
-      prop.charCodeAt(2) < 91) /* Z+1 */
-)
+    prop.charCodeAt(2) < 91 /* Z+1 */
+
+  const inProps = reactProps[prop] === true
+  const inRegex = reactPropsRegex.test(prop)
+
+  return chars || inProps || inRegex
+})
 
 export default isPropValid
--- a/packages/is-prop-valid/src/props.js
+++ b/packages/is-prop-valid/src/props.js
@@ -476,6 +476,4 @@ const props = {
   autofocus: true
 }
 // eslint-disable-next-line import/no-commonjs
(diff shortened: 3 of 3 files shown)


emersonlaurentino (approved):

Andarist:
Could you fork this https://esbench.com/bench/5bfee68a4cd7e6009ef61d23 and add your version to it for comparison?

Andarist:
I've updated that esbench with your proposed changes (added a new test case) and done 2 runs:
- on Chrome your version has resulted in a **slightly** better results (250k vs 240k ops per sec)
- on Firefox the current version won 🤷‍♂️ 

I wouldn't say this is better enough that I would merge this right away. I'm open to discussing this further though. Do you have a version of your site deployed that is using our `@emotion/is-prop-valid` and another version that is using your fork (and this being the only difference between those two)?

Keep in mind that this "huge" regexp is being basically transformed into a lookup table at runtime because we memoize the results.

tlgimenes:
Hello @Andarist, thanks for your quick review !

I have made two versions of my site. One using my webpack alias and another one using `@emotion/is-prop-valid`. The first image is using the current version of emotion and the second one is a version with my modifications. As you can see, in my tests, is-prop-valid went from ~14ms to ~1.3ms.

These metrics were taken using chrome's devTools with `6x CPU throttling` 

image

image

Andarist:
Do u have those deployed publically? I would love to investigate this myself a little bit to understand the issue better.

tlgimenes:
Sure ! I will prepare a deploy preview in Netlify and send here the link.

In the meanwhile, I got your `esbench` link and run the benchmark. As far as I can see, the fair comparison is between `mixed known props regexp with char codes handlers and data aria sensitive regexp` and `mixed known props obj lookup with char codes on and data aria sensitive regexp`. According to my the following print it's a `236,999 ops/s` vs `327,985 ops/s`, i.e. ~38% increase in performance. Am I missing something in here ? Thanks :)

image

tlgimenes:
Ok. So the version using `@emotion/is-prop-valid` is live in [this link](https://deploy-preview-260--marinbrasil.netlify.app/). Also, my version WITHOUT the `@emotion/is-prop-valid` is live in [here](https://marinbrasil.netlify.app/)

To see were `@emotion/is-prop-valid` is being used, you can follow the steps below:

Open Chrome devTools. At the `performance` tab, mark `Fast 3G network` throttling and `6x CPU slowdown`. Run the profiling. You should see something like this
image

Zoom on the red mark. The `anonymous` function is the `is-prop-valid` execution

image

Please let me know if you need more info. Thanks !

tlgimenes:
Hi @Andarist, did you have some time to look into it ? Thanks !

Andarist:
Quite frankly - not. I'm a little bit burned out on Emotion right now and have been slammed with other OSS work :/

Issue #2169, opened 2020-12-11
Using TypeScript's React.ComponentType with Emotion 11 and /** @jsxImportSource @emotion/react */


stuartkeith:
**Description:**

I'm upgrading a project from Emotion 10 to Emotion 11. There are some parts of the code that use a generic Prop type then use React.ComponentType to specify a component that uses those props. Ultimately something like this:

```
function example<Props>(Component: React.ComponentType<Props>, props: Props) {
    return <Component {...props} />;
}
```

With Emotion 11 and `/** @jsxImportSource @emotion/react */`, the props cannot be assigned to the component:

```
Type 'Props' is not assignable to type 'IntrinsicAttributes & WithConditionalCSSProp<PropsWithChildren<Props>> & Props & { children?: ReactNode; }'.
  Type 'Props' is not assignable to type 'WithConditionalCSSProp<PropsWithChildren<Props>>'.ts(2322)
```

From what I can gather, TypeScript can't assign `Props` to `WithConditionalCSSProp<Props>` due to `WithConditionalCSSProp` being a conditional type. TS can't handle something like the following, even though to the reader it seems intuitive `X` would be assignable to `Output<X>`:

```
type Output<T> = T extends string ? T : T;

function example<X>(input: X): Output<X> {
  // Type 'X' is not assignable to type 'Output<X>'
  return input;
}
```

Is there any workaround for this issue? I'm not sure if there's a way I can fix React.ComponentType or there's a fundamentally different approach I should be taking. Thanks!

Andarist:
Could you prepare a runnable repro case for this problem? It would allow me to experiment with it.

stuartkeith:
No problem, here's one: https://codesandbox.io/s/billowing-night-ml4g7

- `App.tsx` uses `/** @jsx jsx */` and has the type error.
- `AppNoEmotion.tsx` doesn't have the comment or the type error.

stuartkeith:
I've also added a demonstration of a conditional type causing the same issue here: https://codesandbox.io/s/billowing-night-ml4g7?file=/src/AppConditionalType.tsx

Andarist:
The conditional type doesn't work without Emotion either, right?

Andarist:
Another note to myself - it seems that changing the `LibraryManagedAttributes` to this is fixing the issue:
```ts
type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<ReactJSXLibraryManagedAttributes<C, P>>
```
[...]

PR #2181, opened 2020-12-20
Fix TypeScript support for custom components with non-standard declarations

101arrowz:
**What**:
This PR makes a tiny modification to #2140 in order to fix support for components that do not use `React.ComponentType` but rather a custom declaration with a generic render function.

**Why**:
A certain package [RMWC](https://rmwc.io) has been causing issues when using `"jsxImportSource": "@emotion/react"`. For instance, the types of the parameters passed to event handlers such as `onClick` were `any`. I investigated the problem and discovered that the components were not declared with `React.ComponentType`; instead, they had a generic render function to add support for a `tag` property that changes the underlying DOM element used (e.g. `tag="div"` can be used to make a button use a div element internally). These changes resolve the issues and fix the typings on event handlers.

**How**:
In all honesty: absolutely no idea. However, it seems to me that `LibraryManagedAttributes` is meant to only specify the attributes that will be handled outside of React, whereas Emotion was previously including P in the intersection no matter what. This may have caused attributes to have been duplicated and merged with a union. For functions, unions make typing the parameters impossible in TypeScript, which would be the cause of the implicit `any` issues. Of course, this is all speculation.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset 

Again, I'm really not sure *how* this works, but I do know that it does work.
CodeSandbox: https://codesandbox.io/s/gallant-snowflake-xt8dy

### DIFF
--- a/.changeset/silent-otters-build.md
+++ b/.changeset/silent-otters-build.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fixes React components with custom/generic render functions
--- a/.changeset/silent-otters-build.md
+++ b/.changeset/silent-otters-build.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-Fixes React components with custom/generic render functions
+Fixed some typing issues with React components with custom generic render functions.
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -1,5 +1,5 @@
 // Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
-// TypeScript Version: 3.2
+// TypeScript Version: 3.4
 
 import { EmotionCache } from '@emotion/cache'
 import {
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -4,9 +4,9 @@ import { Theme } from './index'
 
 type WithConditionalCSSProp<P> = 'className' extends keyof P
   ? string extends P['className' & keyof P]
-    ? P & { css?: Interpolation<Theme> }
-    : P
-  : P
+    ? { css?: Interpolation<Theme> }
+    : {}
+  : {}
 
 // unpack all here to avoid infinite self-referencing when defining our own JSX namespace
 type ReactJSXElement = JSX.Element
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -231,3 +231,19 @@ const anim1 = keyframes`
     ? true
     : false
 }
+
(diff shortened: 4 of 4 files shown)


Andarist:
Very interesting! I can confirm that it is fixing the problem in the linked codesandbox - although I don't understand what happens here either (gonna take another look later, with fresh mind).

Could you try to add a type test for this? The PR that you have linked to has added some so you could take a look at how it has been done there.

101arrowz:
I know how to write a test, but I'm not exactly sure how to create a minimal reproducible example without bringing in RMWC because I don't know what specifically is causing the incompatibility. Perhaps further investigation could help, but I've already spent 3 days debugging this just to find this solution, and I don't really feel like doing any more.

101arrowz:
@Andarist I just added a test by effectively copy-pasting the RMWC definition and removing extraneous additions. As expected, it fails before this PR and works afterwards, but the issue still doesn't quite make sense. I suppose it doesn't really matter, since the patch makes logical sense and definitely works.

Andarist:
Thank you! An interesting fact is that removing the intersection from our `LibraryManagedAttributes` also fixes the issue so the alternative fix for this issue could be:
```diff
diff --git a/packages/react/types/jsx-namespace.d.ts b/packages/react/types/jsx-namespace.d.ts
index db09ddcf..1e45aca5 100644
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -25,8 +25,9 @@ export namespace EmotionJSX {
     extends ReactJSXElementAttributesProperty {}
   interface ElementChildrenAttribute extends ReactJSXElementChildrenAttribute {}
 
-  type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+  type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<
     ReactJSXLibraryManagedAttributes<C, P>
+  >
 
   interface IntrinsicAttributes extends ReactJSXIntrinsicAttributes {}
   interface IntrinsicClassAttributes<T>
```

This is also fixing #2169 but, unfortunately, is causing some other test cases to fail for some reason.

Nevertheless - your fix makes sense and even might come with some perf improvement so I'm going to merge this in and try to figure out the other thing later.

.changeset/silent-otters-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fixes React components with custom/generic render functions
```
Andarist, line 5:
```suggestion
Fixed some typing issues with React components with custom generic render functions.
```

Andarist (approved):

PR #2186, opened 2020-12-21
Fixed an issue with some generic functions not being able to use the parametrized ComponentType

### LINKED ISSUE #2169: Using TypeScript's React.ComponentType with Emotion 11 and /** @jsxImportSource @emotion/react */

Andarist:
fixes #2169, quite frankly I don't really understand the issue so I have no idea how to properly write a good changeset for this

### DIFF
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -25,9 +25,9 @@ export namespace EmotionJSX {
     extends ReactJSXElementAttributesProperty {}
   interface ElementChildrenAttribute extends ReactJSXElementChildrenAttribute {}
 
-  type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<
-    ReactJSXLibraryManagedAttributes<C, P>
-  >
+  type LibraryManagedAttributes<C, P> = P extends any
+    ? WithConditionalCSSProp<ReactJSXLibraryManagedAttributes<C, P>>
+    : never
 
   interface IntrinsicAttributes extends ReactJSXIntrinsicAttributes {}
   interface IntrinsicClassAttributes<T>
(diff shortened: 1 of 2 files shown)


packages/react/types/jsx-namespace.d.ts
```diff
@@ -25,8 +25,9 @@ export namespace EmotionJSX {
     extends ReactJSXElementAttributesProperty {}
   interface ElementChildrenAttribute extends ReactJSXElementChildrenAttribute {}
 
-  type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+  type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<
```
Andarist, line 28: oh, this will clash with this: https://github.com/emotion-js/emotion/pull/1952/files#r463502611 . it has been mentioned that the distribution fixes that problem, might need to apply this here, will take a look later

Andarist:
In the end, I've tracked down a bug in TS: https://github.com/microsoft/TypeScript/issues/42240 . I don't think this is worth fixing here on our side (how we use conditional types creates some problems with generic components but that is not a very popular pattern).

PR #2191, opened 2020-12-23
add incremental prop

magic-akari:
**What**: 
add a Non-standard Properties: `incremental`

**Why**: 
> The Boolean attribute incremental is a WebKit and Blink extension (so supported by Safari, Opera, Chrome, etc.) which, if present, tells the user agent to process the input as a live search. As the user edits the value of the field, the user agent sends search events to the HTMLInputElement object representing the search box. This allows your code to update the search results in real time as the user edits the search.
>
> If incremental is not specified, the search event is only sent when the user explicitly initiates a search (such as by pressing the Enter or Return key while editing the field).
>
> The search event is rate-limited so that it is not sent more frequently than an implementation-defined interval.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] Code complete N/A
- [ ] Changeset  Not sure if it is needed


Andarist:
cc @probablyup

probablyup (approved):

mitchellhamilton (approved):

PR #2194, opened 2020-12-27
Clarify Emotion 11 codemod instructions

josephmarkus:
**What**: Clarify Emotion 11 codemod instructions

**Why**: so that a user sees an easier option before a manual one

**How**: By lifting instructions above the list, so that a user sees an easier option first; adding explicit steps

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset N/A


Andarist:
Thanks!

PR #2197, opened 2021-01-01
Fixed issue with typescript and mutable refs

gabrielbull:
**What**:

Typescript definitions for refs.

**Why**:

When using mutable refs, typescript will throw an error

**How**:

Changed the typescript definitions.

**Checklist**:

- [ ] ~Documentation~ N/A
- [ ] ~Tests~ N/A
- [x] Code complete
- [x] Changeset 

**Example**

This code was not valid in typescript prior to the changes:

```jsx
import { FunctionComponent, useRef } from 'react'
import { View } from 'react-native'
import styled from 'emotion/native'

const MyComponent: FunctionComponent = () => {
  const ref = useRef<View>()
  
  return <MyStyledComponent ref={ref} />
}

const MyStyledComponent = styled(View)`
  background-color: red;
`
```

### DIFF
--- a/packages/native/types/base.d.ts
+++ b/packages/native/types/base.d.ts
@@ -82,7 +82,7 @@ export interface StyledComponent<
   ): StyledComponent<
     ComponentProps & React.ComponentProps<C>,
     {},
-    { ref?: React.Ref<InstanceType<C>> }
+    { ref?: React.Ref<InstanceType<C> | undefined> }
   >
   withComponent<C extends React.ComponentType<React.ComponentProps<C>>>(
     component: C
@@ -156,7 +156,7 @@ export interface CreateStyled {
       as?: React.ElementType
     },
     {},
-    { ref?: React.Ref<InstanceType<C>> },
+    { ref?: React.Ref<InstanceType<C> | undefined> },
     ReactNativeStyleType<React.ComponentProps<C>>
   >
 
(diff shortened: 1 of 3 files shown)


gabrielbull:
Added changes

packages/native/types/base.d.ts
```diff
@@ -82,7 +82,7 @@ export interface StyledComponent<
   ): StyledComponent<
     ComponentProps & React.ComponentProps<C>,
     {},
-    { ref?: React.Ref<InstanceType<C>> }
+    { ref?: React.Ref<InstanceType<C> | undefined> }
```
Andarist, line 85: hm, this should pretty much be the same thing - there is little to no difference between an optional property and a property that accepts undefined (and I wish that there was a difference 😢 )

could you add a test case that would show what exactly has been fixed?

Andarist: Ok, I see you have put one in the description of the PR. I think this works exactly the same as `@types/react`. Take a look at this:
https://codesandbox.io/s/keen-frog-39lxt?file=/src/App.tsx:170-173
and the version that fixes the problem:
https://codesandbox.io/s/ecstatic-meninsky-3wzlh?file=/src/App.tsx:174-177

The key is to provide the initial value of `null` to `useRef`. It reflects React expectation (refs are set to either the host element or null)

gabrielbull: You're right.

PR #2206, opened 2021-01-08
Export `SerializedStyles` from @emotion/react flow types

### LINKED ISSUE #2203: Include `SerializedStyles` in `@emotion/react`'s flow types
**The problem**

We make heavy use of Emotion's `css` prop. One pitfall comes when allowing custom components' styles to be extended by parent components. The main issue with this is that the `css` prop is unknown to our type system (flow) and thus needs to be manually declared in the extensible component's props type. To do this, we currently import the `SerializedStyles` type from `@emotion/utils` which means that we need an explicit dependency on `@emotion/utils` just for this.

Looking at the Typescript typedefs for `@emotion/react`, it seems like `SerializedStyles` is explicitly exported already there:

https://github.com/emotion-js/emotion/blob/a89d4257b0246a1640a99f77838e5edad4ec4386/packages/react/types/index.d.ts#L27-L35

**Proposed solution**

Re-export the `SerializedStyles` type from `@emotion/react` like is currently done for Typescript.

```js
export type {SerializedStyles} from `@emotion/serialize`;
```

**Alternative solutions**

Definitely open to any suggestions of how to more easily combine custom components with the `css` prop and a type system.

**Additional context**

I'm happy to make the change if this sounds agreeable.

jimmydief:
**What**: Export `SerializedStyles` from @emotion/react flow types

**Why**: This will make it easier to properly type the return value of the `css` function in Flow.

**How**: Re-exporting the type.

**Checklist**:

- [ ] Documentation (N/A?)
- [ ] Tests (N/A?)
- [X] Code complete
- [X] Changeset 

Fixes #2203

### DIFF
--- a/.changeset/sweet-seas-add.md
+++ b/.changeset/sweet-seas-add.md
@@ -1,5 +1,5 @@
 ---
-'@emotion/react': minor
+'@emotion/react': patch
 ---
 
-Export `SerializedStyles` from `@emotion/react` flow types
+Added export for `SerializedStyles` Flow type.
--- a/.changeset/sweet-seas-add.md
+++ b/.changeset/sweet-seas-add.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': minor
+---
+
+Export `SerializedStyles` from `@emotion/react` flow types
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -1,5 +1,6 @@
 // @flow
 import pkg from '../package.json'
+export type { SerializedStyles } from '@emotion/utils'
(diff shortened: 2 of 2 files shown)


.changeset/sweet-seas-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': minor
+---
+
+Export `SerializedStyles` from `@emotion/react` flow types
```
Andarist, line 5:
```suggestion
'@emotion/react': patch
---

Added export for `SerializedStyles` Flow type.
```

Andarist:
Thanks!

PR #2214, opened 2021-01-16
Update typescript.mdx

EduVencovsky:
While adding emotion to a typescript project, it took me a few hours to realize why my Theme wasn't the correct type, until I realized that in the docs, the way to import the Material-UI theme was wrong.

I added a comment to make explicitly the fact that how you import the theme might change according to Material-UI version.

### DIFF
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -286,7 +286,8 @@ declare module '@emotion/react' {
 
 // You are also able to use a 3rd party theme this way:
 import '@emotion/react'
-import { MuiTheme } from 'material-ui'
+// Import of the Material-UI theme might change depending on the version you are using 
+import { MuiTheme } from 'material-ui' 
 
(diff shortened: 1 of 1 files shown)


docs/typescript.mdx
```diff
@@ -286,7 +286,8 @@ declare module '@emotion/react' {
 
 // You are also able to use a 3rd party theme this way:
 import '@emotion/react'
-import { MuiTheme } from 'material-ui'
+// Import of the Material-UI theme might change depending on the version you are using 
+import { MuiTheme } from 'material-ui'
```
Andarist, line 290: well, this was just an example - maybe it would be beneficial to just change this to smth like:
```suggestion
import { LibTheme } from 'some-lib' 
```

WDYT?

EduVencovsky: Sounds better!

Issue #2210, opened 2021-01-11
Multiple instances of emotion overwrite each other


knenkne:
**Current behavior:**
We've microservice architecture so each team can have their own emotion instance, in the same time they can even import other libraries that also use emotion. All apps getting loaded asynchronously. And when the next app is loaded, it creates own emotion instance that overwrites previous (at least it looks like so, but old <style /> tags still exist in DOM).
![image](https://user-images.githubusercontent.com/35743791/104240565-51793b00-546d-11eb-9ffb-cd4cb7e00867.png)
Yep, In `dev` mode we're getting the warn, but in `production` build all sheets getting overwritten.

**To reproduce:**
[Main repo](https://github.com/knenkne/emotion-app)
1. `npm i` to install pseudolibraries and deps
2.  `npm run start` to see the warn
2. `nrm run build` to see how one library overwrites another after 2 seconds (`public` folder)

[Library repo 1](https://github.com/knenkne/emotion-11-library-1)
[Library repo 2](https://github.com/knenkne/emotion-11-library-2)

**Expected behavior:**
Allow multiple instances of emotion

**Environment information:**
- Main repo with emotion@10*
- Libraries repos with emotion@11*

Andarist:
In this situation, you really should use separate caches with unique keys - unfortunately `<style/>` elements are shared resources in the global document and we can't easily distinguish which belong to each of those microfrontends. 

The problematic part though is that you are using hard dependencies for `@emotion/react` in each one and thus you load multiple instances of Emotion 11 so their context identities won't be shared. I guess that because you are in this microfrontend architecture this is a hard requirement for you so I would advise you to just export `CacheProvider` + `createCache` from each of your libraries~ and just wrap each React subtree in those CacheProviders.

So in the end you would handle things like this:
```diff
diff --git a/src/index.js b/src/index.js
index b55fafe..c350d6a 100644
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,8 @@ import React from "react"
 import ReactDOM from "react-dom"
 import styled from "@emotion/styled"
 import css from "@emotion/css"
-
+import { CacheProvider } from "@emotion/core"
+import createCache from "@emotion/cache"
 
 const getEmotion11Library1Chunk = async () => {
   return await import("emotion-11-library-1")
@@ -14,13 +15,23 @@ const getEmotion11Library2Chunk = async () => {
 
 setTimeout(() => {
   getEmotion11Library1Chunk().then((chunk) => {
[...]

PR #2222, opened 2021-01-25
Prevent style rules loss by not moving already-hydrated <style> tags

gregjacobs:
**What**: This PR fixes emotion-js/emotion#2210. The issue is that `<style>` elements are moved in the DOM, and that causes them to lose their `sheet.cssRules` array in production ("speedy") mode

**Why**: Multiple micro-frontends can use the same Emotion key, which causes Emotion to accidentally clobber the first micro-frontend's styles by loading the second's

**How**: By simply detecting when we have already processed an existing `<style>` tag within a `StyleSheet` instance, we know never to move that particular `<style>` tag in the DOM again.

Emotion was previously moving the `<style>` tag to the same location in the DOM from the `StyleSheet._insertTag` method, and this caused the loss of styles.

**Checklist**:

- [x] Documentation (N/A)
- [X] Tests
- [X] Code complete
- [X] Changeset

### DIFF
--- a/.changeset/happy-stingrays-accept.md
+++ b/.changeset/happy-stingrays-accept.md
@@ -1,0 +1,6 @@
+---
+'@emotion/css': patch
+'@emotion/sheet': patch
+---
+
+Fix CSS rules loss issue when multiple instances used the same key (#2210)
--- a/.changeset/three-penguins-reflect.md
+++ b/.changeset/three-penguins-reflect.md
@@ -1,6 +1,9 @@
 ---
+'@emotion/cache': patch
 '@emotion/css': patch
+'@emotion/react': patch
 '@emotion/sheet': patch
+'@emotion/styled': patch
 ---
 
 Fix CSS rules loss issue when multiple instances used the same key (#2210)
--- a/packages/cache/__tests__/__snapshots__/hydration.js.snap
+++ b/packages/cache/__tests__/__snapshots__/hydration.js.snap
@@ -5,6 +5,7 @@ exports[`it works 1`] = `
   <head>
     <style
       data-emotion="css 1lrxbo5"
+      data-h=""
       data-s=""
     >
       .css-1lrxbo5{color:hotpink;}
@@ -19,6 +20,7 @@ exports[`rehydrated styles to head can be flushed 1`] = `
   <head>
     <style
       data-emotion="emo 1lrxbo5"
+      data-h=""
       data-s=""
     >
       .emo-1lrxbo5{color:hotpink;}
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@emotion/cache",
-  "version": "11.1.4",
+  "version": "11.1.3",
   "description": "emotion's cache",
   "main": "dist/emotion-cache.cjs.js",
   "module": "dist/emotion-cache.esm.js",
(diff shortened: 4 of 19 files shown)


Andarist:
This solution has a problem - each sheet associated with the same key will "own" style elements and if you flush a sheet the styles will be gone but the cache owning another sheet won't be notified by that and it won't be able to reinsert some styles.

Why can't you use different keys **or** reuse the same instance across your page?

gregjacobs:
@Andarist So I was thinking about the flush() issue too after I went to bed last night and was going to get up this morning and change the implementation, haha

I think the original solution I proposed in https://github.com/emotion-js/emotion/issues/2210#issuecomment-766427504 is best then, where basically only one `StyleSheet` instance will ever own a given `<style>` tag, and `flush()` will work correctly. 

Essentially this will make it so that the first Emotion instance for a given `key` will own any existing SSR `<style>` tag (if there is one), and subsequent instances will just create their own `<style>` tags from scratch. The subsequent instances will potentially duplicate some of the first `<style>` tag's styles, but I think that's okay - the page will still render correctly, and as you said, sharing the same key isn't really a "supported" use case anyway so I think a couple of duplicated style rules is okay. This will just make micro-frontend apps which happen to share the same key render correctly.

----

To answer your question about why we can't use different keys or the same Emotion instance on our page, I actually had written a small comment in https://github.com/emotion-js/emotion/issues/2210#issuecomment-766428070 of why it's important for micro-frontends for Emotion to support this use case in general, but to give a little more color to it, we actually don't own the micro-frontends and there will be many of them published by many different teams. I work for a large company, and in one division alone there are 1200 developers. Imagine trying to get them all to coordinate and stay in sync on Emotion keys, or even implementing a pattern to prevent them from clashing in each of their apps 😅  

The best solutions for us are most-often technical ones rather than human/pattern ones, which is why having Emotion do as-close-to-the-right-thing-as-possible (even if it's not a fully-supported use case) would be best.

gregjacobs:
Ok, so I added the new flag to make sure that all `<style>` tags are always owned by exactly one `StyleSheet` instance. I did this by adding a `data-h` attribute ("hydrated") to all `<style>` tags that have been processed. 

All tests pass, and I added multi-instance tests as well

Was just thinking though: could I possibly leverage the existing `data-s` attribute to determine when a `<style>` tag is already owned by a StyleSheet? What exactly does `data-s` represent? (If not, can leave it as `data-h`)

gregjacobs:
Btw, the changes in source code are just 2 lines. Had to update a bunch of snapshots though 😅

gregjacobs:
Hey @Andarist, any way to get that last CI check to pass? Seems to be stuck? 

Best,
Greg

Andarist:
This still has a problem with Emotion 11 hijacking some Emotion 10 styles as Emotion 10 doesn't put those data markers into the DOM at all. I'll try to figure out how to best handle this, considering all of those scenarios - unfortunately, I'm kinda burned out on Emotion a little bit right now so I'm not sure when I will be able to think this through

gregjacobs:
Believe this has already been fixed (discussion and PR link in https://github.com/emotion-js/emotion/issues/2210). Closing

Andarist:
@gregjacobs thank you for cleaning this up and for filing this PR in the first place. I'm sorry that we didn't get to your PR sooner so we could figure out the final solution and that the other (newer) PR has been pushed forward. It all really boiled down to your PR being created in a less convenient time for us (we had less free time to give to the project at that point in time) and this has slipped our attention.

gregjacobs:
Hey @Andarist. It's all good man, thanks for looking at it in the first place! I had lost track of this PR as well actually as I had worked around the issue on my side by ensuring a single global Emotion instance. Glad someone else was able to take the reins though and come up with a good solution! 

I appreciate the attention, kind words, and follow up, and I'll be back to contribute again in the future if the need should arise!

All the best,
Greg

PR #2233, opened 2021-01-29
Fix/2139 jest snapshots

### LINKED ISSUE #2139: @emotion/jest - Snapshot serializer is not replacing hashes
**Current behavior:**

I have updated project to emotion 11 and i have noticed failing jest tests. When i provide emotion styles as an array: `css={[style1, style2]}`, their hashes in snapshot will not be replaced.

Component with emotion styles: https://github.com/ejdzipi/emotion-jest-serializer-repro/blob/master/src/StyleArrayComponent/StyleArrayComponent.jsx

Resulting snapshot: https://github.com/ejdzipi/emotion-jest-serializer-repro/blob/master/src/StyleArrayComponent/__snapshots__/StyleArrayComponent.test.js.snap
 
**To reproduce:**

I have prepared a repro: https://github.com/ejdzipi/emotion-jest-serializer-repro

**Expected behavior:**

Hashes are correctly replaced

**Environment information:**

- `react` version: 16.14
- `@emotion/react` version: 11.1.1

ajs139:
**What**:

Fix #2139 - Snapshot serializer is not replacing hashes.

**Why**:

When creating snapshots for components that have an array as the css property, the snapshot was outputting some of the  internal properties. Additionally, there was a problem with the serialization when using `mount` on components that had the css properties on child components. 

**How**:

Fixed a bug where the component wasn't being unwrapped properly (so its children were not being modified).
Accounted for the `css` prop being an array.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset


packages/jest/test/__snapshots__/react-enzyme.test.js.snap
```diff
@@ -114,6 +114,33 @@ exports[`enzyme mount theming 1`] = `
...
+  background-color: black;
+  color: white;
+}
+
+<div
+  className="emotion-0"
+>
+  Test content
+  <div
+    className="emotion-1 emotion-2"
```
ajs139, line 135: need to look at why this extra class name is present

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
```diff
@@ -343,6 +368,24 @@ exports[`enzyme shallow theming 1`] = `
...
+  display: -ms-flexbox;
+  display: flex;
+}
+
+<div
+  className="emotion-0"
+>
+  Test content
+  <div
+    className="emotion-1 emotion-2"
```
ajs139, line 384: will look at this also

yarn.lock
```diff
@@ -1866,6 +1866,11 @@
   resolved "https://registry.yarnpkg.com/@emotion/memoize/-/memoize-0.7.3.tgz#5b6b1c11d6a6dddf1f2fc996f74cf3b219644d78"
   integrity sha512-2Md9mH6mvo+ygq1trTeVp2uzAKwE2P7In0cRpD/M9Q70aH8L+rxMLbb3JCN2JoSWsV2O+DdFjfbbXoMoLBczow==
 
+"@emotion/memoize@0.7.4":
```
ajs139, line 1869: will remove this

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
Andarist: It seems that you have already resolved this issue, right?

packages/jest/test/react-enzyme.test.js
```diff
@@ -191,6 +204,25 @@ describe('enzyme', () => {
     cases
   )
 
+  test('child containing css array property', () => {
```
Andarist, line 207: is this the correct test title? no css prop here holds an array

packages/jest/test/react-enzyme.test.js
```diff
@@ -143,6 +143,19 @@ const cases = {
       )
     }
   },
+  'using the css helper': {
```
Andarist, line 146: OTOH this one holds an array so I would assume that maybe the added test titles should be swapped but I'm also not sure what "using the css helper" test would be about - could you make this test title a little bit more descriptive?

packages/jest/test/react-enzyme.test.js
ajs139: updated, as noted the titles got switched

packages/jest/test/react-enzyme.test.js
ajs139: thanks, let me know if the new description is missing anything

.changeset/two-laws-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': patch
+---
+
+Fix issue creating enzyme snapshots for components using an array as the css prop.
```
Andarist, line 5:
```suggestion
---
'@emotion/jest': minor
---

Fixed an issue with Enzyme snapshots for components using an array as the `css` prop - those should be printed OK now.
```

Andarist (approved):

Issue #1892, opened 2020-06-07
Possibility to use class-name property instead of className


khmelevskii:
**The problem**

I created a tiny Emotion wrapper in ClojureScript. I'm going to open-source this wrapper but I have issue with `className` property. In Clojure idiomatic to use kebab-case notation. So, it would be great to use prop `class-name` in styled components instead of `className` but unfortunately I don't have any methods in emotion to change it.
Currently I have only one way, it's wrap styled component to `react.createElement` and convert `className` to `class-name`.

**Proposed solution**

It would be great to add this to options (somethings like `shouldForwardProp`). For example,
```
{
  classNameProps: 'class-name',
}
```
Link to this wrapper (WIP) https://github.com/khmelevskii/emotion-cljs

Andarist:
I'm afraid that this is out of Emotion's scope and we do not plan to make this customizable. The primary target of this library is regular React written in JS in which this works as expected.

Thanks for taking interest in Emotion though!

khmelevskii:
@Andarist yes, I see your point. In ClojureScript world it's very easy to use React and it would be great to have Emotion in our ecosystem too.
I agree that it's a little strange to add not relevant for JS option but maybe it's possible to add some common method, which I will be able to use for my case too.
For example,
```
options: {
  renameForwardProp: (propName) => propName === 'className' ? 'class-name' : propName,
  // or
  attachForwardProps: (props) => ({
    'class-name' => props.className,
  })
}
```
So, if we can think of some good method, which you guys approve and this method solves my case and is useful in JS world, I would love to create PR.

Andarist:
I have no idea about ClojureScript - so please bear with me. Are you reimplementing stuff in your library or just wrapping our JS code and provide "bindings" for the ClojureScript? Haven't you already implemented what you are looking for [here](https://github.com/khmelevskii/emotion-cljs/blob/01be75a1fe9ec7762f28fa8974e37284511639cf/src/emotion/core.cljs#L51)? If yes - what's your concern about the used technique?

khmelevskii:
Sure, no problem. `emotion-cljs` is just a tiny wrapper which provides "bindings" to Emotion. I don't implement any features and just provide idiomatic clojure syntax for working with emotion.
You are right, I handled this problem in my lib but unfortunately I have only one way how I can do it - I need to wrap component to another component where I convert `className` property.
[...]

PR #2235, opened 2021-01-31
Add shouldTransformProp option

khmelevskii:
**What**:
Add new options to be able to transform props #1892

**Why**:
In my case I created [emotion-cljs](https://github.com/khmelevskii/emotion-cljs) wrapper to work with emotion in ClojureScript.
In Clojure it is idiomatic to use kebab-case notation and most libraries support it. For example:
1. https://github.com/reagent-project/reagent
As you can see we use `on-change` instead of `onChange`, `on-mouse-move` instead of `onMouseMove`, `class-name` instead of `className` etc. Also it's possible to use `class` instead of `className` https://cljdoc.org/d/reagent/reagent/1.0.0/doc/tutorials/using-hiccup-to-describe-html#special-interpretation-of-class-attribute
> In JavaScript, class is a reserved keyword, so React uses the className to specify class attributes. Reagent just uses class.
2. https://github.com/lilactown/helix The same in another popular react wrapper.

So, it's idiomatic for clojurists and we would like to use kebab notation that's why it would be great to do the same in emotion-cljs and do it effectively.

With new option `shouldTransformProp` it would be easy to replace all kebab-case props to camelCase and pass them to bare react.

I described my case with ClojureScript, but this option can be potentially useful for another property transformations. If you think that this method shouldn't be in public interface, it would be great to have this option as undocumented one.

**How**:
This implementation shouldn't influence current code and just run transform function in correct place if this function was specified.
**Checklist**:

- [ ] Documentation. Ready to add if you are ready to add this option to documentation
- [ ] Tests. Ready to add if this PR can be approved.
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/packages/styled/src/base.js
+++ b/packages/styled/src/base.js
@@ -32,9 +32,11 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
 
   let identifierName
   let targetClassName
+  let shouldTransformProp;
   if (options !== undefined) {
     identifierName = options.label
     targetClassName = options.target
+    shouldTransformProp = options.shouldTransformProp
   }
 
   const shouldForwardProp = composeShouldForwardProps(tag, options, isReal)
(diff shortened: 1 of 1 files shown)


Andarist:
Closing as per discussion here: https://github.com/emotion-js/emotion/issues/1892#issuecomment-774922750

PR #2246, opened 2021-02-07
fix(eslint-plugin): ignore `css` function when checking for object style

### LINKED ISSUE #2243: [@emotion/eslint-plugin] the syntax-preference rule does not account for `css` function when checking check css props
**Current behavior:**

When using the `css` helper with the `css` prop, `@emotion/syntax-preference` doesn't think it's in object form:

```
ESLint: Styles should be written using objects.(@emotion/syntax-preference)
```

While it does work when you pass the object directly to the `css` prop, my understanding (and experience irl) is that the `css` function allows for stronger type-checking when using TypeScript.

If I set the style to `string`, then it does not error if I use `css` - I suspect this is probably the result of a fallback (i.e it checks "if is object else is string").

**To reproduce:**

```
import { css } from '@emotion/react';

function App() {
  return (
    <div css={appClassName}>
      <Title>Hello CodeSandbox</Title>
      {/* Inline CSS prop example with object notation */}
      <h2 css={css({ fontSize: 20 })}>
        Start editing to see some magic happen!
      </h2>
    </div>
  );
}
```

**Expected behavior:**

There to be no eslint warning on the `css` prop.

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version: 11.1.5
- `@emotion/eslint-plugin` version: 11.0.0

G-Rath:
**What**:

Adjusts `syntax-preference` eslint rule to support "ignoring" usage of the `css` function when checking if the `object` style is being used for the `css` prop.

**Why**:

When using the `css` helper with the `css` prop, `@emotion/syntax-preference` doesn't think it's in object form:

```
ESLint: Styles should be written using objects.(@emotion/syntax-preference)
```

(See #2243)

**How**:

I've adjusted `checkCssPropExpressionPreferringObject` to double-back on itself if the `node` is a `CallExpression` with a single argument, so effectively it "ignores" the function and looks at the argument instead.

I went with checking for a single argument over `callee.name === 'css'` to support if people imported the function under a different name. The potential trade-off is if people have their own wrapper function that takes only a single argument, the rule will complain if that argument isn't an object or array.

I've got no strong issue with either pro/con, so happy to change the check to be the other way around.

**Checklist**:

- [x] Documentation (N/A)
- [x] Tests
- [x] Code complete
- [x] Changeset 

Fixes #2243

### DIFF
--- a/.changeset/long-ladybugs-pull.md
+++ b/.changeset/long-ladybugs-pull.md
@@ -2,4 +2,4 @@
 '@emotion/eslint-plugin': patch
 ---
 
-Adjusts `syntax-preference` eslint rule to support "ignoring" usage of the `css` function when checking the `css` prop for object syntax.
+Improved `syntax-preference` rule to support `css` function and check style types for arguments of `css` & styled calls.
--- a/.changeset/long-ladybugs-pull.md
+++ b/.changeset/long-ladybugs-pull.md
@@ -1,0 +1,5 @@
+---
+'@emotion/eslint-plugin': patch
+---
+
+Adjusts `syntax-preference` eslint rule to support "ignoring" usage of the `css` function when checking the `css` prop for object syntax.
--- a/packages/eslint-plugin/src/rules/syntax-preference.js
+++ b/packages/eslint-plugin/src/rules/syntax-preference.js
@@ -4,6 +4,9 @@
  */
 
 function isStringStyle(node) {
+  if (node.tag.type === 'Identifier' && node.tag.name === 'css') {
+    return true
+  }
   // shorthand notation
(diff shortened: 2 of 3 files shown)


.changeset/long-ladybugs-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/eslint-plugin': patch
+---
+
+Adjusts `syntax-preference` eslint rule to support "ignoring" usage of the `css` function when checking the `css` prop for object syntax.
```
Andarist, line 5:
```suggestion
Improved `syntax-preference` rule to support `css` function and check style types for arguments of `css` & styled calls.
```

PR #2248, opened 2021-02-09
Import css from core to ensure correct import even when emotion/css@11 is hoisted

xanido:
**What**:
This ensures that users of `@emotion/core` and the `babel-plugin-emotion` do not encounter 'broken' transpiles in situations where `@emotion/css@v11` is hoisted.

**Why**:
In a project where the only (non-babel) emotion dependency is `@emotion/core` v10 ([as per docs here](https://5faaafd0bd0f3f0008469537--emotion.netlify.app/docs/introduction#react)), and is build with `babel-plugin-emotion`, this code:

```js
import React from 'react';
import { css } from '@emotion/core';

const ArrayCss = () => (
  <section
    css={[
      css`font-weight: bold;`,
      css`font-style: italic;`,
    ]}
  >
    Hello world
  </section>
);
```

Results in the following transpiled output:

```js
// notice that this import was added - referring to a package
// that is not a direct dependency of the project (but is available,
// as `@emotion/css` is hoisted from `@emotion/core`s deps)
import _css from "@emotion/css";
import React from 'react';
import { css } from '@emotion/core';
import { jsx as ___EmotionJSX } from "@emotion/core";
var _ref = {
  name: "in3yi3",
  styles: "font-weight:bold;"
};
var _ref2 = {
  name: "1msjh1x",
  styles: "font-style:italic;"
};

var ArrayCss = function ArrayCss() {
  return ___EmotionJSX("section", {
    css: /*#__PURE__*/_css([_ref, _ref2])
  }, "Hello world");
};
```

This works as long as the version of `@emotion/css` that's hoisted is v10. If v11 is hoisted this will blow up, as v11 [does not have a default export](https://github.com/emotion-js/emotion/blob/0c31ed05e564c4e68d11862e23d96003c6b7e4b5/packages/css/src/index.js#L4).

A minimal reproduction of such a scenario is available here: https://github.com/xanido/babel-plugin-emotion-issue-demo

**How**:
Modifed the babel plugin to import `css` from `@emotion/core` when performing `cssPropOptimization`. 

`@emotion/core` itself re-exports `css`, which guarantees that the correct version of `@emotion/css` will ultimately be imported in the transpiled code.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/kind-coats-doubt.md
+++ b/.changeset/kind-coats-doubt.md
@@ -1,0 +1,5 @@
+---
+'babel-plugin-emotion': patch
+---
+
+optimiseCssProp results in `css` being imported from @emotion/core rather than @emotion/css
--- a/packages/babel-plugin-emotion/__tests__/source-maps/__snapshots__/index.js.snap
+++ b/packages/babel-plugin-emotion/__tests__/source-maps/__snapshots__/index.js.snap
@@ -66,7 +66,7 @@ const SomeComponent = props => (
 
       ↓ ↓ ↓ ↓ ↓ ↓
 
-import _css from \\"@emotion/css\\";
+import { css as _css } from \\"@emotion/core\\";
 
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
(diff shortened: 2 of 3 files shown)


Andarist:
It makes sense for me as `@emotion/css` had a rather big chance of not being a direct (peer)dependency of the project. Any objections @mitchellhamilton ?

mitchellhamilton:
No objections, getting this in def sounds good. I'll setup publishing on the v10 branch and get this published

mitchellhamilton (approved):
Thanks!

Issue #2257, opened 2021-02-13
Unbearably slow typescript definitions


charrondev:
**Current behavior:**

Some of the built-in types cause type-checking and VSCode intellisense feature to be incredibly, painfully slow.

**To reproduce:**

This only seems to affect large projects. For me that's working on https://github.com/vanilla/vanilla (working specifically with react components in `library/src/scripts`, but it reproduces in other areas as well).

I've recorded some gifs of the performance degradation.

|Situation|Gif|
|---|---|
|Current Behaviour|![2021-02-13 12 43 43](https://user-images.githubusercontent.com/1770056/107856957-a6f09100-6df9-11eb-88fd-f2dad474775b.gif)|
|With a slight modification of the type|![2021-02-13 12 44 16](https://user-images.githubusercontent.com/1770056/107856965-b079f900-6df9-11eb-8533-c76dbd85c0bd.gif)|
 

**Expected behavior:**

Intellisense on a machine as powerful as the one I'm using should be instant. (32Gb of RAM, i9 CPU, lightning fast SSD).

**Environment information:**

- `react` version: 16.9.0
- `@emotion/css` version: 11.1.3

## The slowness cause

Typescript has had a lot of issues recently with various libraries replicating CSS or react types. https://github.com/microsoft/TypeScript/issues/34801#issuecomment-767026973

In this case the cause of the slowness in particular seems to come from this piece of code

https://github.com/emotion-js/emotion/blob/master/packages/serialize/types/index.d.ts#L10-L14

```ts
export type CSSPropertiesWithMultiValues = {
  [K in keyof CSSProperties]:
    | CSSProperties[K]
    | Array<Extract<CSSProperties[K], string>> // The Extract<> here.
}
```

It seems the `Extract<>` on something with as many properties as this `CSSProperties` is very slow. What I've done for my own personal use was to change the code to 

```ts
export type CSSPropertiesWithMultiValues = {
  [K in keyof CSSProperties]:
    | CSSProperties[K]
    | Array<CSSProperties[K]>>
}
```

Now this is **_may not be the correct type_** of what's allowed, as it it then allows multiple values an array of values to be passed to properties that don't support it. My question is until typescript improves the performance of `Extract<>` in a situation like this, what do we do?

charrondev:
Of course immediately after filing the issue, I think I found a more correct solution.

```
export type CSSPropertiesWithMultiValues = {
  [K in keyof CSSProperties]:
[...]

PR #2258, opened 2021-02-13
Fix slowdown in typechecker from `CSSPropertiesWithMultiValues`

charrondev:
**What**:

Fixes https://github.com/emotion-js/emotion/issues/2257

**Why**:

This change greatly improve IDE performance on large projects. It also improves correctness (`Array<never>` isn't the best way to show that you don't want to allow array values).

**How**:

The previous behaviour here was working like this. Here's a slimmed down example.

```ts
type CSSProperties = {
  prop1: number,
  prop2: "val1" | "val2" | string,
}

// After

type CSSPropertiesWithMultiValues = {
  prop1: number | Array<never>,
  prop2: "val1" | "val2" | string | Array<"val1" | "val2" | string>,
}
```

Now it works like this.

```ts
type CSSProperties = {
  prop1: number,
  prop2: "val1" | "val2" | string,
}

// After

type CSSPropertiesWithMultiValues = {
  prop1: number,
  prop2: "val1" | "val2" | string | Array<"val1" | "val2" | string>,
}
```

I'm not actually 100% sure which values don't support strings though. Don't all of the properties support `inherit` or `auto` or some pixel value something along those lines?

I'm assuming they are and this code was written for a reason, but if it's not necessary the conditional type could be removed entirely.

**Checklist**:

- [x] Documentation - N/A
- [x] Tests - N/A
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -9,8 +9,9 @@ export { RegisteredCache, SerializedStyles }
 export type CSSProperties = CSS.PropertiesFallback<number | string>
 export type CSSPropertiesWithMultiValues = {
   [K in keyof CSSProperties]:
-    | CSSProperties[K]
-    | Array<Extract<CSSProperties[K], string>>
+    CSSProperties[K] extends string
+      ? CSSProperties[K] | Array<CSSProperties[K]>
+      : CSSProperties[K]
 }
 
 export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject }


Andarist:
I'm actually not sure why this was needed in the first place - I mean the specialcasing of string properties.

```ts
type A = keyof CSSProperties // some properties listed + ...755 more

type NonNever<T extends {}> = Pick<
  T,
  { [K in keyof T]: T[K] extends never ? never : K }[keyof T]
>
type B = keyof NonNever<{
  [K in keyof CSSProperties]: CSSProperties[K] extends string
    ? never
    : CSSProperties[K]
}> // some properties listed + ...756 more
```

I can't quite find any property that would actually pass this conditional type. This most likely comes from the fact that all of those properties are optional and thus their types include `undefined` and:
```ts
type Test = string | undefined extends string ? 1 : 0 // 0
```

charrondev:
Would it be best if I just removed the conditional and made it always apply the array option?

Like this

```ts
export type CSSPropertiesWithMultiValues = {
  [K in keyof CSSProperties]:
    | CSSProperties[K]
    | Array<CSSProperties[K], string>
}
```

mitchellhamilton:
I'm guessing we don't have type tests for the array syntax and this change doesn't actually work because the array case in the conditional type in this PR will always return false. It previously sort of worked because `Extract` distributes over the union but `CSSProperties[K] extends string` doesn't(since `CSSProperties[K]` is not just a type parameter and distribution in conditional types only happens when the thing is only a type parameter). IMO, the correct behavior would be to use `Exclude<CSSProperties[K], undefined>`.

https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKygXgN4CgpQIYH4BcAzsAE4CWAdgOYDcmUcBx51UAPlBQK4C2cEJOgF906UJCgAxEgHseABQBKKKBiwBtANIQQUSlADWOmQDNYiALr5zCLTotQIAD2AQKAE0JRmlKlFxY8LbaIA4cAIIkJNggADxBdqEAfFDWCSEW6CJi4NAAwlxRbsAAQhAAFtgAbmQyhSpqUIl6FIbGZkFWgYiJYVCR0XEAoi7RAMbA8T0ZADTepL5JSVmi4vkyRRNllTV1JA30zfpGIKY2XTa97P1RMbEjYwA2XO4QU8H2c1weECaUEO4llkgA

Andarist:
An additional playground showcasing why this matters:
https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKygXgN4CgpQIYH4BcAzsAE4CWAdgOYDcmUcBx51UAPlBQK4C2cEJOgF906UJCgAxEgHseABQBKKKBiwBtANIQQUSlADWOmQDNYiALr5zCLTotQIAD2AQKAE0JRmlKlFxY8LbaIA4cAIIkJNggADxBdqEAfFDWCSEW6CJi4NAAwlxRbsAAQhAAFtgAbmQyhSpqUIl6FIbGZkFWgYiJYVCR0XEAoi7RAMbA8T0ZADTepL5JSVmi4vkyRRNllTV1JA30zfpGIKY2XTa97P1RMbEjYwA2XO4QU8H2c1weECaUEO4lischIAMpkHhgR7QZCqQ4hFptU4dSzWS4Za4DO7pezLbJjGQUYhQEyyBSKazSORKA5YbDWADkrmIDJm9Dg1nUDJMMhkrKgABYAEyZfGE4ljQokYrWApFCilCrVWr1WGNelQJkQFlsrAcprc3n84Wi0QEonAKAEzbAWUbaVbJW7VVwumM5nAVnszmGvlzE3A83EwgQqEQazgyHQ2k4d3az26hg+nl+wUi4EAegzUAAtHmzeLLdaHcAhXabdtlXsYxqtTrvQaU8ahV8fn8KADTegg5aQ1GIGWoJGwzW4-W9cmjf6W1Bvq9252skA

OTOH If the correct approach would yield significant performance degradation over the simple approach I'm not sure if it would be worth pursuing it. @charrondev could you check if the correct version works OK in your codebase?

charrondev:
Thanks @Andarist for the example. That illustrates the problem quite well.

A quick testing shows that the "Correct" way defined there seems to still be performant in my codebase, but I'll give it a full workday with that type in there and see how it goes.

charrondev:
After working with it for a bit, I was still having issues. Seeing the example this seems to give the same results though as your correct example.


**[Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKygXigbwFBSgQwPwBcUAzsAE4CWAdgOYDcWUchJ51NUAPlFQK4C2cCGQYBfDBlCQoAMTIB7fgAUASinSMA2gGkIIKNSgBrPfIBmsRAF0i8BDr1WoEAB7AIVACbFWlWlDxLe10QJ24AQTIyHBAAHjsHUIA+KFtERKsMcUlwaABhXiiPYAAhCAALHAA3CnlC9UxsRIMqY1MLOxsgjK4oSOi4gFE3aIBjYHj0kKsAGl92JKSsiSl8+SLxssqaurIGrRCWtpBzIK6E6d7+mNjh0YAbXk8ISeDHOd4vCDNqCE9F5Y5aThe7APLrMgQTYVaq1eqoRpQZqGEwnDrWNJvUJXKI3ABy8ioeN493uODg9xeF0cizEK1yUAAyhR+GBKfsmocUe0zpiehFcXFqclAaNCaQoGYFMoVEQ5IpVBzcEQAOTuUgqmaMOBETQqszyeSaqAAFgATJlsmKqBLRoVIVRgEQCkVHVtYbslThVergJrtbr9YbjebLRJrbaIVCnVBwRtSjCdvCNNhvVA1RANVrsDqkUGjXNQ6LxcBcKC45DxkQQWCo9DtnC9gjGGmM1mA3mDQXTRbizbS8QWWyIERmaz2c3Uz7M37s0xA12Q73sgB6FdQAC0W-DJagYvjZuddYTDc9k+V6d9-pzC+DhbNHy+PyofzDGAjpZw5ePh763-j7pJk2KYXm2s4dnqi73o+zzPq+fYSoO44QL+Y7Dl607tjenZ3j2MHfL8niWkAA)**

```ts
type AltCorrectBehaviour = {
  [Key in keyof Obj]: Obj[Key] | Array<NonNullable<Obj[Key]>>;
}
```


I'll give that a shot.

Andarist:
Using `NonNullable` shouldn't really change any perf characteristics that much as those 2 are defined as follow:
```ts
type Exclude<T, U> = T extends U ? never : T;
type NonNullable<T> = T extends null | undefined ? never : T;
```
So `NonNullable` is pretty much just the shortcut for:
```ts
type NonNullable<T> = Exclude<T, null | undefined>
```

> I think the slow thing is checks against never[].

Right, that is probably the case in the current code but neither Exclude nor NonNullable approach should lead to `never[]` anywhere in this scenario, right?

amcasey:
This seems worthwhile as a correctness fix, but I'm not seeing a significant change in tsc check time.  Are you, @charrondev, or is this mostly affecting the editing experience?  Do you have steps I could follow to reproduce the speedup you're seeing?  (Context: if you've made it faster, I'd like to understand how so we can try to give more users the benefit of the improvement - by documenting best practices, if nothing else.)  Thanks!

charrondev:
For me the slowdown isn't necessarily noticeable in type checking the whole project at once. The main slowdown is intellisense in my IDE.

I'll see if I can make a reproduction.

amcasey:
@charrondev, I do see a pretty noticeable tsc check time improvement with your original suggestion,  `CSSProperties[K] extends string ? CSSProperties[K] | Array<CSSProperties[K]> : CSSProperties[K]` but, if I understand the thread above correctly, that wasn't quite correct and the new fix is `CSSProperties[K] | Array<Exclude<CSSProperties[K], undefined>>`.  Is that right?

Fortunately, I think replacing `ArrayCSSInterpolation` with a type alias (as suggested in the bug) gives the same perf boost, even with the correct `CSSPropertiesWithMultiValues`.

Regardless, I'm still interested in knowing what editor operations you're using as a test case.

p.s.  I'm also seeing a ~20% reduction in check time when upgrading from TS 4.0 to 4.2.  Something to consider. 😉

PR #2261, opened 2021-02-20
Adding reference to React 17's requirements

Robbie-Cook:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/docs/install.mdx
+++ b/docs/install.mdx
@@ -46,6 +46,7 @@ render(
   </SomeComponent>
 )
 ```
+If you are using React 17, you will need to use `/** @jsxImportSource @emotion/react */` as your prefix instead of `/** @jsx jsx */`.
 
 ## With [`styled`](/docs/styled.md)
 


docs/install.mdx
```diff
@@ -46,6 +46,7 @@ render(
   </SomeComponent>
 )
 ```
+If you are using React 17, you will need to use `/** @jsxImportSource @emotion/react */` as your prefix instead of `/** @jsx jsx */`.
```
Andarist, line 49: Well, technically this is not 100% correct - with React 17 you *don't have to* use the `@jsxImportSource` - you just have such an option.

The whole thing is quite hard to document on our side because of how various tools handle this 😭

Robbie-Cook: Fair enough, thanks for this! Do you need `/** @jsxRuntime classic */` on React 17 projects to handle this?

PR #2264, opened 2021-02-22
Fix incorrectly throwing error on the gradient content values in dev

lucasweng:
**What**:
Incorrectly throwing error on gradient content values in dev

**Why**:
This change allows using gradient content values in dev. Per doc https://developer.mozilla.org/en-US/docs/Web/CSS/content#values, `<gradient>` is a valid data type for the content property.

**How**:
Update `contentValuePattern` regex to allow valid gradient function types https://developer.mozilla.org/en-US/docs/Web/CSS/gradient#syntax
```js
let contentValuePattern = /(attr|calc|counters?|url|(((repeating-)?(linear|radial))|conic)-gradient)\(/
```

**Checklist**:

- [x] Documentation - N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/fair-buses-drop.md
+++ b/.changeset/fair-buses-drop.md
@@ -1,0 +1,5 @@
+---
+'@emotion/serialize': patch
+---
+
+Fix incorrectly throwing error on the gradient content values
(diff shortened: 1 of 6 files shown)


mitchellhamilton (approved):
Thanks!

PR #2269, opened 2021-02-25
Fix enzyme serializer for conditional styles

mskelton:
**What**: The change in #2233 introduced a bug where conditional styles passed to the css prop would break the Enzyme snapshot serializer.

**Why**: To allow users who use Enzyme snapshots to use the Emotion snapshot serializer.

**How**: Return early if the style variable is falsy to prevent errors

**Checklist**: Yes

- [ ] Documentation - N/A
- [x] Tests
- [x] Code complete
- [x] Changeset 

**Screenshot**

Screen Shot 2021-02-25 at 9 24 09 AM


packages/jest/test/__snapshots__/react-enzyme.test.js.snap
```diff
@@ -14,6 +14,47 @@ exports[`enzyme mount basic 1`] = `
...
 `;
 
+exports[`enzyme mount conditional styles 1`] = `
+.emotion-0 {
+  background-color: black;
+}
+
+.emotion-11rveqsm-Greeting {
+  background-color: red;
+}
```
mskelton, line 24: I'm struggling with getting this to work right.  When rendering with conditional styles, it is inserting styles from other tests in the spec and I'm not entirely sure why.  I tried just filtering out the falsy values from the css prop first but that didn't help, so I must be doing something wrong.  Help appreciated.

packages/jest/src/create-serializer.js
```diff
@@ -89,15 +89,17 @@ function filterEmotionProps(props = {}) {
   return rest
 }
 
+function getStylesFromCss(css) {
+  return (Array.isArray(css) ? css : [css]).filter(Boolean)
```
Andarist, line 93: `css` can contain nested arrays etc so I think that you should let things through and handle them more granularly - when there is an attempt to operate on an item. That would allow for a natural flow of the program - rather than purifying this recursively upfront.

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
Andarist: Interesting - I might be wrong as I'm just trying to guess what might be the culprit here but my guess is that for some reason one of those return too much:
https://github.com/emotion-js/emotion/blob/fa1c25dae5967e71bc28ee7828ecf6a9d84f09e3/packages/jest/src/create-serializer.js#L222-L223

So to fix this I would inspect what happens in those 2 calls.

packages/jest/src/create-serializer.js
mskelton: I updated my branch per your suggestion.

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
mskelton: Okay, so I fixed this issue and only the styles for the element are being returned.  Yay!  I think there are two remaining issues:

1. `isShallowEnzymeElement` does not correctly identify shallow vs mounted elements who had a css prop that was fasly.  This results in the following when mounting

```html
<div
  className="emotion-0"
>
  <span
    css="unknown styles"
  >
    <span
      className="emotion-1"
    >
      Test content
    </span>
  </span>
</div>
```

2. Conditional styles are getting `css="unknown styles"` since the class name is not added so when the css prop is attempted to be cleaned up it isn't since `hasKnownClass` is false.

Any thoughts you might have on these issues would be greatly appreciated.

Andarist: Could you add a failing test? I have recognized some other problems with shallowly rendered components which I'm investigating right now~ so this would be helpful for me.

Andarist:
@mskelton I've pushed out some changes to the implementation which simplify how `css` prop is handled by the Enzyme serializer and makes the thing actually more correct for a wider range of scenarios. The test case that I've added was not handled correctly before my changes: https://github.com/emotion-js/emotion/pull/2269/files#diff-06a0eb3ce19d24a37e94c9307f74556cf44e9aac4d841bfed1dd467e44ac62a8R159 . Please take a look and feel free to do a review of the approach that I've taken here.

@ajs139 If you could take a look at what I'm doing here I would highly appreciate it. One thing that I know this won't handle are mixed caches in a single tree - but this seems so far-fetched that I don't really want to handle this right now.

packages/jest/src/enzyme-tickler.js
```diff
@@ -0,0 +1,19 @@
+const tickledCssProps = new WeakMap()
+
+export const getTickledClassName = cssProp => tickledCssProps.get(cssProp)
+
+export const tickle = wrapper => {
+  const isShallow = typeof wrapper.dive === 'function'
+
+  wrapper.find('EmotionCssPropInternal').forEach(el => {
```
ajs139, line 8: This is much more surgical than the previous implementation, looks great.

ajs139 (approved):

packages/jest/src/enzyme-tickler.js
```diff
@@ -0,0 +1,19 @@
+const tickledCssProps = new WeakMap()
+
+export const getTickledClassName = cssProp => tickledCssProps.get(cssProp)
```
ajs139, line 3:
```suggestion
export const getTickledClassName = cssProp => {
  if(!tickledCssProps.has(cssProp)) {
    throw new Error('Could not find cssProp, ensure enzyme snapshots are configured correctly - https://emotion.sh/docs/testing.');
  }
  return tickledCssProps.get(cssProp)
};
```

ajs139: Just in case we detect a shallow component, but it wasn't "tickled".

mskelton:
@Andarist Since I opened the PR, I can't approve it, but your updates look awesome and fix all the remaining bugs!  Can't wait for this to be fully completed so I can finish updating my company's apps to Emotion 11.  Great work!

packages/jest/src/enzyme-tickler.js
Andarist: Actually - we'd have to complicate this check because the `css` prop can "resolve" to a null (or other falsy values and:
```js
const wm = new WeakMap()
wm.set(null, {}) // throws, that's why we return for `!cssProp` in `tickle`
wm.get(null) // ok, just returns undefined
```

So I would be in favor of keeping this as-is for simplicity. I don't see when the proposed error would actually throw and it might be friendlier to the user to actually print something, rather than preventing them to work for some kind of an edge case. This is not a crucial production path after all.

ajs139: Sounds good, thanks for looking into it.

PR #2271, opened 2021-02-26
Fix: topbar width causing horizontal scroll

vilaboim:
**What**:

The width of the black topbar in the site breaks the layout when there is a vertical scroll, causing the whole page to scroll horizontally to accommodate the width of the top bar.

This PR allows the topbar to fit the width page so that the top bar doesn't break the width of the page, preventing the entire site layout from scrolling horizontally.

scroll

**Why**:

To prevent horizontal scroll.

**How**:

Changing `width: 100vw` to `width: 100%` on Black Lives Matter div.

**Checklist**:

- [ ] Documentation - N/A
- [ ] Tests - N/A
- [ ] Code complete - N/A
- [ ] Changeset - N/A

### DIFF
--- a/site/src/components/SiteHeader.js
+++ b/site/src/components/SiteHeader.js
@@ -40,7 +40,7 @@ export default function SiteHeader() {
             display: 'flex',
             justifyContent: 'center',
             backgroundColor: 'black',
-            width: '100vw',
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks!

PR #2276, opened 2021-03-05
Make readonly arrays accepted by cx()

garronej:
The type definitions of `ArrayClassNamesArg` that defines what the `cx()` function can accept as argument.

Because: 

image

should not give error. `cx` is pure, the array passed will never be edited. 

Using `ReadonlyArray` instead or `Array`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [ ] Code complete
- [ ] Changeset 

Hi, sorry I give up getting everything checked, I get `yarn install` errors and really need to move on.
As a result I don't expect this to be merged but I submit the PR anyway just to let you know about the issue.

Thank you for your amazing work.

### DIFF
--- a/packages/css/types/create-instance.d.ts
+++ b/packages/css/types/create-instance.d.ts
@@ -14,7 +14,7 @@ export {
 
 export { EmotionCache, Options }
 
-export interface ArrayClassNamesArg extends Array<ClassNamesArg> {}
+export interface ArrayClassNamesArg extends ReadonlyArray<ClassNamesArg> {}
 export type ClassNamesArg =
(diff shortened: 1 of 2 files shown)


garronej:
up?

Andarist:
This one has already been changed in https://github.com/emotion-js/emotion/pull/2057 but I lost track of this PR. I've revived it here: https://github.com/emotion-js/emotion/pull/2295 . Unfortunately, it's failing our type tests right now - I need to investigate it. If you help out with this - I would highly appreciate it.

PR #2279, opened 2021-03-08
Fix incorrectly throwing error on valid content values in dev

lucasweng:
**What**:
Fix incorrectly throwing error on valid content values in dev

**Why**:
This change allows using multiple valid content values like `content: open-quote counter(chapter_counter);` in dev. Per doc https://developer.mozilla.org/en-US/docs/Web/CSS/content, several values can be used simultaneously, except for `normal`, `none` and global values.

**How**:
- Update `contentValuePattern` regex for the cases of multiple content values
- Remove invalid regex pattern `calc`
- Remove invalid content value `counter` in `contentValues`
```js
let contentValuePattern = /(attr|counters?|url|(((repeating-)?(linear|radial))|conic)-gradient)\(|(no-)?(open|close)-quote/
let contentValues = ['normal', 'none', 'initial', 'inherit', 'unset']
```

**Checklist**:

- [x] Documentation - N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/sweet-donkeys-film.md
+++ b/.changeset/sweet-donkeys-film.md
@@ -1,0 +1,5 @@
+---
+'@emotion/serialize': patch
+---
+
+Fix incorrectly throwing error on valid content values
--- a/.changeset/sweet-donkeys-film.md
+++ b/.changeset/sweet-donkeys-film.md
@@ -2,4 +2,4 @@
 '@emotion/serialize': patch
 ---
 
-Fix incorrectly throwing error on valid content values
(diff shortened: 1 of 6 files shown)


.changeset/sweet-donkeys-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/serialize': patch
+---
+
+Fix incorrectly throwing error on valid content values
```
Andarist, line 5:
```suggestion
Fixed the issue with incorrectly throwing errors on valid content values containing `open-quote`.
```

PR #2283, opened 2021-03-11
Fixed an issue with styles containing a trailing css variable not being inserted correctly where the styled were optimized by our Babel plugin to a static opaque object

### LINKED ISSUE #2219: Production build issue when attribute selectors are used with @emotion/styled approach
**Current behavior:**

Attribute selector doesn't work in production build when the styled approach is used. Currently, it can be solved by using css`` approach instead. 

**To reproduce:**

1. Create an empty react CRA project
2. Use the following styled component styles:
`const ComponentStyled = styled.div`
  &[data-placement*='top'] .nested-element-selector {
    <any rules here>
  }
`;
`

**Expected behavior:**

CSS styles should be applied for production build, but it is working only in dev.

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version: 11.1.4
- `@emotion/styled` version: 11.0.0
- `@emotion/css` version: 11.1.3

Andarist:
fixes #2219


packages/cache/src/index.js
```diff
@@ -164,7 +164,9 @@ let createCache = (options: Options): EmotionCache => {
         }
       }
 
-      stylis(selector ? `${selector}{${serialized.styles}}` : serialized.styles)
+      stylis(
+        selector ? `${selector}{${serialized.styles};}` : serialized.styles
```
Andarist, line 168: this is the core of the fix - `}` is allowed in css variables so it was parsed as part of it and the whole thing resulted in incorrect styles as the opaque object does not contain a trailing `;`

packages/babel-plugin/src/index.js
```diff
@@ -12,6 +12,20 @@ import coreMacro, {
...
+const getEmotionBabelPluginPragmaOptions = state => {
+  if (!state.file.ast.comments) {
+    return {}
+  }
+  return state.file.ast.comments
+    .map(comment => comment.value.match(EMOTION_BABEL_PLUGIN_ANNOTATION_REGEX))
+    .filter(Boolean)
+    .map(match => JSON.parse(`{${match[1]}}`))
+    .reduce((a, b) => Object.assign(a, b), {})
+}
```
mitchellhamilton, line 27: I'd rather we didn't have this because people could start relying on it. Maybe we could do the same parsing in the babel test infra?

Andarist: We could - that was my initial intention but I didn't have time to figure out how this should be done properly. Do you have any idea how this could be done without sharing the parsed options through the state (which would basically also allow people to abuse it)? I didn't want to set up any more complicated infra that would sidestep Babel for this.

In the ideal situation, we'd just provide those options through regular options but that can't be done easily to the best of my knowledge (unless we read files in `test` and do our own parsing there).

PR #2287, opened 2021-03-14
Remove babel-plugin from emotion/css

bloodyowl:
**What**:

This PR removes babel-plugin from emotion-css dependencies

**Why**:

babel-plugin brings unnecessary dependencies

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -18,17 +18,20 @@
     "test:typescript": "dtslint types"
   },
   "dependencies": {
-    "@emotion/babel-plugin": "^11.0.0",
     "@emotion/cache": "^11.1.3",
     "@emotion/serialize": "^1.0.0",
     "@emotion/sheet": "^1.0.0",
     "@emotion/utils": "^1.0.0"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0"
+    "@babel/core": "^7.0.0",
+    "@emotion/babel-plugin": "^11.0.0"
   },
   "peerDependenciesMeta": {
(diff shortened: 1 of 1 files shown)


Andarist:
You are right that this is sort-of unnecessary but also - it shouldn't create many problems in practice and having this as regular dep is simpler for those users that intend to use the macro variant.

bloodyowl:
I'm thinking of contexts 
 where you want to keep your node_modules light (e.g. docker images). Given there's no direct link to the dependency, that could save some bandwidth and storage 😊

Andarist:
Well, how many kilobytes does this save you though? I strongly believe that there are a lot of better opportunities to handle first before this one becomes a great offender.

bloodyowl:
Screen Shot 2021-03-15 at 12 25 38

Saves ~60% of the package size (3,9MB), and 75% in disk space on APFS (9,8MB)

Andarist:
@mitchellhamilton u were in favor of not doing this - space savings seem to be pretty big though, should we adjust this?

mitchellhamilton:
This would be a breaking change and we're not really planning a major anytime soon so it's kind of a non-starter right now though I think I'd be mostly okay with moving it to an optional peer dep now that we can express that.

Andarist:
Leaving this open as a reminder for ourselves so we don't forget about it when releasing v12. It won't happen any time soon though.

PR #2294, opened 2021-03-17
Upgrade benchmark setup to Parcel 2 (beta)

Andarist:
Our flow CI job includes building the benchmark app~ and it's failing now. I could have just removed that step from that job as this is not closely related, nor it really validates anything but I've tried to fix this issue by upgrading to Parcel 2. Unfortunately, I came across a bug in their scope hoisting algorithm and the compiled JS won't run.

Leaving this here as a reminder to try this sometime later when they release new versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ site/build
 package-lock.json
 .DS_Store
 .cache
+.parcel-cache
 public/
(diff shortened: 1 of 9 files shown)


Andarist:
Superseded by https://github.com/emotion-js/emotion/pull/2544

PR #2295, opened 2021-03-17
Readonly arrays

Andarist:
It's a continuation of https://github.com/emotion-js/emotion/pull/2057


packages/cache/types/index.d.ts
```diff
@@ -18,20 +18,20 @@ export interface StylisElement {
...
-  children: Array<StylisElement>,
+  children: ReadonlyArray<StylisElement>,
   callback: StylisPluginCallback
 ) => string | undefined
 
 export type StylisPlugin = (
   element: StylisElement,
   index: number,
-  children: Array<StylisElement>,
+  children: ReadonlyArray<StylisElement>,
```
mitchellhamilton, line 28: Changing these to Readonly seem wrong? Aren't these arrays mutable?

packages/styled/types/base.d.ts
```diff
@@ -91,14 +87,10 @@ export interface CreateStyledComponent<
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
    */
-  <AdditionalProps extends {}>(
+  <AdditionalProps extends {} = {}>(
```
mitchellhamilton, line 90: What's this for?

Andarist: It's leftover that I've missed when doing this revert: https://github.com/emotion-js/emotion/pull/2295/commits/7135e61a1807315d624e58005f7105e605236f3e

The other overload has this and this one seemed to have at the time when https://github.com/emotion-js/emotion/pull/2057 was created. It was removed as part of this PR: https://github.com/emotion-js/emotion/commit/0f3a2d340269c5b446fdc8dcf2b0a4e45d998ad2#diff-3fae48ecece75c5d911928f5c96c7292ac893baae5232e9868e68446de341401R93

I'll remove it - nice catch 👍

packages/cache/types/index.d.ts
Andarist: Yep, you are right 👍

mitchellhamilton (approved):

PR #2296, opened 2021-03-17
Avoid creating TypeScript interfaces that extend `Array`

Andarist:


mitchellhamilton (commented):
I agree with this change though this would break for people on TypeScript versions under 3.7 since that's when recursive type aliases like this started being allowed. I'm not sure if I'm comfortable breaking people though maybe it's been enough time that it's reasonable to expect people to be on 3.7 or above? wdyt?

Andarist:
Nice catch! I haven't thought about this. I don't think we are giving any guarantees regarding TS version compatibility and it is quite impractical to tie our major versions to this. I would be OK with shipping this and if it causes too many problems then we could consider using `"typesVersions"`. This would complicate the setup a little bit so it would be unfortunate - the perf improvement here is not dramatic but it's somewhat noticeable from what I understand so it would be great to leverage that.

mitchellhamilton:
Yeah, sounds good. Maybe update the comments that say the lowest supported TS version in the declarations to 3.7 and dtslint should stop failing then?

Andarist:
> Yeah, sounds good. Maybe update the comments that say the lowest supported TS version in the declarations to 3.7 and dtslint should stop failing then?

Updated that in all of our packages

Issue #2289, opened 2021-03-16
Types for react element's CSS property seem too permissive


jraoult:
**Current behavior:**

When using Typescript, it seems it is still possible to assign incompatible values to the `css` property (the main one being assigning a class name by mistake)

```tsx
const Sub = (props: { className?: string }) => {
  return (
    <>
      <div
        // this works 👍
        className={props.className}
        css={[
          // 🙅 this should not compile
          props.className,
          css`
            min-height: 150px;
            background-color: lightblue;
          `,
        ]}
      >
        Hello there
      </div>
    </>
  );
};
```

I traced it back to the [JSX namespace declaration](https://github.com/emotion-js/emotion/blob/830dd0e6d071c98bc0b4b0ecc99dd21a93f057b9/packages/react/types/jsx-namespace.d.ts) where it is declared as an `Interpolation`.

I might be missing something but I would think the type for the `css` should rather be `SerializedStyles | Array<SerializedStyles> | ((theme: Theme) => SerializedStyles)`.

I tried to tweak the definition locally and it indeed caught the issues. Was there a rationale to allow `Interpolation`?

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version: 11.1.5
- TS: 4.2.3

Andarist:
I've created a quick PR for this: https://github.com/emotion-js/emotion/pull/2297 but I don't have time to test this properly (nor in any codebase nor by creating new type tests in this repo). If you could help me with this I would highly appreciate it.

jraoult:
@Andarist thx for your reactivity I'll have a look.

PR #2297, opened 2021-03-17
Add stricter ReactInterpolation type

Andarist:
fixes https://github.com/emotion-js/emotion/issues/2289

### DIFF
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -46,6 +46,24 @@ export function withEmotionCache<Props, RefType = any>(
   func: (props: Props, context: EmotionCache, ref: Ref<RefType>) => ReactNode
 ): FC<Props & ClassAttributes<RefType>>
 
+type ReactInterpolationPrimitive =
+  | null
+  | undefined
+  | boolean
+  | SerializedStyles
+  | CSSObject
+
+type ReactArrayInterpolation<Props> = Array<ReactInterpolation<Props>>
+
+interface ReactFunctionInterpolation<Props> {
+  (props: Props): ReactInterpolation<Props>
+}
+
+export type ReactInterpolation<Props> =
+  | ReactInterpolationPrimitive
+  | ReactArrayInterpolation<Props>
+  | ReactFunctionInterpolation<Props>
+
 export function css(
   template: TemplateStringsArray,
   ...args: Array<CSSInterpolation>
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -1,10 +1,9 @@
 import 'react'
-import { Interpolation } from '@emotion/serialize'
-import { Theme } from './index'
+import { Theme, ReactInterpolation } from './index'
 
 type WithConditionalCSSProp<P> = 'className' extends keyof P
   ? string extends P['className' & keyof P]
-    ? { css?: Interpolation<Theme> }
+    ? { css?: ReactInterpolation<Theme> }
     : {}
   : {}
 
@@ -34,7 +33,7 @@ export namespace EmotionJSX {
 
   type IntrinsicElements = {
     [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
-      css?: Interpolation<Theme>
+      css?: ReactInterpolation<Theme>
     }
   }
 }
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -3,7 +3,7 @@
 
 import * as React from 'react'
 import { ComponentSelector, Interpolation } from '@emotion/serialize'
-import { PropsOf, DistributiveOmit, Theme } from '@emotion/react'
+import { PropsOf, ReactInterpolation, Theme } from '@emotion/react'
 
 export {
   ArrayInterpolation,
@@ -69,7 +69,7 @@ export interface CreateStyledComponent<
    */
   <AdditionalProps extends {} = {}>(
     ...styles: Array<
-      Interpolation<
+      ReactInterpolation<
         ComponentProps &
           SpecificComponentProps &
           AdditionalProps & { theme: Theme }


packages/react/types/index.d.ts
```diff
@@ -46,6 +46,24 @@ export function withEmotionCache<Props, RefType = any>(
...
+
+interface ReactFunctionInterpolation<Props> {
+  (props: Props): ReactInterpolation<Props>
+}
+
+export type ReactInterpolation<Props> =
+  | ReactInterpolationPrimitive
+  | ReactArrayInterpolation<Props>
+  | ReactFunctionInterpolation<Props>
+
```
mitchellhamilton, line 66: Can we call the type parameter `Theme` since that's what it is here?

Andarist: Ive also used it for one of the styled overloads where it is equivalent to Props

packages/react/types/index.d.ts
```diff
@@ -46,6 +46,24 @@ export function withEmotionCache<Props, RefType = any>(
   func: (props: Props, context: EmotionCache, ref: Ref<RefType>) => ReactNode
 ): FC<Props & ClassAttributes<RefType>>
 
+type ReactInterpolationPrimitive =
```
jraoult, line 49: I think fundamentally what we mean is  _falsy_ or something that produce valid style information:

```ts
type ReactInterpolationPrimitive =
| Falsy 
| SerializedStyles
| CSSObject
```
Falsy in TS can be expressed like that (not the `""` and `false` instead of boolean):
```ts
type Falsy = false | "" | 0 | null | undefined
```

Any other value would be probably the result of a programming mistake and would not be interpreted correctly by Emotion anyway, so we might as wall catch them:
```ts
css={"my-custom-class"}
css={true}
```

whereas  the following could be the intended result of a boolean shortcircuit conditional:
```ts
css={""}
css={false}
```

Andarist: Hah, interesting that with `someString && somethingElse` TS actually computes the output as `"" | typeof somethingElse`! TIL

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYghgGwM4igXigM0U6AfKAIkKgIAZSoA7AVwQUpqoBMJMBLKiZgKB9YDGCOACdoAgPZUkwLBIkAuKDJGcA5n0nTZcAQIhhg8ZKgwAKbCaXGUlAN5RJCCSKUr1UAL4BKdAD4oO09NKRkoMSR0OQkoADJYwMcJZ1dorx5dfUMbEDMI7xDtcIgkACYozHkoAH5EpxclSpjPKEb5DL0DIxxciNLvIA

jraoult: @Andarist yes this is pretty neat. From the quick test of my code base (a branch where I'm refactoring to use the `css` property), these types already help me a lot in finding mistakes I made.

In your example, for `res2`, I expected TS to be a little smarter with the ternary operator. I can't really understand what other value `foo` can be in the right branch other then `""` 🤷‍♂️..

So this case would generate a false negative, but to be fair I think this is a pattern that should be very rare when using Emotion. I can't really see what one would want to achieve with this where the shortcircuit conditional would be enough.

Andarist: > In your example, for res2, I expected TS to be a little smarter with the ternary operator. I can't really understand what other value foo can be in the right branch other then "" 🤷‍♂️..

Yep, same - both ternary and `&&` should really behave the same way. This was for illustration purposes - not that relevant for this PR really (I don't feel that it's a show-stopper or anything).

packages/react/types/index.d.ts
mitchellhamilton: This type doesn't seem right for styled, e.g. `styled.div('display:flex;')` is valid.

Also, this change isn't right for the css prop either since `css={['display:flex;']}` is valid. The only thing that's wrong about the current types is that `string` isn't allowed at the root of the css prop.

Andarist: Well, sure - the question is if we want to support this. Types don't always have to match the runtime reality fully. It feels better to guide users to better patterns if they are using a type checker.

jraoult: @mitchellhamilton oh ok, I must have missed these style string patterns in the doc because I based my suggestion on what I saw demonstrated. If we really want to allow this form of string we can go ultra geeky and use template literal types but it seems overengineering at this point.

To be honest, the one thing I'm really interested in is for TS to catch me (and anyone starting with the `css` property) when I'm worngly assigning a class name to the `css` property.

So I'd agree with @Andarist here. It's about nudging and guiding to use some patterns rather that fully typing everything that could work but it's up to you folks.

mitchellhamilton: Sure, let's change this for the css prop. Can you revert the change for @emotion/styled though since having @emotion/styled depend on a new thing from @emotion/react would be a breaking change since @emotion/styled has a peer dependency on @emotion/react? (also for styled, i believe we compile styles to `styled.div('display: ', something, ';')` and making it so our types treat something that we compile to as invalid seems bad)

jraoult:
@Andarist did you need help with this? I'm quite keen to see it merged so tell me if I can do anything.

Andarist:
@jraoult adding type tests and checking what happens with failed CI jobs (the ones on the CircleCI) would certainly help

jraoult:
@Andarist I took some time today to work on this. I think I'm ready to push. How should I proceed? Can you give me write access to this branch? Should I create another PR?

Andarist:
U can create a PR targeting this branch

jraoult:
@Andarist you can find the PR here: https://github.com/emotion-js/emotion/pull/2318

packages/react/types/index.d.ts
Andarist: Ok, I've removed this from `@emotion/styled` and also renamed Props to Theme since this will always be true right now.

This might be a user-facing change (some people might use `Interpolation` type in their code) but I think this is OK if we make this a minor change. Tying major versioning to types (unless changes are really extreme) falls outside of the pragmatic semver for me. What's your opinion on this?

Andarist:
@jraoult @mitchellhamilton passing string prop is actually supported in the codebase: 
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/react/src/emotion-element.js#L66-L74

I guess this was mainly to make the Emotion 9 -> Emotion 10 easier but this code has not been removed in Emotion 11. This logic probably kicks in when providing the `cache` from `@emotion/css` to `CacheProvider`.

jraoult:
> passing string prop is actually supported in the codebase:

Exactly, this is actually what got me into this rabbit hole initially 😉 During refactoring, I noticed one could still give the `css` prop a class name generated by `@emotion/css`, and it would work. I then assumed from this finding that `css` would be able to take multiple class names in an array and that I could remove `cx`. It's only later that I realised it was a wrong assumption and that passing a string to `css` wasn't event documented. 

So my initial proposal for this typing was to make sure no undocumented pattern could be used and help people avoid the pitfall.

PR #2308, opened 2021-03-24
Fix grammar in package-summary.mdx

flugg:
**What**:
Fixed double `which` and converted `customise` to `customize` to be consistent with this PR: https://github.com/emotion-js/emotion/pull/1870

**Why**:
Makes it easier to read.

**How**:
By using the keyboard.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/docs/package-summary.mdx
+++ b/docs/package-summary.mdx
@@ -17,7 +17,7 @@ Depends on @emotion/react
 ## @emotion/css
 
 Contains a framework agnostic version of Emotion. Has different functionality and APIs from @emotion/react and doesn’t work the same way under the hood.
-Depends on @emotion/cache which can be customised by creating a custom instance at @emotion/css/create-instance
+Depends on @emotion/cache which can be customized by creating a custom instance at @emotion/css/create-instance
 
 ## @emotion/cache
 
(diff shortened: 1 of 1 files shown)


flugg:
There also seems to be an incomplete sentence in the `@emotion/cache` paragraph:

```
...can create another instance of @emotion/css and it accepts
Depended upon by @emotion/react and @emotion/css
```

Not sure what it accepts. If it's wanted, feel free to throw me a full sentence and I can add it in this PR.

docs/package-summary.mdx
```diff
@@ -17,11 +17,11 @@ Depends on @emotion/react
 ## @emotion/css
 
 Contains a framework agnostic version of Emotion. Has different functionality and APIs from @emotion/react and doesn’t work the same way under the hood.
-Depends on @emotion/cache which which can be customised by creating a custom instance at @emotion/css/create-instance
+Depends on @emotion/cache which can be customized by creating a custom instance at @emotion/css/create-instance
```
Andarist, line 20: both "customized" and "customised" are correct - the former is used by American English and the latter is British. 

@mitchellhamilton do you have any preference over this? to which flavor Australian is closer?

mitchellhamilton: I don't really have a strong preference, American is probably the better one to use to be consistent with what APIs generally use (e.g. `color` in CSS).

PR #2318, opened 2021-03-30
Stricter react interpolation type [external changes]

jraoult:
See #2297 

@Andarist I just realised that this other PR https://github.com/emotion-js/emotion/pull/2296#issuecomment-801705893 will have an impact on my proposed changes since I had to fix an issue with recursive type alias.

### DIFF
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -46,14 +46,12 @@ export function withEmotionCache<Props, RefType = any>(
   func: (props: Props, context: EmotionCache, ref: Ref<RefType>) => ReactNode
 ): FC<Props & ClassAttributes<RefType>>
 
-type ReactInterpolationPrimitive =
-  | null
-  | undefined
-  | boolean
-  | SerializedStyles
-  | CSSObject
+type Falsy = false | '' | 0 | null | undefined
+type ReactInterpolationPrimitive = Falsy | SerializedStyles | CSSObject
 
-type ReactArrayInterpolation<Props> = Array<ReactInterpolation<Props>>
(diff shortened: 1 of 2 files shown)


jraoult:
Hey @Andarist any thoughts on this one?

packages/react/types/index.d.ts
```diff
@@ -46,14 +46,12 @@ export function withEmotionCache<Props, RefType = any>(
...
-  | boolean
-  | SerializedStyles
-  | CSSObject
+type Falsy = false | '' | 0 | null | undefined
+type ReactInterpolationPrimitive = Falsy | SerializedStyles | CSSObject
 
-type ReactArrayInterpolation<Props> = Array<ReactInterpolation<Props>>
+type ReactArrayInterpolation<Props> = Array<
+  ReactInterpolationPrimitive | ReactFunctionInterpolation<Props>
+>
```
Andarist, line 54: Any reason why this has been split? An additional problem this has that this doesn't allow nested arrays now.

jraoult: Yes so it was creating a cycle with type alias so I had to _break_ the cycle to allow the tests to pass.

But I think as you noticed here: https://github.com/emotion-js/emotion/pull/2296#issuecomment-801705893 will allow avoiding these changes since TS 3.7 brought a change that allows this pattern.

jraoult:
@Andarist is there anything I can do to help to get this over the line 😀 ?

jraoult:
Hey @Andarist I realized I forgot to bump that one. How can I help to get this through?

PR #2333, opened 2021-04-06
chore(styled): let StyledOptions generic argument be optional

antongolub:
We're now migrating from v10 and noticed a strange thing. `StyledOptions` now requires `Props` argument, which does not affect at all. What do you think about either removing the param or at least making it optional?

```ts
declare type PropertyKey = string | number | symbol;
...
export interface StyledOptions<Props> {
  label?: string
  shouldForwardProp?(propName: PropertyKey): boolean
  target?: string
}
```


Andarist:
Please prepare a runnable repro case of your problem or add a type test (in the types directory) that would showcase why this is needed.

antongolub:
Hey, @Andarist,

This is not a _problem_. Once we've bumped emotion version [in our project](https://github.com/qiwi/pijma/pull/405), we just found API diff, which does not add any functionality.
#1643 brought this relaxed type earlier, so now generic seems useless here.

Andarist:
Hm, it should still be possible to add a test type for this - since some code did not work for you and you expect it to work.

antongolub:
Sounds reasonable. I'll do.

packages/native/types/base.d.ts
```diff
@@ -64,7 +64,7 @@ export interface FilteringStyledOptions<
   shouldForwardProp?(propName: PropertyKey): propName is ForwardedProps
 }
 
-export interface StyledOptions<Props> {
+export interface StyledOptions<Props = Record<PropertyKey, any>> {
```
Andarist, line 67: well, this generic stays unused so I think it should be more than OK to just drop it

antongolub: Yep. I suggest using this impl cause it's compatible with both v10 and v11-based app code. Removing generic will be a breaking change for someone.

Andarist: Any change is a breaking change for someone 😅 I think in this case it's a pragmatic choice to not treat it as one - especially give that I don't expect `StyledOptions` to be used explicitly in a lot of codebases.

antongolub: Maybe it's better to use generic argument if passed?

```ts
export interface StyledOptions<Props = null> {
  label?: string
  shouldForwardProp?(propName: Props extends null ? PropertyKey : keyof Props): boolean
  target?: string
}
```

Andarist: Quite honestly, I'm not sure if it's worth it. This type should not be used that much publicly and we'd have to check if this would be compatible when passed to `styled` calls. It would also be good to dig up the PRs by @JakeGinnivan which were related to this and make sure that this would not break because of the reasons mentioned there. Ideally, we would pass `Props` to `StyledOptions` from within `CreateStyledComponent` types.

If you are willing to do the work - then sure, ye, we could try to do that.

antongolub: I have a case where it makes sense. We'd like to reuse some options and corresponding types for our UI kit primitives and rich components.

https://github.com/qiwi/pijma/blob/d884ba8b954b59f190623caf1623c9489c11b550/packages/core/src/primitive/Card.ts#L23
```ts
export interface CardProps extends BoxProps {
  bg?: string
  s?: string
  b?: string
  bt?: string
  br?: string
  bb?: string
  bl?: string
  r?: Value
  rtr?: Value
  rtl?: Value
  rbr?: Value
  rbl?: Value
}

export const CardNonProps = ['bg', 'b', 'bt', 'br', 'bb', 'bl', 'r', 'rtr', 'rtl', 'btr', 'btl', 's'].concat(BoxNonProps)

export const CardOptions: StyledOptions<CardProps> = {
  // expect any prop key strictly belong to CardProps only
  shouldForwardProp: (prop) => !CardNonProps.includes(prop),
}
```

Andarist: The question is - can the resulting type be used when using `styled`?

```ts
type Props = { foo: boolean }

declare const options: StyledOptions<Props>

const test1 = styled.div<Props>({
  color: 'red',
}, options)

const test2 = styled.div({
  color: 'red',
}, options)
```

antongolub: Everything should be ok. There's a minor type issue, which is not related to this PR.
```
TS2769: No overload matches this call.   Overload 1 of 3, '(...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { ...; }>[]): StyledComponent<...>', gave the following error.     Argument of type 'StyledOptions<Props>' is not assignable to parameter of type 'Interpolation<{ theme?: Theme | undefined; as?: ElementType<any> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { ...; }>'.       Type 'StyledOptions<Props>' is not assignable to type 'CSSObject'.   Overload 2 of 3, '(template: TemplateStringsArray, ...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>[]): StyledComponent<...>', gave the following error.     Argument of type '{ color: string; }' is not assignable to parameter of type 'TemplateStringsArray'.       Object literal may only specify known properties, and 'color' does not exist in type 'TemplateStringsArray'.   Overload 3 of 3, '(template: TemplateStringsArray, ...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>[]): StyledComponent<...>', gave the following error.     Argument of type '{ color: string; }' is not assignable to parameter of type 'TemplateStringsArray'.       Object literal may only specify known properties, and 'color' does not exist in type 'TemplateStringsArray'.
```

Andarist: Well, if this is supposed to land then I strongly believe that those 2 cases should not result in any type errors - even if the problem is not related to this PR it IMHO would be preferable to fix it here or prepare a separate fix, land it and then rebase this PR. Without those working it's hard for me to justify merging this PR with that generic being used.

antongolub: Give me a bit more time. I'll fix the issue above.

Andarist: No problem - there is no rush. I appreciate your efforts here :)

antongolub: @Andarist, 

Not sure if I found the best solution, but it seems to work.

antongolub: @Andarist,
Could you take a look?

packages/native/types/base.d.ts
```diff
@@ -103,11 +105,17 @@ export interface CreateStyledComponent<
    * @typeparam AdditionalProps  Additional props to add to your styled component
    */
   <AdditionalProps extends {} = {}>(
-    ...styles: ArrayInterpolation<
-      ComponentProps &
-        SpecificComponentProps &
-        AdditionalProps & { theme: Theme },
-      StyleType
+    ...styles: Array<
+      | StyledOptions<
```
Andarist, line 109: `StyledOptions` shouldn't be allowed here - to the very best of my knowledge. `StyledOptions` can only be provided before styles are accepted (options are passed to the outer~ call):
```js
styled(tag, styledOptions)(...styles)
```

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,40 @@ const Input5 = styled.input`
...
+    foo: boolean
+  }
+  const options: StyledOptions<JSX.IntrinsicElements['div']> = {
+    label: 'foo',
+    target: 'bar',
+    shouldForwardProp: p => true
+  }
+  const test1 = styled.div({ color: 'red' }, options)
+  const test2 = styled.div<Props>({ color: 'red' }, options)
+  const test3 = styled('div', options)({ color: 'red' })
```
Andarist, line 255: Could you also add a similar test but for a composite component (non-HTML)?

packages/native/types/base.d.ts
antongolub: Ok. Reverted.

packages/styled/types/tests.tsx
antongolub: Sure.

packages/styled/types/base.d.ts
```diff
@@ -69,11 +75,17 @@ export interface CreateStyledComponent<
    */
   <AdditionalProps extends {} = {}>(
     ...styles: Array<
-      Interpolation<
-        ComponentProps &
-          SpecificComponentProps &
-          AdditionalProps & { theme: Theme }
-      >
+      // tslint:disable-next-line no-null-undefined-union
+      | StyledOptions<
```
Andarist, line 79: the same thing should be reverted from here as well

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,43 @@ const Input5 = styled.input`
...
+
+  const styledOpts0: StyledOptions = {
+    label: 'foo',
+    target: 'bar',
+    shouldForwardProp: p => true
+  }
+  const styledOpts1: StyledOptsParameterized = {
+    label: 'foo',
+    target: 'bar',
+    shouldForwardProp: (p: 'a' | string) => true
```
Andarist, line 232: what was the intention here? `'a' | string` is an equivalent of just `string`

packages/styled/types/base.d.ts
antongolub: Maybe, but it affects the case [you mentioned above](https://github.com/emotion-js/emotion/pull/2333#discussion_r609606867). 
```js
const test2 = styled.div({
  color: 'red',
}, options) // ← This point
```
Now I think it was just a typo because `styled.div(...args)`  does not allow to pass `StyledOptions` as arg:

```js
export type CreateStyledComponent = <Props>(
  ...args: Interpolations
) => StyledComponent<Props>

export type CreateStyled = {
  <Props>(
    tag: StyledElementType<Props>,
    options?: StyledOptions
  ): (...args: Interpolations) => StyledComponent<Props>,
  [key: string]: CreateStyledComponent,
  bind: () => CreateStyled
}
```
So I'm going to revert related changes and remove the corresponding test.

packages/styled/types/tests.tsx
antongolub: I tried to verify that `p` value type is properly extracted as a string key union. I'll modify it to make it a bit clearer.

antongolub:
@Andarist, any suggestions/objections?

antongolub:
@Andarist, let’s try to move this forward.

Andarist:
Sorry for the delay but I was lately very busy. Please be patient - I'll get to reviewing this PR when I get some time for it.

antongolub:
@Andarist, @mitchellhamilton,

Do you need any help with this? Maybe I should change or add something to make it easier for review?

antongolub:
@Andarist , is any chance to land this?

antongolub:
ping

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,45 @@ const Input5 = styled.input`
...
+    target: 'bar',
+    shouldForwardProp: p => true
+  }
+
+  const test1 = styled('div', options)({ color: 'red' })
+  const test2 = styled('div', options)<Props>({ color: 'red' })
+  const test3 = styled(() => <div className="foo" />, options)({ color: 'red' })
+  const test4 = styled(() => <div className="foo" />, options)<Props>({
+    color: 'red'
+  })
```
Andarist, line 260: Those are basically the most interesting test cases that have been added as part of this PR - what I mean by that is that this is sort of the expected "integration point" and we need to make sure that this integration point works OK. 

Those added tests are somewhat broad - not allowing me to quickly see how you are going to utilize the introduced changes. Unless on of those 4 tests did not work before your changes? Could you add something more resembling your use case? Writing down in a comment (backed with an example) what has not been possible before this PR and what will become possible with the changes in this PR would make it far easier to review this.

Sorry for the delay - I'll make sure to iterate over this quickly now cause you have already been waiting for so long. I need to understand the full context behind this PR though and I somewhat feel like I'm missing it (it might be here, among all of the comments etc but I don't have the energy to build it up from those older comments, sorry for that - a short summary would be highly appreciated).

Andarist: One main problem that I'm having with this PR is that it has been somewhat repurposed (which is not a bad thing in itself, it just makes it slightly harder for me to navigate it and to build that mental context for it). It has started as allowing the generic to be optional but it ended up adding a new feature~ to it (the `propName` can not be constrained to specific keys)

antongolub: The context: we're building sub-primitive factories to generate a deeply customizable set of components. Not just theming. These sets also differ in internals, their behaviour, and their composition approach. It's something like SDK for UI Kits.

The only purpose of this PR is to achieve strict consistency between `shouldForwardProp` filter and component `Props` where it's needed. We want to control props forwarding and we try to ensure that every primitive combination is correct by the type checking.

The current implementation allows any filter to handle any props. Moreover, this feature was available in earlier versions of `emotion` https://github.com/antongolub/emotion/commit/ad77ed24b4bfe62d6c80f0498cd7e76863e2f28e#diff-3fae48ecece75c5d911928f5c96c7292ac893baae5232e9868e68446de341401R34

image

antongolub: I understand, there would be little chance to bring such functionality as new. But I'm just trying to get back what was lost ))

Andarist: > The current implementation allows any filter to handle any props.

This I can see, somewhat. But a test case showcasing how something was not an error before and how it is an error now would be helpful in assessing how this change works now.

> I understand, there would be little chance to bring such functionality as new.

I'm not opposed to bringing any functionality as new - I just want to understand as much of the context behind the change as possible 😉 

> The current implementation allows any filter to handle any props. Moreover, this feature was available in earlier versions of emotion antongolub@ad77ed2#diff-3fae48ecece75c5d911928f5c96c7292ac893baae5232e9868e68446de341401R34

This kinda points to `FilteringStyledOptions` that is still available as part of the Emotion types - so I'm not exactly sure how this was available before but it is not now.

antongolub: wrong line, my fault:

https://github.com/antongolub/emotion/commit/ad77ed24b4bfe62d6c80f0498cd7e76863e2f28e#diff-3fae48ecece75c5d911928f5c96c7292ac893baae5232e9868e68446de341401R40

image

Andarist: Just a note - I would still appreciate compelling test cases to be added to this PR before proceeding further.

antongolub: Hmm... But this is exactly how we're using this right now. We take some `props`, and some `propsForwarder` filter, wrap them into `StyledOptions<typeof props>`, and use this type to describe `styledOptions` exportable constant.
```js
const propsForwarder = (p: 'foo') => true
const styledOptsParameterized: StyledOptions<{ foo: string }> = {
    shouldForwardProp: propsForwarder
  }

  const styledOptsParameterizedBroken: StyledOptions<{ foo: string }> = {
    shouldForwardProp: (p: 'bar') => true // $ExpectError
  }
```

antongolub: Done if I understood suggestions correctly) About `FilteringStyledOptions` — I'm not sure that it should be changed here. This pull request definitely doesn't break anything and can be safely merged.

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,45 @@ const Input5 = styled.input`
...
+
+  // $ExpectError
+  styled('div', { shouldForwardProp: (prop: 'foo') => true })({})
+
+  styled<keyof JSX.IntrinsicElements>('div', {
+    shouldForwardProp: (prop: 'color') => true
+  })({})
+
+  // $ExpectError
+  styled<keyof JSX.IntrinsicElements>('div', { shouldForwardProp: (prop: 'foo') => true })({}) // prettier-ignore
```
Andarist, line 260: Similar note as here: https://github.com/emotion-js/emotion/pull/2333/files#r784689683
```suggestion
```

packages/styled/types/tests.tsx
Andarist: I think we should attempt to apply the same change to `FilteringStyledOptions` (with the appropriate tests). If this will prove to be harder than expected - then this can be put aside and we'll just follow with this PR. It's just that I find it weird that those types don't align with each other and I bet that I will be baffled about this in the future when looking through code.

You could also try to check the inference results with `$ExpectType` as suggested in point 1.

Andarist:
Could you also add a changeset for this change?

I believe that we are close to landing this now :p

antongolub:
Ok, I've tweaked a bit `FilteringStyledOptions`. Have a look.

antongolub:
@Andarist, your turn

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,69 @@ const Input5 = styled.input`
   // $ExpectError
   ;<StyledCompWithoutAs as={Section} />
 }
+
+{
+  // Props forwarding through StyledOptions and FilteringStyledOptions
+
+  const fc: React.FC<{ foo: string }> = props => <div {...props} />
+
+  styled(fc, { shouldForwardProp: (prop: 'foo') => true })({})
```
Andarist, line 226: this is somewhat interesting - `shouldForwardProp` here gets typed through this:
```
StyledOptions<React.PropsWithChildren<{ foo: string; }>>.shouldForwardProp?(propName: "foo" | "children"): boolean
```
So I'm unsure why does this typecheck if this should be an error:
[TS playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygAooAueAckRx3PgB8KwALLCYOVcgSjICNqIIKKgBQI0JFgIU6bHni9iZStW58BQ0ePDQ48MHgDOGfWQwBPAA4gciBWIOpj8YGas27UeAF59QA)

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,69 @@ const Input5 = styled.input`
...
+
+  styled(fc, { shouldForwardProp: (prop: string) => true })({})
+
+  // $ExpectError
+  styled(fc, { shouldForwardProp: (prop: 'bar') => true })({})
+
+  const shouldForwardProp1 = (prop: 'foo') => true
+  styled(fc, { shouldForwardProp: shouldForwardProp1 })({})
```
Andarist, line 234: The same doubt about this one as [here](https://github.com/emotion-js/emotion/pull/2333/files#r795040383)

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,69 @@ const Input5 = styled.input`
...
+
+  // $ExpectError
+  const shouldForwardProp8: FilteringStyledOptions<{
+    foo: string
+  }>['shouldForwardProp'] = (prop: 'unknown'): prop is 'unknown' => true
+
+  const shouldForwardProp9: FilteringStyledOptions<
+    { foo: string; bar: string },
+    'foo'
+  >['shouldForwardProp'] = (prop: 'foo'): prop is 'foo' => true
```
Andarist, line 269: Perhaps interestingly this is types as:
```
const shouldForwardProp9: ((propName: string | number | symbol) => propName is "foo") | undefined
```

Could we add comment above `FilteringStyledOptions` and `StyledOptions` mentioning what are their responsibilities? I am somewhat confused about that. I understand that for filtering variant we accept a type predicate whereas for the other one we accept a simpler function that returns a boolean **but** that is not the only difference between them and I feel like this should be clarified.

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,69 @@ const Input5 = styled.input`
...
+
+  // $ExpectError
+  const shouldForwardProp10: FilteringStyledOptions<
+    { foo: string; bar: string },
+    'foo'
+  >['shouldForwardProp'] = (prop: 'bar'): prop is 'bar' => true
```
Andarist, line 275: Should something like this be allowed?
```js
  const shouldForwardProp10: FilteringStyledOptions<
    { foo: string; bar: string },
    'foo'
  >['shouldForwardProp'] = (prop: 'foo' | 'bar'): prop is 'bar' => true
```

Andarist:
@antongolub your turn 😜

packages/styled/types/tests.tsx
antongolub: Hmm... I think It depends on the direction of comparison. [Type swap fixes the example](https://www.typescriptlang.org/play?ssl=5&ssc=22&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygAooAueAckRx3PgB8KwALLCYOVcgSjICNqIIKKgBQoSLAQp02PPF7Eylatz4ChoseGhx4YPAGcMeshgCeABxA5EBEftRH4wU5eu3e8ALx6gA)

```ts
declare function a(a: 'foo' | 'children'): boolean
declare function b(a: 'foo'): boolean

declare const c: typeof a
const d: typeof b = c
```

packages/styled/types/tests.tsx
antongolub: Seems compatible. The test passes.

packages/styled/types/tests.tsx
antongolub: Smth like this?
```ts
/**
 * Same as StyledOptions but shouldForwardProp must be a type guard (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).
 * Practical meaning: you can declare and reuse atomic `shouldForwardProp` filters that are strictly bound with some `ForwardedProps` type.
 */
```

packages/styled/types/tests.tsx
Andarist: So this comes from the fact that `shouldForwardProp` is currently declared using "object method" syntax, and not using the property syntax (where the value turns out to be a function). This makes the function bivariant instead of contravariant (I've expected the latter here).

Take a look at this example that might showcase the difference:
[TS playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwGcAHcLRLMAMSwgxBgAoBrEATwC54AiRHHLgJScARnwggoqANwAoGaEiwEKdNjzwoYMCCIYAwngwwoAN1hZJGBjl0FOAbxnwkNOjE7M2nAkayoA5gLwALwAfPCiOOKSsgC+QvAmOFjAsjKa2roGqEam5pYMjs7ktPTeJGBkFNSlMAA0MvFpCtBwSGiYuPgZOhgAQlhmMBY51rYOLrWeHIS+AQmR0fjxnEkpaT26A0MjVkWTbuWk5FSu9A1NMkA)

I feel like usually bivariant functions are not the ones that we want so it's a good time to fix that old mistake with this patch:
```diff
diff --git a/packages/styled/types/base.d.ts b/packages/styled/types/base.d.ts
index 4c972110..83949462 100644
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -23,15 +23,15 @@ export interface FilteringStyledOptions<
   ForwardedProps extends keyof Props = keyof Props
 > {
   label?: string
-  shouldForwardProp?(propName: PropertyKey): propName is ForwardedProps
+  shouldForwardProp?: (propName: PropertyKey) => propName is ForwardedProps
   target?: string
 }
 
 export interface StyledOptions<Props = null> {
   label?: string
-  shouldForwardProp?(
+  shouldForwardProp?: (
     propName: Props extends null ? PropertyKey : keyof Props
-  ): boolean
+  ) => boolean
   target?: string
 }
```

Note that this might require some tests to be adjusted

packages/styled/types/tests.tsx
Andarist: ye, i guess that should suffice

packages/styled/types/tests.tsx
Andarist: Hm, even without the proposed patch that would fix the bivariance problem I see this as being an error:
```
Type '(prop: "foo" | "bar") => prop is "bar"' is not assignable to type '(propName: string | number | symbol) => propName is "foo"'.
  Type predicate 'prop is "bar"' is not assignable to 'propName is "foo"'.
    Type '"bar"' is not assignable to type '"foo"'.ts(2322)
```

But perhaps my true question here (I actually don't remember why **exactly** I've asked about this) was about:
```ts
  const shouldForwardProp10: FilteringStyledOptions<
    { foo: string; bar: string },
    'foo'
  >['shouldForwardProp'] = (prop: 'foo' | 'bar'): prop is 'foo' => true
```
and this indeed is OK with the bivariant `shouldForwardProp` but starts erroring if we change this to a regular contravariant function.

Andarist:
So it seems like the only thing to settle on is the variance type for the `shouldForwardProp`. Let me know what do you think about this.

packages/styled/types/tests.tsx
antongolub: My fault. I've tested a bit different case.

```
  const shouldForwardProp9: FilteringStyledOptions<
    { foo: string; bar: string },
    'foo'
  >['shouldForwardProp'] = (prop: 'foo' | 'bar'): prop is 'foo' => true

  // $ExpectError
  const shouldForwardProp10: FilteringStyledOptions<
    { foo: string; bar: string },
    'foo'
  >['shouldForwardProp'] = (prop: 'foo' | 'bar'): prop is 'bar' => true
```

packages/styled/types/base.d.ts
```diff
@@ -13,19 +17,24 @@ export {
...
 export { ComponentSelector, Interpolation }
 
-/** Same as StyledOptions but shouldForwardProp must be a type guard */
+/**
+ * Same as StyledOptions but shouldForwardProp must be a type guard (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).
+ * Practical sense: you can define and reuse atomic `shouldForwardProp` filters that are strictly bound with some `ForwardedProps` type.
+ */
 export interface FilteringStyledOptions<
-  Props,
+  Props = Record<PropertyKey, any>,
```
Andarist, line 25: q: why the `Props` have a different default between `FilteringStyledOptions` and `StyledOptions`?

Andarist:
> So it seems like the only thing to settle on is the variance type for the shouldForwardProp. Let me know what do you think about this.

👋

Andarist:
FYI just 2 small questions to be answered here and we should be able to land this soon. No rush, of course 😅

packages/styled/types/base.d.ts
antongolub: > Props = Record<PropertyKey, any>,

This is just an artifact of the first impl. I'll fix.

> So it seems like the only thing to settle on is the variance type for the `shouldForwardProp`.
 
It _seems_ that `strictFunctionTypes` option is not applied to object `method?: () {}` signature. So if we rewrite the code as suggested above we break everything.

```ts
export interface FilteringStyledOptions<
  Props = AnyRecord,
  ForwardedProps extends keyof Props = keyof Props
> {
  label?: string
  shouldForwardProp?: <P extends UnionToIntersection<ForwardedProps>>(propName: P) => propName is P
  target?: string
}

export interface StyledOptions<Props = AnyRecord> {
  label?: string
  shouldForwardProp?: <P extends keyof UnionToIntersection<Props>>(propName: P) => boolean
  target?: string
}

type AnyRecord = Record<PropertyKey, any>

export type UnionToIntersection<U, K = any> = (
  U extends K ? (k: U) => void : never
  ) extends (k: infer I) => void
  ? I
  : never
```

The main problem: I don't know how to infer param type inside generic, so I cannot add a type assertion that impl fn param extends HTML element attributes.
```
type F = <P extends any extends infer P ? P : never>(p: P) => boolean
// P is unknown
const f: F = (prop: 'style') => true
```
I also tried to handle this in another way: to represent `shouldForwardProp` as a union of fns with all possible signatures. I looks syntactically correct and automatable with recursive mapping, but it requires a quantum computer.

```ts
type T = 'a' | 'b' | 'c'
type F = (p: 'a') => true | (p: 'b') => true | (p: 'c') => true 
| (p: 'a' | 'b') => true | (p: 'b' | 'c') => true | (p: 'a' | 'c') => true 
| (p: 'a' | 'b' | 'c') => 
```

For ~300 attributed for every HTML element type we get: `C(n,k) = n! / k!(n-k)!, k = 1, n=300, C =~3E615`.
https://www.numberempire.com/factorialcalculator.php?number=300

PS Sorry for the late reply.

Andarist: > It seems that strictFunctionTypes option is not applied to object method?: () {} signature

If I understand this correctly then with `strictFunctionTypes: false` every function is bivariant but with `strictFunctionTypes: true` **methods** are still bivariant but every other function becomes contravariant.

I think you might have a typo in this example as it should be `method?(): void` (or similar) to fit the rest of the sentence. I've just wanted to clarify that we have a common understanding of this stuff (where I'm not 100% sure if my understanding is fully correct).

> So if we rewrite the code as suggested above we break everything.

Does "as suggested" mean "change to a function type and thus make it contravariant"? Could you give me an example of what this would break? If it breaks something important then I think we should have a type test for this.

>   shouldForwardProp?: <P extends UnionToIntersection<ForwardedProps>>(propName: P) => propName is P

I'm unsure what's happening here. I understand what `UnionToIntersection` does but I don't see why it was applied here. The result of this type is an intersection and `P` definitely should be constrained to a union of strings (or to just a string type). Could you explain what was the goal here?

antongolub: I've pushed several snapshot branches so you can observe my poor efforts:
* [**snap#3**](https://github.com/antongolub/emotion/tree/snap2) All gneric singatures use `Record<PropertyKey, any>` as default type. Filters types exposes as object fields. The onl[ working.
* [**snap#2**](https://github.com/antongolub/emotion/tree/snap2) `shouldForwardProp` type declared as a generic function + `UnionToIntersecition`
* [**snap#1**](https://github.com/antongolub/emotion/tree/snap1) same as 2 but w/o  `UnionToIntersecition`.

`UnionToIntersection` was placed here to ensure that keyof type includes all the keys of all types of `Props` union:
`ClassAttributes<HTMLDivElement> | HTMLAttributes<HTMLDivElement> → keyof ClassAttributes<HTMLDivElement> | keyof HTMLAttributes<HTMLDivElement>`. Just for experiment.

antongolub: It seems that we ran into some limitation (in understanding?) of the current TS. What I want to get:
```ts
// Declare filter signature that operates the keys of set 'a' | 'b' ... 'z'
type F {
  filter: (p: 'a' | 'b' ... 'z') => bool
}

// Introduce a specific filter, that uses a subset p type
const f: F = {
    filter: (p: 'a' | 'c') => bool
}
// 'a' | 'c' extends 'a' | 'c' ... 'z'
```

Andarist: > I've pushed several snapshot branches so you can observe my poor efforts:

Thank you for those - I'll try to review them this week to get a better feeling of what we are dealing with here.

> UnionToIntersection was placed here to ensure that keyof type includes all the keys of all types of Props union:

Do we need to care about `ClassAttributes` though? This interface only contains `key` and `ref` properties and those can't really b received by `shouldForwardProp` as those are special props in React and it won't ever handle them to us.

> It seems that we ran into some limitation (in understanding?) of the current TS. What I want to get:

Ye, this is impossible - mainly because now you could make such a call `f.filter('z')` and it could potentially crash at runtime as the supplied implementation is not capable of handling `'z'`.

So basically, here, assigning to this type is not valid. Such things could be expressed with some function helpers and stuff, like here:
[TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYlC8UDeAoKUBmBLANsCATgFxQAUYJA5AIaVQA+UlARnY5QMZtMAmlAlAgB8UZgHsxOCNQB2KAL4oUoSFACCMkDACuMjsCxiZCMgDpz1AgHMAziVkgA2gF1B8EQ6UroANWo4sHmp8AB4AFSgIAA98GR4bZExcfGJ1TR09AyMoeRFEGEcAImw8QkLnSJiIOISwopKU8qgAfigIkhkIADdCJR4IDhxLaAxdfUNjLv9A4IgYZMJwytj4xIbCEg0tMazjXNJLKxI-AKDQsKF+EgilDiMbYEwTKdPZ+dKCUlR0dHXU8iotAYTFYbhEwAI2ggCn4Sgwpj+pBoAhQ8MRLBRAHpMVAAALAGwAWmikH0xIIBDEBFRCIWn04WJx+KJJIGwHJlOpaLpSL4sJQdxkD0wACZntMznMed8fti8QTiVFSezCJy0D8-iQAUwgewIAJhFAIVCYSggA)

The gist of this is that we **preserve** the type of the implementation function but we check if it's valid according to our rules. However, I'm not sure when this would be needed/helpful in our case - probably a full example involving those filters and a `styled` call would help me to see this (maybe it's already in those branches of yours, I didn't yet have a chance to look there though)

packages/native/types/tsconfig.json
```diff
@@ -6,6 +6,10 @@
     "lib": ["es6", "dom"],
     "module": "commonjs",
     "noEmit": true,
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
```
Andarist, line 12: all of those should already be enabled by the `strict: true` that we already have here, was this change required for anything?
```suggestion
```

packages/styled/types/tsconfig.json
```diff
@@ -7,6 +7,10 @@
     "lib": ["es6", "dom"],
     "module": "commonjs",
     "noEmit": true,
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
```
Andarist, line 13:
```suggestion
```

packages/styled/types/tests.tsx
srmagura: It does seem like this line of code 

```ts
styled(fc, { shouldForwardProp: (prop: 'foo') => true })({})
```

should error. You should not be allowed to pass in a `shouldForwardProp` that does not handle all props of the component.

It sounds like @Andarist's patch will fix this.

packages/styled/types/base.d.ts
```diff
@@ -13,19 +17,22 @@ export {
 
 export { ComponentSelector, Interpolation }
 
-/** Same as StyledOptions but shouldForwardProp must be a type guard */
+/**
+ * Same as StyledOptions but shouldForwardProp must be a type guard (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).
+ * Practical sense: you can define and reuse atomic `shouldForwardProp` filters that are strictly bound with some `ForwardedProps` type.
```
srmagura, line 22: I'm not sure what this line is saying... maybe rewording it or including an example would be helpful.

packages/styled/types/base.d.ts
```diff
@@ -13,19 +17,22 @@ export {
 
 export { ComponentSelector, Interpolation }
 
-/** Same as StyledOptions but shouldForwardProp must be a type guard */
+/**
+ * Same as StyledOptions but shouldForwardProp must be a type guard (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).
```
srmagura, line 21: I don't think this link to the TS handbook is necessary, and it will eventually become outdated if they release a new version of the handbook.

packages/styled/types/base.d.ts
```diff
@@ -2,7 +2,11 @@
 // TypeScript Version: 3.2
 
 import * as React from 'react'
-import { ComponentSelector, Interpolation } from '@emotion/serialize'
+import {
```
srmagura, line 5: Can revert this change

Andarist:
I will wait for the CI results and I still need to add some changesets here but other than that this is almost ready to land.

If you have any objections to the final shape of this PR - this is the time to raise them. The situation here is quite convoluted and also the ultra-wide type (`string`) for the `shouldForwardProp`'s parameter is not ideal but given how our types are structured I feel like it's the only way to match type-safety expectations.

srmagura (commented):
It looks good to me, though I am not super confident since this is a complex change. My only comments were about cleaning up some of the comments in the code.

packages/styled/types/tests.tsx
```diff
@@ -217,3 +217,87 @@ const Input5 = styled.input`
...
   ;<StyledCompWithoutAs as={Section} />
 }
+
+{
+  // Props forwarding through StyledOptions and FilteringStyledOptions
+
+  const fc: React.FC<{ foo: string }> = () => null
+
+  // we can't accept a "known" prop here because we need to include `AdditionalProps` and those aren't available yet
+  // `Props` represent the actual props of a component while `AdditionalProps` represent props used only for styling purposes
```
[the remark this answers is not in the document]
Andarist: `AdditionalProps` are declared here:
https://github.com/emotion-js/emotion/blob/65f444405562f7e46e548286514ccd330327b1c7/packages/styled/types/base.d.ts#L72

The problem is this:
```ts
styled<GenericsThatDontHaveAccessToAdditionalProps>('div', {
  shouldForwardProp: (prop: ThisHasToBeAbleToAcceptAdditionalProps) => true
})<AdditionalProps>({
  color: (props) => props.myAwesomeAdditionalColor
})
```

As we might see here - `AdditionalProps` are provided to what is being returned from `styled` but yet `shouldForwardProp` has only access to what is given to `styled`.

Issue #2158, opened 2020-12-04
Dynamic global styles with SSR issue


zce:
**Current behavior:**

style tags generated by `<Global />` components appear in front of style tags generated by SSR: 

![image](https://user-images.githubusercontent.com/6166576/101144413-40a9fe00-3653-11eb-97e8-6cc94291fb3c.png)

As a result, the dynamically calculated global styles cannot override the styles in the SSR:

![image](https://user-images.githubusercontent.com/6166576/101144619-8e266b00-3653-11eb-8111-b0cf32c324dc.png)

**To reproduce:**

https://codesandbox.io/s/snowy-violet-tq704

**Expected behavior:**

The style generated by SSR should appear first:

![image](https://user-images.githubusercontent.com/6166576/101144787-cf1e7f80-3653-11eb-9726-9bed1dbe3b8c.png)

**Environment information:**

- `react` version: 17
- `@emotion/react` version: 11

marcel0ll:
I am having similar problems while using gatsby + emotion

Andarist:
If you would not use the `extractCritical` API and just let the default approach to do its work then it would work correctly as you expect it to do so. That being said the reported situation is not ideal and I'm thinking about how it can be fixed. It's also highly related to the problem reported here: https://github.com/emotion-js/emotion/issues/2049

> I am having similar problems while using gatsby + emotion. Is there a way to control de order that emotion style tags are injected into the HTML?

There is only `prepend` option for caches. You could work around this issue by implementing an alternative version of `extractCritical`. The basic requirement to make this work is to split regular rules and global ones in the rendered SSR output.

zce:
At present, my solution is to exclude `<Global />` from the `CacheProvider`. 

```javascript
export const wrapRootElement = ({ element }) => (
  <ThemeProvider theme={theme}>
    <CacheProvider value={global}>
      <Global styles={styles} />
    </CacheProvider>
    <CacheProvider value={cache}>{element}</CacheProvider>
  </ThemeProvider>
)
```

`<Global />` will generate a style tag during SSR, and it will move to the front of the first emotion style tag when running

oliviertassinari:
I was about to open the same issue with this reproduction: https://github.com/oliviertassinari/mui-repro-global-styles. The bug was surfaced in https://github.com/mui-org/material-ui/issues/24440.

Andarist:
[...]

Issue #2049, opened 2020-10-21
Global component + advanced SSR doesn't hydrate as expected


jdb8:
**Current behavior:**

When using the "advanced approach" (https://emotion.sh/docs/ssr#advanced-approach) to SSR in combination with the `<Global>` component, hydration doesn't appear to be working as expected.

All non-global styles seem to be rehydrated properly, but styles included via `<Global>` are injected into to the page when the client bundle hydrates instead of rehydrating the existing server-provided global styles.

This leads to issues due to specificity where the global rules now have a chance to override non-globals, causing visual bugs.

From looking at the source, this appears to be because `InnerGlobal` checks if the global styles are already included on the page for rehydration purposes here: https://github.com/emotion-js/emotion/blob/c85378a204613885a356eaba1480c5151838c458/packages/core/src/global.js#L76

Unfortunately, this selector doesn't match anything that would be produced by the "advanced approach" SSR guide, which suggests including:

```html
<style data-emotion-css="${ids.join(' ')}">${css}</style>
```

on the page. So if the `InnerGlobal` has `this.props.serialized.name` of `deadb33f`, it's expecting to see:

```html
<style data-emotion-css="deadb33f">...</style>
``` 

on the page to know where to hydrate, which isn't the case: the reality will look more like:

```html
<style data-emotion-css="deadb33f s0meId an0th3rOne">...</style>
```

**To reproduce:**

TODO: cannot use the codesandbox since it's not set up for SSR. Will make a repo if this helps, but want to post this first since I do think I know roughly why it's happening.

**Expected behavior:**

The styles included in `<Global>` should be rehydrated in the same way as non-globals: i.e. any styles that are already included on the page via SSR should not be re-inserted.

I need to dig into how this is working for non-global styles, since presumably the logic is set up there to scan through the IDs includes as part of the `data-emotion-css` attribute rather than expecting a single id per style tag.

I'm also not sure if there's an existing way to use `extractCritical` to split up the css into multiple `<style>` tags - it seems like that would also fix it from the other direction, although the docs would likely need to be updated if that's the way forward.

**Environment information:**

- `react` version: 16.13.1
- `@emotion/core` version: 10.0.35

----

Let me know if there's any other info you need from me. I'll try and spin up a repo if my snippets above don't paint the full picture.

Andarist:
Thank you for the detailed report and for looking through the source code as part of your investigation (I truly believe that most people don't do this enough)!

[...]

PR #2334, opened 2021-04-06
feat(server): add extractCriticalToChunks

mnajdova:
**What**:

This PR aims to fix https://github.com/emotion-js/emotion/issues/2158 and possibly https://github.com/emotion-js/emotion/issues/2049

**Why**:

The new implementation of `extractCritical` should allow developers to solve issues with SSR, mostly with the order of injecting the global css as well as fix dynamic re-writes of the global styles.

**How**:

It is implemented by adding a new `extractCriticalToChunks` utility that will return HTML and an array of styles objects each containing `css` string and an array of ids. Each `injectGlobal` or `<Global />` component will be one item in the array, so that it can be treated independently. The css for the regular styles is added to one item of the list.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Code complete
- [ ] Changeset 

---


packages/server/src/create-instance/extract-critical2.js
```diff
@@ -0,0 +1,38 @@
...
+
+  o.ids = Object.keys(cache.inserted).filter(id => {
+    if (
+      (ids[id] !== undefined ||
+        cache.registered[`${cache.key}-${id}`] === undefined) &&
+      cache.inserted[id] !== true
+    ) {
+      if (cache.registered[`${cache.key}-${id}`]) {
+        o.css += cache.inserted[id]
+      } else {
```
mnajdova, line 28: Based on emotion-js#2049 (comment) seems like it is enough to check if the styles are registered in order to be able to conclude if they are global or not.

docs/ssr.mdx
```diff
@@ -41,6 +41,47 @@ import { renderToString } from 'react-dom/server'
...
+  .header('Content-Type', 'text/html')
+  .send(`<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>My site</title>
+    <style data-emotion="global-css">${globalCss.reduce((acc, curr) => acc + curr)}</style>
+    <style data-emotion="${key} ${ids.join(' ')}">${css}</style>
```
mnajdova, line 67: I believe this should be enough...

mnajdova:
I am not sure if this is all that we need, to be honest. I think that one problem may be that the ids still contain the ids of the global styles too, not sure if this will be problematic.

mnajdova:
TODO: Looks like Iam not resetting something in the tests. Running only the new tests returns the expected output.

packages/server/src/create-instance/extract-critical2.js
```diff
@@ -0,0 +1,38 @@
...
+        o.globalCss.push(cache.inserted[id])
```
Andarist, line 29: Since we can end up with multiple `<Global/>` or `injectGlobal` calls, I think we need to return a new object for each global style (since we should construct a style element per each global style to isolate them and make them flushable independently).

I imagine that `extractCritical2` would return `Array<{ key: string, ids: string[], css: string }>`. Not completely sure yet if we need both key and ids or if this could be simplified without diving more into the code. By adding tests for independent flushing this should become more apparent what is needed here. We might not need to differentiate between regular and global "chunks" (elements of the proposed return type) since the returned info should be enough for the HTML to be generated but I might be wrong on this one.

mnajdova: Ah I see that makes sense, I believe that for the regular `css` we should have an array of ids, but for the global items, it would theoretically be one id. Let me try to refactor and make sure that that would work. Will add test cases for independent flushing.

docs/ssr.mdx
mnajdova: Added `constructStyleTags`, but after the refactoring, it turned out to be trivial to add these (as all styles are an array, where the global styles are added first, and the rest of the regular styles last).. Let me know if you still think it's useful to have it.

packages/server/src/create-instance/extract-critical2.js
mnajdova: Updated, let me know if this is closer to what you had in mind

mnajdova:
@Andarist one question, just to make sure that I completely understand what's happening :) 

When using the `<Global />` with SSR, the styles from the `<Global />` component are being added in two `style` tags (simple example):
```
<style data-emotion="css-global" data-s="">*{color:white;}</style> <!-- added by the <Global /> component -->
<style id="emotion-server-side" data-emotion="css 1xldrws cwfy8f" data-s="">
  *{color:white;}
  .css-cwfy8f{background:pink;width:200px;height:200px;}
</style> <!-- all styles, global + styles from some styled() component -->
```

Now if the `<Global />` component is removed, the first `<style />` tag with `data-emotion="css-global"` that was automatically created will be correctly removed, but the styles are still added in the second `<style />` tag, which was generated with the result from `extractCritical`. This mean that `extractCritical` on the second re-render still contain the global classes although **SHOULD BE** removed? My hope is that with the new `extractCritical2`, when generating the `styles` array, the regular CSS chunk will never add the global styles together with the rest of the CSS, so if some id of the global styles doesn't exist, that `<style />` tag will never be added. I am mostly asking because I am not sure how the input param of the `extractCritical` is being created.

Andarist (commented):
We'd need to a test like this:

1. SSR `<Global/><Global/><div css={{}}/>` using `extractCritical2`
2. rehydrate that
3. unmount one of `<Global/>`s

The expected outcome is that only those styles would be removed and that they would be removed successfully (which I might sometimes refer to as flushing).

I think that this might require adding `key` to the items returned by `extractCritical2` as `<Global/>` styles have keys suffixed with `*-global` (from what I recall). 

A second concern is that we should check how this is going to work when rehydrating with `@emotion/css`. It would be great if we could add the ability to unmount `injectGlobal` styles:
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/css/src/create-instance.js#L96 . However, right now `@emotion/css` assumes a single sheet so `flush` is easy and straightforward:
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/css/src/create-instance.js#L112-L116
The ability to unmount `injectGlobal` styles would require maintaining multiple sheets (this is what [`<Global/>`](https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/react/src/global.js#L94) kinda does) and this might be a little bit tricky/unclear for `@emotion/css`

`@emotion/css` stuff is not a must here but would be great if we could at least take a closer look at it to assess how problematic implementing this for it would be.

packages/server/src/create-instance/extract-critical2.js
```diff
@@ -0,0 +1,46 @@
...
+  let regularCss = ''
+
+  Object.keys(cache.inserted).forEach(id => {
+    if (
+      (ids[id] !== undefined ||
+        cache.registered[`${cache.key}-${id}`] === undefined) &&
+      cache.inserted[id] !== true
+    ) {
+      if (cache.registered[`${cache.key}-${id}`]) {
+        // regular css can be added in one style tag (by convention the first item in the array)
```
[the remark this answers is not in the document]
mnajdova: The comment is outdated will update

packages/server/test/util.js
```diff
@@ -163,7 +163,28 @@ export const prettifyCritical = ({
...
+}: {
+  html: string,
+  styles: Array<{ css: string, ids: Array<string> }>
+}) => {
+  return {
+    styles: styles.map<{ css: string, ids: Array<string> }>(
+      (item): { css: string, ids: Array<string> } => {
+        return { css: prettify(item.css || ''), ids: item.ids }
+      }
+    ),
```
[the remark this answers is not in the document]
mnajdova: I think I was fixing a flow issue with this, but will check again

packages/server/src/create-instance/extract-critical2.js
```diff
@@ -0,0 +1,46 @@
...
+        regularCssIds.push(id)
+        regularCss += cache.inserted[id].toString()
```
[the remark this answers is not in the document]
mnajdova: This was definitely a flow issue as the value is `string | bool`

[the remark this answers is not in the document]
Andarist: let's put a `// $FlowFixMe` above this line then

packages/server/src/create-instance/construct-style-tags.js
```diff
@@ -0,0 +1,20 @@
...
+) => (criticalData: {
+  html: string,
+  styles: Array<{ ids: Array<string>, css: string }>
+}) => {
+  return criticalData.styles.reduce((acc, item) => {
+    return (
+      acc +
+      generateStyleTag(cache.key, item.ids.join(' '), item.css, nonceString)
+    )
+  }, '')
```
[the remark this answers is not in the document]
mnajdova: Refactored to use `forEach`.

packages/server/src/create-instance/inline.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import type { EmotionCache } from '@emotion/utils'
 
-function generateStyleTag(
+export function generateStyleTag(
```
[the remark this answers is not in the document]
mnajdova: Extracted the commonly used method to utils.

mnajdova:
I'll need some help with the test.

> We'd need to a test like this:
> 
> 1. SSR <Global/><Global/><div css={{}}/> using extractCritical2
> 2. rehydrate that
> 3. unmount one of <Global/>s
> 
> The expected outcome is that only those styles would be removed and that they would be removed successfully (which I might sometimes refer to as flushing).
> 

I've tried implementing this with https://github.com/emotion-js/emotion/pull/2334/commits/d7e7f8c88b92522d187abc0f85f16782a1d508fd (I've updated the components), but I am missing the step 2 
>  2. rehydrate that

Will it be enough if I just call `emotion.hydrate(ids)`? Currently the result of the second re-rendering is identical with the first one, which is because of the input (cache) coming in `extractCritical2`.

packages/server/test/util.js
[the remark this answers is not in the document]
mnajdova: I am getting this erro without it 

```
D:\workspace\emotion>yarn flow check
yarn run v1.22.5
$ flow check
Error ------------------------------------------------------------------------------ packages/server/test/util.js:181:13

Missing type annotation for `U`. `U` is a type parameter declared in function type [1] and was implicitly instantiated
at call of method `map` [2]. [missing-annot]

   packages/server/test/util.js:181:13
                    v--------------------
   181|     styles: styles.map(item => {
   182|       return { css: prettify(item.css), ids: item.ids }
   183|     }),
            -^ [2]

References:
   C:\Users\mnajd\AppData\Local\Temp\flow\flowlib_1d9a0d97\core.js:270:5
   270|     map<U>(callbackfn: (value: T, index: number, array: Array<T>) => U, thisArg?: mixed): Array<U>;
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]



Found 1 error
```

Added `// $FlowFixMe`

mnajdova:
I've created [this branch](https://github.com/oliviertassinari/mui-repro-global-styles/pull/1/files) for testing, the project is taken from the repro in https://github.com/emotion-js/emotion/issues/2158#issuecomment-761550869

I still think we just need to **not bundle** the global styles in the returned string in the `css` coming from `extractCrtitical`, and the issue would be solved, but I may be missing something

Andarist:
> I've tried implementing this with d7e7f8c (I've updated the components), but I am missing the step 2

I've first done [this](https://github.com/emotion-js/emotion/commit/ee704e10d4c95000bb6c06609e08b59cdce5c618#diff-8a41a3bb909094190325495242aab95bc553c6769d9060da0f18c4d3fafcc86dR176) but this wasn't quite "e2e" just yet so I've followed up with [this](https://github.com/emotion-js/emotion/commit/104d539c4829b245204fe37cbe613958a43b0aa3#diff-8a41a3bb909094190325495242aab95bc553c6769d9060da0f18c4d3fafcc86dR201). In the process I've fixed some small~ issues that were in there - so even if my test is somewhat complicated, it was still worth doing this. I've pushed those commits to your branch so please fetch them before you continue working on this :)

I'll leave some additional comments in a moment but those will just be about things I've noticed so far - gonna try to follow up with a full review in the following days.

packages/react/__tests__/rehydration.js
```diff
@@ -166,3 +197,131 @@ test('initializing another Emotion instance should not move already moved styles
     </head>
   `)
 })
+
+test('xxx', () => {
```
Andarist, line 201: we need to write a proper title here 😅

packages/react/__tests__/rehydration.js
```diff
@@ -166,3 +197,131 @@ test('initializing another Emotion instance should not move already moved styles
...
+    }
+  })
+
+  safeQuerySelector('head').innerHTML = styles
+  safeQuerySelector('body').innerHTML = `<div id="root">${app}</div>`
+  expect(safeQuerySelector('html')).toMatchInlineSnapshot(`
+    <html>
+      <head>
+        <style
+          data-emotion="mui-global l6h"
```
Andarist, line 230: this is somewhat surprising - I don't think I've ever seen such a short generated hash in Emotion but also while I was working on this I haven't really noticed anything wrong here and this seems to be OK, might still be worth checking this case in the `@emotion/hash`, just to be sure

packages/react/__tests__/rehydration.js
```diff
@@ -166,3 +197,131 @@ test('initializing another Emotion instance should not move already moved styles
...
+        
+        body{color:white;}
+      </style>
+      <style
+        data-emotion="mui-global"
+        data-s=""
+      >
+        
+        html{background:red;}
+      </style>
```
Andarist, line 290: SSRed global style elements were replaced here by client-rendered ones, this all happens very quickly so this might not be noticeable (although I'm slightly worried about unmounting `@font-face` and stuff like that, even for a brief moment) but I would still try to check what could be done about it

Why does this happen? SSR style is immediately flushed here:
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/react/src/global.js#L126-L131
since we already have some rehydrated `sheet.tags` coming from here: 
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/react/src/global.js#L108-L110

I believe that the original intention was to flush this on updates but right now we are also flushing it when rehydrating 😕

packages/server/src/create-instance/construct-style-tags.js
```diff
@@ -0,0 +1,26 @@
...
+  nonceString: string
+) => (criticalData: {
+  html: string,
+  styles: Array<{ ids: Array<string>, css: string }>
+}) => {
+  let styleTagsString = ''
+
+  criticalData.styles.forEach(item => {
+    styleTagsString += generateStyleTag(
+      item.key || cache.key,
```
Andarist, line 16:
```suggestion
      item.key,
```

I've thought that this was also used by `createRenderStylesToString` and I've left `cache.key` here as a fallback so I could investigate this later but I've now rechecked this and it's not used there. So this should be safe to be removed.

packages/server/src/create-instance/inline.js
```diff
@@ -38,7 +28,12 @@ const createRenderStylesToString = (
   }
 
   if (globalStyles !== '') {
-    result = generateStyleTag(cssKey, globalIds, globalStyles, nonceString)
+    result = generateStyleTag(
+      cssKey,
+      globalIds.substring(1),
```
Andarist, line 33: I think this removes leading whitespace but I have not checked this. Either way - I've just hoisted this substring call from `generateStyleTag` so this should be totally equivalent so maybe it's not worth re-checking this.

packages/server/types/create-instance.d.ts
```diff
@@ -10,10 +10,16 @@ export interface EmotionCritical {
   css: string
 }
 
+export interface EmotionCritical2 {
+  html: string
+  styles: Array<{ ids: Array<string>; css: string }>
```
Andarist, line 15:
```suggestion
  styles: Array<{ key: string; ids: Array<string>; css: string }>
```

packages/react/__tests__/rehydration.js
```diff
@@ -166,3 +197,131 @@ test('initializing another Emotion instance should not move already moved styles
...
+  expect(safeQuerySelector('html')).toMatchInlineSnapshot(`
+    <html>
+      <head>
+        <style
+          data-emotion="mui-global l6h"
+        >
+          body{color:white;}
+        </style>
+        <style
+          data-emotion="mui-global 10q49a4"
```
Andarist, line 235: I've also noticed that global styles (with `-global` suffix) are not put on the inserted cache while rehydrating because this only checks the literal key:
https://github.com/emotion-js/emotion/blob/4d7efcb33c5e70f6f1732a96023a25f5270e72dd/packages/cache/src/index.js#L90-L96

I don't know what consequences this have - we'll have to check this.

packages/react/__tests__/rehydration.js
mnajdova: Ah, this may be the reason for the flickering here - https://deploy-preview-25690--material-ui.netlify.app/ The text is black for a split second, then it changes to white, which is coming from the global styles. Although on this example, I am just omitting the global styles coming from `extractCritical2`, so it may be a different issue. Will have to double-check.

packages/react/__tests__/rehydration.js
mnajdova: I don't think is really likely to have a collision, and I don't think it's worth even looking into changing the hashing algorithm based on one string that generates a shorter hash. I couldn't spot anything suspicious. It is funny on the other hand that it came up in styles we added in tests 😅

packages/react/__tests__/rehydration.js
mnajdova:
```suggestion
test('dynamic global styles in SSR should work as expected', () => {
```

Maybe something like this?

mnajdova:
> I've first done this but this wasn't quite "e2e" just yet so I've followed up with this. In the process I've fixed some small~ issues that were in there - so even if my test is somewhat complicated, it was still worth doing this. I've pushed those commits to your branch so please fetch them before you continue working on this :)
> 
> I'll leave some additional comments in a moment but those will just be about things I've noticed so far - gonna try to follow up with a full review in the following days.

Great, thanks for helping out with the test, I believe it reflects the scenario we want to test for. I will update also the test project I have on MUI to see how it would work with the latest changes. I've polished the PR to get a green build, will wait for the next round of review and will post any problems I find in the mean time.

mnajdova:
I've updated https://github.com/mui-org/material-ui/pull/25690. Seems like everything is working as expected, I could not spot any issues right now - https://deploy-preview-25690--material-ui.netlify.app/

One thing I noticed is that, we may want to also export method for generating a list of style tags instead of just the string, as I needed the styles in that format (probably if available I could have used something like `generateStyleTag` that would return the jsx element instead of string).

packages/react/__tests__/rehydration.js
Andarist: Ye - nothing to worry about. I've still checked this out as I was curious.

At the very end of the hashing function:
https://github.com/emotion-js/emotion/blob/6c4a9e50f177900f54f7b3368a55b1a7d5e3c002/packages/hash/src/index.js#L63
we get `h` equal to `73383811246905` which in binary is `10000101011111000000000000000000110101100111001`. The first operation that is performed on this there is a bitwise right shift (`>>>`). Since this is 64bit number the left 32bit is just dropped, leaving us with `0000000000000000110101100111001` which only then gets shifted and as we shift by 15 positions we end up with `0`. XORing (`^`) the original `h` by `0` also discards the first 32 bit and only performs XOR after that which brings us to the `27449` result. So in a sense - stars just aligned nicely here and the left side of those 32 final bits are all `0`, leaving us with a small-ish result and thus the computed hash is short.

packages/react/__tests__/rehydration.js
Andarist: Hm, I can only see the `color:white` declaration in the SSRed HTML 🤔 Has this site been updated since then?

packages/react/__tests__/rehydration.js
Andarist: cc @mitchellhamilton , the inserted/registered stuff is fuzzy to me, so your input would be very much appreciated on this one

Andarist:
> One thing I noticed is that, we may want to also export method for generating a list of style tags instead of just the string, as I needed the styles in that format (probably if available I could have used something like generateStyleTag that would return the jsx element instead of string).

@mnajdova could you give an example where you have needed that?

packages/react/__tests__/rehydration.js
oliviertassinari: ⚠️ this was marked as resolved, but doesn't seem to be

oliviertassinari (commented):
@mnajdova I assume the borders we see in https://deploy-preview-25690--material-ui.netlify.app/components/grid/ are linked to a different issue? Global leaks between different pages.

mnajdova:
> @mnajdova I assume the borders we see in https://deploy-preview-25690--material-ui.netlify.app/components/grid/ are linked to a different issue? Global leaks between different pages.


No, sorry I’ve updated the example, we should see the borders after 5s too as that Global component stays mounted. Only the color should change.

packages/react/__tests__/rehydration.js
mnajdova: I cannot spot the flickering too this one time.. I’ve updated the example, we should see the borders after 5s too as that Global component stays mounted. Only the color should change to black (default)

mnajdova:
> > One thing I noticed is that, we may want to also export method for generating a list of style tags instead of just the string, as I needed the styles in that format (probably if available I could have used something like generateStyleTag that would return the jsx element instead of string).
> 
> @mnajdova could you give an example where you have needed that?

Here it is - https://github.com/mui-org/material-ui/blob/75e59afc24e881d9897c634d63b82c3295cb6724/docs/pages/_document.js#L155

packages/react/__tests__/rehydration.js
mitchellhamilton: This is correct. When on the client, global styles(as in Global component, not injectGlobal) are not put in the inserted cache. The global styles are rehydrated by the Global component rather than the cache https://github.com/emotion-js/emotion/blob/6c4a9e50f177900f54f7b3368a55b1a7d5e3c002/packages/react/src/global.js#L92-L110

mitchellhamilton (commented):
Could we give this a more descriptive name? Maybe `extractCriticalToChunks` or something idk

mnajdova:
> Could we give this a more descriptive name? Maybe `extractCriticalToChunks` or something idk

@mitchellhamilton done 👍

packages/react/__tests__/rehydration.js
Andarist: ah, ye - I forgot to push out the change I've applied locally 😅 good, vigilant eye!

packages/react/__tests__/rehydration.js
Andarist: @mitchellhamilton I would like to fix this issue with flushing and reinjecting Global styles when rehydrating. The easiest fix for that would be to just merge those 2 layout effects. Any objections to that? The first effect just avoids recreating StyleSheet (from what I can tell) based on the `serialized.name` but since `sheet.flush` is called anyway when it changes and highly dynamic global styles are rather rare I don't think we must maintain that split between two effects 🤔

packages/react/__tests__/rehydration.js
Andarist: Hm... it was actually asked for that `injectGlobal` styled could be removable. This is, of course, not a concern of this PR - but I was also thinking that we could support this in the future.

mnajdova:
> @mnajdova I assume the borders we see in https://deploy-preview-25690--material-ui.netlify.app/components/grid/ are linked to a different issue? Global leaks between different pages.

@oliviertassinari can you still repro this? I wasn’t able to repro it locally too after the latest updates.

packages/react/__tests__/rehydration.js
mitchellhamilton: Sounds good

Andarist: I've implemented this change - could you take a look @mitchellhamilton ? I doubt that this can break anything but better to recheck this

docs/ssr.mdx
```diff
@@ -41,6 +41,46 @@ import { renderToString } from 'react-dom/server'
...
+</head>
+<body>
+    <div id="root">${html}</div>
+
+    <script src="./bundle.js"></script>
+</body>
+</html>`);
+```
+
+#### Deprecated
```
Andarist, line 76: For now, this won't get deprecated - the recommendation would be that:
- `extractCriticalToChunks` should be used when using `@emotion/react`
- `extractCritical` should be used when using `@emotion/css`

packages/react/__tests__/compat/server.js
```diff
@@ -41,7 +41,7 @@ test('extracted rules have correct keys when dealing with multiple caches', () =
   )
 
   let cache2 = createCache({ key: 'ssr-second-key' })
-  let { extractCritical: extractCritical2 } = createEmotionServer(cache2)
+  let { extractCritical: extractCriticalToChunks } = createEmotionServer(cache2)
```
Andarist, line 44: in here it actually should be kept as `extractCritical2` :)

packages/server/test/util.js
```diff
@@ -163,7 +163,29 @@ export const prettifyCritical = ({
...
 }) => {
-  return { css: prettify(css), ids, html }
+  return {
+    css: prettify(css),
+    ids,
+    html
+  }
+}
+
+export const prettifyCritical2 = ({
```
Andarist, line 173: as this critical2 was kinda renamed this should be adjusted as well

packages/server/types/create-instance.d.ts
```diff
@@ -10,10 +10,16 @@ export interface EmotionCritical {
...
+export interface EmotionCriticalToChunks {
+  html: string
+  styles: Array<{ key: string; ids: Array<string>; css: string }>
+}
+
 export interface EmotionServer {
   extractCritical(html: string): EmotionCritical
   renderStylesToString(html: string): string
   renderStylesToNodeStream(): NodeJS.ReadWriteStream
+  constructStyleTags(html: string): EmotionCriticalToChunks
```
Andarist, line 22: This is not correct. The return type of this is a string but the interface definition is missing a declaration for `extractCriticalToChunks` - for which this return type would be correct.

wdyt about renaming `constructStyleTags` to `constructStyleTagsFromChunks`? That way it would be more clear that those 2 APIs can be used together (`extractCriticalToChunks` and `constructStyleTagsFromChunks`)

packages/react/__tests__/compat/server.js
mnajdova: Good catch :)))

docs/ssr.mdx
mnajdova: Updated, you can re-check the wording.

mnajdova:
@Andarist I've tried resolving last comments with https://github.com/emotion-js/emotion/pull/2334/commits/124cf3fc899917207f86a528d255d564f1efb2ba, https://github.com/emotion-js/emotion/pull/2334/commits/88de62deec589a9d1f221e2b4951c53088743be1

mnajdova:
@Andarist I've upgraded our test branch on Material-UI, and I got the error below when navigating from one page to another. I think it's related to https://github.com/emotion-js/emotion/pull/2334/commits/00ce87be39b241e286395558efed0b259ae10999. It fails on `sheet.flush()`. I haven't seen the issue before that commit.

```
Unhandled Runtime Error
TypeError: Cannot read property 'removeChild' of null

Call Stack
<unknown>
node_modules\@emotion\sheet\dist\emotion-sheet.browser.esm.js (133:0)
Array.forEach
<anonymous>
StyleSheet.flush
/_next/static/chunks/pages/_app.js (2676:15)
<unknown>
node_modules\@emotion\react\dist\emotion-react.browser.esm.js (158:0)
HTMLUnknownElement.callCallback
```

Repro steps:

1. Visit: https://deploy-preview-25690--material-ui.netlify.app/branding/about/
2. Click the Docs link on the top of the page
3. Error is thrown 

We use different `<Global />` components on these two pages.

packages/react/__tests__/rehydration.js
mitchellhamilton: Hmmm, reading the code, it seems as though the change doesn't preserve the behavior of inserting the new style tag in the same position as the previous one. Could you add a test based on https://codesandbox.io/s/still-leaf-invs2?file=/src/App.js?

Andarist:
@mnajdova I believe that I've fixed the issue & added tests for this - I've also confirmed that in your repository.

packages/react/src/global.js
```diff
@@ -101,35 +98,24 @@ export let Global: React.AbstractComponent<
...
         `style[data-emotion="${key} ${serialized.name}"]`
       )
-
       if (cache.sheet.tags.length) {
         sheet.before = cache.sheet.tags[0]
       }
       if (node !== null) {
+        // clear the hash so this node won't be recognizable as rehydratable by other <Global/>s
+        node.setAttribute('data-emotion', key)
         sheet.hydrate([node])
```
Andarist, line 107: it turns out that `.hydrate` **moves** those global styles in `<head/>` anyway (but it doesn't remove them, even temporarily) so maybe those changes in the `<Global/>` were not needed at all as there is a high chance that moving styles behaves very similar to flushing+reinjecting 

there is also a possibility that it doesn't (I have not observed any problems with either approach), I feel though that the implementation is somewhat simpler now as it's easier to reason about a single effect than about two, so I would be probably in favor of keeping this implementation, thoughts? cc @mitchellhamilton

packages/react/__tests__/rehydration.js
```diff
@@ -166,3 +198,401 @@ test('initializing another Emotion instance should not move already moved styles
...
+        data-emotion="muii 1lrxbo5"
+        data-s=""
+      >
+        .muii-1lrxbo5{color:hotpink;}
+      </style>
+    </head>
+  `)
+})
+
+test('duplicated global styles can be removed safely after rehydrating HTML SSRed with zero config approach', () => {
```
Andarist, line 466: I thought that this case could be failing so I've added a test for this but it turned out that it wasn't failing.

I was under the impression that this query would select the same node twice when looking for the same hash:
https://github.com/emotion-js/emotion/blob/9596d32302f389979716852bb63e6e85f11a90c7/packages/react/src/global.js#L98-L100
but:
1. we SSR those style elements as `[0, 1]`
2. they are moved automatically to head when initializing, and the order is preserved: `[0, 1]`
3. when rehydrating the first global it grabs the element 0 and rehydrates it which **moves** that element before all other tags from the main cache so we end up temporarily with `[1 , 0, ...mainCache]` order
4. the second global thus now grabs the element 1 and also moves It before other tags from the main chance and we end up with `[0, 1, ...mainCache]`

I think this works OK and I couldn't find a situation in which it would break but maybe @mitchellhamilton would have some idea for an edge case here that could be tested.

mnajdova:
> @mnajdova I believe that I've fixed the issue & added tests for this - I've also confirmed that in your repository.

Great! I double-checked all failing scenarios we had, no issues found 🚀

packages/react/src/global.js
mitchellhamilton: I don't mind how it is implemented as long as it preserves the insert position it got when mounted which I don't believe we have a test for currently and I don't think the changed implementation does it correctly. https://codesandbox.io/s/still-leaf-invs2?file=/src/App.js

Andarist: Yes, the changed implementation doesn't cover this scenario correctly. However, if we add updates to the mix then the current stable release doesn't handle it correctly either: https://codesandbox.io/s/optimistic-browser-0inm1?file=/src/App.js

Scenario:
1. edit second input
2. toggle the first one twice

The premise of React is that the UI is always derived from state and this breaks that rule anyway. I'm not really sure if that's worth supporting unless we would figure out stable insertion points but I don't think we can (without complicating things too much).

mitchellhamilton: Yeah, I know that's not supported and I wouldn't have expected it to be but "The premise of React is that the UI is always derived from state and this breaks that rule anyway" isn't really accurate, preserving state between rerenders but not remounts is expected and just how things work in React e.g. https://codesandbox.io/s/jovial-greider-f4cgq. I don't think we should make the case you proposed work but we shouldn't regress from the current behavior.

Andarist: > preserving state between rerenders but not remounts is expected and just how things work in React

This one is different though as what gets reset is self-contained whereas `<Global/>` isn't and affects the outside world by rendering styles in a mount/remount/etc-order dependent fashion. I still stand by the "premise of React" being broken here anyway since for any given set of props, state, and context **values** the visual output should be the same.

That being said - I also don't see a point in supporting what I've showcased in the sandbox but I see how this could be used as an argument against supporting your use case. If things are not the same 100% of the time I don't think it's really worth supporting the 90% here as it gives a false sense of safety. But, at this point in time, it's not really worth bikeshedding this anyway as this would be sort of a breaking change so it's much better to just revert to the previous solution. I'm just wondering if it's worth implementing SSRed style preservation on top of that previous solution.

Andarist: I've reverted the merging of those 2 layout effects - although I have not added a test for the mentioned case because I'm not super convinced that this is something that we should support in the long run and I didn't quite have the time to add it right now. The use case seems very esoteric to me and we cannot easily produce stable order of global styles throughout the tree anyway.

For now, I've also ensured that the rehydrated global styles are not flushed and reinjected but I'm not super sure If this is required. I don't think we had any reports about this causing problems, so maybe we should just revert this small change?

mitchellhamilton: I don't mind about the flushing and reinjecting change.

(also, i agree about the preserving order in Global not being a thing we should care about, it's really just about not breaking things for people now)

Andarist:
Ok, I'll be taking a last look at this in the following days but I think it's in a good shape and that there are no major blockers against merging this. If you, @mitchellhamilton, have some spare time I would also appreciate your review here since this is touching some core logic that we need to get right.

packages/server/src/create-instance/extract-critical-to-chunks.js
```diff
@@ -0,0 +1,53 @@
...
+    if (
+      (ids[id] !== undefined ||
+        cache.registered[`${cache.key}-${id}`] === undefined) &&
+      cache.inserted[id] !== true
+    ) {
+      if (cache.registered[`${cache.key}-${id}`]) {
+        // regular css can be added in one style tag
+        regularCssIds.push(id)
+        // $FlowFixMe
+        regularCss += cache.inserted[id].toString()
```
mitchellhamilton, line 35:
```suggestion
        regularCss += cache.inserted[id]
```

mitchellhamilton (approved):

Andarist (approved):

PR #2337, opened 2021-04-08
fix bug when parent does not exist

zhaoyao91:
for some case such as in shadow dom, if the parnet is cleaned by other mechanism, this statement will error.

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/sheet/src/index.js
+++ b/packages/sheet/src/index.js
@@ -155,7 +155,12 @@ export class StyleSheet {
 
   flush() {
     // $FlowFixMe
-    this.tags.forEach(tag => tag.parentNode.removeChild(tag))
+    this.tags.forEach(tag => {
+      const parent = tag.parentNode;
+      if (parent) {
+        tag.parentNode.removeChild(tag)
+      }
+    })
     this.tags = []
     this.ctr = 0
     if (process.env.NODE_ENV !== 'production') {


Andarist:
If the parent gets cleaned up by other mechanisms then I would say that you shouldn't use `flush`. It sounds like one or another should be chosen.

zhaoyao91:
> If the parent gets cleaned up by other mechanisms then I would say that you shouldn't use `flush`. It sounds like one or another should be chosen.

In some situation we both need to destroy the dom structure but also flush emotion instance but they are happened parallelly and the order is not under control.

Andarist:
If everything is contained in the shadow dom then i dont see why do u need to flush - removing the container should be enough.

zhaoyao91:
> If everything is contained in the shadow dom then i dont see why do u need to flush - removing the container should be enough.

well, my tool as general lib may work not only in shadow dom but also use head or other element as container, so flush at the end when the usage scope is destroied is an assurance and closed cycle.

Andarist:
I would assume that in this case you need to expose some kind of a cleanup function and the consumer calls that. If that's the case - why can't you guarantee the order in which things happen? If that's not the case - why? It's usually a good practice to expose such function and expect the user to call that if they want to perform cleanup correctly

PR #2340, opened 2021-04-08
Fixed an issue with extracting labels from obj literal properties and added support for extracting from object methods

Andarist:


packages/babel-plugin/__tests__/styled-macro/__fixtures__/label.js
```diff
@@ -3,30 +3,4 @@
 import styled from '@emotion/styled/macro'
 import { jsx } from '@emotion/react'
 
-class Thing {
```
Andarist, line 6: I've split this file into several, smaller ones. Each new test added to this has been changing generated source maps for each test case (so it has affected unaffected cases) and it was really noisy.

packages/babel-plugin/src/utils/label.js
```diff
@@ -53,6 +53,21 @@ export function getLabelFromPath(path: *, state: *, t: *) {
...
+    (!t.isObjectProperty(path) && !t.isObjectMethod(path)) ||
+    path.node.computed
+  ) {
+    return null
+  }
+  if (t.isIdentifier(path.node.key)) {
+    return path.node.key.name
+  } else {
+    // literals
+    return String(path.node.key.value).replace(/\s+/g, '-')
```
Andarist, line 67: This handles numeric and bigint literals but I think that I should revert this and only handle string literals here. Thoughts?

mitchellhamilton: Yeah, I'd just do string literals

Andarist: done

mitchellhamilton (approved):

PR #2349, opened 2021-04-14
feat: re-exporting keyframes from @emotion/react when using importMap

### LINKED ISSUE #2218: Re-exporting `keyframes` from @emotion/react using importMap throws
**Current behavior:**

Hello! 👋 

I'm trying to re-export `keyframes` from `@emotion/react` using the babel-plugin and it throws an error with: `transformer is not a function`.

**To reproduce:**

Define the `importMap` as follows:
```json
[
  "@emotion/babel-plugin",
  {
    "importMap": {
      "myPackageName": {
        "css": {
          "canonicalImport": ["@emotion/react", "css"],
        },
        "keyframes": {
          "canonicalImport": ["@emotion/react", "keyframes"],
        }
      }
    }
  }
]
```

**Expected behavior:**

The build to be succesfull, but instead it throws with: `transformer is not a function`.

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version:  11.1.2
- `@emotion/babel-plugin` version: 11.1.4

**Additional notes:**
I did look over the plugin code-base and indeed it seems that the `keyframes` transformer is not in there: https://github.com/emotion-js/emotion/blob/master/packages/babel-plugin/src/core-macro.js#L191-L193. Was this intentional? Or is just wasn't requested before?

Thank you so much for all you do! 🙏

iChenLei:
**What**: To fix #2218

**Why**: User can't write these example code blow at current time.
```js
[
  "@emotion/babel-plugin",
  {
    "importMap": {
      "myPackageName": {
        "css": {
          "canonicalImport": ["@emotion/react", "css"],
        },
        "keyframes": {
          "canonicalImport": ["@emotion/react", "keyframes"],   👈🏻  // can't use keyframes
        }
      }
    }
  }
]
```
```diff
// Throw Error
- transformer is not a function
```

**How**: Re-export keyframes when using `@emotion/babel-plugin` and `@emotion/react`'s  `keyframes`

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

/cc @Andarist  @mitchellhamilton I am not understanding this issue 100%, so maintainer's ocde review is very important for me, thanks, the awesome `emotion-js` creator !

### DIFF
--- a/packages/babel-plugin/__tests__/import-mapping/__snapshots__/import-mapping.js.snap
+++ b/packages/babel-plugin/__tests__/import-mapping/__snapshots__/import-mapping.js.snap
@@ -1,5 +1,94 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`import mapping emotion-css-keyframes 1`] = `
+"import { keyframes } from 'package-three'
+
+keyframes\`
+  from, 20%, 53%, 80%, to {
+    transform: translate3d(0,0,0);
+  }
+
+  40%, 43% {
+    transform: translate3d(0, -30px, 0);
+  }
+
+  70% {
+    transform: translate3d(0, -15px, 0);
+  }
+
+  90% {
+    transform: translate3d(0,-4px,0);
+  }
+\`
+
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+function _EMOTION_STRINGIFIED_CSS_ERROR__() { return \\"You have tried to stringify object returned from \`css\` function. It isn't supposed to be used directly (e.g. as value of the \`className\` prop), but rather handed to emotion so it can handle it (e.g. as value of \`css\` prop).\\"; }
+
+import { keyframes } from 'package-three';
+
+/*#__PURE__*/
+keyframes(process.env.NODE_ENV === \\"production\\" ? {
+  name: \\"w3sfyc\\",
+  styles: \\"from,20%,53%,80%,to{transform:translate3d(0,0,0);}40%,43%{transform:translate3d(0, -30px, 0);}70%{transform:translate3d(0, -15px, 0);}90%{transform:translate3d(0,-4px,0);}\\"
+} : {
+  name: \\"w3sfyc\\",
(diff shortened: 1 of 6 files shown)


packages/babel-plugin/__tests__/import-mapping/__snapshots__/import-mapping.js.snap
```diff
@@ -1,5 +1,94 @@
...
+
+function _EMOTION_STRINGIFIED_CSS_ERROR__() { return \\"You have tried to stringify object returned from \`css\` function. It isn't supposed to be used directly (e.g. as value of the \`className\` prop), but rather handed to emotion so it can handle it (e.g. as value of \`css\` prop).\\"; }
+
+import { keyframes } from 'package-three';
+
+/*#__PURE__*/
+keyframes(process.env.NODE_ENV === \\"production\\" ? {
+  name: \\"w3sfyc\\",
+  styles: \\"from,20%,53%,80%,to{transform:translate3d(0,0,0);}40%,43%{transform:translate3d(0, -30px, 0);}70%{transform:translate3d(0, -15px, 0);}90%{transform:translate3d(0,-4px,0);}\\"
+} : {
```
Andarist, line 35: It's not that simple - you would have to produce slightly different output here. Notice that the simplified object for `keyframes` is different than this:
https://github.com/emotion-js/emotion/blob/f3c51e8aca287170e43110bee2e4527eacfcada4/packages/react/src/keyframes.js#L13-L22

iChenLei: Got, I try to fix it. Thanks 😄

iChenLei:
@Andarist To be honestly, I am very confused about this `PR` and your advices.
https://github.com/emotion-js/emotion/blob/f3c51e8aca287170e43110bee2e4527eacfcada4/packages/babel-plugin/__tests__/vanilla-emotion-macro/__snapshots__/index.js.snap#L497-L506

Some existing unit test snapshot show us that we don't need add `@keyframes` wrap manually. 😭

Andarist:
You shouldnt wrap this with keyframes but rather produce objects with different shapes than standard ones because keyframes object has different properties than the css object.

Different properties have to be created for both `devNode` and `prodNode` in this file: https://github.com/emotion-js/emotion/blob/main/packages/babel-plugin/src/utils/transform-expression-with-styles.js

PR #2352, opened 2021-04-16
feat: useTheme support typescript generic

### LINKED ISSUE #2123: support for dynamic Theme type composition
useTheme in v10 used to be a generic useTheme<T>() where T would be the Theme's type. In v11 however it is now required to use module declaration.

The problem is that I'm using a monorepo with multiple apps that reuse many components. That means I have no fixed theme type for the app and the final Theme type is constructed depending on which components are being used and the Theme type fragments are being passed up the component tree to compose a final Theme type.

My only option right now would be to explicitly cast the Theme to desired type when using useTheme, which isn't ideal.

iChenLei:
**What**: make `useTheme` support typescript generic to fix #2123

**Why**: community user's feature request

**How**:  make `useTheme` support typescript generic

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset 

/cc @Andarist @mitchellhamilton

### DIFF
--- a/.changeset/breezy-trains-fly.md
+++ b/.changeset/breezy-trains-fly.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+useTheme support typescript generic
--- a/docs/theming.mdx
+++ b/docs/theming.mdx
@@ -192,6 +192,22 @@ render(
 )
 ```
 
+`useTheme` support typescript generic, you can use `useTheme` as follows.
+
+```jsx
+import { useTheme } from '@emotion/react'
+
+interface ITheme {
+  color: string
+}
+
+function SomeText(props) {
+  const theme = useTheme<{ color: string }>();
+  // const theme = useTheme<ITheme>();
+  return <div className={{ color: theme.color }} {...props} />
+}
+```
+
 ## Credits
 
 Thanks goes to the [styled-components team](https://github.com/styled-components/styled-components) and [their contributors](https://github.com/styled-components/styled-components/graphs/contributors) who designed this API.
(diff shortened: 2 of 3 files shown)


mitchellhamilton:
I don't really think we should do this. It's fundamentally not type-safe(I mean nor is any of the theming things but this even less so). If you want to do this, you can already cast to the type you want, the generic isn't doing anything very useful and you can wrap the cast into your own hook if you want. Ideally, you would use createContext for each theme you have and it would be properly typed.

iChenLei:
To be honestly, I just scan `emotion-js/emotion`'s issues page and found issues which i can solve, then I create `PR` to fix(`bug`) or implement(`feature request`).

For this `PR`, I know it isn't `type safe`  and just for `type hint`.
```ts
const theme = useTheme<ICustomTheme>();
// same as
const theme = useTheme() as ICustomTheme
```
Yep, we can use `as` keyword do the same thing, but it also means that disable `<>`operator usage.
From my perspective， if it useful and not cause breaking change, it can be accepted. Of course, it's just my opinion。

### Appended: feel free to close this pr

/cc @webican (issue 2123 author) @dylanvorster (issue 2123 commenter) @Andarist

Andarist:
> To be honestly, I just scan emotion-js/emotion's issues page and found issues which i can solve, then I create PR to fix(bug) or implement(feature request).

The help is very much appreciated (it really is!). It takes you some time to dig into those issues and prepare the PRs - so maybe to save you from situations like this in the future it would best if you would first comment on the issue that you are interested in handling it and then we would tell you if it's worth the trouble?

Andarist:
While I've closed it right now - @mitchellhamilton might still respond to that and reopen if needed.

Issue #2070, opened 2020-11-02
Implement automatic adding of jsxImportSource pragma definition - eslint-plugin-emotion


Emiliano-Bucci:
**What**:
Implement the possibility - when using the new fragma `jsxImportSource` - to automatically add the new pragma definition to the top of the file page, like already is implemented with the "old" pragma `/* @jsx jsx */`

**Why**:
This way we don't have to manually add it every time.

osdiab:
agreed - relevant to #2467 , to opt into that toolset i'd need to in the status quo use the JSX pragma, but it's super inconvenient for NextJS since it does set the runtime to automatic, so it appears I do need to use `jsxImportSource` instead.

PR #2353, opened 2021-04-16
feat: Implement automatic adding of jsxImportSource pragma definition

iChenLei:
**What**: To fix https://github.com/emotion-js/emotion/issues/2070

**Why**: when using the new fragma jsxImportSource - to automatically add the new pragma definition to the top of the file page, like already is implemented with the `old` pragma /* @jsx jsx */

**How**: Add eslint rule support.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset 

### jsx-runtime
react official blog -> https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
babel support -> https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#customizing-the-automatic-runtime-import

https://github.com/emotion-js/emotion/blob/main/packages/react/src/jsx-dev-runtime.js
https://github.com/emotion-js/emotion/blob/main/packages/react/src/jsx-runtime.js

```diff
- "@emotion/jsx-import": "error"
+ "@emotion/jsx-import": [2, "jsxImportSource"]
```
```js
// .eslintrc.js
module.exports = {
  root: true,
  plugins: [ '@emotion'],
  rules: {
    "@emotion/jsx-import": [1, "jsxImportSource"]
   }
}
```
It works as blow:

emotion-eslint-02

### Add jsxImportSource custom jsx-runtime option
```diff
- "@emotion/jsx-import": "error"
+ "@emotion/jsx-import": [2, { jsxImportSource: "customreact/jsx-runtime"  }]
```
```js
// .eslintrc.js
module.exports = {
  root: true,
  plugins: [ '@emotion'],
  rules: {
    "@emotion/jsx-import": [2, { jsxImportSource: "customreact/jsx-runtime"  }]
   }
}
```
It works as blow:

emotion-eslint-01

/cc @Andarist @mitchellhamilton

### DIFF
--- a/packages/eslint-plugin/src/rules/jsx-import.js
+++ b/packages/eslint-plugin/src/rules/jsx-import.js
@@ -1,4 +1,5 @@
 const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/
+const JSX_IMPORT_SOURCE_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/
 
 // TODO: handling this case
 // <div css={`color:hotpink;`} />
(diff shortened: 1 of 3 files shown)


mitchellhamilton (commented):
Could you change the config to be like this?

The default should be this:

```json
{
  "runtime": "classic"
}
```

and to use the new runtime, it should be this

```json
{
  "runtime": "automatic"
}
```

and with a custom importSource, it should be this:

```json
{
  "runtime": "automatic",
  "importSource": "whatever"
}
```

iChenLei:
Thanks for your code review, your suggestion make config be more make sense. I will change my origin implemention later.

iChenLei:
After I changed the implemention:
#### { "runtime": "automatic" }
emotion-eslint-02

#### { "runtime": "automatic", "importSource": "whatever" }
emotion-eslint-o1

If your config is not oneOf [{ "runtime": "xx" }, { "runtime": "xx", "importSource": "yy" }], your code will be treated by the previous @emotion/eslint-plugin. `jsxImportSource` support is a special deal.

osdiab:
Looking forward to using this!

asteroidice (approved):
I'm not qualified to approve this PR. But I would love to be able to use the code in here

packages/eslint-plugin/src/rules/jsx-import.js
```diff
@@ -7,9 +8,66 @@ const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/
...
+          if (node.name.name !== 'css') {
+            return
+          }
+          let hasJsxImportSource = false
+          let sourceCode = context.getSourceCode()
+          let pragma = sourceCode
+            .getAllComments()
+            .find(node => JSX_IMPORT_SOURCE_REGEX.test(node.value))
+          hasJsxImportSource =
+            pragma && pragma.value.match(JSX_IMPORT_SOURCE_REGEX)
```
mitchellhamilton, line 51: Can you check that the importSource is the one that the user specified?

iChenLei: Thanks for code review 🤟  , apply your suggestion later.

iChenLei: @mitchellhamilton I apply your suggestion now, code review please, thanks!

avertin:
I really appreciate this PR. For now, I've copied this as a local eslint rule for my project. Hopefully it can be integrated soon. 

I noticed one difference between classic and automatic options that I was hoping to understand better. For classic, once the jsx import rule is satisfied, these additional rules are also satisfied ( 'jsx' is defined but never used - no-unused-vars AND 'React' must be in scope when using JSX  react/react-in-jsx-scope). With automatic, after the jsx import rule is satisfied, these errors are still present. I'm new to eslint rules and I was wondering what accounts for this difference?

Issue #2358, opened 2021-04-26
Rewrite to TypeScript


Andarist:
Since Flow is not actively used by any core contributor and its adoption in the community has dropped significantly (especially in comparison to its alternative) it's time to refactor the codebase from Flow to TypeScript.

We could use some help with that. It's best to start from the small, leaf packages. We already have TS types for our packages, they are just not generated from the source.

iChenLei:
Cooool !  This is a big migration work, we need a detailed roadmap to implement it.

Andarist:
For now, I would try to convert a simple package like `@emotion/memoize` to see what exact steps will be required there. When we know that then similar steps will have to be replicated in other packages.

rjdestigter:
I took a stab at the _utils_ package: https://github.com/emotion-js/emotion/pull/2359

with-heart:
I opened #2360 to try to handle some of the tooling/config stuff needed to support the rest of the conversions. As of right now, it's mostly focused on `tsconfig.json` files and `eslint` config stuff.

`eslint` is able to auto-fix quite a few of the existing issues (didn't include the auto-fixes in the PR yet to keep it readable), but there are still 155 errors and 62 warnings. I need to go through those and figure out what's caused by missing config stuff versus what's caused by flow usage/will be resolved by actual TypeScript conversion.

michaeljaltamirano:
> I've created a [ts-migration](https://github.com/emotion-js/emotion/tree/ts-migration) branch. It seems to me that the easiest approach to do this would be to actually:
> 
> 1. drop Flow entirely - from the source code, CI configs, tooling
> 2. add TS tooling (partially already done in #2360 and #2359
> 3. migrate packages one by one
> 
> This would simplify both the whole process and individual PRs.

I think this is a great idea 🙂 Thanks for setting it up!

JoshuaKGoldberg:
> 1. drop Flow entirely - from the source code, CI configs, tooling

I've started a PR in parallel to do so against the `ts-migration` branch here: https://github.com/emotion-js/emotion/pull/2385

If a maintainer could grant my builds to run, that'd be very helpful. 😄

Andarist:
> If a maintainer could grant my builds to run, that'd be very helpful. 😄

I hate that GitHub "feature" >.< Can I somehow grant you permission to run builds? Or will I have to approve them manually until we merge your PR? I believe that the restrictions get lifted if you already have some commits in the repo.

[...]

PR #2359, opened 2021-04-26
TypeScript migration of @emotion/utils

rjdestigter:
**What**:
Migrating @emotion/utils from Javascript + Flow to TypeScript.

This was an initial stab at migrating a package to TypeScript. I'm primarily running into problems with linting. I had to add a line to _.eslintignore_ due to ESLint not understanding optional properties on types and interfaces. This first commit also used `--no-verify` as the commit hook script also failed for similar reasons.

I kept the old declaration for now to use in the `test.ts` file to make sure that the auto generated function signatures matched the older declaration.

Build and tests do succeed on this branch atm.

 Todo:
- [ ] Figure out linting
- [ ] Figure out proper TypeScript version?
- [ ] Commit hook linting fails
- [ ] Remove old declaration files
- [ ] Is the TSLint file still relevant? Should the repo's ESLint configuration be updated first maybe?

**Why**:
https://github.com/emotion-js/emotion/issues/2358

**How**:
- Added the latest TypeScript as a dependency to @emotion/utils
- Moved existing _tsconfig.json_ to root of package
- Moved `*.js` files to `*.ts` and removed references to Flow

**Checklist**:

More todos:
- [ ] Documentation N/A?
- [ ] Tests (They are running atm)
- [ ] Code complete
- [ ] Changeset


Andarist:
Thank you for starting this! I've already played a little bit with your branch locally and trying to wrestle the config-related stuff. It might take me some days - since I have to pause working on this and won't be available much in the following days but I will get back to this ASAP and get this through the finish line!

packages/utils/src/index.ts
```diff
@@ -44,7 +43,7 @@ export const insertStyles = (
   }
   if (cache.inserted[serialized.name] === undefined) {
     let stylesForSSR = ''
-    let current = serialized
+    let current: SerializedStyles | undefined = serialized
```
MattSchiller, line 46: Is this even actually possible for `serialized` to be `undefined` here? I would've expected `if (cache.inserted[serialized.name] === undefined) {` to error from accessing `undefined.name` if so.

If that's actually possible, would it make sense to introduce optional chaining into the if statement's syntax?

MattSchiller (commented):
As a casual consumer of this repo, I love seeing this work starting! Left a comment and gonna see what I can do to contribute here, too 🙇

packages/utils/src/index.ts
michaeljaltamirano: I think it is `current`-specific typing in advance because `current.next` is possibly undefined and we're checking against it as part of the `do`/`while`:

https://github.com/emotion-js/emotion/blob/3f66787ad41923515ae72d8b03d881a3b83efdd6/packages/utils/src/index.ts#L46-L58

MattSchiller: Ah, gotcha, that makes sense. I haven't seen a while loop like this _in a while_ :pun-dog: so it threw me. Thanks for the quick update!

Andarist:
@rjdestigter ive merged in TS tooling setup into `ts-migration` branch. If you would rebase on top of that it would be sweat. Let me know if you are still interested in working on that

PR #2360, opened 2021-04-27
Add TypeScript tooling and configs

with-heart:
**What**:
This PR is an attempt to add TypeScript-related tooling and config required to complete the conversion from Flow to TypeScript. 

First step was getting `eslint` to run without any configuration errors (which @rjdestigter was running into in #2359), which seems to working for the most part. Also created a root `tsconfig.json` and tsconfig files for each package which extend from the root, since all of the existing were already duplicating the same config.

**Why**:
#2358

**How**:

- Created a root `tsconfig.json` which specifies the `compilerOptions` shared by all of the existing `tsconfig.json` files
	- Set `include` to `["packages", "scripts", "site", "test", "playgrounds"]`
	- Removed a few rules specifiers already enabled by `"strict": true`
- Created new and moved existing `tsconfig.json` files at root of each package
	- Specified `extends` for each to root `tsconfig.json`
	- Removed `compilerOptions` from all as they now inherit these from the root
	- Set `include` for each to `"src"`
- Configured `eslint` to support ts
	- Added `@typescript-eslint/parser` & `@typescript-eslint/eslint-config`
	- Switched `eslint-config-standard` to [`eslint-config-standard-with-typescript`](https://github.com/standard/eslint-config-standard-with-typescript)
	- Set `parserOptions.project` to root `tsconfig.json`
	- Disabled `no-use-before-define` and enabled `@typescript-eslint/no-use-before-define` as `warn`
- Added `@babel/preset-typescript` package and added to `presets` in `babel.config.js`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset N/A?

I haven't even attempted to build anything yet or actually convert files to ts. It seems like a lot of the base tooling- and config-related stuff probably needs to be dealt with first in order to support the conversion of actual packages.

I'd like to expand the checklist to include remaining items to handle, but I need some feedback and to explore a little more to determine what else there is to do.


rjdestigter:
Nice! I rebased (locally) my PR off your branch. It looks like there is a conflict between ESLint and Prettier when defining types and interfaces:

ESLint will complain about commas and semicolons: `Unexpected separator (;).eslint@typescript-eslint/member-delimiter-style`

While prettier seems to think we have bad syntax when there is neither: `(property) SerializedStyles.styles: string`

image

package.json
```diff
@@ -51,7 +53,7 @@
   ],
   "eslintConfig": {
     "extends": [
-      "standard",
+      "standard-with-typescript",
```
Andarist, line 56: I think we can just drop the standard plugin since we are using Prettier.

package.json
```diff
@@ -27,6 +27,8 @@
     "release": "yarn build && changeset publish"
   },
   "resolutions": {
+    "**/@typescript-eslint/eslint-plugin": "^4.22.0",
+    "**/@typescript-eslint/parser": "^4.22.0",
```
Andarist, line 31: any particular reason for those resolutions here? 🤔

babel.config.js
```diff
@@ -10,7 +10,7 @@ let isTestFile = filename =>
 module.exports = api => {
   api.cache(true)
   return {
-    presets: ['babel-preset-emotion-dev'],
+    presets: ['babel-preset-emotion-dev', '@babel/preset-typescript'],
```
Andarist, line 13: isn't this conflicting with the flow plugin that is used at the same time?

package.json
```diff
@@ -81,7 +86,9 @@
       "react/prop-types": 0,
       "react/react-in-jsx-scope": 0,
       "standard/computed-property-even-spacing": 0,
-      "@emotion/pkg-renaming": 2
+      "@emotion/pkg-renaming": 2,
+      "no-use-before-define": 0,
+      "@typescript-eslint/no-use-before-define": 1
```
Andarist, line 91: interesting - isn't it one of the goals of TypeScript ESLint project to hide this from consumers? 🤔 I mean - don't they automatically switch to TS-oriented variants of plugins like this?

packages/babel-plugin/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
Andarist, line 1: For the time being, I would leave the existing tsconfig.json files as they are targeting **test** directories and not really the source files. Keeping them together with type tests might actually be a requirement of DTslint or something - but this has to be investigated later.

As to the project itself - I think a single `tsconfig.json` in the root will work great for us. This is how TS is set up in Changesets repo.

package.json
with-heart: `gatsby` has a dependency on `^2.x` for each, and it was causing linting to fail. I can investigate more to see if there's a better way, this was just the quick and easy solution I Google'd to avoid the conflict.

package.json
with-heart: I think we might need to enable the `recommended` rules from `@typescript-eslint/eslint-config` for that. I didn't want to enable it since I wasn't sure how well it'd play together with `standard`. Since we're dropping that, I'll enable the recommended rules and see how that goes.

babel.config.js
with-heart: Probably but I haven't tried to build anything yet. Not sure how to coordinate that part exactly if do conflict. Just remove `flow` and assume we won't merge all of these branches till everything is converted? 

Added because `preconstruct` uses it.

packages/babel-plugin/tsconfig.json
with-heart: Good point. I'll put those back!

babel.config.js
mitchellhamilton: We should use https://babeljs.io/docs/en/options#overrides to make run @babel/preset-typescript on `.ts` and `.tsx` and `@babel/preset-flow` on `.js`. Ideally we should merge these incrementally.

with-heart:
@rjdestigter Yeah, I noticed that as well! I think it has to do with `prettier` being unsure which parser to use. Need to investigate more.

package.json
Andarist: Hm, why does Gatsby depend on this anyway? It shouldn't be concerned with linting by default 🤔

babel.config.js
Andarist: note that `@babel/preset-typescript` already has `overrides` in its implementation: https://github.com/babel/babel/blob/672a58660f0b15691c44582f1f3fdcdac0fa0d2f/packages/babel-preset-typescript/src/index.js#L35 . I'm not completely sure how well this composes with other stuff though.

with-heart:
I made a few changes that moved this forward:

- removed `standard` plugins
- added `import`, `react`, and `@typescript-eslint` as `eslint` plugins
- added an `eslint` files override for `**/packages/**/*.tsx?` that sets the `prettier` parser to `typescript`

This puts us at 38 errors and 61 warnings. The warnings are all for `@typescript-eslint/no-use-before-define`, and the errors are split between the parser being unable to deal with `flow` and `parserOptions.project` confusion for root `*.js` config files and `**/__mocks__/*.js` files. Not sure of the best way to resolve that last issue.

with-heart:
Hoping to take a look at the review comments later today.

package.json
```diff
@@ -60,6 +60,8 @@
       "prettier",
       "flowtype",
       "@emotion",
+      "import",
```
Andarist, line 63: I wonder - was it required for this PR? what problem does it solve here?

package.json
```diff
@@ -110,6 +110,20 @@
...
+      {
+        "files": [
+          "**/packages/**/*.ts",
+          "**/packages/**/*.tsx"
+        ],
+        "rules": {
+          "prettier/prettier": [
+            "error",
+            {
+              "parser": "typescript"
```
Andarist, line 124: since we are going to use Babel for parsing TS maybe it would be better to use `babel-ts` here instead?

with-heart: That sounds fine! We'd need to upgrade `prettier` in order to use it though. Is that fine to include?

package.json
with-heart: It was needed for this override: https://github.com/emotion-js/emotion/blob/main/package.json#L99-L107

But [the docs for that rule state](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-commonjs.md) that it's meant to be temporary for transitioning to pure ES6 modules, so I think it's fine to remove this plugin and the override.

package.json
with-heart: Yeap, `extends:@typescript-eslint/eslint-recommended" removed the need for this.

babel.config.js
with-heart: It doesn't appear to conflict. `build` works fine whether the preset is listed or not.

package.json
with-heart: I removed the `resolutions` and `node_modules` and reinstalled and it seems to be working fine now. I'm not sure what happened there exactly, though `gatsby` fixed that issue in February so maybe I had an outdated version installed.

with-heart:
I made a few more changes this morning. `build` works fine and linting now results in 32 errors related to `@typescript-eslint/eslint-parser` being unable to parse `flow`.

If I add an `overrides` section for `**/packages/**/*.jsx?" with `"parser": "babel-eslint"`, this drops to only 5 linting errors related to `flow`. I don't know enough about `flow` to know how to resolve those though, and I'm not sure if we'd want to include that override temporarily while the conversion to ts happens.

Thoughts?

package.json
Andarist: ye, go for it

Andarist: actually - maybe leave that off for a followup PR

packages/babel-preset-css-prop/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
michaeljaltamirano, line 1: Can't leave a comment on the diff but `packages/babel-plugin-jsx-pragmatic/tsconfig.json` being empty looks to be the cause for [the Flow CI issue](https://github.com/emotion-js/emotion/pull/2360/checks?check_run_id=2504489068).

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "jsx": "react",
+    "lib": ["es6", "dom"],
+    "module": "commonjs",
+    "noEmit": true,
+    "strict": true,
+    "target": "es5"
+  },
+  "include": ["packages", "scripts", "site", "test", "playgrounds"],
+  "exclude": ["node_modules"]
+}
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Well never mind then 😄

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "forceConsistentCasingInFileNames": true,
+    "isolatedModules": true,
+    "jsx": "react",
+    "lib": ["es6", "dom"],
+    "module": "commonjs",
+    "noEmit": true,
+    "strict": true,
```
JoshuaKGoldberg, line 9: Just a suggestion: I get a lot of errors around conflicting types, generally around react-native and the `dom` types. 

```suggestion
    "skipDefaultLibCheck": true,
    "strict": true,
```

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "target": "es5"
+  },
+  "include": ["packages", "scripts", "site", "test", "playgrounds"],
+  "exclude": ["node_modules"]
```
JoshuaKGoldberg, line 13: I've found myself doing this in my own integration branch to get TypeScript to not complain about all the `$ExpectError` dts tests:

```suggestion
  "exclude": ["node_modules", "packages/*/types/asdf*"]
```

JoshuaKGoldberg: I guess, more broadly, what is the strategy for `.d.ts` generation? Should a build step like `build:types` be added that runs `tsc --declaration` into `types/` folders?

package.json
```diff
@@ -264,6 +270,7 @@
     "react-test-renderer": "16.8.6",
     "svg-tag-names": "^1.1.1",
     "through": "^2.3.8",
+    "typescript": "^4.2.4",
```
JoshuaKGoldberg, line 273: I think `prettier` is going to need to be upgraded as well. `prettier@1.14.3` is pretty old and crashes on new TypeScript syntax such as `import type ...`.

```
$ C:\Code\emotion\node_modules\.bin\prettier packages\utils\src\index.ts
[error] packages\utils\src\index.ts: SyntaxError: '=' expected. (1:13)
[error] > 1 | import type { RegisteredCache, EmotionCache, SerializedStyles } from './types'
[error]     |             ^
```

michaeljaltamirano: Could the syntax omit `type` until prettier is upgraded? Or are there other blockers present too.

tsconfig.json
Andarist: > I guess, more broadly, what is the strategy for .d.ts generation? Should a build step like build:types be added that runs tsc --declaration into types/ folders?

This one is covered by Preconstruct which we are using to build all packages. You can check out example output containing TS declarations here:
https://unpkg.com/browse/@changesets/cli@2.14.0/dist/

package.json
Andarist: We could just upgrade Prettier - probably best to do it in a separate, dedicated, PR after removing flow types.

JoshuaKGoldberg: https://github.com/emotion-js/emotion/pull/2387 for reference.

My two cents: a Prettier upgrade is generally a nice thing and doesn't relate to the rest of the TypeScript stuff. Could it come in separately, then get brought into the `ts-migration` as a merge from master?

JoshuaKGoldberg: Merged 🎊

with-heart:
Welp, I've updated the branch but I was being mindless and rebased. My bad for throwing a wrench into things, wasn't really thinking.

JoshuaKGoldberg:
This all looks good either way 😄 , I'll get my branch updated hopefully end of this week!

JoshuaKGoldberg:
This all looks good either way 😄 , I'll get my branch updated hopefully end of this week!

@with-heart, other than updating https://github.com/JoshuaKGoldberg/emotion/pull/1, is there anything you'd like me to do to help? I think once that's updated it would be ready to be merged in here?

packages/babel-plugin-jsx-pragmatic/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
Andarist, line 1: Could u try removing those per-directory tsconfigs? Do they change anything in how things work now? In Changesets, this is all we have: https://github.com/atlassian/changesets/blob/c426035565cfac518238c8bf32f3c496c66c0657/tsconfig.json

Andarist:
@with-heart friendly 🏓 would you mind taking a look at what's up with CI?

with-heart:
I fixed the issue with the `eslint` parser for `prettier` and removed the per-package `tsconfig.json` files. Are you able to re-reun the workflows for me @Andarist?

Andarist (approved):

Andarist:
@with-heart thank you for this! ❤️

Issue #2209, opened 2021-01-11
Missing styles in production (CRA + Emotion 11)


green-arrow:
**Current behavior:**

Successfully updated codebase to emotion 11 (from 9) and using the `css` prop in conjunction with craco to override the webpack config and use the `@emotion/babel-preset-css-prop` preset. When running locally in development, all styles are applied properly and everything looks good. When building for production however, it seems like there are styles missing throughout the app.

**To reproduce:**

I don't have a repro case as this is in app, I'm hoping I've just missed a step or something else obvious.

**Environment information:**

- `react` version: 17.0.1
- `@emotion/react` version: 11.1.4

green-arrow:
To give an example, this image shows elements with generated CSS class names, but nothing showing up in the debugger and the style tags for emotion in the `head` are empty. Although there are some styles throughout that seem to get applied and appear and the debugger, and those point to the empty style tags in the `head`.

I sincerely appreciate any assistance :)

tyom:
Are you using `CacheProvider` from `@emotion/react`? In my case, the use of `CacheProvider` (for custom AMP implementation in Next.js) disabled the automatic SSR injection of Emotion `style` blocks. Removing `CacheProvider` wrapper for WWW renders fixed the issue.

green-arrow:
@tyom - no, I’m not currently using `CacheProvider`. Just `css` from `@emotion/react`, with a single usage from vanilla `@emotion/css` for a special use case.

Andarist:
It looks like if you would load more than one instance of Emotion - the other one touching style elements if the first ine and thjs them losing injected styles. From what I remember those elements dont lose styles in FireFox - if this works for u in FF it would validate my diagnosis. Could u check this?

green-arrow:
@Andarist - unfortunately I get the same error in firefox.

Something I noticed though is that the styles for our header seem to be applied initially for about a second, and then everything goes missing. We do use code-splitting, and this could be correlated but not the cause, but it seems like when bundle for the first page loads in (after the shell loads in), that's when the styles seem to break / get removed. Not sure if it would be possible for the "multiple instances of emotion" to get loaded in that way.

green-arrow:
@Andarist - it looks like you are right with the multiple versions. I can't believe I didn't try this before, but I removed the one use of `import { css as vanillaCss } from '@emotion/css';` from the codebase (which is on that initial page that gets loaded I mentioned), and it's back to working.

[...]

PR #2361, opened 2021-04-28
Fix moving of client-side inserted style tags from Emotion 10 when intending to hydrate Emotion 11 styles resulting in losing styles

danieldelcore:
Closes: #2210 
Related: #2209

**What**:

- `@emotion/cache` currently modifies style tags 'belonging' to other `cacheProviders`.
- `@emotion/cache` also currently searches for style tags in both the head and body and remounts them to the head. 

**Why**:

- To fix issues when running multiple emotion instances & multiple cache providers 

**How**:

- We limited the selector to search for tags only in the body to ensure that the head is not modified (and styles mounted via CSSOM are not lost) 
- We made sure that we only modify elements with a cache key that is relevant to that instance

**Checklist**:

- [x] Documentation N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset

**Reproduction**

https://github.com/seancurtis/emotion-concurrency

https://user-images.githubusercontent.com/3030010/116354642-0debe680-a83c-11eb-8780-744b0c5ed6f0.mov

https://user-images.githubusercontent.com/3030010/116635542-6806bc00-a9a2-11eb-9cd0-de7bc49df9eb.mp4


mitchellhamilton:
I've updated this so it fixes the issue. The issue is very specific to v10 and v11 on the same page, multiple Emotion 11 caches already don't suffer from this problem. (ofc they will suffer from other issues but nothing as problematic as destroying all previously inserted client-side inserted styles from Emotion 10)

> We limited the selector to search for tags only in the body to ensure that the head is not modified (and styles mounted via CSSOM are not lost)

This wouldn't have solved the issue completely because it would still have the issue if someone used Emotion 10 with a custom container that was the body element or inside the body.

> We made sure that we only modify elements with a cache key that is relevant to that instance

That code very intentionally targets all Emotion 11 server rendered style elements(and unintentionally before this change, it targeted client-side inserted Emotion 10 styles which caused the problem you were having) because they have to be moved before the React hydration starts. If we did it on a cache-specific basis and you created a cache inside the render of a React component, the style elements would be moved too late and React would warn and break things, e.g. https://codesandbox.io/s/blissful-fog-r73qb.

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
   cache.sheet.flush()
   expect(document.documentElement).toMatchSnapshot()
 })
+
+test('flushing rehydrated styles in the head only affect styles matching the cache key', () => {
```
Andarist, line 39: I'm wondering - was this supposed to test a fix or is it just a new test for something that was not tested up until now?

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+  const css = `color:hotpink;`
+  const hash = hashString(css)
```
Andarist, line 41: nit: those are not used in the test

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+        >
+          .css-qweqwee{color:red;}
+        </style>
+      </head>
+      <body />
+    </html>
+  `)
+})
+
+test('Should only hydrate style elements matching the cache key', () => {
```
Andarist, line 88: same q as here: https://github.com/emotion-js/emotion/pull/2361/files#r624680916

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+  let css = `color:hotpink;`
+  let hash = hashString(css)
+
+  safeQuerySelector(
+    'body'
+  ).innerHTML = `<style data-emotion="emo ${hash}">.css-${hash}{${css}}</style>`
```
Andarist, line 94:
```suggestion
  ).innerHTML = `<style data-emotion="emo ${hash}">.emo-${hash}{${css}}</style>`
```

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+
+  expect(cache.inserted).toEqual({})
+  expect(document.documentElement).toMatchInlineSnapshot(`
+    <html>
+      <head>
+        <style
+          data-emotion="emo 1lrxbo5"
+          data-s=""
+        >
+          .css-1lrxbo5{color:hotpink;}
```
Andarist, line 106:
```suggestion
          .emo-1lrxbo5{color:hotpink;}
```

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+
+test('Should only hydrate style elements matching the cache key', () => {
+  let css = `color:hotpink;`
+  let hash = hashString(css)
+
+  safeQuerySelector(
+    'body'
+  ).innerHTML = `<style data-emotion="emo ${hash}">.css-${hash}{${css}}</style>`
+
+  const cache = createCache({ key: 'css' })
```
Andarist, line 96: `"css"` is reserved for the default cache and we discourage using a custom cache with such key - so maybe worth using a different key here?

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+  })
+
+  afterEach(() => {
+    process.env.NODE_ENV = 'test'
+  })
+
+  test('Existing head styles from Emotion 10 should not be moved', () => {
+    safeQuerySelector('head').innerHTML = [
+      '<style data-emotion="css-global"></style>',
+      '<style data-emotion="css"></style>',
```
Andarist, line 143: maybe worth adding a comment here that those are specific to Emotion 10 because they do not have `data-s` attribute?

packages/cache/src/index.js
```diff
@@ -87,7 +103,12 @@ let createCache = (options: Options): EmotionCache => {
         const attrib = ((node.getAttribute(`data-emotion`): any): string).split(
           ' '
         )
-        if (attrib[0] !== key) {
+        if (
+          attrib[0] !== key ||
+          // this means we will ignore elements which don't have a space in them which
+          // means that the style elements we're looking at are only Emotion 11 server-rendered style elements
+          attrib.length === 1
+        ) {
```
Andarist, line 111: Instead of this if condition we could leverage the browser capabilities and query for this:
```js
document.querySelectorAll(`[data-emotion^="${key} "]`)
```

packages/cache/__tests__/hydration.js
```diff
@@ -35,3 +35,145 @@ test('rehydrated styles to head can be flushed', () => {
...
+    const css = `color:hotpink;`
+    const hash = hashString(css)
+
+    safeQuerySelector(
+      'body'
+    ).innerHTML = `<style data-emotion="css ${hash}">.css-${hash}{${css}}</style>`
+
+    createCache({ key: 'css' })
+
+    expect(document.documentElement).toMatchInlineSnapshot(`
```
Andarist, line 156: based on this assertion it's rather hard to see if this asserts the correct thing - maybe it would be better to render those Emotion 10 styles to the body? that way we'd know that they have not been moved to the head

Andarist:
> ofc they will suffer from other issues but nothing as problematic as destroying all previously inserted client-side inserted styles from Emotion 10

Do you have any particular problems in mind? We probably should document or fix potential problems.

---

I think the possible permutations of this problem are:

| First  | Second   | Assessment   |
|---|---|---|
| SSRed 10 | init different 10 | I think we can just skip over this since this has been long enough in production that people worked around this |
| client 10 | init different 10|Same as above|
| SSRed 10 | init 11| This is not a problem since the SSR format of Emotion 10 was slightly different (key in the attribute key and not in the value) |
| client 10 | init 11| This is being fixed by this PR |
| SSRed 11 | init different 11|Handled OK thanks to `data-s`|
| client 11 | init different 11|Same as above|
| SSRed 11 | init 10 |This is not a problem since the SSR format of Emotion 10 was slightly different (key in the attribute key and not in the value)|
| client 11 | init 10 |Same as above|

We can't deal with clashing `key`s in any combination and our recommendation is to always use unique keys if there is a chance of a clash (like for example if somebody develops a browser extension or an embeddable widget).

Have I missed something?

danieldelcore:
This is fantastic! Thanks so much for jumping in @mitchellhamilton @Andarist ❤️. 
I'll work on that feedback today 👍

packages/cache/__tests__/hydration.js
danieldelcore: Hey 👋 yeah it was a bit of both, without the fix  for filtering the cache key (`data-emotion="emo"`), both style tags would be removed. AFAIK that's why we were having issues using the CacheProvider before

packages/cache/src/index.js
danieldelcore: Great call-out!

mitchellhamilton:
> Do you have any particular problems in mind? We probably should document or fix potential problems.

What I meant was using the same key is always going to have some problems and they are:
- Using stylisPlugins - you absolutely **MUST** use a different key when using custom stylis plugins because we hash the styles before they go through stylis so without the hashing so without using a different key, you will have different styles 
- Flushing after SSR - it's unclear which cache owns the styles(i believe how it works right now, multiple caches can own it?) so it might be removed
- custom containers after SSR(+prepend after SSR i think?) - things might be moved to the wrong container after SSR because it's unclear which cache owns the styles

For people who don't care about these things(which is the vast majority of people imo), multiple caches with the same key should work fine. I'm not saying people should intentionally use the same cache key but they will probably unintentionally use multiple caches with the same key because different versions of emotion will be run in the same app.

> SSRed 10 | init different 10 | I think we can just skip over this since this has been long enough in production that people worked around this

What's the problem you're referencing here?

packages/cache/__tests__/hydration.js
danieldelcore: Yea this test is a bit of a weird one tbh. It's intended to assert the cause of our original issue which is essentially style tags on the head being re-mounted to the head when a cache provider instantiates itself, resulting in CSSOM styles being lost. 

Since we're not able to test CSSOM styles in jest, I feel like it's probably a good idea just to remove it instead. 
What do you think?  😄

danieldelcore:
Hey folks, just pushed up some changes to address your feedback. LMK if there's anything else you'd like me to look at 👍

MatthewCharlton:
Are there any other reasons why 'css' was used as the default key and not some kind of unique ID generated at runtime?

mitchellhamilton:
> Are there any other reasons why 'css' was used as the default key and not some kind of unique ID generated at runtime?

We need keys to match from a server-render to a client-render.

packages/cache/__tests__/hydration.js
mitchellhamilton: This test doesn't test any changed behavior, the styles in this test are both from server-rendering which was already handled correctly.

packages/cache/__tests__/hydration.js
mitchellhamilton: This is the only test that tests behavior that has changed, I've updated it so it's more clear

Andarist:
> What's the problem you're referencing here that people have worked around(as in specifically this row)?

I'm not aware of any problems there - I've just skipped thinking about those entirely but included them in the table for completeness of the listed combinations 😅

jacek-leoniec:
Any timeline when this will be released?

PR #2385, opened 2021-05-21
Removed Flow

JoshuaKGoldberg:
**What**:

Removes all Flow tooling from source code.

* Comments out most type annotations to prepare them for the next step, TypeScript! 🚀 (https://github.com/emotion-js/emotion/pull/2360).
* Type annotations that don't add anything, like `*` and `<empty>`, are removed altogether.
* `$FlowFixMe`s are removed ~; only one of their comments is kept for not being related to the Flow type checker~

**Why**:

Following the strategy mentioned by @Andarist in https://github.com/emotion-js/emotion/issues/2358#issuecomment-838310438.

**How**:

I did a find-all for strings such as `: ` and `Component<`, then continuously ran `yarn lint`, `yarn build`, and `yarn test` until there were no more errors. Also `yarn test -u` to clean up the auto-generated snapshots.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset 

**Question**: This also removes the Flow docs page, because Flow types are no longer generated. Is that right for this stage? My guess is that they could be generated from TypeScript types later?


Andarist:
> This also removes the Flow docs page, because Flow types are no longer generated. Is that right for this stage? My guess is that they could be generated from TypeScript types later?

That's a good question - I think we can remove them since atm we target the `ts-migration` branch anyway. Something should be decided about this later though.

package.json
```diff
@@ -67,12 +64,11 @@
       "no-template-curly-in-string": 0,
       "prefer-const": 0,
       "no-unused-vars": 0,
-      "flowtype/define-flow-type": 2,
       "import/no-duplicates": 0,
       "prettier/prettier": [
         "error",
         {
-          "parser": "flow"
+          "parser": "typescript-eslint"
```
Andarist, line 71: Interesting - I would have assumed that javascript/typescript/babel-ts parsers would work here. I understand the difference between them but I don't immediately see why there is a need for Prettier's typescript-eslint to exist in the first place. Is this just about making sure what both use the same baseline of features? So if something can be parsed using typescrip-teslint then it can also be parsed using Prettier? I suspect that otherwise we could end up in a situation with a single file being parseable by one without being parseable by the other one.

packages/jest/src/utils.js
```diff
@@ -298,10 +299,10 @@ export function getMediaRules(rules: Array<Object>, media: string): Array<any> {
   )
 }
 
-export function isPrimitive(test: any) {
+export function isPrimitive(test) {
   return test !== Object(test)
 }
 
-export function hasIntersection(left: any[], right: any[]) {
+export function hasIntersection(left /* [] */, right /* [] */) {
```
Andarist, line 306:
```suggestion
export function hasIntersection(left /* any[] */, right /* any[] */) {
```

packages/react/src/theming.js
```diff
@@ -1,13 +1,15 @@
-// @flow
 import * as React from 'react'
 import weakMemoize from '@emotion/weak-memoize'
 import hoistNonReactStatics from './isolated-hoist-non-react-statics-do-not-use-this-in-your-code'
 
-export const ThemeContext = /* #__PURE__ */ React.createContext<Object>({})
+export const ThemeContext = /* #__PURE__ */ React.createContext < Object > {}
```
Andarist, line 5:
```suggestion
export const ThemeContext = /* #__PURE__ */ React.createContext({})
```

site/src/components/Box.js
```diff
@@ -47,8 +47,12 @@ type Props = $Shape<{
   justify: string,
   align: string
 }>
-const Box = styled.div<Props>(
-  display,
+*/
+
+const Box =
```
Andarist, line 52: this went wrong - the generic type should be put in the comment

site/src/components/Title.js
```diff
@@ -1,17 +1,18 @@
...
-export default styled<Props>(markdownComponents.h1)(
-  mq({
-    paddingTop: 0,
-    marginTop: 0,
-    marginBottom: constants.space[2],
-    fontWeight: 700,
-    color: [colors.hightlight, colors.hightlight, colors.color]
-  })
-)
+export default styled <
```
Andarist, line 8: similarly here - the generic param should be commented out

site/src/utils/markdown-styles.js
```diff
@@ -12,9 +11,9 @@ const textStyles = mq({
   lineHeight: '1.7'
 })
 
-export const p = styled.p<empty>(textStyles)
+export const p = styled.p < empty > textStyles
```
Andarist, line 14:
```suggestion
export const p = styled.p(textStyles)
```

site/src/utils/markdown-styles.js
```diff
@@ -25,39 +24,48 @@ export const code = styled.code<empty>(
   })
 )
 
-export const pre = styled.pre<empty>({
-  '& > code': {
-    padding: 0
+export const pre =
```
Andarist, line 27: multiple generic type params were adjusted in a wrong way in this file

site/src/utils/markdown-styles.js
```diff
@@ -90,32 +96,29 @@ export const h2 = styled.h2`
...
 `
 
-// $FlowFixMe(flow@0.100.0): tagged templates don't support generics
 export const h6 = styled.div`
   font-size: 12;
 `
 
-export const a = styled.a<empty>(
+export const a = styled.a(
+  /* <empty> */
```
Andarist, line 121: according to the PR description all `<empty>` should be removed

Andarist (commented):
You have mentioned that:
> $FlowFixMes are removed; only one of their comments is kept for not being related to the Flow type checker

I believe that you are referring to the comment in `packages/react/__tests__/theme-provider.js` - I can't find a reason why this would be kept. I'm not sure why it's there but after this PR lands, it will be completely obsolete anyway.

---

Excellent work 🚀 I love how you have left the flow types as comments for now. As soon as this PR gets ✅ we'll be able to merge.

package.json
JoshuaKGoldberg: Yeah, they're pretty equivalent. For now it's all vanilla JavaScript, which TypeScript supports. I'd think long-term it'd be easier to just have one parser for all of them.

site/src/utils/markdown-styles.js
JoshuaKGoldberg: Ha, I missed some... will do another pass.

site/src/utils/markdown-styles.js
JoshuaKGoldberg: Hmm yeah, I'll revert this back and retry -- and any other page with `< empty` or `empty >`.

site/src/components/Box.js
Andarist: @JoshuaKGoldberg I believe that you have missed this one comment

PR #2387, opened 2021-05-23
chore: bump Prettier and Prettier accessories to 2.x

JoshuaKGoldberg:
**What**:

Bumps the `prettier`-related packages in `package.json` to their latest versions.

**Why**:

Newer versions of Prettier support newer TypeScript syntax things such as template type strings and type imports.

**How**:

`yarn prettier "./*" --write`

Excluded Markdown files because some of them had a lot of noise.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset 

I realize it is a little weird for me to send this kind of touch-many-files PR, as someone who has yet to meaningfully contribute to Emotion. But it's going to be very useful for #2360. No worries if someone else takes it over in a new PR 😄


.prettierrc.yaml
```diff
@@ -1,9 +1,11 @@
+arrowParens: avoid
 semi: false
 singleQuote: true
+trailingComma: none
```
JoshuaKGoldberg, line 4: Personally I agree with the Prettier defaults here, but for the sake of touching fewer files added the overrides to preserve this repo's old behavior. Thoughts?

```suggestion
semi: false
singleQuote: true
```

Andarist: Ye, keeping the output as close to what we have in the repo currently is preferable (although I also agree that we could switch defaults of those 2 later on)

docs/flow.mdx
```diff
@@ -18,14 +18,20 @@ alternatives:
 import styled from '@emotion/styled'
 
 // Option A
-const A = styled<Props>('div')`
+const A =
```
Andarist, line 21: this has been broken - since we haven't dropped Flow on the main branch just yet... would you be willing to fix this?

Andarist: Maybe it would be enough to just revert this single file? I don't think we run Prettier automatically over docs files so it should pass the CI

site/CHANGELOG.md
```diff
@@ -4,9 +4,4 @@
 
 ### Minor Changes
 
-- [8c3fb9d9](https://github.com/emotion-js/emotion/commit/8c3fb9d9) [#1401](https://github.com/emotion-js/emotion/pull/1401) Thanks [@jordanoverbye](https://github.com/jordanoverbye)! - 1. Update document title when the user changes route
-  2. Update markup to me more semantic (nav, aside, main etc instead of just divs)
-  3. Fix overflow grid issue
-  4. Fix broken link on CacheProvider page
-  5. Remove h1 from pageheader, so there is only 1 h1 on all pages
-  6. Add more vertical spacing between nav groups in sidebar
+- [8c3fb9d9](https://github.com/emotion-js/emotion/commit/8c3fb9d9) [#1401](https://github.com/emotion-js/emotion/pull/1401) Thanks [@jordanoverbye](https://github.com/jordanoverbye)! - 1. Update document title when the user changes route 2. Update markup to me more semantic (nav, aside, main etc instead of just divs) 3. Fix overflow grid issue 4. Fix broken link on CacheProvider page 5. Remove h1 from pageheader, so there is only 1 h1 on all pages 6. Add more vertical spacing between nav groups in sidebar
```
Andarist, line 7: the list here was already broken but it's even more so now 😅

docs/flow.mdx
JoshuaKGoldberg: Ah, I thought I'd excluded `.md` & `.mdx` files, sloppy -- reverted them in a6a05c11!

Andarist:
We'll have to merge this into the `ts-migration` branch now, the merge conflicts might be... brutal 😅

Andarist:
@JoshuaKGoldberg I've handled the merge, it wasn't that bad after all - luckily there were no real code changes, just formatting. 

Next step: get this PR merged - https://github.com/emotion-js/emotion/pull/2360

Issue #2423, opened 2021-07-08
React has detected a change in the order of Hooks called by EmotionCssPropInternal


tills13:
**Current behavior:**

When using a conditional `css` which consumes theme, the following React error is produced when the condition changes from false to true: 

```React has detected a change in the order of Hooks called by EmotionCssPropInternal```

**To reproduce:**

https://codesandbox.io/s/emotion-issue-template-forked-ettgx?file=/src/index.js

In particular, the line

```
<Title css={toggle && toggledCss}>
```

when `toggle` changes from `false` to `true`, the error above is produced. It only happens the first time it's toggled, I assume this is because the css or resulting wrapped component is cached somewhere? -- not sure and not sure if relevant. I did some digging in Emotion and found the offending line to be [this one](https://github.com/emotion-js/emotion/blob/23f43ab9f24d44219b0b007a00f4ac681fe8712e/packages/react/src/emotion-element.js#L91). This line also hinted at a potential solution: 

 ```
<Title css={[toggle && toggledCss]}>
```

Now... I'm not really sure which _styles_ of `css` usage are supported. The documentation is lacking in this regard and much of our usage is inferred from TypeScript types. So really, my question is: if this is invalid, how _should_ I write something like this? A less contrived example might be conditionally applying `hover`, `active`, etc states depending on whether or not a button is `disabled`.

**Expected behavior:**

If this style is in fact supported, no React warning should be produced.

**Environment information:**

- `react` version: 16.8.6+
- `@emotion/react` version: 11.0+

Andarist:
Well, your code looks reasonable and valid. We are trying to minimize how many components "subscribe" to the theme context in the implementation and this has backfired here. I wonder - maybe this is a premature optimization anyway? @mitchellhamilton thoughts?

mitchellhamilton:
Yeah, unconditionally using the theme context sgtm

tills13:
I wouldn't mind putting a PR together for this unless either of you would like to handle it yourselves.

mitchellhamilton:
A PR would be great

AlexanderArvidsson:
I just ran into this issue in our platform, except for me, the same problem occurs regardless if I use an array and always on initial render, even if the condition does not change (I tested with a constant).

Quickly looking at the line that @tills13 found, it seems to me that this line breaks the first [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html), namely that hooks should never be called conditionally.

[...]

PR #2424, opened 2021-07-09
unconditionally use ThemeContext in EmotionElement

tills13:
Fixes: https://github.com/emotion-js/emotion/issues/2423

**What**: remove conditional use of ThemeContext for serializing styles.

**Why**: prevents a React error (warning) in the following code: https://codesandbox.io/s/emotion-issue-template-forked-ettgx?file=/src/index.js

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests
- [x] Code complete
- [x] Changeset


tills13:
I figure this would be a major given it's a behaviour change that should be tested by dependants? LMK if you want me to change it to a minor or a patch.

packages/react/src/emotion-element.js
```diff
@@ -88,9 +88,7 @@ let Emotion = /* #__PURE__ */ withEmotionCache<any, any>(
     let serialized = serializeStyles(
       registeredStyles,
       undefined,
-      typeof cssProp === 'function' || Array.isArray(cssProp)
```
Andarist, line 91: Could you also apply the same fix [here](https://github.com/emotion-js/emotion/blob/dcc72d06ace804330fd285a76c8574f3a89001f9/packages/react/src/global.js#L43-L45)? It would also be sweat if you could add some tests for this fix

.changeset/light-swans-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': major
```
Andarist, line 2:
```suggestion
'@emotion/react': patch
```

tills13:
The tests are kind of boring and specific (I typically don't write tests for _the_ behaviour, as opposed testing behaviour changes in different code paths), but they will prevent someone from accidentally reverting the behaviour or thinking they could optimize back to what it was before.

packages/react/__tests__/element.js
```diff
@@ -0,0 +1,126 @@
+// @flow
```
Andarist, line 1: The added tests focus on the implementation details (`serializeStyles` calls etc). It would be IMHO much better to just test what you have reported and what you have reproed in [this codesandbox](https://codesandbox.io/s/emotion-issue-template-forked-ettgx?file=/src/index.js). The test would look like this:
1. render with `css={false}`
2. render with `css={() => {}}`
3. assert that it didn't crash

Andarist:
> I figure this would be a major given it's a behaviour change that should be tested by dependents? LMK if you want me to change it to a minor or a patch.

At the end of the day, this allows more situations to be rendered without crashing and doesn't really impact any situations that were previously rendered OK. So I would just stick with classifying this as a patch bump (a bug fix)

packages/react/__tests__/element.js
tills13: kk thanks for the feedback I will address tonight.

tills13:
@Andarist this is ready to go, I think.

.changeset/light-swans-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+unconditional ThemeContext when serializing styled in EmotionElement
```
Andarist, line 5:
```suggestion

Use theme context when rendering components at all times. This removes a conditional usage of a React hook that could break [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) in some scenarios.
```

PR #2426, opened 2021-07-10
is-prop-valid: use array instead of object

behnammodi:
### package: `is-prop-valid`

- use array instead of object
- remove `Object.keys`

So this PR effect on size and performance


Andarist:
This code is running through a codegen:
https://github.com/emotion-js/emotion/blob/83893978b753ad0638d8fe8f59d78ed835165d48/packages/is-prop-valid/src/index.js#L6
So there is no functional/size/performance difference between both approaches. If I would write this from scratch then I would probably choose your version - however, this is just such a small difference that I prefer keeping the git history cleaner.

Given that - I'm going to close this PR, hope you understand.

behnammodi:
@Andarist thank you for review this PR, keeping git history clean is not matter for another library that uses this package, but for keeping git history we can have a new file like `props.new.js` and `props.old.js`, and also we have different sizes, additionally we removed a heavy function `Object.keys`, as you know less work is better than more work

Andarist:
The point is that there is actually no `Object.keys` call in what we ship to the users - it's compiled away. Take a look at the final file:
https://unpkg.com/@emotion/is-prop-valid@1.1.0/dist/emotion-is-prop-valid.cjs.dev.js

behnammodi:
@Andarist Awesome, I got it 👍🏻

PR #2427, opened 2021-07-12
[memoize] Migrate to TypeScript

sarayourfriend:
**What**:

Migrates the `memoize` package to TypeScript. This is a simple migration really. The main difference from the flow types was just needing to declare a name for the callback function's argument.

**Why**:

Part of #2358 

**How**:

I renamed the index.js file to `index.ts` and then uncommented the Flow types and then updated them to TypeScript.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset 

I _think_ the best way to test this is by running `yarn build memoize` and then verifying that the `dist/declarations/src/index.d.ts` file is correct.


sarayourfriend:
I think I may have messed up the changeset 😞 Is a changeset necessary for each individual migration PR or will we just have a single changeset when the `ts-migration` branch is merged?

packages/memoize/tsconfig.json
```diff
@@ -14,5 +14,5 @@
     "typeRoots": ["../"],
```
Andarist, line 14: I wouldn't move this file into the root of the `memoize` package. This `tsconfig.json` is intended for the type tests in the `types` directory. The `packages/memoize/src` should be controlled by the very root `tsconfig.json` and not the `packages/memoize/tsconfig.json`

Andarist:
> I think I may have messed up the changeset 😞 Is a changeset necessary for each individual migration PR or will we just have a single changeset when the ts-migration branch is merged? @Andarist

Probably best to create a changeset per individual migration PR - you deserve the attribution in the generated (based on those changesets) changelog! 😉

packages/memoize/tsconfig.json
sarayourfriend: Ah gotcha! I'm used to composite projects is why I did that, I'll move it back, thanks!

sarayourfriend: Done in c31cc38

.changeset/hip-moons-play.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/memoize': minor
+---
+
+Migrate to TypeScript and emit declarations based on the code
```
Andarist, line 5:
```suggestion
Source code has been migrated to TypeScript. From now on type declarations will be emitted based on that, instead of being hand-written.
```

PR #2431, opened 2021-07-13
[sheet] Convert to TypeScript

sarayourfriend:
**What**: Converts `sheet` to TypeScript

**Why**: Part of #2358 

**How**: Renamed `index.js` to `index.ts` and then added back type definitons. Most of them were the same as the flow type save some `void -> undefined`. Part of that also changed the signature for `sheetForTag` to return `undefined` (it technically could) so we need to add a null check. I decided to create a utility called `assertDefined` similar to the one used by WordPress's `dom` package: https://github.com/WordPress/gutenberg/blob/af7c07ce630dff9177dd1e25c1459d7e25e836e2/packages/dom/src/utils/assert-is-defined.ts

This allows us to just call `assertDefined` and have a runtime check that the variable is defined. This runtime check is disabled in production to prevent a performance regression.

I also had to add the private `_alreadyInsertedOrderInsensitiveRule` attribute to the StyleSheet class. `_insertTag` is also marked private to match the original .d.ts files.

Also had to update one of the type tests that was erroring out. Not sure why it passed previously as `options` 100% needs to be defined or the constructor for `StyleSheet` will throw a TypeError (it doesn't check at all that options is defined).

Overall I think this PR actually adds some type safety/sanity to `sheet` that flow was missing.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [x] Changeset 

Will depend on changes from #2427 for the jest tests to run by figured I'd get it ready anyway


packages/sheet/src/index.ts
```diff
@@ -62,15 +59,18 @@ function createStyleElement(
...
+  tags: HTMLStyleElement[]
+  container: HTMLElement
+  key: string
+  nonce: string | undefined
+  prepend: boolean | undefined
+  before: Element | null
+
+  private _alreadyInsertedOrderInsensitiveRule: boolean | undefined
+
+  constructor(options: Options) {
```
Andarist, line 73: the previous definition was:
```js
constructor(options?: Options);
```

which was incorrect because `options` are required. 

I'm glad that the whole migration can fix some minor issues like this as isn't done only for the sake of migrating 🎉

Andarist:
@sarayourfriend thank you for all your hard work on those migrations PRs ❤️ I'm sorry that I didn't get to them sooner - which usually is not a problem - but I have had much less time for OSS lately and my backlog has overwhelmed me.

sarayourfriend:
@Andarist No worries, I totally understand! Glad I can help 😄

PR #2432, opened 2021-07-14
[is-prop-valid] Convert to TypeScript

sarayourfriend:
**What**: Converting `is-prop-valid` to TypeScript (mostly)

**Why**: Part of #2358 

**How**: Converted the `index.js` to `index.ts`. `props.js` I think needs to stay as regular JavaScript for `codegen` to work.. at least I couldn't figure out how to configure codegen to work with TypeScript. Maybe @Andarist you'll be able to figure that out if you're more familiar with that plugin.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


packages/is-prop-valid/src/props.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * This module needs to remain pure JavaScript for codegen to work on it
```
Andarist, line 2: this makes sense 👍 making this go through something like `ts-node` would be overengineered, I'm OK with this being a JS file

PR #2441, opened 2021-07-23
expose __unsafe_useEmotionCache from @emotion/react

garronej:
**What**:

Expose `__unsafe_useEmotionCache()` from `@emotion/react`

**Why**:

Needed for seamless integration of [tss-react](https://github.com/garronej/tss-react) with `@emotion/react`.  
[tss-react](https://github.com/garronej/tss-react) will be the new [`makeStyles`](https://material-ui.com/styles/basics/#hook-api) in `@material-ui` v5.

**How**:

I exposed a proxy to `useContext(EmotionCacheContext)`  

**Checklist**:

- [ ] Documentation N.A? Being an interoperability feature, I don't think it has to be documented. But does it? 
- [x] Tests
- [x] Code complete
- [x] Changeset Semantically it should be a minor but tell me if you feel that it should be a patch, I will rebase.

Hi @Andarist,
I implemented `makeStyles` on top of `@emotion/react` as [you discussed](https://github.com/mui-org/material-ui/issues/24109#issuecomment-780855014) with @oliviertassinari.  
It's [tss-react](https://github.com/garronej/tss-react).  
In v5 of material-ui `makeStyle` will be dropped and tss-react [will be mentioned](https://github.com/garronej/tss-react/issues/3#issuecomment-879022879) in the v4 to v5 migration guide for the teams that don't want to make the switch to the styled API.  
It would be great if I could have access to the cache, `tss-react` have just been released and [it has already been requested](https://github.com/garronej/tss-react/issues/6) that it supports custom cache.  
Of course I could expose my own provider but it would be much better if `tss-react` could blend in transparently with `@emotion/react`.

### DIFF
--- a/.changeset/forty-pumas-refuse.md
+++ b/.changeset/forty-pumas-refuse.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-expose \_\_unsafe_useEmotionCache from @emotion/react, see https://github.com/emotion-js/emotion/pull/2441
+Exposed `__unsafe_useEmotionCache` which can be used to access the current Emotion's cache in an easier way than before. Using this might break 0-config SSR and is not recommended to be used unless there you know what you are doing and you are OK with the mentioned downside.
--- a/.changeset/forty-pumas-refuse.md
+++ b/.changeset/forty-pumas-refuse.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+expose \_\_unsafe_useEmotionCache from @emotion/react, see https://github.com/emotion-js/emotion/pull/2441
--- a/packages/react/src/context.js
+++ b/packages/react/src/context.js
@@ -20,6 +20,11 @@ let EmotionCacheContext: React.Context<EmotionCache | null> =
 
 export let CacheProvider = EmotionCacheContext.Provider
 
+export let __unsafe_useEmotionCache =
+  function useEmotionCache(): EmotionCache | null {
+    return useContext(EmotionCacheContext)
+  }
+
 let withEmotionCache = function withEmotionCache<Props, Ref: React.Ref<*>>(
   func: (props: Props, cache: EmotionCache, ref: Ref) => React.Node
 ): React.AbstractComponent<Props> {
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -1,7 +1,11 @@
 // @flow
 import pkg from '../package.json'
 export type { SerializedStyles } from '@emotion/utils'
-export { withEmotionCache, CacheProvider } from './context'
+export {
+  withEmotionCache,
+  CacheProvider,
+  __unsafe_useEmotionCache
+} from './context'
 export { jsx } from './jsx'
 export { jsx as createElement } from './jsx'
 export { Global } from './global'


packages/react/src/index.js
```diff
@@ -1,7 +1,11 @@
 // @flow
 import pkg from '../package.json'
 export type { SerializedStyles } from '@emotion/utils'
-export { withEmotionCache, CacheProvider } from './context'
+export {
+  withEmotionCache,
+  CacheProvider,
+  useEmotionCacheContext
+} from './context'
```
oliviertassinari, line 8: The naming here feels a bit strange:

- Why do we have `CacheProvider ` not mentioning "emotion" but `withEmotionCache` does?
- Why do we have `withEmotionCache` not mentioning "context" but `useEmotionCacheContext` does?

garronej: Agreed 😞 I'm stupid for putting `'Context'`,
However `withEmotionCache` and `CacheProvider` were named this way already. It's in the diff only because prettier rearranged the statement because it was becoming too long for a single line.  
I assume it is out of the question to introduce breaking change for fixing name inconsistency.  
That in mind what would you see as a better name for the hook?  
- `useCache` or
- `useEmotionCache`  
I personally prefer `useEmotionCache`, it is more consistent with `withEmotionCache` with which it shares the same goal of providing access to the cache.

garronej: I have rebased to export `useEmotionCache` instead of `useEmotionCacheContext`

oliviertassinari: Ok, in any case, I'm not a maintainer of this GitHub repository

thieryw:
It would be a goodness if this could be merged!

Andarist:
@mitchellhamilton what are your thoughts about this? It definitely shouldn't be encouraged to use this because there are not a lot of use cases this allows and they are mainly related to generating string class names that break 0-config SSR. OTOH, it's already possible to get access to the cache leveraging the public `withEmotionCache` - it's only way more problematic to do so (which we could treat as intentional). In the vein of being pragmatic over purist, I'm inclined to consent this - I'm pondering though if it shouldn't be exported with some kind of `__unsafe` prefix.

garronej:
Hi @Andarist, 
Thank you for taking the time to have a look.  
I agree with your views. It's a shame to sacrifice 0-config SSR but unfortunately [there is no other ways](https://github.com/mui-org/material-ui/issues/24109#issuecomment-780855014).  
I am encline to use the `__unsafe` prefix, I think it's a good idea, I rebased.

packages/react/src/context.js
```diff
@@ -20,6 +20,11 @@ let EmotionCacheContext: React.Context<EmotionCache | null> =
 
 export let CacheProvider = EmotionCacheContext.Provider
 
+export let __unsafe_useEmotionCache =
+  function __unsafe_useEmotionCache(): EmotionCache | null {
+    return useContext(EmotionCacheContext)
+  }
+
```
Jack-Works, line 27:
```suggestion
export function __unsafe_useEmotionCache(): EmotionCache | null {
    return useContext(EmotionCacheContext)
  }

```

garronej: I know, this is what I would write but I am trying to respect the repo conventions.  

image

Andarist: I wouldn't really bother about the existing convention here. The main reason why `withEmotionCache` is defined like that is because of the fact that it might be reassigned a few lines below - this doesn't apply to the `__unsafe_useEmotionCache`

Andarist:
> Thank you for taking the time to have a look.

No problem, I will still wait on @mitchellhamilton's opinion though, before moving this forward.

mitchellhamilton:
I'm fine with exposing this as `__unsafe_useEmotionCache`

packages/react/src/context.js
garronej: Ah yes I see!  
However I rebased to do it like this:  
image
Because it is meant to be imported like this:  
import { __unsafe_useEmotionCache as useEmotionCache } from "@emotion/react";

oliviertassinari:
Seeing this export make me think about something else related to emotion's cache & React context.
In `@material-ui/styled-engine`, the package has a direct dependency on `@emotion/cache` so it can access `createCache`. However, it seems to come with a few downsides: 

1. All the developers that use Material-UI with styled-components download `@emotion/cache` unless they use an npm package alias. Proof: [install](https://github.com/mui-org/material-ui/blob/next/examples/nextjs-with-styled-components-typescript/README.md) and check the lock file. 
3. The package size reported by bundlephobia is harder to follow, it does show how the dependencies are deduplicated once installed. Proof: https://bundlephobia.com/package/@material-ui/styled-engine@5.0.0-beta.1

Would it make sense to reexport the API of the packages that `@emotion/react` depends on? Now, maybe it doesn't make sense. For instance, for Material-UI, we have talked about making `@emotion/styled` a direct dependency instead of a peer dependency as it has no singletons.
So I guess I wished there we were a single emotion package to install and use, not multiples.

garronej:
> I'm fine with exposing this as `__unsafe_useEmotionCache`

@Andarist Is there any remaining blocking for this PR to be merged? 
I hope I don't came out as being too pushy 😣, I promised emotion cache support to tss-react users weeks ago and I am embarrassed not to have delivered yet.

Andarist:
Sorry for not handling this yet. Gonna go through all pending PRs tomorrow and ship a new version

Jack-Works:
> > I'm fine with exposing this as `__unsafe_useEmotionCache`
> 
> @Andarist Is there any remaining blocking for this PR to be merged?
> I hope I don't came out as being too pushy 😣, I promised emotion cache support to tss-react users weeks ago and I am embarrassed not to have delivered yet.

Don't feel embarrassed! I do not so hurry for that issue to be resolved. Thanks for all of your work!

garronej:
> Sorry for not handling this yet. Gonna go through all pending PRs tomorrow and ship a new version

No need to apologize, I can't imagine the amount of work required to maintain a project as widely used as emotion.  

>  Don't feel embarrassed! I do not so hurry for that issue to be resolved. Thanks for all of your work!  

Thanks for your understanding 😊 You are welcome.

.changeset/forty-pumas-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+expose \_\_unsafe_useEmotionCache from @emotion/react, see https://github.com/emotion-js/emotion/pull/2441
```
Andarist, line 5:
```suggestion
Exposed `__unsafe_useEmotionCache` which can be used to access the current Emotion's cache in an easier way than before. Using this might break 0-config SSR and is not recommended to be used unless there you know what you are doing and you are OK with the mentioned downside.
```

garronej:
Thank you 😊

ciampo:
Hey @garronej and @Andarist , thank you for working on this! Would it be possible to include `__unsafe_useEmotionCache()` in the the TypeScript types for the `@emotion/react` package?

Thank you!

garronej:
> Hey @garronej and @Andarist , thank you for working on this! Would it be possible to include `__unsafe_useEmotionCache()` in the the TypeScript types for the `@emotion/react` package?
> 
> Thank you!

Since the use of this feature is discouraged if not for interoperability, I doubt @Andarist will be encline to add the types definition for it.  
What about using [module augmentation](https://github.com/garronej/tss-react/blob/c15efff74660367dc30a7d1f5c210004c59976e5/src/%40emotion/react.ts#L3-L5) to import it?

ciampo:
> > Hey @garronej and @Andarist , thank you for working on this! Would it be possible to include `__unsafe_useEmotionCache()` in the the TypeScript types for the `@emotion/react` package?
> > Thank you!
> 
> Since the use of this feature is discouraged if not for interoperability, I doubt @Andarist will be encline to add the types definition for it.
> What about using [module augmentation](https://github.com/garronej/tss-react/blob/c15efff74660367dc30a7d1f5c210004c59976e5/src/%40emotion/react.ts#L3-L5) to import it?

Sounds good, thank you for the advice!

PR #2447, opened 2021-08-03
Prevent flush from raising error on detached tag

mdesantis:
It prevents `flush()` from raising the following error: `TypeError: tag.parentNode is null`. Such error happens when the `tag` that is being flushed had already been detached from its parent node by a third party library, which may be application code or a library that handles the page routing—I'm experiencing this issue with [Turbolinks](https://github.com/turbolinks/turbolinks).

**What**: it prevents `flush()` from raising the following error: `TypeError: tag.parentNode is null` when the style tag had been already detached by its parent node

**Why**: because it makes play better with third party code that manipulate the style tags somehow

**How**: it skips the `.removeChild(tag)` call on the tag's parent node when there is no parent node

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests
- [x ] Code complete
- [ ] Changeset N/A 

Should I add a test?

### DIFF
--- a/packages/sheet/src/index.js
+++ b/packages/sheet/src/index.js
@@ -155,7 +155,7 @@ export class StyleSheet {
 
   flush() {
     // $FlowFixMe
-    this.tags.forEach(tag => tag.parentNode.removeChild(tag))
+    this.tags.forEach(tag => tag.parentNode && tag.parentNode.removeChild(tag))
     this.tags = []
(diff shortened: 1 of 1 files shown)


Andarist:
Thank you for the PR - this has been raised in the past. I would like to dig up the old conversations and evaluate this in the upcoming week, thanks for being patient!

PR #2456, opened 2021-08-16
Support typings for Jest v27

billyjanitsch:
**What**:

`@emotion/jest` is currently only compatible with Jest 26 when using TypeScript because it optionally peer-depends on `@types/jest@26`:

https://github.com/emotion-js/emotion/blob/4c7b6de1ff294e45d1e9757dbe37b961cded211b/packages/jest/package.json#L26-L27

This PR widens the peer dependency range from `^26.0.14` to `^26.0.14 || ^27.0.0` such that both 26 and 27 are supported.

**Why**:

See above.

**How**:

See above.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/ninety-bugs-tie.md
+++ b/.changeset/ninety-bugs-tie.md
@@ -1,0 +1,5 @@
+---
+'@emotion/jest': minor
+---
+
+Support typings for Jest v27
--- a/packages/jest/package.json
+++ b/packages/jest/package.json
@@ -24,7 +24,7 @@
     "stylis": "^4.0.3"
   },
   "peerDependencies": {
-    "@types/jest": "^26.0.14",
+    "@types/jest": "^26.0.14 || ^27.0.0",
     "enzyme-to-json": "^3.2.1"
   },
   "peerDependenciesMeta": {


chyzwar:
I am npm user but this would work with npn7? We would also want to see these relesead, @billyjanitsch  if your are busy we caise another pr, rebased with master?

billyjanitsch:
>I am npm user but this would work with npn7?

Yes, this change will remove the peer dependency install error that currently prevents you from running `npm i @emotion/jest@11.3.0 jest@27 @types/jest@27`.

>@billyjanitsch if your are busy we caise another pr, rebased with master?

There's no reason to rebase the branch now. It can be done by a maintainer before merge. This PR is just awaiting review. Please be patient as this project is a volunteer effort.

---

@Andarist whenever you have a moment, could you let me know if you see anything that prevents this from being an easy merge? I don't believe there were any changes to the v27 typings that would affect `@emotion/jest`. But if you have any concerns, I'm happy to look into them to reduce the burden for you.

Andarist:
> This PR is just awaiting review. Please be patient as this project is a volunteer effort.

❤️ 

> I don't believe there were any changes to the v27 typings that would affect @emotion/jest. But if you have any concerns, I'm happy to look into them to reduce the burden for you.

Hard to assess - we don't have a lot of actual dependency on those types so hopefully this shouldn't be problematic. We'll see after a release, in the worst-case scenario we'll have to do some minor adjustments to make it work with both (that's the hope, at least 😅 ).

Issue #1875, opened 2020-05-20
Wrong style rule used


rflopes:
I'm trying to test some styles on my component and despite all tries I can't get the test to pass.

After invetigating a little bit, I found that the way you are getting the component rules is what makes the test to fail.

**Current behavior:**
Check this codesandbox: https://codesandbox.io/s/bold-glitter-h8rkb

When testing a styled component, the jest-emotion matcher validates against the wrong style rule.

**Expected behavior:**
The jest-emotion matcher validates against the correct style rule.

**More information:**
At the Checkbox test, I'm trying to validate that the hover state has the correct border, but as is, the matcher will validate the component style against the `&:checked &:hover` style rule, because it is the last style rule that has the `:hover`.

I debugged the matcher code locally and found the following:
### Selectors for the component (and for the `:hover`state)
```
[
  '.css-ih6inj-StyledCheckbox:hover'
  '.css-ih6inj-StyledCheckbox:checked:hover',
]
```
### Declarations
```
[
  {
    type: 'declaration', // This one is for the first selector `:hover`
    property: 'border-color',
    value: '#7E7E7E',
    position: Position { start: [Object], end: [Object], source: undefined }
  },
  {
    type: 'declaration', // This is for the second selector `:hover:checked`
    property: 'border-color',
    value: '#9E9E9E',
    position: Position { start: [Object], end: [Object], source: undefined }
  }
]
```

### The declaration that was used to validate the component style
```
 {
  type: 'declaration',
  property: 'border-color',
  value: '#9E9E9E',
  position: Position { start: [Object], end: [Object], source: undefined }
 }
```

Which is the last declaration for the `:hover` state despite it is only applied when the component is `:checked` (and the component being unchecked on the test)

If I change the order of the style declaration on the component from:
```
&:hover ...

&:checked &:hover ...
```

to
```
&:checked &:hover ...

&:hover ...
```

the test passes but then if I test the style for the `:hover` when the input is `:checked`it will fail because the matcher will compare against the last declared style for hover.

Andarist:
While I understand that this has been a problem for you and that it's quite easy to get yourself into such a problem when using this matcher against pseudo-classes, I'm not sure if we can classify this as a bug.

The problem is that we don't know in what state a particular element is - in the very same way we don't know that it is in the `:hover` state so you have to tell us that it is. This `target` is not even limited to the root element - it would work for styles such as `& span:hover` - which makes this even more complicated. 

[...]

PR #2457, opened 2021-08-17
add RegExp as typing for target

jcleefw:
**What**:
Following the conversation described in [this issue](https://github.com/emotion-js/emotion/issues/1875), the suggested solution is to insert regex patterns to ensure matching criteria

**Why**:
This quick fix will allow exact matching of pseudo-css selector with `regex patterns`. The existing implementation will rise issues in typesscript type checking where the function is expecting `string` only.

**How**:
Added typing to receive RegExp 

**Checklist**:
- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [ ] Changeset


Andarist:
Is there already a test that covers a regexp being used as `target`?

jcleefw:
> Is there already a test that covers a regexp being used as `target`?

Yes, [here](https://github.com/jcleefw/emotion/blob/b8476e08af4a2e8de94a112cb0daf6e8e4d56fe1/packages/jest/test/matchers.test.js#L133)

jcleefw:
> Is there already a test that covers a regexp being used as `target`?

no. i'm happy to write them, but unfortunately i have difficulty install or building the package. Yarn using yarn2. The readme arent really helpful.

Andarist:
> yes i believe it's here

cool! it's good to have this validated

packages/jest/src/matchers.js
```diff
@@ -43,7 +43,7 @@ function toHaveStyleRule(
   received: *,
   property: *,
   value: *,
-  options?: { target?: string, media?: string } = {}
+  options?: { target?: string | RegExp, media?: string } = {}
```
Andarist, line 46: I suspect that you might have wanted to fix in the TS types which are, currently, located here:
https://github.com/emotion-js/emotion/blob/639b276f4c749e2e50b21d9e50a68d0dbb570cb9/packages/jest/types/index.d.ts#L12

You have only changed the Flow type here - and we are in the process of replacing Flow with TS.

jcleefw: changes made!

PR #2500, opened 2021-10-08
Fix error in alternative JS environments

Jarred-Sumner:
**Use `globalThis` instead of `window || global` when checking if multiple copies of `@emotion/react` are running**

**Why**:

Bun.js, a new JS runtime environment I'm working on - one quirk is it reads from the `"browser"` field in `package.json`. This improves compatibility for many npm packages, but sometimes breaks stuff.

`@emotion/react` is one case where that doesn't work, but the fix is simple and also improves `@emotion/react`'s compatibility with [web workers](https://developer.mozilla.org/en-US/docs/Web/API/Worker) (if there is a usecase for that)

CleanShot 2021-10-07 at 18 39 17@2x

**How**:

[`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is the global context identifier available in all major JavaScript engines newer than Internet Explorer 11. It works in Deno, Node.js 12.0.0+, web workers, web browsers, etc. [caniuse](https://caniuse.com/?search=globalThis)

For maximum compatibility, the code checks `globalThis` is not `undefined` before falling back to either `window` or `global` (as before). 

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests (manually)
- [x] Code complete
- [x] Changeset 

Also I'm not actually sure if a changeset is appropriate here

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -22,8 +22,9 @@ if (process.env.NODE_ENV !== 'production') {
   if (isBrowser && !isJest) {
     // globalThis has wide browser support - https://caniuse.com/?search=globalThis, Node.js 12 and later
     const globalContext =
+      // $FlowIgnore
       typeof globalThis !== 'undefined'
-        ? globalThis
+        ? globalThis // eslint-disable-line no-undef
         : isBrowser
         ? window
         : global


Jarred-Sumner:
> Error:   26:11  error  'globalThis' is not defined  no-undef

Looks like eslint is unaware of `globalThis`. I'll google how to fix that

Issue #2037, opened 2020-10-08
Ability to insert styles at a specific position in <head>


jdb8:
**The problem**

I see that `@emotion/cache` [provides a `container` option](https://emotion.sh/docs/@emotion/cache#container), which seems like it's useful for having some control over the DOM element that emotion inserts its `<style>` nodes into. Unfortunately, I can't find a way to use this for my purpose, which is to control the position of the inserted nodes *in the document head*.

The reason I need this is to have more fine-grained control over rule ordering in a page that has existing styles (which are sadly often composed with the newer Emotion-generated classnames). I'd like Emotion (both on the server and client) to insert all of its `<styles>` (both those initially on the page as well as those dynamically inserted later on) at a specific position in the `<head>`.

Initially I thought I could use the `container` option, but that seems to have two problems here:

* it deals only with nesting *inside* a container, not adjacent to, and...
* it doesn't seem like there are any compatible container elements that accept `<style>` children in the `<head>`

Together, these two issues make it hard to find a way to do what I want here with any existing Emotion apis.

**Proposed solution**

Webpack's `style-loader` [provides an `insert` option](https://webpack.js.org/loaders/style-loader/#insert) which can be given a function to be run in order to determine the position of any inserted styles. Something like this might work here too.

**Alternative solutions**

Another option might be to introduce a more flexible version of `container` - not a full function like `style-loader`, but an `insertAdjacentTo`, `replaceCommentWith` or something similar to allow for more control over the placement?

**Additional context**

Also, based on my problem statement - if there's any existing way to do what I want in Emotion 10, it goes without saying that I'd be very grateful to hear so :)

Andarist:
> it doesn't seem like there are any compatible container elements that accept <style> children in the <head>

That's not quite true - probably some validators might yell at you, but all browsers supports div elements within the head. 

> Also, based on my problem statement - if there's any existing way to do what I want in Emotion 10, it goes without saying that I'd be very grateful to hear so :)

It's not possible with Emotion 10 and we do not plan to introduce such an option. However, Emotion 11 - which is just around the corner - has a new `prepend` option that should fix your problem. Yes - it's a boolean so either your styles are appended or prepended to the container but I don't think much more control is needed over this.

jdb8:
>That's not quite true - probably some validators might yell at you, but all browsers supports div elements within the head.

[...]

PR #2521, opened 2021-10-20
Add `insertionPoint` option in `EmotionCache`

mnajdova:
**What**:

Fixes https://github.com/emotion-js/emotion/issues/2037 (it is closed, but the feature request haven't been implemented before). 

It adds an option to the `EmotionCache` for specifying an insertion point - dom node after which the style rules will be inserted.

**Why**:

It is necessary feature for applications that have third-party styles coming from one place, which should be injected first, then the app's styles coming from emotion, for example, coming from a UI library (as @mui/material) that needs to be injected next, and then any other overrides for the default styles.

**How**:

It's pretty much self explanatory, the style rules are being injected after the specified node. This option wins over `prepend`, if both are being specified, as I would say is more specific.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/sixty-balloons-build.md
+++ b/.changeset/sixty-balloons-build.md
@@ -5,18 +5,19 @@
 
 Add insertionPoint option to the EmotionCache, to insert rules after the specified element.
 
-```
+```jsx
 const head = document.querySelector('head')
 
-const firstStyle = document.createElement('style')
-const lastStyle = document.createElement('style')
-head.appendChild(firstStyle)
-head.appendChild(lastStyle)
+head.innerHTML = `
+  <meta name="emotion-insertion-point" content="" />
+`
 
-// the emotion sheets should be inserted between the first and last style nodes
+const metaTag = document.querySelector("meta[name='emotion-insertion-point']")
+
+// the emotion sheets should be inserted right after the meta tag
 const cache = createCache({
   key: 'my-app',
-  insertionPoint: firstStyle
+  insertionPoint: metaTag
 })
 
 function App() {
--- a/.changeset/sixty-balloons-build.md
+++ b/.changeset/sixty-balloons-build.md
@@ -8,16 +8,15 @@ Add insertionPoint option to the EmotionCache, to insert rules after the specifi
 ```jsx
 const head = document.querySelector('head')
 
-head.innerHTML = `
-  <meta name="emotion-insertion-point" content="" />
-`
+const emotionInsertionPoint = document.createElement('meta')
+emotionInsertionPoint.setAttribute('name', 'emotion-insertion-point')
 
-const metaTag = document.querySelector("meta[name='emotion-insertion-point']")
+head.appendChild(emotionInsertionPoint)
 
 // the emotion sheets should be inserted right after the meta tag
 const cache = createCache({
   key: 'my-app',
-  insertionPoint: metaTag
+  insertionPoint: emotionInsertionPoint
 })
 
 function App() {
--- a/.changeset/sixty-balloons-build.md
+++ b/.changeset/sixty-balloons-build.md
@@ -8,8 +8,10 @@ Add insertionPoint option to the EmotionCache, to insert rules after the specifi
 ```jsx
 const head = document.querySelector('head')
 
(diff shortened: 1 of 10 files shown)


Andarist (approved):

.changeset/sixty-balloons-build.md
```diff
@@ -0,0 +1,31 @@
...
+
+Add insertionPoint option to the EmotionCache, to insert rules after the specified element.
+
+```jsx
+const head = document.querySelector('head')
+
+// <meta name="emotion-insertion-point" content="">
+const emotionInsertionPoint = document.createElement('meta')
+emotionInsertionPoint.setAttribute('name', 'emotion-insertion-point')
+emotionInsertionPoint.setAttribute('content', '')
```
Andarist, line 14: FYI: I've checked the spec and it seems that `content` attribute is required in this case:
> If either name, http-equiv, or itemprop is specified, then the content attribute must also be specified. Otherwise, it must be omitted.

Andarist:
Thank you! ❤️

.changeset/sixty-balloons-build.md
power-f-GOD: Kindly, does this work with SSR, 'cause `document` seems to be undefined in my case? What do you suggest I do?
Thanks.

Andarist: You need to create such a tag (or similar) in your SSR response and then select it on the client's side and pass it as the `insertionPoint` to `createCache`

power-f-GOD: Hi, @Andarist. Thanks for your response. I saw and did as you suggested and it worked. Thank you.

But then, there's another issue which I'm noticing. I'm using Framer Motion's page transition feature--AnimatePresence--with MUI, and it happens that on route change start, the styles of the unmounting page/component gets removed from the DOM causing a FOUC of the unmounting page (before it transitions out).

Is there a way to preserve the styles of the pages as they get mounted, so they don't have to get removed and re-added everytime they get mounted/unmounted. I hope you understand what I mean?

Andarist: What styles are removed on the route change? By default Emotion doesn't unmount any styles - except global ones when used through `<Global/>`. Framer Motion also works in a way that it temporarily preserve components that are unmounted in the tree (they just render exiting children for longer than you are rendering them - they can do that cause they just cache previously received children). So even if some styles would get unmounted during the process they should be unmounted after the component goes completely out of the page so there should be no FOUC.

So while I understand what you are observing there, I don't understand how this can happen. Do you have a repro case (even a complex one, like a full deployed page) that I could take a look at?

power-f-GOD: Thanks for your response(s) again. I do appreciate them.

Ok. I don't have a deployed page yet of the app, as it's not fully ready to be deployed, but I can make a short video and send. Not sure if GitHub accepts accepts videos in the comments.

I'll try or probably send a link.

I'll do that in a moment.

Thank you.

power-f-GOD: What it's supposed to look like...
https://user-images.githubusercontent.com/28790485/152666236-7758fccd-d279-46e4-bc16-6ff7472d4cc7.mov

power-f-GOD: ...versus what it currently looks like in prod:

https://user-images.githubusercontent.com/28790485/152666255-2a1b4684-bfb5-4c2e-96ad-bed96d099bbe.mov

:

power-f-GOD: @Andarist , so you'll notice the FOUC in the production (build) video.

power-f-GOD: Hi, @Andarist . Seen this yet?

Andarist: Could you perhaps record this with https://www.replay.io/ ? I'd like to debug this and can't do that with just a standard recording of the problem.

power-f-GOD: Alright, I will. Thanks for your support. 🤲🏼

PR #2542, opened 2021-11-09
fix(styled): Ensure no hydration mismatch with React.useId

eps1lon:
**What**:

Fixes https://github.com/reactwg/react-18/discussions/111#discussioncomment-1581040

**Why**:

If `React.useId` is used then emotion currently causes hydration mismatches

**How**:

Render same amount of siblings on client and server (`<style />` is replaced with `<React.Fragment />` on the client). 
Pattern is copied from https://github.com/vercel/next.js/pull/31102/files#diff-63632e6e580e69692941c89bab0a0e0436ade3f9543ef610925f795928d9d5fdR579-R612.
Fix is verified in https://codesandbox.io/s/friendly-bassi-6buos?file=/src/App.js:410-429

**Checklist**:

- ~[ ]~ Documentation Don't think this needs documentation
- [x] Tests Unclear if you have a hydration test suite.
   Broken with `@emotion/styled@11.3.0`: https://codesandbox.io/s/sad-breeze-qznnl?file=/src/Sidebar.js
   Fixed with this PR: https://codesandbox.io/s/serene-platform-r4eyr?file=/src/Sidebar.js
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/strange-kids-change.md
+++ b/.changeset/strange-kids-change.md
@@ -1,5 +1,6 @@
 ---
-'@emotion/styled': patch
+'@emotion/react': minor
+'@emotion/styled': minor
 ---
 
-Fix hydration mismatches if `React.useId` is used
+Fixed hydration mismatches if `React.useId` (the upcoming API of the React 18) is used within a tree below our components.
--- a/.changeset/strange-kids-change.md
+++ b/.changeset/strange-kids-change.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fix hydration mismatches if `React.useId` is used
--- a/packages/jest/src/create-enzyme-serializer.js
+++ b/packages/jest/src/create-enzyme-serializer.js
@@ -3,8 +3,63 @@ import type { Options } from './create-serializer'
 import { createSerializer as createEmotionSerializer } from './create-serializer'
 import * as enzymeTickler from './enzyme-tickler'
 import { createSerializer as createEnzymeToJsonSerializer } from 'enzyme-to-json'
+import { isEmotionCssPropElementType, isStyledElementType } from './utils'
 
-const enzymeSerializer = createEnzymeToJsonSerializer({})
+const enzymeToJsonSerializer = createEnzymeToJsonSerializer({
+  map: json => {
+    if (typeof json.node.type === 'string') {
+      return json
+    }
+    const isRealStyled = json.node.type.__emotion_real === json.node.type
+    if (isRealStyled) {
+      return {
+        ...json,
+        children: json.children.slice(-1)
+      }
+    }
+    return json
+  }
+})
+
+// this is a hack, leveraging the internal/implementation knowledge about the enzyme's ShallowWrapper
+// there is no sane way to get this information otherwise though
+const getUnrenderedElement = shallowWrapper => {
+  const symbols = Object.getOwnPropertySymbols(shallowWrapper)
+  const elementValues = symbols.filter(sym => {
+    const val = shallowWrapper[sym]
+    return !!val && val.$$typeof === Symbol.for('react.element')
+  })
+  if (elementValues.length !== 1) {
+    throw new Error(
+      "Could not get unrendered element reliably from the Enzyme's ShallowWrapper. This is a bug in Emotion - please open an issue with repro steps included:\n" +
+        'https://github.com/emotion-js/emotion/issues/new?assignees=&labels=bug%2C+needs+triage&template=--bug-report.md&title='
+    )
+  }
+  return shallowWrapper[elementValues[0]]
+}
+
+const wrappedEnzymeSerializer = {
+  test: enzymeToJsonSerializer.test,
+  print: (enzymeWrapper, printer) => {
+    const isShallow = !!enzymeWrapper.dive
+
+    if (isShallow) {
+      const unrendered = getUnrenderedElement(enzymeWrapper)
+      if (
+        (isEmotionCssPropElementType(unrendered) ||
+          isStyledElementType(unrendered)) &&
+        enzymeWrapper.getElement().type === Symbol.for('react.fragment')
+      ) {
+        return enzymeToJsonSerializer.print(
+          enzymeWrapper.children().last(),
+          printer
+        )
+      }
+    }
+
+    return enzymeToJsonSerializer.print(enzymeWrapper, printer)
+  }
+}
 
 export function createEnzymeSerializer({
   classNameReplacer,
@@ -16,7 +71,7 @@ export function createEnzymeSerializer({
   })
   return {
     test(node: *) {
-      return enzymeSerializer.test(node) || emotionSerializer.test(node)
+      return wrappedEnzymeSerializer.test(node) || emotionSerializer.test(node)
     },
     serialize(
       node: *,
@@ -26,9 +81,9 @@ export function createEnzymeSerializer({
       refs: *,
       printer: Function
     ) {
-      if (enzymeSerializer.test(node)) {
+      if (wrappedEnzymeSerializer.test(node)) {
         const tickled = enzymeTickler.tickle(node)
-        return enzymeSerializer.print(
+        return wrappedEnzymeSerializer.print(
           tickled,
           // https://github.com/facebook/jest/blob/470ef2d29c576d6a10de344ec25d5a855f02d519/packages/pretty-format/src/index.ts#L281
           valChild => printer(valChild, config, indentation, depth, refs)
--- a/packages/jest/src/create-enzyme-serializer.js
+++ b/packages/jest/src/create-enzyme-serializer.js
@@ -3,7 +3,11 @@ import type { Options } from './create-serializer'
 import { createSerializer as createEmotionSerializer } from './create-serializer'
 import * as enzymeTickler from './enzyme-tickler'
 import { createSerializer as createEnzymeToJsonSerializer } from 'enzyme-to-json'
-import { isEmotionCssPropElementType, isStyledElementType } from './utils'
+import {
+  isEmotionCssPropElementType,
+  isStyledElementType,
+  unwrapFromPotentialFragment
+} from './utils'
 
 const enzymeToJsonSerializer = createEnzymeToJsonSerializer({
   map: json => {
@@ -43,15 +47,14 @@ const wrappedEnzymeSerializer = {
   print: (enzymeWrapper, printer) => {
     const isShallow = !!enzymeWrapper.dive
 
-    if (isShallow) {
+    if (isShallow && enzymeWrapper.root() === enzymeWrapper) {
       const unrendered = getUnrenderedElement(enzymeWrapper)
       if (
-        (isEmotionCssPropElementType(unrendered) ||
-          isStyledElementType(unrendered)) &&
-        enzymeWrapper.getElement().type === Symbol.for('react.fragment')
+        isEmotionCssPropElementType(unrendered) ||
+        isStyledElementType(unrendered)
       ) {
         return enzymeToJsonSerializer.print(
-          enzymeWrapper.children().last(),
+          unwrapFromPotentialFragment(enzymeWrapper),
           printer
         )
       }
--- a/packages/jest/src/utils.js
+++ b/packages/jest/src/utils.js
@@ -62,7 +62,7 @@ function getClassNamesFromEnzyme(selectors, nodeWithFragment) {
   // TODO: Does approach work with older Emotion versions? Does it cause false positives with components named "Fragment" that aren't `React.Fragment`?
   const node =
     nodeWithFragment.name() === 'Fragment'
-      ? nodeWithFragment.children()
+      ? nodeWithFragment.children().at(1)
       : nodeWithFragment
   // We need to dive in to get the className if we have a styled element from a shallow render
   const isShallow = shouldDive(node)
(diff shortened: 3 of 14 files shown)


Andarist:
Thank you for taking a look at this - since your comment last week I was planning to get to it but you beat me to it :) I was actually thinking if adding a fake element like this would fix the issue - I'm glad that we are able to resolve this without ditching the zero-config approach 👍

packages/jest/src/utils.js
```diff
@@ -58,7 +58,12 @@ function getClassNameProp(node) {
   return (node && node.prop('className')) || ''
 }
 
-function getClassNamesFromEnzyme(selectors, node) {
+function getClassNamesFromEnzyme(selectors, nodeWithFragment) {
+  // TODO: Does approach work with older Emotion versions? Does it cause false positives with components named "Fragment" that aren't `React.Fragment`?
+  const node =
+    nodeWithFragment.name() === 'Fragment'
+      ? nodeWithFragment.children()
+      : nodeWithFragment
```
Andarist, line 66: I'll take a look into this issue and resolve this TODO comment before merging this

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
```diff
@@ -446,11 +446,13 @@ exports[`enzyme shallow displayName 1`] = `
 `;
 
 exports[`enzyme shallow empty styled 1`] = `
-<div
-  className="emotion-0 emotion-1"
->
-  Hello
-</div>
+<Fragment>
```
Andarist, line 449: I wonder if this shouldn't render `<>` but I'm not sure at the moment if this is under our control, gonna check this out

packages/styled/src/base.js
```diff
@@ -132,27 +132,31 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
         newProps.ref = ref
 
         const ele = React.createElement(finalTag, newProps)
+        let possiblyStyleElement = <></>
```
Andarist, line 135: Do u happen to know if there is any potential perf difference between using a fragment like this and using a `Noop` component like in the referenced Next.js PR? The intuition tells me that if there is any then probably the fragment approach is better but it probably also doesn't matter

packages/styled/src/base.js
```diff
@@ -132,27 +132,31 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
+              {...{
+                [`data-emotion`]: `${cache.key} ${serializedNames}`,
+                dangerouslySetInnerHTML: { __html: rules },
+                nonce: cache.sheet.nonce
+              }}
+            />
           )
         }
-        return ele
+        // Need to return the same number of siblings or else `React.useId` will cause hydration mismatches.
```
Andarist, line 153: the same fix, most likely, should also be applied in those places:
- https://github.com/emotion-js/emotion/blob/516fe458058c9ec8218740472b301e935801ebbc/packages/react/src/class-names.js#L142
- https://github.com/emotion-js/emotion/blob/516fe458058c9ec8218740472b301e935801ebbc/packages/react/src/emotion-element.js#L144
- https://github.com/emotion-js/emotion/blob/516fe458058c9ec8218740472b301e935801ebbc/packages/react/src/global.js#L137 (maybe this one can be skipped as `<Global/>` renders either `null` or `<style/>`

packages/styled/src/base.js
eps1lon: I noticed this as well when going over the referenced PR again. Maybe Fragments are less robust if they're implemented differently in various JSX runtimes. A no-op (`() => null`) is probably safer.

But even if there's no difference, let's follow nextjs (first come, first serve) to establish a de-facto standard.

packages/styled/src/base.js
```diff
@@ -132,27 +132,31 @@ let createStyled: CreateStyled = (tag: any, options?: StyledOptions) => {
...
+        return (
+          <>
+            {possiblyStyleElement}
+            {ele}
+          </>
```
Andarist, line 158: @mitchellhamilton should we even render those style elements for compat caches? I mean - right now we need to because `extractCritical` "analyzes" the rendered HTML. Is this mainly to handle the default cache of the vanilla Emotion?

With `@emotion/react` (and `@emotion/styled`) each render gets its unique cache by default (well, each top emotion-aware component gets one by default but let's skip over that detail for now). I wonder if in this context we shouldn't skip rendering those style elements altogether (presumably we'll only be able to do that with the new SSR APIs that we've been talking about)? Basically, when using any of our SSR APIs we can expect a custom cache to be used (when using `@emotion/react`) and we should be able to get access to that cache and just grab its content (instead of retrieving that from the rendered HTML). Or am I missing something?

I think the goal with the new SSR APIs is to rely roughly on a similar approach as outlined above, right? We can't fully backport this to the old APIs (they will simply get dropped in the next major version anyway) but I wonder if there is any reason why this has not been done closer to this approach from the start.

.changeset/strange-kids-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fix hydration mismatches if `React.useId` is used
```
Andarist, line 5: I wonder - would you be open to creating a PR targeting this PR that would add tests for this (this would require bumping React version)? I would love to have a way to verify changes like this but I'm also hesitant to land React 18 on our main branch before it hits its stable mark.

eps1lon: A robust hydration test suite is pretty involved since you really need two separated processes: one for SSR and one for hydrating. 

> I would love to have a way to verify changes like this but I'm also hesitant to land React 18 on our main branch before it hits its stable mark.

The linked codesandbox in the issue description should be sufficient for manual verification. It has a hydration mismatch with the latest version on npm but none with the codesandbox build from this PR:

Broken with `@emotion/styled@11.3.0`: https://codesandbox.io/s/sad-breeze-qznnl?file=/src/Sidebar.js
Fixed with this PR: https://codesandbox.io/s/serene-platform-r4eyr?file=/src/Sidebar.js

packages/styled/src/base.js
eps1lon: `Global` seems safe. Fixed class-names and emotion-element.

.changeset/strange-kids-change.md
Andarist: > A robust hydration test suite is pretty involved since you really need two separated processes: one for SSR and one for hydrating.

Ye, I'm not for something like that now. It's possible to just mock whatever `isBrowser` checks before server and client renders to get a good representation of the real situation. For example [this PR](https://github.com/emotion-js/emotion/pull/2334) has introduced such tests [here](https://github.com/emotion-js/emotion/pull/2334/files#diff-8a41a3bb909094190325495242aab95bc553c6769d9060da0f18c4d3fafcc86d) (IIRC)

eps1lon: Ah, thanks for the pointers. Will look into it.

> I would love to have a way to verify changes like this but I'm also hesitant to land React 18 on our main branch before it hits its stable mark.

I can install it behind an alias and let jest use it for specific tests.

Andarist: Oh, that would be cool - I have never used aliases like this. Hopefully, this won't be too problematic ;p Having that in place would be awesome ❤️

eps1lon: Test illustrating the behavior without this change: [`862729d` (#2542)](https://github.com/emotion-js/emotion/pull/2542/commits/862729d103ebd78791198191201c6f42a9cd356f)

packages/react/__tests__/rehydration.js
```diff
@@ -592,3 +594,73 @@ test('duplicated global styles can be removed safely after rehydrating HTML SSRe
...
+    jest
+      .mock('react', () => {
+        return jest.requireActual('react18')
+      })
+      .mock('react-dom', () => {
+        return jest.requireActual('react18-dom')
+      })
+      .mock('react-dom/server', () => {
+        return jest.requireActual('react18-dom/server')
+      })
```
Andarist, line 610: neat

packages/styled/src/base.js
mitchellhamilton: > should we even render those style elements for compat caches?

I'm pretty sure we don't right now. We retrieve the content from the cache, not the html (we look at the html to see _what_ is used because it was originally made from the case where you're just using a single cache for multiple renders, we don't look at the html for the content of the styles though)

Andarist: Yep - this is kinda what I've meant. That with the new API we wouldn't have to look at the rendered HTML because we could rely on caches being exclusive to renders.

packages/jest/src/utils.js
Andarist: @eps1lon i've tweaked this and left the comment about the chosen solution, what do you think - is relying on `Symbol.for` OK here or should we add `react-is` as a dependency and use that?

eps1lon: `react-is` comes with its own problems (e.g. if it's a dependency people might have a different version of React installed. See https://github.com/facebook/react/issues/20099 for more details).

I'd say pick the one that you're most comfortable with and see if it works for the scenarios that it's used with.

Andarist: thanks for the feedback, I've noticed later that we are actually already using `Symbol.for` for different checks, so I'm just going to stick with this strategy

packages/jest/test/__snapshots__/react-enzyme.test.js.snap
Andarist: I've fixed the serialization tests here - but I have a feeling that our test suite is missing some cases that should currently fail. I will do some further investigation soon.

packages/jest/src/create-enzyme-serializer.js
```diff
@@ -3,8 +3,66 @@ import type { Options } from './create-serializer'
...
 import { createSerializer as createEnzymeToJsonSerializer } from 'enzyme-to-json'
+import {
+  isEmotionCssPropElementType,
+  isStyledElementType,
+  unwrapFromPotentialFragment
+} from './utils'
 
-const enzymeSerializer = createEnzymeToJsonSerializer({})
+const enzymeToJsonSerializer = createEnzymeToJsonSerializer({
+  map: json => {
```
Andarist, line 13: Note for me, we probably should try to map this in our implementation - just how we do it for CSS prop elements. I've always got it working with this diff:
<details>
<summary>diff</summary>

```diff
diff --git a/packages/jest/src/create-enzyme-serializer.js b/packages/jest/src/create-enzyme-serializer.js
index 52ec73be..47787dbe 100644
--- a/packages/jest/src/create-enzyme-serializer.js
+++ b/packages/jest/src/create-enzyme-serializer.js
@@ -9,21 +9,7 @@ import {
   unwrapFromPotentialFragment
 } from './utils'
 
-const enzymeToJsonSerializer = createEnzymeToJsonSerializer({
-  map: json => {
-    if (typeof json.node.type === 'string') {
-      return json
-    }
-    const isRealStyled = json.node.type.__emotion_real === json.node.type
-    if (isRealStyled) {
-      return {
-        ...json,
-        children: json.children.slice(-1)
-      }
-    }
-    return json
-  }
-})
+const enzymeToJsonSerializer = createEnzymeToJsonSerializer()
 
 // this is a hack, leveraging the internal/implementation knowledge about the enzyme's ShallowWrapper
 // there is no sane way to get this information otherwise though
diff --git a/packages/jest/src/create-serializer.js b/packages/jest/src/create-serializer.js
index 75b984c3..a89b5cce 100644
--- a/packages/jest/src/create-serializer.js
+++ b/packages/jest/src/create-serializer.js
@@ -13,7 +13,9 @@ import {
   getKeys,
   flatMap,
   isPrimitive,
-  hasIntersection
+  hasIntersection,
+  isStyledElementType,
+  isStyledComponentType
 } from './utils'
 
 function getNodes(node, nodes = []) {
@@ -149,6 +151,12 @@ const createConvertEmotionElements = (keys: string[]) => (node: any) => {
       type: node.props.__EMOTION_TYPE_PLEASE_DO_NOT_USE__
     }
   }
+  if (isStyledComponentType(node.node)) {
+    return {
+      ...node,
+      children: node.children.slice(-1)
+    }
+  }
   if (isReactElement(node)) {
     return copyProps({}, node)
   }
diff --git a/packages/jest/src/utils.js b/packages/jest/src/utils.js
index 6eb7dbd1..72f655d1 100644
--- a/packages/jest/src/utils.js
+++ b/packages/jest/src/utils.js
@@ -98,12 +98,17 @@ export function isEmotionCssPropElementType(val: any): boolean {
   )
 }
 
+export function isStyledComponentType(val: any): boolean {
+  if (!val) return false
+  const { type } = val
+  return type.__emotion_real === type
+}
+
 export function isStyledElementType(val: any): boolean {
   if (val.$$typeof !== Symbol.for('react.element')) {
     return false
   }
-  const { type } = val
-  return type.__emotion_real === type
+  return isStyledComponentType(val)
 }
 
 export function isEmotionCssPropEnzymeElement(val: any): boolean {

```
</details>

The only thing missing here is that this should flatten multiple (two?) levels of our "wrappers". After unwrapping the CSS prop element we should also try to unwrap the styled element.

I decided that I don't care about this that much now though, the current solution is working fine so I prefer to land this sooner than later and maybe revisit this in the future.

PR #2543, opened 2021-11-09
[serialize]: Convert to Typescript

danilofuchs:
**What**:

Migrates the `serialize` package to Typescript. This was not that simple as the internal functions deal with a large variety of possible types for each argument (`Interpolation`), and some were undocumented.

**Why**:

Part of #2358 

**How**:

- Convert Flow type annotations
- Copy .d.ts definitions to main file
- Fix whatever breaks

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


danilofuchs:
Note: This isn't finished just yet, some tests are still breaking and it's missing the changeset

packages/serialize/src/index.ts
```diff
@@ -203,30 +283,39 @@ function handleInterpolation(
...
-              `css\`${replaced}\``
+            `\`keyframes\` output got interpolated into plain string, please wrap it with \`css\`.
+
+Instead of doing this:
+
+${[...matched, `\`${replaced}\``].join('\n')}
+
+You should wrap it with \`css\` like this:
+
+css\`${replaced}\``
```
danilofuchs, line 294: I rewrote this string interpolation as semantic coloring broke on VSCode due to the complex ' and ` usage
Screen Shot 2021-11-08 at 17 10 03

packages/serialize/src/index.ts
```diff
@@ -296,34 +389,33 @@ function createStringFromObject(
...
+  registered: RegisteredCache,
+  mergedProps?: Props
+): SerializedStyles {
   if (
     args.length === 1 &&
-    typeof args[0] === 'object' &&
-    args[0] !== null &&
-    args[0].styles !== undefined
+    !Array.isArray(args[0]) &&
+    isSerializedStyles(args[0])
```
Andarist, line 410: Like you have mentioned - this package is quite specific as it might be hard for TS to understand all of the type refinements etc.

I would prefer for this PR to stick away from runtime changes as much as possible (if something is absolutely required then I'm open to considering it). This is a very hot path and I don't want to retest performance after those changes :wink:

In cases like this, it should be OK to just use type assertion within the given block, without introducing type predicates.

packages/serialize/src/index.ts
```diff
@@ -345,7 +437,7 @@ export const serializeStyles = function (
   }
   let sourceMap
 
-  if (process.env.NODE_ENV !== 'production') {
+  if (process.env.NODE_ENV !== 'production' && sourceMapPattern !== undefined) {
     styles = styles.replace(sourceMapPattern, match => {
```
Andarist, line 441: Would this work?
```suggestion
  if (process.env.NODE_ENV !== 'production') {
    styles = styles.replace(sourceMapPattern!, match => {
```

packages/serialize/src/index.ts
```diff
@@ -235,8 +324,12 @@ function createStringFromObject(
...
   } else {
     for (let key in obj) {
-      let value = obj[key]
-      if (typeof value !== 'object') {
+      let value = (obj as any)[key] as Interpolation<Props>
+      if (
+        typeof value === 'string' ||
+        typeof value === 'number' ||
+        typeof value === 'symbol'
+      ) {
```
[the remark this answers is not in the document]
Andarist: Similarly here -> this should just be casted within the `if` block

packages/serialize/src/index.ts
```diff
@@ -203,30 +283,39 @@ function handleInterpolation(
...
+  const cached = registered[asString]
+  return cached !== undefined ? cached : asString
 }
 
-function createStringFromObject(
-  mergedProps /*: void | Object */,
-  registered /*: RegisteredCache | void */,
-  obj /*: { [key: string]: Interpolation } */
-) /*: string */ {
+function createStringFromObject<Props>(
```
Andarist, line 310: this is an internal function - i wonder if this needs to accept a generic, couldn't we get away with something simpler here? we could try to treat `Props` as `unknown` here, this should still allow us to accept a type parameterized with a generic, while using `ArrayInterpolation<unknown>` (and similar) "internally".

packages/serialize/src/index.ts
```diff
@@ -203,30 +283,39 @@ function handleInterpolation(
...
+
+css\`${replaced}\``
           )
         }
       }
       break
   }
 
   // finalize string values (regular strings and functions interpolated into css calls)
+  const asString = interpolation as string
```
Andarist, line 302: casts like this are OK 👍

packages/utils/src/types.ts
```diff
@@ -9,6 +9,7 @@ export type SerializedStyles = {
   styles: string
   map?: string
   next?: SerializedStyles
+  toString: () => string
```
Andarist, line 12: could you point me to a line that required this change?

packages/serialize/src/index.ts
danilofuchs: Sure!
I came to the same conclusion after tests started breaking, I will revert these changes

packages/serialize/src/index.ts
danilofuchs: We could do it, sure.
In my projects I avoid it by default as it could lead to a runtime error if misused. But here it is well contained so I see no problem

packages/serialize/src/index.ts
danilofuchs: Ok!
We don't actually manipulate the `Props` type internally so it should not affect anything

packages/utils/src/types.ts
danilofuchs: https://github.com/emotion-js/emotion/pull/2543/files#diff-ec2de171b3fd05a04be89b82f3a9f0beb0768838544583edc37162fcf9762c62R201

Andarist: maybe it would be enough to change that call-site to `String(interpolation)`?

danilofuchs:
Hi @Andarist, I've updated the PR with the requested changes. I made sure no runtime code changes were made.

However, some tests are now breaking. Apparently `Interpolation` could also be a `Symbol`. When adding it to the interoplated string, if throws a TypeError. Specifically, `Symbol(react.forward_ref)`.

Perhaps something in the old Flow build output is now different, causing this unexpected bug. Either way, the code seems to not expect a `Symbol` value at all, and will break if called with it.

Should we apply a patch on the main branch before migrating to TS?

packages/serialize/src/index.ts
```diff
@@ -107,41 +166,43 @@ if (process.env.NODE_ENV !== 'production') {
...
     if (
       process.env.NODE_ENV !== 'production' &&
-      interpolation.toString() === 'NO_COMPONENT_SELECTOR'
+      String(componentSelector) === 'NO_COMPONENT_SELECTOR'
     ) {
       throw new Error(
         'Component selectors can only be used in conjunction with @emotion/babel-plugin.'
       )
     }
-    return interpolation
```
Andarist, line 126: This missing return is causing the errors mentioned here: https://github.com/emotion-js/emotion/pull/2543#issuecomment-968197924

danilofuchs: You are right! I removed it accidentally.

It was implicitly being converted to a String, so I added an explicit cast.

packages/serialize/src/index.ts
```diff
@@ -370,6 +453,7 @@ export const serializeStyles = function (
       styles,
       map: sourceMap,
       next: cursor,
+      // @ts-ignore: SerializedStyles does not have toString method, and we don't want to add it
```
danilofuchs, line 456: I used the same comment from the original Flow code

Andarist:
@danilofuchs the code looks OK (i've pushed some additional, minor, changes but they were mostly nits) but some type tests are failing. I've looked briefly into failing tests for `@emotion/serialize` and I'm pretty sure that this is caused by older TS versions being tripped on `import/export type` syntax.

The problem is that we need this syntax because we compile stuff with `isolatedModules`. I've tried changing the minimal TS version header for this package but it turns out that we are using very old version of `dtslint` and it only supports up to TS 3.4 and that syntax has been introduced in 3.8 🙈 

I recall that I've tried to upgrade dtslint like a year ago or something and I've bumped into some problems, didn't have enough time to dig into this back then but I guess I might have to do that after all.

PR #2552, opened 2021-11-14
SSR Client Docs Incorrect

ashblue:
It looks like the client SSR documentation for the advanced setup was missing the key set for the cache. I've added this in so future users don't get stuck with a giant mystery error while following the guide (took me an hour to debug).

**What**:

Added missing documentation cache key for client side SSR.

**Why**:

The current example triggers a key not found error and breaks on the client-side.

**How**:

By using the current example

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code completeN/A
- [ ] Changeset N/A

### DIFF
--- a/docs/ssr.mdx
+++ b/docs/ssr.mdx
@@ -120,7 +120,11 @@ res
 
 ```jsx
 // Hydration of the ids in `data-emotion-css` will automatically occur when the cache is created
-const cache = createCache()
+const cache = createCache({ 
+  // The key must match your corresponding cache name set earlier
+  key: 'custom'
+})
+
 ReactDOM.hydrate(
   <CacheProvider value={cache}>
     <App />


docs/ssr.mdx
```diff
@@ -120,7 +120,11 @@ res
 
 ```jsx
 // Hydration of the ids in `data-emotion-css` will automatically occur when the cache is created
-const cache = createCache()
+const cache = createCache({ 
+  // The key must match your corresponding cache name set earlier
```
Andarist, line 124: what do you mean by "corresponding cache name set earlier"?

docs/ssr.mdx
```diff
@@ -120,7 +120,11 @@ res
 
 ```jsx
 // Hydration of the ids in `data-emotion-css` will automatically occur when the cache is created
```
Andarist, line 122: It's unrelated change but this part is also slightly incorrect and could be fixed together with the missing `key`
```suggestion
// Hydration of the ids in `data-emotion` will automatically occur when the cache is created
```

Andarist:
Could you tell me how the problem has manifested for you? What kind of an error have you seen or what kind of a broken behavior have you experienced? Maybe we could make this more obvious somehow. When I think about it... we already have an error in place for cases when you forget to pass the `key` explicitly:
https://github.com/emotion-js/emotion/blob/2bac69b6c058d007cfc190e94490c0cb225ca40c/packages/cache/src/index.js#L49-L54
Have you seen this error?

docs/ssr.mdx
ashblue: In the example docs above the key was set on line 87 - 88 to `const key = 'custom';`. So it has to match whatever somebody set to that. I would assume you could end up with multiple keys for whatever reason, but I could be wrong.

docs/ssr.mdx
ashblue: Will I be able to squash (safe to quick merge that in) or should I amend my commit with that?

Andarist: I will squash this when merging to the main branch, it's fine to just commit this using the GitHub's UI button

ashblue:
> Could you tell me how the problem has manifested for you? What kind of an error have you seen or what kind of a broken behavior have you experienced? Maybe we could make this more obvious somehow. When I think about it... we already have an error in place for cases when you forget to pass the `key` explicitly:
> 
> https://github.com/emotion-js/emotion/blob/2bac69b6c058d007cfc190e94490c0cb225ca40c/packages/cache/src/index.js#L49-L54
> 
> 
> Have you seen this error?

That would be a much better error than I'm seeing with the missing key.

image

ashblue:
@Andarist I think this is good to go. If you want me to I can break off https://github.com/emotion-js/emotion/pull/2552#issuecomment-970440058 into a separate ticket.

PR #2555, opened 2021-11-19
TS migration

emmatown:
Just wanted a PR open to more easily see the progress of this


Andarist:
I see that you have done a little bit more than just checking out the progress 😉 Could we move those changes to a separate PR that would target this? That way we could review this more easily but we could also create a changeset that would point to that other PR (I feel like we should **merge** this PR at the end, instead of squashing everything into a single commit).

Andarist:
We might be currently blocked on the `dtslint` update, see [here](https://github.com/emotion-js/emotion/pull/2543#issuecomment-974501224). I also mention in that comment that we might have to require at least TS 3.8 after this migration. It doesn't feel ideal but this version has been released 2 years ago - since TS doesn't follow semver I feel like it's OK for us to introduce such a requirement in a minor version too (it's not like we'd change the requirement to 4.4 or smth)

danilofuchs:
I agree TS 3.8 seems like a safe choice. TS rarely introduces breaking changes, so users of older versions might be able to upgrade easily.

TS does allow emitting multiple type declarations, each for a specific version, but I'm not sure it's viable.

Andarist:
> TS does allow emitting multiple type declarations, each for a specific version, but I'm not sure it's viable.

I'm not sure if emitting for specific versions is supported. The only thing that I'm aware of is `package.json#typesVersion` but that is only intended as a hint for the consumer - so they can load typedefs targeting their matching TS version. AFAIK such different typedefs must have to be written, mostly, by hand

mitchellhamilton:
There is https://github.com/sandersn/downlevel-dts but not sure it's really worth doing

Andarist:
Ye, I'm not sure if it's worth it - I would probably consider doing this after the release, if we get some issue reports about this.

PR #2558, opened 2021-11-20
Migrate @emotion/css to TypeScript

mitchellhamilton:
Pretty simple migration, just finding where the current types are wrong 🙃


packages/css/src/create-instance.ts
```diff
@@ -1,24 +1,80 @@
...
+    template: TemplateStringsArray,
+    ...args: Array<CSSInterpolation>
+  ): void
+  injectGlobal(...args: Array<CSSInterpolation>): void
+  keyframes(
+    template: TemplateStringsArray,
+    ...args: Array<CSSInterpolation>
+  ): string
+  keyframes(...args: Array<CSSInterpolation>): string
+  sheet: CSSStyleSheet
```
Andarist, line 63:
```suggestion
  sheet: StyleSheet
```

packages/css/src/create-instance.ts
```diff
@@ -31,61 +87,29 @@ function merge(
...
-  getRegisteredStyles
-}
-*/
-
-let createEmotion = (options) /*: Emotion */ => {
+let createEmotion = (options: Options): Emotion => {
   let cache = createCache(options)
 
-  cache.sheet.speedy = function (value /*: boolean */) {
+  ;(cache.sheet as any).speedy = function (value: boolean) {
```
Andarist, line 93:
```suggestion
  ;(cache.sheet as CSSStyleSheet).speedy = function (value: boolean) {
```

Andarist (approved):

PR #2559, opened 2021-11-20
Fix some TS things

mitchellhamilton:
Some misc fixes to things:

- The unitless thing shouldn't be `as const`, the keys aren't what you want from the types.
- Running tsc on CI
- The native package isn't part of the root TS project since @types/react-native conflicts with DOM types
  - If we convert the native package, that'll need it's own TS config (but only the native package, no other packages should need their own TS configs)
- The type tests aren't part of the root TS config, tbh I kinda think we should just remove DTSLint and use `@ts-expect-error` for when we expect errors and probably just a function like `<T>(t: T) => void` to assert types (i know that's not really correct but it's probably good enough for our cases)


tsconfig.json
```diff
@@ -8,8 +8,14 @@
...
-    "target": "es5"
+    "target": "es5",
+    "types": []
   },
   "include": ["packages", "scripts", "site", "test", "playgrounds"],
-  "exclude": ["node_modules", "packages/*/types/test/*"]
+  "exclude": [
+    "node_modules",
+    "packages/*/types/test/*",
+    "packages/*/types/test*",
```
Andarist, line 18: what the second one is doing here? 🤔

Andarist:
> The type tests aren't part of the root TS config, tbh I kinda think we should just remove DTSLint and use @ts-expect-error for when we expect errors and probably just a function like <T>(t: T) => void to assert types (i know that's not really correct but it's probably good enough for our cases)

The main reason I see to stick with `dtslint` is to test types against multiple TS versions - and to test them against `next` version. Since some of our types are a little bit more advanced the latter seems especially nice. 

I'm using the strategy that you have mentioned in other repos but mostly because I was too lazy to setup dtslint - since that is already done here, what's the cost of keeping it around?

tsconfig.json
mitchellhamilton: Not all of the type tests are in a test directory, maybe the first one is unnecessary, idk how TS's globs work

mitchellhamilton:
> since that is already done here, what's the cost of keeping it around?

If we can easily upgrade, sure I guess

PR #2560, opened 2021-11-22
v10: Drop usage of deprecated React types

eps1lon:
**What**:

Fixes breakage with planned React 18 types e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/runs/4279099932?check_suite_focus=true#step:6:2266

**Why**:

We plan to remove deprecated types in React 18. Since Emotion v10 peer dependencies will allow React 18, the types should not break.
Planned changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691
Current progress: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

**How**:

Replace `SFC` with `FC`

**Checklist**:

- ~[ ]~ Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/wicked-ducks-rest.md
+++ b/.changeset/wicked-ducks-rest.md
@@ -1,7 +1,8 @@
 ---
-'@emotion/core': patch
-'emotion-theming': patch
-'@emotion/styled-base': patch
+'@emotion/core': minor
+'emotion-theming': minor
+'@emotion/styled': minor
+'@emotion/styled-base': minor
 ---
 
-Drop usage of deprecated React types
+Dropped usage of a deprecated `SFC` React type in favor of `FC`. The `FC` type has been introduced in `@types/react@16.7.2` so this version of this package is now a minimum requirement for TypeScript users.
--- a/.changeset/wicked-ducks-rest.md
+++ b/.changeset/wicked-ducks-rest.md
@@ -1,0 +1,7 @@
+---
+'@emotion/core': patch
+'emotion-theming': patch
+'@emotion/styled-base': patch
+---
+
+Drop usage of deprecated React types
--- a/docs/typescript.mdx
+++ b/docs/typescript.mdx
@@ -148,7 +148,7 @@ const Image3 = styled.div<ImageProps>`
 ### React Components
 
 ```tsx
-import React, { SFC } from 'react'
+import React, { FC } from 'react'
 import styled from '@emotion/styled'
 
 type ComponentProps = {
@@ -156,7 +156,7 @@ type ComponentProps = {
   label: string
 }
 
-const Component: SFC<ComponentProps> = ({
+const Component: FC<ComponentProps> = ({
   label,
   className
 }) => <div className={className}>{label}</div>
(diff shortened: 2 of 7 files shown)


Andarist:
The `FC` type has been introduced in `@types/react@16.7.2` and that has been released 3 years ago. It should be OK for us to release this but there is a small breaking change here - so I'm mentioning this for completeness.

I wonder - is this blocking MUI anyhow? The PR is targeting the v10 line - I will probably backport all (most?) React 18 work when it actually gets released as a stable version but nothing has been done towards this goal yet.

Taking up on the occasion - are you considering doing [this](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691#issuecomment-957411343) as part of the React 18 types?

eps1lon:
> The FC type has been introduced in @types/react@16.7.2 and that has been released 3 years ago. It should be OK for us to release this but there is a small breaking change here - so I'm mentioning this for completeness.

Note that you unfortunately don't declare `@types/react`. However, your last v10 release was tested with `@types/react@16.8` so you were already assuming existence of `React.FC`.

> I wonder - is this blocking MUI anyhow? The PR is targeting the v10 line - I will probably backport all (most?)

It's blocking the DefinitelyTyped repo. Some types for Storybook related addons pull in Emotion V10. 

> Taking up on the occasion - are you considering doing [this](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691#issuecomment-957411343) as part of the React 18 types?

There isn't enough information there to draw up an actual  plan. What is the current behavior? What is the expected behavior? What are you doing right now? I would suggest opening a dedicate issue that goes into more detail why the current behavior is a problem and how we could remedy that. And what the implications are for existing code.

Andarist:
> There isn't enough information there to draw up an actual plan. What is the current behavior? What is the expected behavior? What are you doing right now? I would suggest opening a dedicate issue that goes into more detail why the current behavior is a problem and how we could remedy that. And what the implications are for existing code.

For completeness of this offtopic discussion here - I've created a discussion in the DefinitelyTyped repo here: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/57377

.changeset/wicked-ducks-rest.md
```diff
@@ -0,0 +1,7 @@
+---
+'@emotion/core': patch
+'emotion-theming': patch
+'@emotion/styled-base': patch
```
Andarist, line 4:
```suggestion
'@emotion/core': minor
'emotion-theming': minor
'@emotion/styled': minor
'@emotion/styled-base': minor
```

.changeset/wicked-ducks-rest.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Drop usage of deprecated React types
```
Andarist, line 7:
```suggestion
Dropped usage of a deprecated `SFC` React type in favor of `FC`. The `FC` type has been introduced in `@types/react@16.7.2` so this version of this package is now a minimum requirement for TypeScript users.
```

PR #2562, opened 2021-11-23
feat(eslint-plugin): allow eslint v8

G-Rath:
**What**:

Currently npm v7 will error due to the peer dependencies not matching.

**Why**:

I added ESLint v8 as an allowed version - I've not upgraded the codebase since that'd be significant more work and could involve breaking changes, but will attempt to do so later when I have the bandwidth (probably this weekend).

I've been using the plugin fine with v8 (by forcing the resolution), and the plugin itself is not doing anything super special so its pretty safe to allow v8.

**How**:

Adding 8 to the range constraint in the peer dependencies section.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Changeset

### DIFF
--- a/.changeset/serious-pianos-smoke.md
+++ b/.changeset/serious-pianos-smoke.md
@@ -2,4 +2,4 @@
 '@emotion/eslint-plugin': minor
 ---
 
-allow ESLint v8
+ESLint 8 has been added to the peer dependency range (ESLint 6 and ESLint 7 are still being supported).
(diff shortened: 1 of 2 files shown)


.changeset/serious-pianos-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/eslint-plugin': minor
+---
+
+allow ESLint v8
```
Andarist, line 5:
```suggestion
ESLint 8 has been added to the peer dependency range (ESLint 6 and ESLint 7 are still being supported).
```

Andarist:
Thank you!

PR #2566, opened 2021-11-25
Fix hydration mismatches with `React.useId`

Andarist:


.changeset/strange-kids-change.md
```diff
@@ -0,0 +1,10 @@
+---
+'@emotion/core': minor
+'@emotion/styled': minor
+'@emotion/styled-base': minor
```
mitchellhamilton, line 4: Why `minor` instead of `patch`? Don't mind strongly either way, just curious

.changeset/strange-kids-change.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+author: @eps1lon
+author: @Andarist
+
+Fixed hydration mismatches if `React.useId` (the upcoming API of the React 18) is used within a tree below our components.
```
mitchellhamilton, line 10:
```suggestion
Fixed hydration mismatches if `React.useId` (an upcoming API in React 18) is used within a tree below our components.
```

mitchellhamilton (approved):

.changeset/strange-kids-change.md
Andarist: I've decided to do the same for the types-related change. Kinda doesn't matter but maybe puts less pressure on people to immediately update their code. At the time being they **don't** have to do the update - they will only have to do that if they start using React 18 and that has not been even released yet. 

Dunno, I just sometimes use minor bumps for some additional context than a "new feature". The problem here is that this might affect some people - for example, the serialized output of `jest-emotion` might change (like it has changed in our tests on this branch here). So this leaves a little bit of breathing room for people so they can stay on the previous version if they want.

mitchellhamilton: Sounds good

PR #2570, opened 2021-11-27
feat(babel-plugin-jsx-pragmatic): migrate to typescript

G-Rath:
**What**:

Migrates `@emotion/babel-plugin-jsx-pragmatic` to TypeScript.

**Why**:

Part of #2358

**How**:

Renamed the js file to ts, and made changes to make TS happy.

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Code complete
- [x] Changeset 

I've not converted the tests because I don't know how to handle `babel-tester` since it's apparently coming in as some magical workspace thing (I've not worked with workspaces a lot 😅).

I'm sure there will be a way to tell TypeScript that it exists, but that probably will need either a type definition written or the package converted to TypeScript, so have left it for now.

I'll be making a few follow-up PRs to DefinitelyTyped that'll remove the need for some of the types, but they probably won't land for at least a month.

### DIFF
--- a/.changeset/wet-boxes-raise.md
+++ b/.changeset/wet-boxes-raise.md
@@ -2,4 +2,4 @@
 '@emotion/babel-plugin-jsx-pragmatic': minor
 ---
 
-Convert to TypeScript. No breaking changes
+Source code has been migrated to TypeScript. From now on type declarations will be emitted based on that, instead of being hand-written.
--- a/.changeset/wet-boxes-raise.md
+++ b/.changeset/wet-boxes-raise.md
@@ -1,0 +1,5 @@
+---
+'@emotion/babel-plugin-jsx-pragmatic': minor
+---
+
+Convert to TypeScript. No breaking changes
--- a/yarn.lock
+++ b/yarn.lock
@@ -5594,6 +5594,17 @@
     "@types/babel__template" "*"
     "@types/babel__traverse" "*"
 
+"@types/babel__core@^7.1.16":
+  version "7.1.16"
+  resolved "https://registry.yarnpkg.com/@types/babel__core/-/babel__core-7.1.16.tgz#bc12c74b7d65e82d29876b5d0baf5c625ac58702"
+  integrity sha512-EAEHtisTMM+KaKwfWdC3oyllIqswlznXCIVCt7/oRNrh+DhgT4UEBNC/jlADNjvw7UnfbcdkGQcPVZ1xYiLcrQ==
+  dependencies:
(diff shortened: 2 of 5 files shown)


G-Rath:
btw this should also allow me to do `@emotion/preset-css-prop`, since that uses this package

srmagura (commented):
@G-Rath Looks good. Was there any progress on getting some of these types into DefinitelyTyped so they can be removed from the Emotion code?

G-Rath:
> Ideally the tests for babel-plugin-jsx-pragmatic should be converted to TypeScript too for this change.

I don't mind doing that, but would prefer to address it in another PR if possible because I think it is a larger change: it's types will probably need some crafting, and at the least `babel-check-duplicated-nodes` doesn't have types so someone will need to define them.

srmagura:
OK, since it is a bigger change, doing it in a separate PR is fine 👍

@Andarist  I believe this is ready to be merged.

srmagura (approved):

yarn.lock
```diff
@@ -5594,6 +5594,17 @@
...
+"@types/babel__core@^7.1.18":
+  version "7.1.18"
+  resolved "https://registry.yarnpkg.com/@types/babel__core/-/babel__core-7.1.18.tgz#1a29abcc411a9c05e2094c98f9a1b7da6cdf49f8"
+  integrity sha512-S7unDjm/C7z2A2R9NzfKCK1I+BAALDtxEmsJBwlB3EzNfb929ykjL++1CK9LO++EIp2fQrC8O+BwjKvz6UeDyQ==
+  dependencies:
+    "@babel/parser" "^7.1.0"
+    "@babel/types" "^7.0.0"
+    "@types/babel__generator" "*"
+    "@types/babel__template" "*"
+    "@types/babel__traverse" "*"
```
Andarist, line 5606: sweet, sweet little lies 😅

.changeset/wet-boxes-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/babel-plugin-jsx-pragmatic': minor
+---
+
+Convert to TypeScript. No breaking changes
```
Andarist, line 5:
```suggestion
Source code has been migrated to TypeScript. From now on type declarations will be emitted based on that, instead of being hand-written.
```

Andarist:
@G-Rath I've prepared `babel-tester` migration here: https://github.com/emotion-js/emotion/pull/2688

PR #2572, opened 2021-11-29
fix(serialize): registered parameter of serializeStyles should be optional

otakustay:
**What**:

According to [its implementation](https://github.com/emotion-js/emotion/blob/26ded6109fcd8ca9875cc2ce4564fee678a3f3c5/packages/serialize/src/index.js#L311) and [how css function uses it](https://github.com/emotion-js/emotion/blob/26ded6109fcd8ca9875cc2ce4564fee678a3f3c5/packages/react/src/css.js#L7), the `registered` paramter of `serializeStyles` should be optional

**Why**:

To implement a custom `css` function working with frameworks other than react, current type is generating type erorrs like:

```
Expected 2-3 arguments, but got 1.ts(2554)
index.d.ts(67, 3): An argument for 'registered' was not provided.
(alias) serializeStyles<P>(args: (TemplateStringsArray | Interpolation<P>)[], registered: RegisteredCache, props?: P | undefined): SerializedStyles
import serializeStyles
```

**How**:

I added a single optional placeholder to this parameter, so that custom bindings like emotion-to-solid.js can work strong typed.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/.changeset/friendly-carrots-buy.md
+++ b/.changeset/friendly-carrots-buy.md
@@ -1,0 +1,5 @@
+---
+"@emotion/serialize": patch
+---
+
+Changed the `registered` paramater to be declared as optional in the TypeScript definition to match the runtime requirement.
--- a/.changeset/friendly-carrots-buy.md
+++ b/.changeset/friendly-carrots-buy.md
@@ -2,4 +2,4 @@
 "@emotion/serialize": patch
 ---
 
-Changed the `registered` paramater to be declared as optional in the TypeScript definition to match the runtime requirement.
+Changed the `registered` parameter to be declared as optional in the TypeScript definition to match the runtime requirement.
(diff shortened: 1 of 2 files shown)


danilofuchs:
@Andarist if this is correct, we must update the serialize TS migration as well

otakustay:
Is there anything I can do to make it merged?

srmagura:
Seems legit to me, @Andarist can you take a look at this?

.changeset/friendly-carrots-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@emotion/serialize": patch
+---
+
+Changed the `registered` paramater to be declared as optional in the TypeScript definition to match the runtime requirement.
```
Andarist, line 5:
```suggestion
Changed the `registered` parameter to be declared as optional in the TypeScript definition to match the runtime requirement.
```

Andarist:
Thanks ❤️  and sorry for not getting to this sooner

PR #2577, opened 2021-12-07
Export Keyframes type from @emotion/react

Methuselah96:
**What**:

Export `Keyframes` type from @emotion/core for v10.

**Why**:

In order to reference `Keyframes` without having to add a dependency to `@emotion/serialize`. See https://github.com/storybookjs/storybook/pull/16905 for more context.

**How**:

N/A

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Changeset

### DIFF
--- a/.changeset/old-chefs-repeat.md
+++ b/.changeset/old-chefs-repeat.md
@@ -1,5 +1,5 @@
 ---
-"@emotion/react": patch
+"@emotion/react": minor
 ---
 
 Export Keyframes type from @emotion/core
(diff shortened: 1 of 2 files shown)


.changeset/old-chefs-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@emotion/react": minor
+---
+
+Export Keyframes type from @emotion/react
```
Andarist, line 5: ---
"@emotion/react": patch
---

Export `Keyframes` type to avoid TypeScript inserting `import("@emotion/serialize").Keyframes` references into declaration files emitted based on a source files exporting `keyframes` result. This avoids issues with strict package managers that don't allow accessing undeclared dependencies.

PR #2585, opened 2021-12-09
Fix/babel plugin emotion function object crash

kddc:
**What**:
Fix the babel plugin crashing when facing objects with multiple emotion classes where some of them are arrow functions and adding the correct labels to the classes. 

**Why**:
We have something like this in our code

```js
const { useStyles } = {
  useStyles: () => {
    return {
      color: css`
        color: hotpink;
      `,
      colorFn: () => css`
        color: hotpink;
      `,
    }
  }
}
```

When the babel plugin processes the arrow function inside of it, it crashes with `TypeError: Cannot read properties of undefined (reading 'charAt')`, as it can not find a declaratorName for the function in the `label.js` file.

**How**:
The `getDeclaratorName` method checked for parents that could be of type `ArrowFunctionExpression`, but this was not handled inside the body of this function. I check the parents parent in this case using the existing `getDeclaratorName` method.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


packages/babel-plugin/src/utils/label.js
```diff
@@ -137,7 +148,7 @@ function getDeclaratorName(path, t) {
   }
 
   let variableDeclarator = parent.findParent(p => p.isVariableDeclarator())
-  if (!variableDeclarator) {
+  if (!variableDeclarator || !variableDeclarator.get('id').isIdentifier()) {
```
Andarist, line 151: @kddc this was the more appropriate fix. The `isArrowFunctionExpression` block that you have introduced has covered the introduced test case but I believe that it was more about improving label extraction for this specific case than about fixing a crash (the latter was a byproduct of the former though, in that scenario).

By patching the underlying issue here I've managed to also fixed the weird test case introduced as `no-label-obj-pattern-computed-property`

Andarist:
@kddc thank you very much for the PR! this means a lot ❤️

packages/babel-plugin/src/utils/label.js
kddc: @Andarist Awesome - that's even better then :)

Issue #2554, opened 2021-11-19
Add support for Cloudflare Worker environment in @emotion/cache, @emotion/styled, @emotion/react, and @emotion/server


tbrockman:
Some discussion of the issue in the [emotion Slack channel](https://emotion-slack.slack.com/archives/C64DA51B2/p1637122063055500).

**Current behavior:**

Attempting to render a React application using emotion as guided by [the documentation](https://emotion.sh/docs/ssr#on-server) will result in the Cloudflare Worker immediately crashing, as described [here](https://github.com/emotion-js/emotion/issues/2446).

This is because `stream` is not present in a Worker environment. 

After removing the code in `@emotion/server` ~which imports~ which indirectly imports `stream` through `through`, the Worker will no longer immediate crash on startup, but will throw an error on rendering:

```bash
ReferenceError: document is not defined
    at createCache (./index.mjs:1173:38)
    at handleRequest (./index.mjs:2241:19)
    at Object.fetch (./index.mjs:2236:22) at line 1172, col 36
```

This is due to the Worker bundler targeting the `browser`. Rollup (webpack untested) will optimize away `isBrowser` checks (as it *should* always be true), which still leaves references to `document`, which is undefined in the Worker environment. This issue is also described [here](https://github.com/emotion-js/emotion/issues/1455). Hardcoding `isBrowser = false` results in successful rendering.

**To reproduce:**

Clone [this Github repository](https://github.com/tbrockman/example-ssr-worker) and follow the README.

**Expected behavior:**

Ideally, `emotion` + `react` + SSR should work within Workers.

**Environment information:**

- `react` version: 17.0.2
- `@emotion/react` version: 11.6.0
- `@emotion/server` version: 11.4.0
- `@emotion/cache` version: 11.6.0
- `@emotion/styled` version: 11.6.0

Andarist:
>  Hardcoding isBrowser = false results in successful rendering.

I wonder - in which file do you patch this variable? with `nodeResolve({ browser: true })` I would expect it to pick up a file that doesnt have this in its code because a browser file already inlines such checks as those file target browser and thus this doesnt have to be checked at runtime :thinking:

tbrockman:
Not sure if I understand the question entirely, but I made the changes directly in `emotion/utils` and the other `emotion` packages with similar `let isBrowser = typeof document !== 'undefined'` references, then built and published versions to my local registry.

Andarist:
[...]

PR #2589, opened 2021-12-10
Feat: generate bundles for "worker" style environement

nicksrandall:
**What**:

This is PR uses a ~~*hopefully* soon-to-be-released~~ newly released version of preconstruct that support building "worker" compatible builds. The primary use-case for this is to support Cloudflare workers. 

**Why**:
See #2554 

**How**:
I've made changes to each package in the monorepo and extended the package.json to include an "exports" field with custom conditionals for "browser" and "worker". 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/docs/package-summary.mdx
+++ b/docs/package-summary.mdx
@@ -17,11 +17,11 @@ Depends on @emotion/react
 ## @emotion/css
 
 Contains a framework agnostic version of Emotion. Has different functionality and APIs from @emotion/react and doesn’t work the same way under the hood.
-Depends on @emotion/cache which can be customised by creating a custom instance at @emotion/css/create-instance
+Depends on @emotion/cache which can be customized by creating a custom instance at @emotion/css/create-instance
 
 ## @emotion/cache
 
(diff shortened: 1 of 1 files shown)


ItsWendell:
This is also super relevant for the Next.js Edge Runtime now (which is essentially built Cloudflare Worker): https://github.com/emotion-js/emotion/issues/2777, https://github.com/emotion-js/emotion/issues/2554

nicksrandall:
@mitchellhamilton @Andarist I have updated this PR to use the new v2.1.8 version of preconstruct and attempted to build with the new "worker" condition enabled but I must be missing something because that specific condition is not being generated. (The browser conditions is working). Any chance you could take a quick look and let me know what I did wrong?

mitchellhamilton:
`envConditoins` should be `envConditions`. Also, could you remove browser & worker builds for things that don't use `typeof window` or etc. (e.g. `memoize` but there are some others it looks like). Also, could you set `preconstruct.exports` to true at the root level so there should only be `preconstruct.exports` config on packages that need worker & browser builds.

nicksrandall:
Didn't mean to close, re-created in #2819

PR #2590, opened 2021-12-10
Stylis upgrade

Andarist:
FYI @thysultan - this is what i've ended up with as to changes introduced to accommodate 4.0.12. I've decided to ignore 4.0.11 and removed the fix that I had for it ( https://github.com/emotion-js/emotion/commit/26b3e2d28d5bd1b91c79d4074d69eeee469ae72e ) since that version was out only briefly and we didn't actually get any bug reports about this.

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -19,7 +19,7 @@
     "@emotion/sheet": "^1.1.0",
     "@emotion/utils": "^1.0.0",
     "@emotion/weak-memoize": "^0.2.5",
-    "stylis": "^4.0.10"
+    "stylis": "4.0.11"
   },
   "devDependencies": {
     "@emotion/hash": "*",
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -19,7 +19,7 @@
     "@emotion/sheet": "^1.1.0",
     "@emotion/utils": "^1.0.0",
     "@emotion/weak-memoize": "^0.2.5",
-    "stylis": "4.0.12"
+    "stylis": "4.0.13"
   },
   "devDependencies": {
     "@emotion/hash": "*",
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -19,7 +19,7 @@
     "@emotion/sheet": "^1.1.0",
     "@emotion/utils": "^1.0.0",
     "@emotion/weak-memoize": "^0.2.5",
-    "stylis": "4.0.11"
+    "stylis": "4.0.12"
   },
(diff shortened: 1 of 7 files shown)


packages/cache/src/stylis-plugins.js
```diff
@@ -101,14 +101,6 @@ export let compat = element => {
...
 
   while (parent.type !== 'rule') {
-    if (
-      // @media
-      parent.type.charCodeAt(1) !== 109 &&
-      // @supports
-      parent.type.charCodeAt(1) !== 115
-    ) {
-      return
-    }
```
Andarist: This was a fix for `stylis@4.0.11` that was out briefly but Sultan has patched the "regression" in a timely manner, and released `stylis@4.0.12`. So I've decided not to include this fix.

packages/cache/src/stylis-plugins.js
```diff
@@ -89,8 +89,9 @@ export let compat = element => {
   if (
     element.type !== 'rule' ||
     !element.parent ||
-    // .length indicates if this rule contains pseudo or not
-    !element.length
+    // positive .length indicates that this rule contains pseudo
+    // negative .length indicates that this rule has been already prefixed
+    element.length < 1
```
Andarist, line 94: this fixes prefixed rules (like `.css-hash::-webkit-input-placeholder {}`) - with the latest `stylis@4.0.11` and `stylis@4.0.12` those were replaced with their non-prefixed versions

packages/cache/package.json
```diff
@@ -19,7 +19,7 @@
     "@emotion/sheet": "^1.1.0",
     "@emotion/utils": "^1.0.0",
     "@emotion/weak-memoize": "^0.2.5",
-    "stylis": "^4.0.10"
+    "stylis": "4.0.12"
```
Andarist, line 22: I've decided to pin the version of this dependency as it's super crucial for Emotion and changes to it can affect our code in subtle ways. We want to keep this up-to-date with every release of Stylis - but we should always verify that our test suite passes before we can bump this dependency range

packages/cache/src/stylis-plugins.js
thysultan: Is stylis's prefixing of `::-webkit-input-placeholder` incorrect as it stands?

Andarist: No, its prefixing it alright - we were screwing up the prefixed rules here though, before this change

mitchellhamilton (approved):

PR #2599, opened 2021-12-20
Fixed test setup of one of the SSR-related tests

Andarist:


packages/css/test/instance/inline.test.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment node
+ */
```
Andarist, line 3: this is the core of the fix, the generated snapshot without this was incorrect as Emotion was running in client-mode, so stuff from other tests slipped into another one

PR #2600, opened 2021-12-20
`useInsertionEffect` when it's available

Andarist:


packages/css/test/instance/__snapshots__/inline.test.js.snap
```diff
@@ -91,52 +91,6 @@ exports[`hydration only inserts rules that are not in the critical css 3`] = `
   box-shadow: -15px -15px 0 0 aqua,-30px -30px 0 0 cornflowerblue;
 }
 
-@font-face {
```
Andarist, line 94: those were duplicated in the snapshot - I'm not entirely sure why, but the test setup here (and in other 3 tests that very similar) was a little bit convoluted, with using JSDOM manually etc. I've migrated those tests to `disableBrowserEnvTemporarily` strategy and this has "fixed" the problem accidentally.

packages/css/test/instance/__snapshots__/inline.test.js.snap
```diff
@@ -157,9 +111,7 @@ exports[`renderStylesToString renders large recursive component 1`] = `
 >
   .some-key-127stik{color:hotpink;}
 </style>
-<div class="some-key-127stik"
-     data-reactroot
```
Andarist, line 161: there is a comment somewhere about this - React 18 is not using this attribute any longer (I'm still confirming this with React 18 Working Group), so I'm normalizing this to avoid snapshot mismatched between versions

packages/react/src/emotion-element.js
```diff
@@ -49,7 +54,32 @@ export const createEmotionProps = (type: React.ElementType, props: Object) => {
   return newProps
 }
 
-const Noop = () => null
+const Insertion = ({ cache, serialized, isStringTag }) => {
+  registerStyles(cache, serialized, isStringTag)
+
+  const rules = useInsertionEffectMaybe(() =>
```
Andarist, line 60: One thing to consider - it would be cool to flatten the React tree cause this `Insertion` component is not needed. Or rather, it is **somewhat** needed:
- it creates a "slot" so we avoid SSR mismatches when `useId` is used. But that is only a requirement when relying on 0-config and I think that a lot of people are not actually using this feature. With most SSR-oriented APIs we don't have to render `style` elements "inline" and in those setups there is no need for this extra "slot"
- it preserves injection order. We know that injection order is not something that people should rely on but we'd like to make it as easy as possible for them to migrate

It seems that we can't easily work around the second "requirement' unless we provide a way for people to configure `injectWhenever`. I think the plan is to just make this a default in the next major version though - so the question is, should we allow people to flatten their React trees with a flag or something?

packages/css/test/no-babel/index.test.js
```diff
@@ -129,22 +136,30 @@ describe('css', () => {
...
-    expect(() => {
-      const fontSize = '20px'
-      const H1 = styled('h1')({ fontSize })
-      const Thing = styled('div')({
-        display: 'flex',
-        [String(H1)]: {
-          color: 'green'
-        }
-      })
+  test.only('component as selectors (object syntax)', () => {
```
mitchellhamilton, line 139:
```suggestion
  test('component as selectors (object syntax)', () => {
```

packages/react/src/emotion-element.js
mitchellhamilton: I think for now, let's leave it out.

mitchellhamilton (approved):
Looks good to me, just needs changesets

Issue #2638, opened 2022-02-12
The style is `.css-4oa7uv-Header.Space` rather than `.css-4oa7uv-Header\.Space`


PeterlitsZo:
**Current behavior:**

It make the className as `.css-4oa7uv-Header.Space`, and it also make a `.css-4oa7uv-Header.Space` style rules. But it should be `.css-4oa7uv-Header.Space` rule, it should be `.css-4oa7uv-Header\.Space` rule.

Here is my code:

``` typescript
// Header.Space = function HeaderSpace({ children }: HeaderSpaceProps) {
Header.Space = ({ children }: HeaderSpaceProps) => {
  return (
    <div
      sx={{
        py: mRV({ _: '6rem', sm: '8rem', md: '8rem' }),
        display: 'flex',
        flexDirection: 'column',
        alignItems: 'center',
        gap: '2rem',
      }}
    >
      {children}
    </div>
  )
};
```

When I export `Header`, I find that `Header.Space` component doesn't work well.
I check what happens. Well, there is a `.css-4oa7uv-Header.Space` style rule!
But the right name should be `.css-4oa7uv-Header\.Space`. It forget `\` sign!
I read the emtion source file. I find it will get the label name from function call stack if the NODE_ENV is development.
I try to rewrite the jest's test file. But I cannot make it again in test file...
Could you help me? And do you need a runable demo? If you pointer out how to let it show again in test, I think I can help you with a PR.

**Expected behavior:**

The style rule should be for `.css-4oa7uv-Header\.Space`.

**Environment information:**

- `react` version: `^17.0.2`
- `@emotion/react` version: `^11.7.1`

Andarist:
Please always try to share a repro case in a runnable form - either by providing a git repository to clone or a codesandbox. OSS maintainers usually can't afford the time to set up the repro, even if exact steps are given.

PeterlitsZo:
OK.

Andarist:
I see this in the devtools:
<img width="489" alt="Screenshot 2022-02-13 at 21 02 15" src="https://user-images.githubusercontent.com/9800850/153772629-a5966720-4488-43d3-8fb2-fea164266cd9.png">
and it renders OK:
<img width="300" alt="Screenshot 2022-02-13 at 21 03 50" src="https://user-images.githubusercontent.com/9800850/153772665-b5a33d14-39a0-4a08-a4de-45479846a482.png">

In what browser do you experience this problem?

Andarist:
[...]

PR #2658, opened 2022-02-21
Add more test about sub component (#2638)

PeterlitsZo:
**What**:

I add more test for issue #2638 

**Why**:

Thanks for [srmagura](https://github.com/srmagura)'s help, the issue #2638 is fixed, but I find that there looks like the test for issue #2638 is lack(?). So I add it.

**How**:

I have some source code in www.github.com/PeterlitsZo/FUCK/project-for-emotion. I just put it in the `playgrounds` and using the error's stack of it to make sure the function label is right.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/react/__tests__/get-label-from-stack-trace.js
+++ b/packages/react/__tests__/get-label-from-stack-trace.js
@@ -719,3 +719,35 @@ updateContainer@`
 
   // No SSR stack trace since this comes from a Gatsby project
 })
+
+/**
+ * See github.com/PeterlitsZo/FUCK/project-for-emotion
+ *
+ * I put it in playgrounds as a package. I cannot get the same error stack in
+ * cra, I am not sure that it depend on vite or not.
+ *
+ * I rename the `App.Space` with new name `App.MyComponent$9`.
+ */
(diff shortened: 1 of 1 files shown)


srmagura (changes requested):
I don't think this additional test coverage is super necessary, but it is true that this case is not explicitly tested.

If you do want to pursue this, it needs to match the format used for the other tests:
- Make a `describe` block that includes cases for all 4 environments: Chrome, Firefox, Safari, and SSR. You can collect the stack traces via the Next.js playground. Let me know if you don't have access to a Mac for Safari.
- Add a block comment above the `describe` block that includes the code used to produce the stack trace.
- Remove the existing comment since the information there is not relevant to future readers of this code.
- Remove the last half (or so) of the stack trace to save space.

PeterlitsZo:
OK! I will finish it for Firefox and Chrome firstly. I do not have Mac now. And I do not know SSR now but I will try my best.

Thanks for your reply.

srmagura:
Closing since this hasn't been updated in a while.

PR #2659, opened 2022-02-22
Fix the description of the cascade problem in the docs

Andarist:


docs/composition.mdx
```diff
@@ -27,7 +27,7 @@ render(
...
 
-For example, we have some base styles and a danger style, we want the danger styles to have precedence over the base styles but because `base` is in the stylesheet after `danger` it has higher [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). In regular CSS, you might do something to make `danger` have a higher specificity than `base` like moving the `danger` class, so it's more specific than `base`, use `!important` or abandon composition and rewrite the styles each time you need them.
+For example, we have some base styles and a danger style, we want the danger styles to have precedence over the base styles but because `base` is in the stylesheet after `danger` its styles will override the styles defined in `danger`, as per [the cascade rules](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#the_cascade). In regular CSS, you might do something to make `danger` have a precedence over `base` like moving the `danger` class, so it's more defined later than `base`, or use `!important` or abandon composition and rewrite the styles each time you need them.
```
srmagura, line 30: Minor suggestions:
"have a precedence over" -> "take precedence over"
"so it's more defined later" -> "so it's defined later"

Andarist: @srmagura could u create a suggestion here so I could just accept it?

srmagura, line 30:
```suggestion
For example, we have some base styles and a danger style, we want the danger styles to have precedence over the base styles but because `base` is in the stylesheet after `danger` its styles will override the styles defined in `danger`, as per [the cascade rules](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#the_cascade). In regular CSS, you might do something to make `danger` take precedence over `base` like moving the `danger` class so it's defined later than `base`, or use `!important`, or abandon composition and rewrite the styles each time you need them.
```

srmagura: Well I accidentally but it in a different comment thread but yes 👍

Issue #2675, opened 2022-03-07
Custom StyleSheet


Jack-Works:
https://github.com/emotion-js/emotion/blob/ce9873faabef8baf4eeebf8fab0d7b1a46ff74c7/packages/react/src/global.js#L94

Hi! I found emotion in creating a new Sheet based on the old sheet. Is it possible to get rid of this behavior, or use the same constructor as the `cache.sheet` has?

I'm handling a very special environment and I must implement my own `StyleSheet` class to insert rules. It's be like this:

```js
    const emotionCache = createEmotionCache({ key: keyA })
    const styleSheet = new MySpecialStyleSheet(keyA, shadow)
    emotionCache.sheet = styleSheet
```

It works well besides the global styles. Global styles using the emotion's original implementation therefore I cannot control the insertRule behavior.

If I can provide my own StyleSheet implementation, or at least emotion can write this to use my StyleSheet:

```js
let sheet = new cache.sheet.constructor({
    ...
})
```

it will be great! thanks!

Andarist:
Hi there, 

nice to see you here - thank you for your great OSS work! I can't wait for throw types to happen 😉 

Using `cache.sheet.constructor` within `Global` seems like something that we could do. Considering it's a simple change and that it wouldn't break anything and doesn't really introduce any new options/configuration etc this is a solution that I would prefer. Especially that you could only tweak your cache once and the rest of the thing would "just work".

If you prepare a PR for such a change with tests then I would happily merge it.

> I'm handling a very special environment and I must implement my own StyleSheet class to insert rules.

Out of my curiosity - could you describe this in more detail? I love learning about quirky use cases :P

Jack-Works:
> nice to see you here - thank you for your great OSS work! I can't wait for throw types to happen 😉

thanks! but it's likely not to happen because the TypeScript team doesn't approve that feature.

> Out of my curiosity - could you describe this in more detail? I love learning about quirky use cases :P

Our application is rendered in the Shadow Root.

When a Dialog is opened, it needs to be rendered into another ShadowRoot (in order to have the correct DOM hierarchy to be the frontest DOM). Therefore, our application actually needs to render the same StyleSheet in multiple places.

As you can see, this component sits inside a ShadowRoot.

![image](https://user-images.githubusercontent.com/5390719/157019360-2983e097-f450-4c77-ad4c-40c8409efd7d.png)

And this dialog is in the out-most DOM to show in the top-most.

[...]

PR #2677, opened 2022-03-09
chore: custom sheet

Jack-Works:
**What**: change `new StyleSheet` to `new cache.sheet.constructor`

**Why**: close https://github.com/emotion-js/emotion/issues/2675

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/packages/react/__tests__/custom-cache.js
+++ b/packages/react/__tests__/custom-cache.js
@@ -2,7 +2,8 @@
 /** @jsx jsx */
 import 'test-utils/next-env'
 import createCache from '@emotion/cache'
-import { jsx, CacheProvider } from '@emotion/react'
+import { jsx, CacheProvider, Global } from '@emotion/react'
+import { StyleSheet } from '@emotion/sheet'
 import renderer from 'react-test-renderer'
 
 function stylisPlugin(element) {
@@ -37,4 +38,42 @@ test('with custom plugins', () => {
       className="emotion-0"
     />
   `)
+})
+
+test('with custom sheet', () => {
+  // https://github.com/emotion-js/emotion/issues/2675
+  let cache = createCache({
+    key: 'test',
+    speedy: false
+  })
+
+  class GlobalSheet extends StyleSheet {
+    insert(rule) {
+      return super.insert('/** global */' + rule)
+    }
+  }
+  class MySheet extends StyleSheet {
+    constructor(options) {
+      super(options)
+      if (Reflect.get(options.container, 'sheet')) {
+        return new GlobalSheet(options)
+      } else {
+        Reflect.set(options.container, 'sheet', this)
+      }
+    }
+
+    insert(rule) {
+      super.insert(`/** ${this.key} */${rule}`)
+    }
+  }
+  cache.sheet = new MySheet({ key: 'test', container: document.head })
+
+  expect(
+    render(
+      <CacheProvider value={cache}>
+        <div css={{ display: 'flex', color: 'blue' }} />
+        <Global styles={{ body: { width: '0' } }} />
+      </CacheProvider>
+    )
+  ).toMatchInlineSnapshot()
 })
--- a/packages/react/__tests__/custom-cache.js
+++ b/packages/react/__tests__/custom-cache.js
@@ -57,7 +57,10 @@ test('Global should "inherit" sheet class from the cache', () => {
       super.insert(`/** ${this.key} */${rule}`)
     }
   }
-  cache.sheet = new MySheet({ key: 'test', container: document.head })
+  cache.sheet = new MySheet({
+    key: 'test',
+    container: safeQuerySelector('head')
+  })
 
   render(
     <CacheProvider value={cache}>
--- a/packages/react/__tests__/custom-cache.js
+++ b/packages/react/__tests__/custom-cache.js
@@ -1,10 +1,11 @@
 // @flow
 /** @jsx jsx */
-import 'test-utils/next-env'
 import createCache from '@emotion/cache'
-import { jsx, CacheProvider, Global } from '@emotion/react'
+import { CacheProvider, Global, jsx } from '@emotion/react'
 import { StyleSheet } from '@emotion/sheet'
 import renderer from 'react-test-renderer'
+import { safeQuerySelector } from 'test-utils'
+import 'test-utils/next-env'
 
 function stylisPlugin(element) {
   if (element.type === 'decl' && element.value.startsWith('color:')) {
@@ -15,6 +16,11 @@ function stylisPlugin(element) {
 function render(ele) {
   return renderer.create(ele).toJSON()
 }
+
+beforeEach(() => {
+  safeQuerySelector('head').innerHTML = ''
+  safeQuerySelector('body').innerHTML = ''
+})
 
 test('with custom plugins', () => {
   let cache = createCache({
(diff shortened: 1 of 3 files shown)


Jack-Works:
hi! I cannot run `jest -u` to update the snapshot cause it complains


```
 FAIL  packages/react/__tests__/custom-cache.js
  ● Test suite failed to run

    SyntaxError: C:\Users\Jack\Workspace\emotion\packages\react\__tests__\custom-cache.js: Support for the experimental 
syntax 'jsx' isn't currently enabled (27:7):

      25 |   expect(
      26 |     render(
    > 27 |       <CacheProvider value={cache}>
         |       ^
      28 |         <div css={{ display: 'flex', color: 'blue' }} />
      29 |       </CacheProvider>
      30 |     )

    Add @babel/preset-react (https://git.io/JfeDR) to the 'presets' section of your Babel config to enable transformation.
    If you want to leave it as-is, add @babel/plugin-syntax-jsx (https://git.io/vb4yA) to the 'plugins' section to enable parsing.

      at Parser._raise (node_modules/jest-cli/node_modules/@babel/parser/src/parser/error.js:147:45)
```

packages/react/src/global.js
```diff
@@ -91,7 +91,8 @@ export let Global: React.AbstractComponent<GlobalProps> =
     useInsertionEffect(() => {
       const key = `${cache.key}-global`
 
-      let sheet = new StyleSheet({
+      // use case of https://github.com/emotion-js/emotion/issues/2675
+      let sheet = new cache.sheet.constructor({
```
Andarist, line 95: Could you remove this import?
https://github.com/emotion-js/emotion/blob/e638c6d844db8c5f33250a830f37c0cd91fdc1e7/packages/react/src/global.js#L8

packages/react/__tests__/custom-cache.js
```diff
@@ -38,3 +39,41 @@ test('with custom plugins', () => {
...
+    }
+  }
+  class MySheet extends StyleSheet {
+    constructor(options) {
+      super(options)
+      if (Reflect.get(options.container, 'sheet')) {
+        return new GlobalSheet(options)
+      } else {
+        Reflect.set(options.container, 'sheet', this)
+      }
```
Andarist, line 62: I wonder - any particular reason why `Reflect` is used here?

Andarist:
> hi! I cannot run jest -u to update the snapshot cause it complains

That is somewhat weird but don't worry - I will update them before merging

packages/react/__tests__/custom-cache.js
Jack-Works: To bypass the flow type check. `sheet` is not a property on the `HTMLElement`

Andarist:
I've wanted to push out changes to this branch - but it seems like I don't have sufficient permissions to push to your fork.

You can check how to allow this for future PRs here - https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork . At the moment, you can try applying this patch:
```diff
diff --git a/packages/react/__tests__/custom-cache.js b/packages/react/__tests__/custom-cache.js
index cab71fc3..86f6d89f 100644
--- a/packages/react/__tests__/custom-cache.js
+++ b/packages/react/__tests__/custom-cache.js
@@ -1,10 +1,11 @@
 // @flow
 /** @jsx jsx */
-import 'test-utils/next-env'
 import createCache from '@emotion/cache'
-import { jsx, CacheProvider, Global } from '@emotion/react'
+import { CacheProvider, Global, jsx } from '@emotion/react'
 import { StyleSheet } from '@emotion/sheet'
 import renderer from 'react-test-renderer'
+import { safeQuerySelector } from 'test-utils'
+import 'test-utils/next-env'
 
 function stylisPlugin(element) {
   if (element.type === 'decl' && element.value.startsWith('color:')) {
@@ -16,6 +17,11 @@ function render(ele) {
   return renderer.create(ele).toJSON()
 }
 
+beforeEach(() => {
+  safeQuerySelector('head').innerHTML = ''
+  safeQuerySelector('body').innerHTML = ''
+})
+
 test('with custom plugins', () => {
   let cache = createCache({
     key: 'custom-plugins',
@@ -40,40 +46,42 @@ test('with custom plugins', () => {
   `)
 })
 
-test('with custom sheet', () => {
+test('Global should "inherit" sheet class from the cache', () => {
   // https://github.com/emotion-js/emotion/issues/2675
   let cache = createCache({
     key: 'test',
     speedy: false
   })
-
-  class GlobalSheet extends StyleSheet {
-    insert(rule) {
-      return super.insert('/** global */' + rule)
-    }
-  }
   class MySheet extends StyleSheet {
-    constructor(options) {
-      super(options)
-      if (Reflect.get(options.container, 'sheet')) {
-        return new GlobalSheet(options)
-      } else {
-        Reflect.set(options.container, 'sheet', this)
-      }
-    }
-
     insert(rule) {
       super.insert(`/** ${this.key} */${rule}`)
     }
   }
   cache.sheet = new MySheet({ key: 'test', container: document.head })
 
-  expect(
-    render(
-      <CacheProvider value={cache}>
-        <div css={{ display: 'flex', color: 'blue' }} />
-        <Global styles={{ body: { width: '0' } }} />
-      </CacheProvider>
-    )
-  ).toMatchInlineSnapshot()
+  render(
+    <CacheProvider value={cache}>
+      <div css={{ color: 'hotpink' }} />
+      <Global styles={{ body: { width: '0' } }} />
+    </CacheProvider>
+  )
+
+  expect(safeQuerySelector('head')).toMatchInlineSnapshot(`
+    <head>
+      <style
+        data-emotion="test-global"
+        data-s=""
+      >
+        
+        /** test-global */body{width:0;}
+      </style>
+      <style
+        data-emotion="test"
+        data-s=""
+      >
+        
+        /** test */.test-1lrxbo5{color:hotpink;}
+      </style>
+    </head>
+  `)
 })

```

Andarist:
> hi! I cannot run jest -u to update the snapshot cause it complains

That was weird and I've also experienced that. The problem is when you are using `.toMatchInlineSnapshot()` in this file - Jest tries to automatically parse the file and insert the generated snapshot as the argument to this call and it fails in the process. I didn't ever have this problem in this repo and I'm not sure if we have changed something in our configs recently or what. I'm pretty sure that I was using `.toMatchInlineSnapshot()` pretty extensively here in the past.

I didn't solve this problem, I just have provided a dummy value as a snapshot manually and then constructed the correct value manually based on the CLI report. I'll have to look into this later but that's a completely separate issue from this PR.

packages/react/__tests__/custom-cache.js
```diff
@@ -38,3 +45,46 @@ test('with custom plugins', () => {
...
   `)
 })
+
+test('Global should "inherit" sheet class from the cache', () => {
+  // https://github.com/emotion-js/emotion/issues/2675
+  let cache = createCache({
+    key: 'test',
+    speedy: false
+  })
+  class MySheet extends StyleSheet {
```
Andarist, line 55: I've noticed that you were constructing different sheet classes in the constructor of your custom implementation but perhaps that's not really needed. You can easily differentiate hashed/main and global sheets by just checking the `options.key.endsWith('-global')`. Based on that you can just assign some private property and branch the logic in the `insert` method. This way you only have to write a single `StyleSheet` class which might simplify your implementation.

Jack-Works: oh, I need to get the un-global version of the sheet so that's the easiest way, any way, thanks!

Jack-Works:
> That was weird and I've also experienced that.

Maybe because I'm using Windows? 🤔

Andarist:
I have meant that the same thing has happened to me - so it was definitely not specific to your machine etc.

PR #2678, opened 2022-03-10
docs: fix cache-provider demo rendering

pengx17:
**What**:
The demo playground in https://emotion.sh/docs/cache-provider is not rendered properly. 

**Why**:
The playground (react-live) requires the demo to include `stylis` as the dependency.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/site/src/components/Playground.js
+++ b/site/src/components/Playground.js
@@ -31,6 +31,8 @@ export const scope = {
         return require('@emotion/is-prop-valid')
       case 'facepaint':
         return require('facepaint')
+      case 'stylis':
+        return require('stylis')
       default:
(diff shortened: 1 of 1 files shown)


docs/cache-provider.mdx
```diff
@@ -14,7 +14,7 @@ import { prefixer } from 'stylis'
 const myCache = createCache({
   key: 'my-prefix-key',
   stylisPlugins: [
-    customPlugin,
+    // customPlugin,
```
Andarist, line 17: Instead of removing this here, we can declare this before calling `createCache`:
```js
const customPlugin = () => {}
```

PR #2688, opened 2022-03-18
Migrate `babel-tester` to TypeScript

Andarist:


srmagura (changes requested):
All code looks good 🙂. **The only thing is, I think you need to run a `yarn install`.** When I did, it added 2 entries to the `yarn.lock`.

This issue is unrelated to your PR, but when I run `yarn build`, I'm getting one new file (`babel-plugin-jsx-pragmatic/src/index.d.ts`), and formatting changes to 9 other `.d.ts` files. Should I commit these changes and make a new PR?

When I have done TypeScript libraries in the past, I have added the `.d.ts` files to `.gitignore` since they are compiler output. Maybe I should do that instead? I don't understand preconstruct so I'm a bit confused here.

srmagura (approved):

Andarist:
> This issue is unrelated to your PR, but when I run yarn build, I'm getting one new file (babel-plugin-jsx-pragmatic/src/index.d.ts), and formatting changes to 9 other .d.ts files. Should I commit these changes and make a new PR?

Hm, it doesn't happen to me 🤔 

> When I have done TypeScript libraries in the past, I have added the .d.ts files to .gitignore since they are compiler output. Maybe I should do that instead? I don't understand preconstruct so I'm a bit confused here.

Yeah, those should be gitignored but preconstruct generates them in `dist` directories and those are already gitignored.

PR #2700, opened 2022-03-30
Allow `var()` in content property

arturmuller:
**What**: The `var()` css function is now allowed in the content property.

**Why**: It is valid (and sometimes very useful!) to use CSS variables in the content property. Currently, Emotion throws and error (in development) when `var` is encountered, making its use impossible.

**How**: `var` has been added to `contentValuePattern` regex.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset N/A (I am not sure if this should release a new package?)

### DIFF
--- a/.changeset/young-deers-cheat.md
+++ b/.changeset/young-deers-cheat.md
@@ -1,0 +1,7 @@
+---
+"@emotion/css": patch
+"@emotion/react": patch
+"@emotion/serialize": patch
+---
+
+Fixed a false positive warning for `content` properties that included `var()`.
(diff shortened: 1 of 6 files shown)


Andarist:
Could you just prepare a simple jsfiddle or something that would showcase how exactly you are using this with plain CSS?

arturmuller:
@Andarist Sure thing. Here's a CodePen: https://codepen.io/artmllr/pen/zYpErqJ

Issue #557, opened 2018-02-05
Should be able to exclude styles in jest-emotion serializer output


staylor:
- `emotion` version: 9.0.0-2
- `react` version: 16.2

**Relevant code.**

```javascript
import * as emotion from 'emotion';
import { createSerializer } from './jest';

expect.addSnapshotSerializer(createSerializer(emotion));
```
**What you did:**

I forked the logic of `jest-emotion` to be able to output snapshots with stable classes, but without styles. https://github.com/staylor/reformation/blob/master/src/emotion/jest.js

**What happened:**

This allows snapshots to remain stable when logic changes OR when styles change. 

**Reproduction:**

I can simply do this in my tests with emotion components:

```
import 'emotion/serializer'
```

**Problem description:**

I fully expect styles and classNames to be constantly changing, so I would like stable classNames. Snapshots  are still expected to churn when I add more components and logic.

**Suggested solution:**

I think an `includeStyles: true` option should be added to `createSerializer`, that way, I can set it to `false` in my implementation, and get stable classnames and no styles in my snapshot output.

mitchellhamilton:
I'd happily accept a PR that adds this as long as `includeStyles` defaults to true

tujoworker:
@staylor Did du send in a PR on this?

thompsongl:
Created #2709 to implement the request made here after coming to the same conclusion about wanting to exclude styles from snapshots

PR #2709, opened 2022-04-04
[@emotion/jest] Add `includeStyles` option to `createSerializer`

thompsongl:
**What**: Added `includeStyles` option to `createSerializer` to optionally disable snapshot style insertion

**Why**: As described in #557, we would like to exclude styles from snapshots

**How**:

* Added an `includeStyles` boolean option to `createSerializer` with a default of `true` to maintain existing functionality
* Used the `includeStyles` option to optionally bypass `getPrettyStylesFromClassNames`, resulting in `''` getting passed to `replaceClassNames` and no styles being inserted

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/polite-icons-judge.md
+++ b/.changeset/polite-icons-judge.md
@@ -2,4 +2,4 @@
 '@emotion/jest': minor
 ---
 
-Added `includeStyles` option to `createSerializer` to optionally disable snapshot style insertion
+Added `includeStyles` option to `createSerializer` to optionally disable styles printing.
--- a/.changeset/polite-icons-judge.md
+++ b/.changeset/polite-icons-judge.md
@@ -1,0 +1,5 @@
+---
+'@emotion/jest': minor
+---
+
+Added `includeStyles` option to `createSerializer` to optionally disable snapshot style insertion
--- a/packages/jest/test/printer.test.js
+++ b/packages/jest/test/printer.test.js
@@ -98,6 +98,33 @@ describe('jest-emotion with DOM elements disabled', () => {
   })
 })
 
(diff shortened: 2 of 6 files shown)


packages/jest/test/printer.test.js
```diff
@@ -98,6 +98,33 @@ describe('jest-emotion with DOM elements disabled', () => {
   })
 })
 
+describe('jest-emotion with style insertion disabled', () => {
```
Andarist, line 101: could you flatten this to a single `it` block? the problem with putting `describe` in the middle is that now the "test report" won't match the order in the file and it's often confusing for me, since this doesn't group more than a single test i would prefer to just use a simple `it` here

packages/jest/test/printer.test.js
```diff
@@ -98,6 +98,31 @@ describe('jest-emotion with DOM elements disabled', () => {
   })
 })
 
+test('jest-emotion with style insertion disabled does not insert styles into snapshots', () => {
```
Andarist, line 101:
```suggestion
test("allows to opt-out from styles printing", () => {
```

.changeset/polite-icons-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': minor
+---
+
+Added `includeStyles` option to `createSerializer` to optionally disable snapshot style insertion
```
Andarist, line 5:
```suggestion
Added `includeStyles` option to `createSerializer` to optionally disable styles printing.
```

thompsongl:
Thanks for the quick review, @Andarist!

PR #2715, opened 2022-04-07
[@emotion/jest] Pass through `includeStyles` option in `createEnzymeSerializer`

thompsongl:
**What**: Add the missing `includeStyles` option to `createEnzymeSerializer`

**Why**: The option was added to `createSerializer` in #2709, but was not added to `createEnzymeSerializer`

**How**: The option is accepted and passed through to `createEmotionSerializer`

- [ ] Documentation N/A
- [ ] Tests  N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/hot-seals-promise.md
+++ b/.changeset/hot-seals-promise.md
@@ -1,0 +1,5 @@
+---
+'@emotion/jest': patch
+---
+
+Fixed `createEnzymeSerializer` missing the `includeStyles` option
--- a/.changeset/hot-seals-promise.md
+++ b/.changeset/hot-seals-promise.md
@@ -2,4 +2,4 @@
 '@emotion/jest': patch
 ---
 
-Fixed `createEnzymeSerializer` missing the `includeStyles` option
+Fixed an issue with `createEnzymeSerializer` not handling the recently added `includeStyles` option.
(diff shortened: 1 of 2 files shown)


packages/jest/src/create-enzyme-serializer.js
```diff
@@ -66,11 +66,13 @@ const wrappedEnzymeSerializer = {
 
 export function createEnzymeSerializer({
   classNameReplacer,
-  DOMElements = true
+  DOMElements = true,
+  includeStyles = true
```
Andarist, line 70: We might also need to add this in the `.d.ts` file

thompsongl: `createEnzymeSerializer` uses the same `CreateSerializerOptions` interface as `createSerializer` (updated with `includeStyles` in #2709). In `enzyme.d.ts`, imported from `index.d.ts`. Or are you referring to a different `d.ts` file?

Andarist: No, this is fine. I was reviewing from mobile and I've assumed that there is another type for this one.

.changeset/hot-seals-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/jest': patch
+---
+
+Fixed `createEnzymeSerializer` missing the `includeStyles` option
```
Andarist, line 5:
```suggestion
Fixed an issue with `createEnzymeSerializer` not handling the recently added `includeStyles` option.
```

PR #2723, opened 2022-04-11
POC: Eslint Codemod utils added to eslint plugin ** this is still WIP **

DarkPurple141:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,5 +267,8 @@
     "unist-util-visit": "^1.2.0",
     "webpack-bundle-analyzer": "3.3.2",
     "worker-loader": "2.0.0"
+  },
+  "devDependencies": {
+    "eslint-codemod-utils": "^0.1.2"
   }
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -267,5 +267,8 @@
     "unist-util-visit": "^1.2.0",
     "webpack-bundle-analyzer": "3.3.2",
     "worker-loader": "2.0.0"
+  },
+  "devDependencies": {
+    "eslint-codemod-utils": "^0.1.2"
```
Andarist, line 272: If anything this could be forked from the `ts-migration` branch. I don't give any guarantees that this will end up being merged - I'm rather hesitant to start relying on an experimental library like this.

DarkPurple141: No worries - was just looking to see its use / failings in a real world plugin. May or may not abandon this PR after further experimentation. Cheers!

PR #2728, opened 2022-04-16
Allow any Node, including ShadowRoot, as container

Peeja:
**What**:

* Expands the type of `container` in `@emotion/sheet` and `@emotion/cache` to `Node`.
* Tests the use of a `ShadowRoot` as a `container`.

**Why**:

When using Shadow DOM, stylesheets are scoped to the shadow DOM tree they appear in. The most common place to put them is at the shadow root. But a `ShadowRoot`, though it contains `HTMLElement`s, is not itself an `HTMLElement`, though it is a `Node`. The types only allowed `HTMLElement`s as the `container`, but as the code is written, any `Node` should work. The added test validates that.

**How**:

Broadens a type in a few places and adds one test.

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] Code complete
- [X] Changeset 
  - Note: I've made this a patch bump because it's "just" types. Arguably it could be a minor bump because it's a new feature, according to SemVer. Happy to switch it if you'd prefer.


packages/sheet/src/index.js
```diff
@@ -66,7 +66,7 @@ export class StyleSheet {
   isSpeedy: boolean
   ctr: number
   tags: HTMLStyleElement[]
-  container: HTMLElement
+  container: Node
```
srmagura, line 69:
```suggestion
  // Using Node instead of HTMLElement since container may be a ShadowRoot
  container: Node 
```

.changeset/five-balloons-sneeze.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/cache': patch
+'@emotion/sheet': patch
+---
+
+The cache's `container` can be a `ShadowRoot`, or any other kind of `Node`
```
srmagura, line 6:
```suggestion
`@emotion/cache` and `@emotion/sheet` now allow `container` to be a `ShadowRoot`, or any other kind of `Node`. This change only affects the TypeScript types.
```

srmagura (changes requested):
Great work 👏. Left two suggestions.

Peeja:
Thanks! I've committed those suggestions.

srmagura (approved):
LGTM 🙏. Looks like ESLint is complaining about a trailing space in `packages/sheet/src/index.js`. Can you run Prettier on that file and push the change? I would push to your branch but I don't have write access to the repo.

@Andarist ready for you to review

packages/sheet/src/index.js
```diff
@@ -66,7 +66,8 @@ export class StyleSheet {
   isSpeedy: boolean
   ctr: number
   tags: HTMLStyleElement[]
-  container: HTMLElement
+  // Using Node instead of HTMLElement since container may be a ShadowRoot
+  container: Node
```
Andarist, line 70:
```suggestion
  container: Node
```

packages/sheet/types/index.d.ts
```diff
@@ -15,7 +15,7 @@ export class StyleSheet {
   isSpeedy: boolean
   ctr: number
   tags: Array<HTMLStyleElement>
-  container: HTMLElement
+  container: Node
   key: string
   nonce?: string
   before?: Element | null
```
Andarist, line 21: This probably should be adjusted as well, thoughts?
```suggestion
  before?: ChildNode | null
```

srmagura: Agree with this change. Without this change we will probably get TS errors in the `ts-migration` branch because `Node['firstChild']` is `ChildNode`, not `Element`.

Peeja:
Good catch, @Andarist, thanks! Added.

Andarist (approved):

PR #2733, opened 2022-04-24
chore: do not escape strings in snapshots

SimenB:
**What**:

I noticed while going through some of your snapshots that they're harder to read than they need to be. Turning of string escaping (which will probably be the default in a future Jest major version) makes it easier

**Why**:

Easier to read snapshots makes them easier to maintain

**How**:

Added jest config and ran `yarn jest -u`. I had to manually mess around in `packages/styled/__tests__/warnings.js` and `packages/css/test/no-babel/warnings.test.js` due to some weird syntax errors (either with flow annotations or jsx) - not sure why.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete
- [ ] Changeset


jest.config.js
```diff
@@ -17,5 +17,8 @@ module.exports = {
   coveragePathIgnorePatterns: [
     '/node_modules/',
     '<rootDir>/packages/babel-plugin/test/util.js'
-  ]
+  ],
+  snapshotFormat: {
+    escapeString: false
```
Andarist, line 22: While we are at it - could you update all snapshots with this? Unless you have a preference *not* to use this option?
```suggestion
    escapeString: false,
    printBasicPrototype: false
```

SimenB: Nah, that config will be default in jest 29 probably 🙂

PR #2734, opened 2022-04-24
Upgrade Preconstruct

Andarist:


srmagura:
Yes, this does fix the `.d.ts` files being generated in the wrong place!

Please merge https://github.com/emotion-js/emotion/pull/2736 into `preconstruct-upgrade` first, then this PR is good to go. 🙂

BTW, when we try to merge the Preconstruct upgrade into `ts-migration`, we are going to get weird TS errors from `packages/eslint-plugin`. I can tackle those once this is merged. I forgot the copy the TypeScript error message, but it's the one that occurs when the `.d.ts` file would include an "unnamed type" from an external module...

PR #2749, opened 2022-05-15
Added a dedicated error for a situation when `toHaveStyleRule` receives an array

Andarist:
improves DX around a situation reported here: https://github.com/emotion-js/emotion/issues/1463#issuecomment-1126149825

### DIFF
--- a/.changeset/silly-hornets-chew.md
+++ b/.changeset/silly-hornets-chew.md
@@ -1,0 +1,5 @@
+---
+'@emotion/jest': patch
+---
+
+Added a dedicated error for a situation when `toHaveStyleRule` receives an array.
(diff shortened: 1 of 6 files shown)


srmagura (approved):
Looks good. Weird about the jest issue btw...

Andarist:
I've checked Jest 26 and that parser was relying more heavily on Prettier itself back then - that's why it probably worked. I suspect that it just handles JSX by default. Right now they are using `babelCore.parseSync` there and it comes with the whole baggage of config loading - that's probably why they have changed that, they have wanted to parse non-standard syntax etc. Appropriately loading configs, especially in monorepos, is quite cumbersome though - if you want to do it automatically, as there are many ways people could configure it.

mitchellhamilton (approved):

PR #2751, opened 2022-05-17
Fix ignoreFlag not works bug

### LINKED ISSUE #2478: Potentially Unsafe Server-Rendering flag works on the first instance only
**Current behavior:**

When using the ignoreFlag for `nth-child` being potentially unsafe with server rendering, the first instance of the flag works while the second does not (workaround provided). 

**To reproduce:**

Create a component that includes multiple nth-child selectors and attempt to disable the warning with the appropriate ignoreFlag. The issue appears on both `@emotion/styled` and `@emotion/react`.

```ts
const Component = styled("div")`
  > *:nth-child(1)  /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */ {
    color: red;
  }

  > *:nth-child(2) /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */ {
    color: red;
  }
`
```

The workaround I found was to include the comment in the previous block. Based on some quick debugging, it seems like all the comments are getting put into the root element. I imagine this is stylis behavior. When the second block checks for the comment in the previous element, it finds the `> *:nth-child(1)` block has no comment. Putting the comment at the end of the previous element instead seems to solve the issue. The reasoning is just an educated guess 😅

```ts
const Component = styled("div")`
  > *:nth-child(1)  /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */ {
    color: red;
    /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */
  }

  > *:nth-child(2) {
    color: red;
  }
`
```

**Expected behavior:**

The error should not throw when the comment is present before the block or documentation could be provided on how to use this correctly. I couldn't find any relevant documentation about this flag in general. For reference, the positioning of the comment was based off of a comment in this issue: https://github.com/emotion-js/emotion/issues/1105#issuecomment-546299711

**Environment information:**

- `react` version: 17.0.2
- `@emotion/react` version: 11.4.1
- `@emotion/styled` version: 11.3.0

kyoncy:
**What**:
Fix #2478

> When using the ignoreFlag for `nth-child` being potentially unsafe with server rendering, the first instance of the flag works while the second does not (workaround provided).

**Why**:
To ignore unsafe selector warning.

**How**:
The current implementation only compared the previous element, so all elements are now compared.

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset N/A


srmagura (changes requested):
Nice work! This looks good to me, but let's wait for @Andarist's approval.

Can you add a changeset to record this patch-level change to `@emotion/cache`?

kyoncy:
@srmagura 
Thanks for reviewing!

> Can you add a changeset to record this patch-level change to @emotion/cache?

Add a changeset commit!

> Can you refactor the for..of to a classic for loop or a .forEach()?

I used classic for loop because it is returnable!

.changeset/witty-frogs-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/cache': patch
+---
+
+Fix ignoreFlag not works bug
```
srmagura, line 5:
```suggestion
Fix `/* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */` not disabling the warning when there are multiple blocks of CSS.
```

srmagura (approved):
Apart from making the changeset more descriptive, this looks good to me. The changesets end up in our release notes so they need to clearly explain what was changed.

packages/react/__tests__/__snapshots__/warnings.js.snap
```diff
@@ -52,6 +66,20 @@ exports[`unsafe pseudo classes does not warn when using with flag: /* emotion-di
 />
 `;
 
+exports[`unsafe pseudo classes does not warn when using with flag: /* emotion-disable-server-rendering-unsafe-selector-warning-please-do-not-use-this-the-warning-exists-for-a-reason */ ":first-child :nth-child(3) /* [flag] */" in a style multiple selectors 1`] = `
```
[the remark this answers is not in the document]
Andarist: ah, actually the full flag is part of the `describe`'s title and is literally passed in there: https://github.com/emotion-js/emotion/blob/7a107696709e63e2b65110ba265aa2c9a46ecbe9/packages/react/__tests__/warnings.js#L88

So we don't need to change anything here and that was the original intention (perhaps we could change it separately, but it doesn't matter)

packages/cache/src/stylis-plugins.js
```diff
@@ -167,9 +167,11 @@ export let createUnsafeSelectorsAlarm = cache => (element, index, children) => {
...
   if (unsafePseudoClasses && cache.compat !== true) {
-    const prevElement = index > 0 ? children[index - 1] : null
-    if (prevElement && isIgnoringComment(last(prevElement.children))) {
-      return
+    for (let i = 0; i < children.length; i++) {
+      const element = children[i]
+      if (element && isIgnoringComment(last(element.children))) {
+        return
+      }
     }
```
Andarist, line 175: Could you describe the change? What was the assumption of the previous code? How the fixed case is different and why it didn't work with the previous logic? What was the thought process behind the introduced change? I would appreciate if you could show the AST layouts of different cases so it would be easier to reason about this fix.

Andarist:
Could you check if this would fix the recently reported issue: https://github.com/emotion-js/emotion/issues/2763 ?

kyoncy:
> Could you check if this would fix the recently reported issue: #2763 ?

@Andarist 
Sorry for the late reply.
I applied the modifications and deployed the same code as in the [codesandbox example](https://codesandbox.io/s/awesome-khayyam-6u4ljy?file=/src/App.js)

deployed site: https://kyoncy.github.io/emotion-test/
repository: https://github.com/kyoncy/emotion-test/tree/main

packages/cache/src/stylis-plugins.js
kyoncy: In the previous logic, only the first ignoreFlag in a style worked.
Therefore, CI passed, but if there were multiple styles and the second or later style used unsafe pseudo classes (ex. `:first-child`), an error would be displayed even if the ignoreFlag was given.

packages/react/__tests__/__snapshots__/warnings.js.snap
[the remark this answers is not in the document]
kyoncy: > There is `.replace` call somewhere that perhaps doesn't replace all occurrences of that flag in the input string

Replaced by `/* [flag] */` here.

https://github.com/emotion-js/emotion/blob/d39379cdaec57379b44baeb9b11c5f85faa11625/packages/react/__tests__/warnings.js#L117-L120

Do you need to modify or add test cases?

Andarist:
@kyoncy thanks for taking a look. I've pushed out some commits to make this comment detection more strict - we should be able to land this soon. I will just wait a couple of days to give a chance to other maintainers to review this

srmagura (approved):

swernerx:
Hi! I know this is not a fully valid bug report. Still it seems that updating from v11.10.1 to v11.10.2 breaks in our application. I receive the error: `Cannot read properties of undefined (reading 'type') at isIgnoringComment2`. Possibly `element` can be null/undefined?

swernerx:
Seems to be related to:

```js
var isIgnoringComment = function isIgnoringComment(element) {
  if (!element) {
    console.warn("XXX:", element)
  }
  return element && element.type === 'comm' && element.children.indexOf(ignoreFlag) > -1;
};
```

which can be fixed using:

```js
var isIgnoringComment = function isIgnoringComment(element) {
  if (!element) {
    console.warn("XXX:", element)
  }
  return element && element.type === 'comm' && element.children.indexOf(ignoreFlag) > -1;
};
```

swernerx:
The code involved producing the issue is:

```
  terminatedContract: {
    '& .contractIcon, tr:nth-child(1) td:nth-child(1), tr:nth-child(n+2) td': {
      opacity: 0.4
    }
  }
```

I guess it's related to the multiple usage of `:nth-child`. When I migrate that code to:

```
  terminatedContract: {
    '& .contractIcon, tr:nth-of-type(1) td:nth-of-type(1), tr:nth-of-type(n+2) td': {
      opacity: 0.4
    }
  }
```

the issue disappears.

Andarist:
Thank you for the report - I've already merged in the fix for this and gonna release a new version in a second.

PR #2778, opened 2022-06-07
Migration docs change: import { css } from '@emotion/core'

jonlong:
**What**:
A one-line change in the migration documentation:

```diff
- import { css } from 'emotion'
+ import { css } from '@emotion/core'
```

**Why**:

I was migrating an older emotion project to version 11 and was running into an error with my `css` calls. Looks like a typo in the docs — once I imported from `@emotion/core` everything was 👍.

**How**:

Quick PR directly on GitHub.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset  N/A

### DIFF
--- a/docs/migrating-to-emotion-10.mdx
+++ b/docs/migrating-to-emotion-10.mdx
@@ -71,7 +71,7 @@ import styled, { css } from 'react-emotion'
 // ↓ ↓ ↓ ↓ ↓ ↓
 
 import styled from '@emotion/styled'
-import { css } from 'emotion'
+import { css } from '@emotion/core'
 ```
 
 - Add a `css` call to the `css` prop when a template literal is used.


docs/migrating-to-emotion-10.mdx
```diff
@@ -71,7 +71,7 @@ import styled, { css } from 'react-emotion'
 // ↓ ↓ ↓ ↓ ↓ ↓
 
 import styled from '@emotion/styled'
-import { css } from 'emotion'
```
Andarist, line 74: Actually, I think this was correct - it matches the code [here](https://github.com/emotion-js/emotion/blob/62dd16b645a55687ff094487e6c6319e1dd69f95/packages/eslint-plugin/src/rules/import-from-emotion.js#L32). The difference is really nuanced though 🤷‍♂️ As part of the incremental migration you were able to provide the cache from `'emotion'` package - that's because in Emotion 9 actual class names were returned from `css` and it wasn't an uncommon situation to leverage that.

jonlong: Thanks for the note! I think this may not have worked as expected for me because I skipped straight to v11, but used the v10 docs to help with migration — it looks like the packages were renamed in 11, so that's likely where I got tripped up.

Appreciate the clarification :)

Issue #2742, opened 2022-04-30
Typescript@beta moduleResolution nodenext strategy breaks @emotion


noahehall:
**Current behavior:**
- setting tsconfig.json.moduleResolution = node: everything is fine
- setting tsconfig.json.moduleResolution = nodenext: throws error

```sh
./src/app.tsx 13:2-15:21
[tsl] ERROR in /somedir/nirv/src/app.tsx(13,3)
      TS2307: Cannot find module '@emotion/react/jsx-runtime' or its corresponding type declarations.
```

**Expected behavior:**

- [@emotion works with typescript@beta nodenext moduleResolution strategy](https://www.typescriptlang.org/tsconfig#moduleResolution)

**Environment information:**

- `react` version:18.1
- `@emotion/react` version:18.1
- typescript: beta
- tsc: Version 4.7.0-dev.20220430
- pnpm: 6.32.9

** tsconfig **

```json

{
  "include": ["src"], 
  "exclude": ["src/**/*.test.ts"],
  "compileOnSave": false,
  "compilerOptions": {
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "allowUnreachableCode": false,
    "allowUnusedLabels": false,
    "alwaysStrict": true,
    "baseUrl": "./", 
    "checkJs": false,
    "esModuleInterop": true,
    "exactOptionalPropertyTypes": false,
    "forceConsistentCasingInFileNames": true,
    "isolatedModules": true,
    "jsx": "react-jsx",
    "jsxImportSource": "@emotion/react",
    "lib": ["dom", "dom.iterable", "esnext"],
    "module": "esnext", 
    "moduleResolution": "nodenext", <-------- breaks @emotion 
    "noFallthroughCasesInSwitch": false,
    "noImplicitAny": true,
    "noImplicitOverride": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noPropertyAccessFromIndexSignature": false,
    "outDir": "dist",
    "paths": {},
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "target": "es2017", 
    "typeRoots": ["./src/@types", "./node_modules/@types"]
  }
}
```

srmagura:
Thanks for the report, I can confirm this is an issue. Minimal repro: https://github.com/srmagura/emotion-nodenext-repro

Andarist:
[...]

PR #2785, opened 2022-06-18
fix: Add `jsx-runtime` exports field to `@emotion/react`

koistya:
Add `exports` field to `@emotion/react` in order to fix `Cannot find module 'react/jsx-runtime'` when TypeScript is pre-configured with `moduleResolution: "NodeNext"` introduced in TypeScript `v4.7`.

**Ref** https://github.com/emotion-js/emotion/issues/2742, https://github.com/microsoft/TypeScript/issues/33079

https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing
https://nodejs.org/api/packages.html#conditional-exports

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -7,6 +7,50 @@
     "./dist/emotion-react.cjs.js": "./dist/emotion-react.browser.cjs.js",
     "./dist/emotion-react.esm.js": "./dist/emotion-react.browser.esm.js"
   },
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./types/index.d.ts",
+        "browser": "./dist/emotion-react.browser.esm.js",
+        "default": "./dist/emotion-react.esm.js"
+      },
+      "require": {
+        "types": "./types/index.d.ts",
+        "browser": "./dist/emotion-react.browser.cjs.js",
+        "default": "./dist/emotion-react.cjs.js"
+      }
+    },
+    "./macro": {
+      "types": "./macro.d.ts",
+      "default": "./macro.js"
+    },
+    "./jsx-runtime": {
+      "import": {
+        "types": "./types/jsx-runtime.d.ts",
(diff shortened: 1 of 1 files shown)


Andarist:
This was being worked on [here](https://github.com/emotion-js/emotion/pull/2589) but we didn't have time lately to finish the job there.

srmagura:
I tested this out [here](https://github.com/srmagura/emotion-2785-test). The `main` branch compiles to CommonJS, and the `esm` branch compiles to ESM with `"type": "module"` in `package.json`. (I manually applied the changes from this PR to `@emotion/react`'s `package.json` in `node_modules`.)

CommonJS: works!

ESM:
```text
file:///home/srmagura/Documents/oss/bugs/emotion-2785-test/src/index.js:1
import { jsx as _jsx } from "@emotion/react/jsx-runtime";
         ^^^
SyntaxError: Named export 'jsx' not found. The requested module '@emotion/react/jsx-runtime' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@emotion/react/jsx-runtime';
const { jsx: _jsx } = pkg;
```

Which is weird because the `exports` stuff looks correct. I'm using Node 16 — that should work right?

Also a few questions for you:

1. Is it correct that we need to do this for all `@emotion` packages? Or only packages that have multiple entry points?
2. Can you think of an automated way to test this? Both for confirming that it is correct and preventing regressions going forward.

hanson-so:
Is there a workaround for this?

srmagura:
@hanson-so For now, just use `"moduleResolution": "node"` in TSConfig.

nicksrandall:
This change is also included in #2819 which uses latest version of precontruct to automatically generate the exports conditions.

PR #2791, opened 2022-06-22
Add @emotion/babel-plugin importMap example

guymoon:
**What**:

@emotion/babel-plugin importMap example.

**Why**:

I think docs example  is not enough. So I need to read plugin code. I'd like to add more explanation.

**How**:

**Checklist**:

- [x] Documentation
- [N/A] Tests
- [N/A] Code complete
- [N/A] Changeset 

Thank you!


packages/babel-plugin/README.md
```diff
@@ -301,6 +301,13 @@ This option assumes that you are using something to make `@emotion/react`'s `jsx
 
 This option allows you to tell @emotion/babel-plugin what imports it should look at to determine what it should transform so if you re-export Emotion's exports, you can still use the Babel transforms
 
+An example.js
```
Andarist, line 304:
```suggestion
An example file:
```

PR #2799, opened 2022-06-27
Fixed an issue with the lack of `/*#__PURE__*/` annotations before styled calls transformed from tagged template expressions

### LINKED ISSUE #2619: Babel plugin doesn't add PURE comment in latests babel
**Current behavior:**

The current babel plugin seems to work fine for everything except the /* #__PURE__ */ comment that should be seen.

**To reproduce:**

1. See custom repo: https://github.com/paul-sachs/emotion-babel-bug
2. Run `npm install` and `npm run build`
3. Output will not have PURE comment prefixed to styled component

**Expected behavior:**

All styled blocks should be prefixed with PURE comment

**Environment information:**

- "react": "^17.0.2",
- "@emotion/babel-plugin": "^11.7.2"
- "@emotion/core": "^11.0.0",
- "@emotion/styled": "^11.6.0",

Andarist:
fixes #2619


Andarist:
@mitchellhamilton would love a second pair of eyes on this one.

The build of the site has failed here but it's rather for an unrelated reason. There is a PR that rewrites the site from Gatsby to Next.js - I hope it will resolve such issues (and it will be easier to maintain for us).

packages/babel-plugin/src/styled-macro.js
```diff
@@ -80,36 +79,42 @@ export let styledTransformer = ({
...
     reference.replaceWith(getStyledIdentifier())
+    createStyledComponentPath = reference.parentPath
   }
 
-  if (reference.parentPath && reference.parentPath.parentPath) {
-    const styledCallPath = reference.parentPath.parentPath
+  const styledCallLikeWithStylesPath = createStyledComponentPath
+    ? createStyledComponentPath.parentPath
+    : reference.parentPath.parentPath.isTaggedTemplateExpression()
+    ? reference.parentPath.parentPath
```
mitchellhamilton, line 96: When will this case happen? If it's a valid styled call, won't `createStyledComponentPath` always be set?

Andarist: It was indeed redundant, some leftover from a WIP - I had to add this to workaround a bug in the code leading to this but it seems that I later fixed the underlying issue. I've cleaned up this redundant code and everything looks OK, thanks for noticing!

mitchellhamilton (approved):

PR #2813, opened 2022-07-07
fix: Node.js ESM support

steve-taylor:
**What**:

* Added Node.js support for ESM imports using conditional exports
* Added ESM support to `@emotion/server`

**Why**:

This monorepo contains hybrid packages, which are packages whose pacakge.json contains `"main"` and `"module"` fields that point to CommonJS and ESM entrypoints respectively. The `"module"` field is non-standard; it's generally recognized by bundlers, but not by Node.js.

**How**:

[Conditional exports](https://nodejs.org/docs/latest-v16.x/api/packages.html#packages_conditional_exports)

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] Code complete
- [X] Changeset 

This is motivated by https://github.com/styled-components/styled-components/pull/3768 as Styled Components depends on some Emotion packages.

### DIFF
--- a/.changeset/three-baboons-suffer.md
+++ b/.changeset/three-baboons-suffer.md
@@ -1,0 +1,24 @@
+---
+'@emotion/babel-plugin': patch
+'@emotion/babel-plugin-jsx-pragmatic': patch
+'@emotion/babel-preset-css-prop': patch
+'@emotion/cache': patch
+'@emotion/css': patch
+'@emotion/css-prettifier': patch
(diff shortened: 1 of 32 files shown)


Andarist:
This is a delicate issue that has many angles to it. We didn't fully decide how to approach this on our side but the first step that we are going to make is to merge this PR: https://github.com/emotion-js/emotion/pull/2589 . We can think through the next steps when that happens.

PR #2819, opened 2022-07-11
Feat: generate bundles for "worker" style environement

### LINKED ISSUE #2413: `@emotion/styled/base` imports don't work with native ES modules
**Current behavior:**

I'm trying to use [native ECMAScript modules in Node.JS](https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules) in a project that uses `@emotion/styled` and `@emotion/babel-preset-css-prop`.

When `@emotion/babel-preset-css-prop`  runs it injects an import from `@emotion/styled/base` which looks like this:

```js
import _styled from "@emotion/styled/base";
```

Since this is a directory import, and directory imports only work with CommonJS, it fails with an `ERR_UNSUPPORTED_DIR_IMPORT'` error.

**To reproduce:**

You can reproduce the issue with [this Github repository](https://github.com/matt-allan/emotion-styled-base-es-modules).

The main points are:

1. [Enable native ES modules](https://github.com/matt-allan/emotion-styled-base-es-modules/blob/b380ed7d63cfc0a6ca567c1d206059df663a69a3/package.json#L2)
2. [Use `@emotion/styled`](https://github.com/matt-allan/emotion-styled-base-es-modules/blob/b380ed7d63cfc0a6ca567c1d206059df663a69a3/src/index.js#L3)
3. [Transpile with `@emotion/babel-preset-css-prop`](https://github.com/matt-allan/emotion-styled-base-es-modules/blob/b380ed7d63cfc0a6ca567c1d206059df663a69a3/.babelrc#L11) and make sure [`@babel/preset-env` is configured with `{modules: false}`](https://github.com/matt-allan/emotion-styled-base-es-modules/blob/b380ed7d63cfc0a6ca567c1d206059df663a69a3/.babelrc#L6)
4. Try to run the resulting file with node: `node lib/index.js`

**Expected behavior:**

The import works.

**Environment information:**

- `react` version: 17.0.2
- `@emotion/react` version: 11.4.0
- `@emotion/styled` version: 11.3.0
- `@emotion/babel-preset-css-prop` version: 11.2.0
- `@emotion/babel-plugin` version: 11.3.0

I _think_ this is probably the same issue as #2121 since [Next.JS also sets `{modules: false}` for `@babel/preset-env`](https://nextjs.org/docs/advanced-features/customizing-babel-config) and is presumably trying to run Emotion in Node for SSR.

### LINKED ISSUE #2554: Add support for Cloudflare Worker environment in @emotion/cache, @emotion/styled, @emotion/react, and @emotion/server

### LINKED ISSUE #2730: Issues using emotion/styled with ES Modules
**Current behavior:**

When using `import styled from "@emotion/styled"`, in a package setup to use esm, styled has a single key `default`

**To reproduce:**

1. Clone https://github.com/maddijoyce/emotion-esm-issue
2. Run npm test

**Expected behavior:**

As in the above example, using `import styled from "@emotion/styled"`, styled should have styled.h1, styled.h2, etc

**Environment information:**

- node version: 16.13.2
- @emotion/styled version: 11.8.1

**Some additional background:**

Issue first discovered here - https://github.com/facebook/jest/issues/12571

nicksrandall:
**What**:

This is PR uses a ~~*hopefully* soon-to-be-released~~ newly released version of preconstruct that support building "worker" compatible builds. The primary use-case for this is to support Cloudflare workers. 

**Why**:
fixes #2554 
fixes #2730
fixes #2413

**How**:
I've made changes to each package in the monorepo and extended the package.json to include an "exports" field with custom conditionals for "browser" and "worker". 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset


nicksrandall:
@mitchellhamilton I fixed the spelling issue and only added the "worker" condition to packages that need it. I was sure that the spelling issue was the reason that the extra fields were not being generated. However, I have run `preconstruct fix` and precontruct build` but the worker conditions are still not being generated. 

I want to be respectful of your time so I'll keep trying to debug on my end. Thanks again for all your help with this!

mitchellhamilton:
> However, I have run preconstruct fix and precontruct build` but the worker conditions are still not being generated.

You might not be running preconstruct from the root of the repo?

nicksrandall:
@mitchellhamilton I deleted `node_modules` and then re-installed and it seems to be working now. Thanks again for all your help!

srmagura:
Looks great 😃. Question: have you only generated `exports` for specific packages? It seems like we should do this for all of our packages.

nicksrandall:
> It seems like we should do this for all of our packages.

I'm open to making this change if others agree that it would be useful. The reason I didn't initially include it is because it doesn't not actually change anything. Preconstruct builds (and therefore emotion builds) specifically target bundlers and most bundlers know how to read from the `main`, `module` and `browser` fields of package.json so when using only those conditions, it doesn't not add any value to also include the `exports` field. I only added the `exports` field where there was a *new* condition that we wanted to use and `exports` is the only way to use it (worker). 

I imagine that someday, preconstruct will be extended to support Node ESM (ie the `imports` condition) and then it would be valuable to add the exports fields to all packages.

srmagura:
Your first paragraph makes sense to me! I'm not 100% sure what you mean by:

> I imagine that someday, preconstruct will be extended to support Node ESM (ie the imports condition)

I tested your changes and came up with this patch:

```diff
diff --git a/packages/react/package.json b/packages/react/package.json
index de9470c1..7dcf755f 100644
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -8,7 +8,8 @@
   },
   "exports": {
     ".": {
-      "module": {
+      "types": "./types/index.d.ts",
+      "import": {
         "worker": "./dist/emotion-react.worker.esm.js",
         "browser": "./dist/emotion-react.browser.esm.js",
         "default": "./dist/emotion-react.esm.js"
@@ -16,7 +17,8 @@
       "default": "./dist/emotion-react.cjs.js"
     },
     "./jsx-runtime": {
-      "module": {
+      "types": "./types/jsx-runtime.d.ts",
+      "import": {
         "worker": "./jsx-runtime/dist/emotion-react-jsx-runtime.worker.esm.js",
         "browser": "./jsx-runtime/dist/emotion-react-jsx-runtime.browser.esm.js",
         "default": "./jsx-runtime/dist/emotion-react-jsx-runtime.esm.js"
@@ -24,7 +26,7 @@
       "default": "./jsx-runtime/dist/emotion-react-jsx-runtime.cjs.js"
     },
     "./_isolated-hnrs": {
-      "module": {
+      "import": {
         "worker": "./_isolated-hnrs/dist/emotion-react-_isolated-hnrs.worker.esm.js",
         "browser": "./_isolated-hnrs/dist/emotion-react-_isolated-hnrs.browser.esm.js",
         "default": "./_isolated-hnrs/dist/emotion-react-_isolated-hnrs.esm.js"
@@ -32,13 +34,15 @@
       "default": "./_isolated-hnrs/dist/emotion-react-_isolated-hnrs.cjs.js"
     },
     "./jsx-dev-runtime": {
-      "module": {
+      "types": "./types/jsx-dev-runtime.d.ts",
+      "import": {
         "worker": "./jsx-dev-runtime/dist/emotion-react-jsx-dev-runtime.worker.esm.js",
         "browser": "./jsx-dev-runtime/dist/emotion-react-jsx-dev-runtime.browser.esm.js",
         "default": "./jsx-dev-runtime/dist/emotion-react-jsx-dev-runtime.esm.js"
       },
       "default": "./jsx-dev-runtime/dist/emotion-react-jsx-dev-runtime.cjs.js"
     },
+    "./macro": "./macro.js",
     "./package.json": "./package.json"
   },
   "types": "types/index.d.ts",
```

Description of changes:

- The `"types"` key was necessary for TypeScript to find the type declarations — I'm testing with `"moduleResolution": "nodenext"`. 
- Changed `"module"` condition to `"import"`. I reviewed the [relevant Node.js documentation](https://nodejs.org/api/packages.html#conditional-exports) and did not see `"module"` listed as a valid condition. Maybe I'm missing something here?
- Added missing export for `@emotion/react/macro`.

Assuming what I did is correct, these same changes will need to be made to the other `package.json`s you modified too.

mitchellhamilton:
> I'm not a maintainer of emotion so I can not speak for the group but my understanding is that emotion will likely not support ESM node (ie the imports condition) until it is supported in preconstruct. This also means that they will likely not support moduleResolution: "nodenext" until then as well.

We should make it possible for emotion to be imported _by_ Node ESM (including with TypeScript's `"moduleResolution": "nodenext"`) but we definitely won't be _providing_ Node ESM any time soon (because shipping Node ESM isn't really beneficial for consumers)

nicksrandall:
@mitchellhamilton what would that look like? Do you feel like that should be included with this PR? Should I manually add the “imports” and “types” conditions to all packages?

mitchellhamilton:
> Do you feel like that should be included with this PR?

Yes - not doing so would be a regression from the current behavior

> Should I manually add the “imports” and “types” conditions to all packages?

We should not use the `import` condition. Using it is not necessary whatsoever for importing into Node ESM.

Rather than adding the types condition (because otherwise you'd have to duplicate all of the conditions because `preconstruct.extra` will replace not merge conditions), I'd say we should move the .d.ts files to be next to the appropriate source .js files and Preconstruct will do the right things with them (i just made this work in Preconstruct - https://github.com/preconstruct/preconstruct/releases/tag/%40preconstruct%2Fcli%402.2.0)

@Andarist do you have any thoughts on the breaking change-ness of adding the exports field for emotion? Should we just not care if people are importing stuff from dist? Should we keep the files in `/types` and add them to the exports field (probably at least will need to for `@emotion/react/types/css-prop`)? What were you thinking for the TypeScript migration in regards to the types directories?

srmagura:
I'm not an expert on this stuff but here's my 2 cents:

> do you have any thoughts on the breaking change-ness of adding the exports field for emotion? Should we just not care if people are importing stuff from dist? 

It seems fine to do this as a minor version bump. AFAIK we've never documented importing stuff directly from `dist` so it's not part of our "public API".

> Should we keep the files in /types and add them to the exports field (probably at least will need to for @emotion/react/types/css-prop)? What were you thinking for the TypeScript migration in regards to the types directories?

I vote for killing the `/types` directories, with an exception for `@emotion/react/types/css-prop` and any other `.d.ts` files people may be referencing explicitly. 

The `/types` directories still exist in the `ts-migration` branch, which feels weird to me since it's usually just

```ts
export * from '..';
```

mitchellhamilton:
> Unless we reexport what is in /types from within /src

This is what I was thinking we would do - though probably not include them in the exports field

wight554:
You might need to use .mjs extensions for esm for better compatibility, see https://github.com/sheremet-va/dual-packaging

Andarist:
I've pushed some changes to this branch. I think that the only leftover to fix is the `types` issue. I've upgraded the Preconstruct already but didn't yet have time to add those reexporting `src/*.d.ts` files and play around with it. Gonna take a look at this later.

Are there any other blockers that come to your mind @mitchellhamilton ?

mitchellhamilton (approved):

nicksrandall:
@mitchellhamilton @Andarist FWIW, I have tested this version of emotion with Cloudflare workers specifically and it works with no issues!

I think we may be ready to merge! 🎉

Andarist:
Yep, I'm a little bit scared of this PR but I'll be ripping this bandaid off soon.

PR #2835, opened 2022-07-22
feat: update @emotion/server types to support all exported members

manishco:
Looking at the index.js file within emotion/server it does export extractCriticalToChunks however they are not added to the exported types. Is this intentional as we should not use it or can we update the types to support them?

**What**:
Update types for @emotion/server 

**Why**:

To prevent typescript from complaining when I try to use `import { extractCriticalToChunks } from '@emotion/server';` 
image

**How**:

Followed the same approach as the other exported types. 

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/silent-baboons-press.md
+++ b/.changeset/silent-baboons-press.md
@@ -1,0 +1,5 @@
+---
+'@emotion/server': patch
+---
+
+Update emotion server types to support all exported members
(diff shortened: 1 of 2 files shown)


.changeset/silent-baboons-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/server': patch
+---
+
+Update emotion server types to support all exported members
```
Andarist, line 5:
```suggestion
Added missing TypeScript declarations for `extractCriticalToChunks` and `constructStyleTagsFromChunks` exports.
```

PR #2850, opened 2022-08-08
Optimizing `<Insertion />` component ✈

sanjaiyan-dev:


Andarist:
Have you checked if it actually helps with performance in any meaningful way? In most cases, the `serialized` object won't be the same and thus `memo` won't prevent any rerenders.

sanjaiyan-dev:
> Have you checked if it actually helps with performance in any meaningful way? In most cases, the `serialized` object won't be the same and thus `memo` won't prevent any rerenders.

Hi,
since `<Insertion />` is a inner component used for inserting styled to browser  I thought it would improve the performance slightly by memoizing the component. I am extremely sorry if I made any mistakes 🙇🏽

sanjaiyan-dev:
> Have you checked if it actually helps with performance in any meaningful way? In most cases, the `serialized` object won't be the same and thus `memo` won't prevent any rerenders.

And, I thought it will prevent re render if parent component re render due another reason (not changing prop values)

Andarist:
I feel like this is most likely on the verge of being a microoptimization - there is also a non-zero chance that this might actually be worse for performance (remember that using `React.memo` also adds some computation overhead). Therefore I'm going to close this because I don't see us merging this. However, I'm happy to be proven otherwise - so if you come up with a benchmark proving that this is worth it then I would reconsider this.

sanjaiyan-dev:
> I feel like this is most likely on the verge of being a microoptimization - there is also a non-zero chance that this might actually be worse for performance (remember that using `React.memo` also adds some computation overhead). Therefore I'm going to close this because I don't see us merging this. However, I'm happy to be proven otherwise - so if you come up with a benchmark proving that this is worth it then I would reconsider this.

Ok & Sorry 💪🙌

Issue #2745, opened 2022-05-02
Issue with @emotion/babel plugin - variable reported as not defined


adrianosmond:
**Current behavior:**

Given a very specific set of styles:
```
const StyledRoot = styled.div`
  ${fullWidth}

  button {
    @media (min-width: 768px) {
      color: red;
    }
  }
`;
```
the `fullWidth` variable which has been imported from a different file is reported as not defined when using @emotion/babel-plugin inside a Jest test. The styles work when @emotion/babel-plugin is not used. Making minor adjustments to the styles, like removing the `@media` rule, or changing the order of the nesting of `button` and `@media` mean also result in the variable no longer being reported as not defined.

**To reproduce:**

1. Clone https://github.com/adrianosmond/emotion-babel-plugin-issue-repro
2. `npm i`
3. `npm run test`

The tests fail like so

```
 FAIL  src/Component.test.jsx
  ● Test suite failed to run

    ReferenceError: fullWidth is not defined

      4 |
      5 | const StyledRoot = styled.div`
    > 6 |   ${fullWidth}
        |     ^
      7 |
      8 |   button {
      9 |     @media (min-width: 768px) {
```

**Expected behavior:**

The test should pass. The variable `fullWidth` has been defined - it is imported on line 3 of src/Component.jsx

**Environment information:**

- `react` version: 17..0.2
- `@emotion/styled` version: 11.8.1
- `@emotion/babel-plugin` version: 11.9.2

Andarist:
Thanks for filing a report - I've narrowed down the issue to the [`toInputTree`](https://github.com/emotion-js/emotion/blob/ae0f650b5ad8ef658b61edca6157fe208be23b43/packages/babel-plugin/src/utils/minify.js#L21-L39) function and I **think** that I've also recognized how the logic is broken. Gonna be working on a fix soon.

adrianosmond:
Awesome - thanks for all your hard work!

abel-varga-ekata:
Is there an update on this? We've bumped into this exact issue. Our workaround is to fall back and use the old `babel-plugin-emotion` when running the tests, if anyone's interested.

Andarist:
Sorry, this somehow slipped my mind back in May and I've forgotten about this issue. I've prepared a fix for this: https://github.com/emotion-js/emotion/pull/2858

PR #2858, opened 2022-08-21
Fixed an issue in the styles minifier that caused interpolations to be sometimes included more than once in the output

Andarist:
fixes https://github.com/emotion-js/emotion/issues/2745

### DIFF
--- a/packages/babel-plugin/src/utils/minify.js
+++ b/packages/babel-plugin/src/utils/minify.js
@@ -5,18 +5,10 @@ const haveSameLocation = (element1, element2) => {
   return element1.line === element2.line && element1.column === element2.column
 }
 
-const isAutoInsertedRule = element => {
-  if (element.type !== 'rule' || !element.parent) {
-    return false
-  }
-
-  let parent = element
-  do {
-    parent = parent.parent
-  } while (parent && parent.type !== 'rule')
-
-  return !!parent && haveSameLocation(element, parent)
-}
+const isAutoInsertedRule = element =>
+  element.type === 'rule' &&
+  element.parent &&
+  haveSameLocation(element, element.parent)
 
 const toInputTree = (elements, tree) => {
   for (let i = 0; i < elements.length; i++) {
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/utils/minify.js
```diff
@@ -5,18 +5,10 @@ const haveSameLocation = (element1, element2) => {
...
 
-const isAutoInsertedRule = element => {
-  if (element.type !== 'rule' || !element.parent) {
-    return false
-  }
-
-  let parent = element
-  do {
-    parent = parent.parent
-  } while (parent && parent.type !== 'rule')
```
Andarist, line 16: No idea why I've included this loop in the past - all tests pass without it. We now directly compare locations of `element` and `element.parent`.

The issue was that we temporarily replace interpolations with strings that shouldn't appear in real styles (can be seen [here](https://github.com/emotion-js/emotion/blob/ae0f650b5ad8ef658b61edca6157fe208be23b43/packages/babel-plugin/src/utils/minify.js#L133-L149)). Such output string is being processed by Stylis (that's why we replace them in the first place, so we can compile styles with Stylis).So Stylis never knows about our little hack and it could, accidentally, treat our string as part of the selector for a rule.

This was happening for input like this:
```js
  ${fullWidth}

  button {
```
Since it was transformed by us to:
```
  xxx0:xxx

  button {
```

And if we had an at-rule within such block then for a rule in it we didn't recognize a rule to be auto-inserted. This information is being utilized by [`toInputTree`](https://github.com/emotion-js/emotion/blob/ae0f650b5ad8ef658b61edca6157fe208be23b43/packages/babel-plugin/src/utils/minify.js#L21) which tries to convert the AST back to a shape representing the input string to make it possible to print it back to as a minified string.

srmagura (commented):
Looks reasonable to me but I'm going to be honest and say I don't understand the Babel plugin code at all 😂. Sorry I cannot be of more help here.

PR #2864, opened 2022-08-23
Fixed a regression that could cause a crash when looking for an ignoring comment for unsafe pseudo-classes

### LINKED ISSUE #2863: Emotion/Cache throws exeption with newest patch version (related to #2751)
**Current behavior:**

```
const isIgnoringComment = element =>
  element.type === 'comm' && element.children.indexOf(ignoreFlag) > -1
```

of `StylisPlugin` throws an exception: Cannot access `.type`; `element` is undefined.

**To reproduce:**

Code with multiple `:nth-child` in one line might produce an `undefined` element.

```js
 terminatedContract: {
    '& .contractIcon, tr:nth-child(1) td:nth-child(1), tr:nth-child(n+2) td': {
      opacity: 0.4
    }
  }
```

**Expected behavior:**

Checking whether the `element` is not null fixes the issue:

```
const isIgnoringComment = element =>
  element && element.type === 'comm' && element.children.indexOf(ignoreFlag) > -1
```

**Environment information:**

- `react` version: 18.2.0
- `@emotion/react` version: 11.10.0
- `@emotion/cache` version 11.10.2

Andarist:
fixes #2863

### DIFF
--- a/packages/react/__tests__/warnings.js
+++ b/packages/react/__tests__/warnings.js
@@ -143,7 +143,7 @@ describe('unsafe pseudo classes', () => {
           )
           .toJSON()
       ).toMatchSnapshot()
-      expect(console.error.mock.calls).toMatchInlineSnapshot(`
+      expect((console.error: any).mock.calls).toMatchInlineSnapshot(`
         [
           [
             "The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".",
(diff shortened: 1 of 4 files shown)


packages/react/__tests__/warnings.js
```diff
@@ -184,7 +307,7 @@ describe('unsafe pseudo classes', () => {
       expect(console.error).not.toBeCalled()
     })
 
-    test.only('does not warn when using the flag on the rule that preceeds a declaration', () => {
```
Andarist, line 187: My bad - I forgot to uncomment this and I've missed broken cases (tests added as part of this PR are new though, the crash didn't occur in the existing test suite but some expectations around those warnings were broken)

Issue #2738, opened 2022-04-27
Issues in React 17 with terser and minifiers (rewrite property access using dot notation) (useInsertionEffect)


hezi-bigpanda:
In the following PR:
https://github.com/emotion-js/emotion/pull/2651/files

was suggested a fix the following issue:
https://github.com/emotion-js/emotion/issues/2649

It is a great mitigation but some tools tools are rewriting it again to dot notation (for instance, terser by default)
I needed as mitigation to disable this feature, however, I think it needs to be approached better at emotion.

I will do a PR with a suggestion to fix, but maybe you would like to approach it yourself.

I think the ordinal intention is problematic since ecmascript is a *static* module system and you want to use it as dynamic.

@Andarist I am tagging you because you found the issue and maybe you will have an idea

Andarist:
Could you prepare a full runnable repro case for the problem though? I would like to check out how different tools interact in your setup

hezi-bigpanda:
@Andarist , please paste following code:
var m = [];
var t = m['f' + 'y'];
console.log(t);

Here:
https://xem.github.io/terser-online/

Look at the output and you will see my point

hezi-bigpanda:
I suggest to replace it to something like:
const useInsertionEffect = React['useInsertion'.concat('Effect')]
This way, terser can't statically minify it. Of course, it is not semantically equal since string.prototype.concat might be overridden.

I think the best would be to avoid this feature completely, but I guess it is something you won't accept

What do you think ? I will do a PR after I will test it

Andarist:
I know how Terser can optimize this - I just wonder what's your exact build pipeline that makes this a problem. Like - what happens with the Terser's output in your case? Is it handed to a bundler? Is it executed directly in a browser?

Andarist:
My usual recommendation is to **not** bundle your dependencies, if your client decides to use Emotion on their own in their projects they might end up with hard to debug problems. Any particular reason why do you chose this approach?

hezi-bigpanda:
Good point.

I can't assume that the application client bundles itself, so there is a performance consideration.
in addition, for the usage of plugins like svg loader

I believe for *internal* libraries it is a reasonable approach while for public libraries it might be better to just transpile like you suggest.

Obviously, frameworks like react and react-dom are not part of the bundle and being set as externals.

hezi-bigpanda:
[...]

PR #2867, opened 2022-08-26
Move conditional `useInsertionEffect` declarations into separate package

Andarist:
**Why?**

Sometimes people bundle Emotion in their libraries~ (for whatever reasons, one such example would be `@storybook/theming`) and while doing so their bundlers/minifiers replace our carefully crafted `React['useInsertion' + 'Effect']` with just `React.useInsertionEffect`. That, in turn, can create problems when consuming such bundles with webpack and some other tools (it's also the reason why we are using this weird pattern in the first place). Most notably, webpack errors with:
> require function is used in a way in which dependencies cannot be statically extracted

It's hard to control bundlers/minifiers to leave this pattern alone, so I think that what we can offer to people with such setups is this package that they will be able to externalize in their setups. The main reason for bundling Emotion etc is to isolate its context (create its copy) to avoid conflicts with 3rd party code and stuff like that - and it's not to actually make a copy of everything.

fixes https://github.com/emotion-js/emotion/issues/2738


packages/use-insertion-effect-with-fallbacks/src/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+const syncFallback = create => create()
+
+const useInsertionEffect = React['useInsertion' + 'Effect']
+  ? React['useInsertion' + 'Effect']
+  : false
+
+export const useInsertionEffectAlwaysWithSyncFallback = !isBrowser
+  ? syncFallback
+  : useInsertionEffect || syncFallback
```
Andarist, line 13: Please be extra careful when reviewing this as the shape of this has slightly changed when I moved this. I think that the current code is functionally equivalent to the old version though.

packages/use-insertion-effect-with-fallbacks/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@emotion/use-insertion-effect-with-fallbacks",
```
Andarist, line 2: Should this be `*-with-fallback` or `*-with-fallbacks`? :sweat_smile: #realproblems

gyhaLabs:
I have this very similar issue in our project. We are using a common UI library (of our own) - which uses material UI with @emotion/react - which may be a very common combo. (See the installation doc for the material UI: https://mui.com/material-ui/getting-started/installation/).

When I am including this common/shared UI package in the main project, it won't do the build/deployment and the GitLab pipline fails with the error below.

image

This is a valid use-case, right? It is a similar issue with the above mentioned, right @Andarist?

Thanks.

Andarist:
@gyhaLabs I think this is very much related. My strong suspicion is that you are **bundling** Emotion (effectively creating its copy) and that, together with your UI library, is later consumed together by your consumers.

packages/use-insertion-effect-with-fallbacks/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "version": "1.0.0",
```
mitchellhamilton, line 3:
```suggestion
  "version": "0.0.0",
```

+ a changeset

mitchellhamilton (approved):

PR #2870, opened 2022-08-29
Upgrade Yarn

Andarist:


package.json
```diff
@@ -23,7 +23,7 @@
     "preinstall": "node ./scripts/ensure-yarn.js",
     "postinstall": "preconstruct dev && manypkg check",
     "changeset": "changeset",
-    "version-packages": "changeset version && yarn --no-immutable",
+    "version-packages": "changeset version && yarn --mode=\"update-lockfile\"",
```
Andarist, line 26: It has been suggested as the faster/better option for this use case (which still isn't 100% safe as it could potentially change unrelated sections of the lockfile but there is no solution for this problem)

package.json
```diff
@@ -244,7 +244,7 @@
     "unified": "^6.1.6",
     "webpack-bundle-analyzer": "3.3.2"
   },
-  "packageManager": "yarn@3.2.2",
+  "packageManager": "yarn@3.2.3",
```
Andarist, line 247: This is supposed to fix [our patch](https://github.com/emotion-js/emotion/tree/f2608b3a542164a3622453f306e2b782929a3020/.yarn/patches) not being correctly applied. Note that the patch itself could be removed if we'd upgrade Jest.

srmagura (approved):

Issue #2878, opened 2022-09-03
eslint: Unknown property 'css' found react/no-unknown-property


chrisweb:
**Current behavior:**

eslint react plugin (eslint-plugin-react) emits error that the "css" property is unknown

the rule documentation is here: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md

the error appears if you use a version of eslint-plugin-react > 7.31.1

**To reproduce:**

https://codesandbox.io/s/interesting-raman-2ko96i?file=/src/App.tsx

I did not use the emotion codesandbox template but created a new codesandbox with nodejs as environment, to have a terminal and to be able to run the "npm run lint" command in that terminal (I think you need to be logged in codesanbox to run commands in the terminal)

In the App.tsx I listed the steps I did to create this example

**Expected behavior:**

no linting error "Unknown property 'css' found react/no-unknown-property"

**Environment information:**

    "@emotion/css": "11.10.0",
    "@emotion/react": "11.10.4",
    "@emotion/styled": "11.10.4",
    "react": "18.2.0",
    "react-dom": "18.2.0",
    "eslint": "8.23.0",
    "eslint-plugin-react": "7.31.4",

As suggested by ljharb, in the [issue 3384](https://github.com/jsx-eslint/eslint-plugin-react/issues/3384#issuecomment-1236136983) over at eslint-plugin-react, here is a ticket for emotion js

ljharb:
Namely, that emotion should at least support, and ideally require, the attribute name to have a data- prefix, instead of just “css”.

srmagura:
You should configure that ESLint rule to ignore the `css` prop. The rule is not smart enough to know that you are using `@emotion/react` to handle the `css` prop before it reaches the DOM element.

> Namely, that emotion should at least support, and ideally require, the attribute name to have a data- prefix, instead of just “css”.

Allowing/requiring users to write `data-css={...}` would be misleading, because the value of the `css` prop is not placed on the DOM element as a data attribute.

ljharb:
It’s already misleading, because DOM elements can’t have a css attribute.

chrisweb:
[...]

PR #2882, opened 2022-09-04
adds documentation for eslint plugin react

chrisweb:
**What**:

adds documentation for eslint plugin react

**Why**:

To help users that use both emotion and eslint with the react plugin (> 7.31.1) to fix the linting problem they will encounter, more about it in this ticket: https://github.com/emotion-js/emotion/issues/2878#issuecomment-1236384580

**How**:

After filling a ticket https://github.com/emotion-js/emotion/issues/2878#issuecomment-1236384580 to demonstrate and discuss the problem, I created the different eslint configuration files locally and tested them in my project to verify the described solution works, then I wrote the documentation based on them

**Checklist**:

- [x] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/docs/eslint-plugin-react.mdx
+++ b/docs/eslint-plugin-react.mdx
@@ -1,0 +1,36 @@
+---
+title: 'ESLint Plugin React'
+---
+
+The eslint rule react/no-unknown-property from the [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react) package, with a version > 7.31.1, got a fix which now triggers an eslint error if the emotion css property is used on an element.
+
+To fix this linting error one solution is to add the following rule to your eslintrc.* configuration file:
+
+eslintrc.js example:
+
+```js
+    "rules": {
+        "react/no-unknown-property": ['error', { ignore: ['css'] }]
+    }
+```
+
+eslintrc.json example:
+
+```json
+    "rules": {
+        "react/no-unknown-property": ["error", { "ignore": ["css"] }]
+    },
+```
+
+eslintrc.yaml example:
+
+```yaml
+rules:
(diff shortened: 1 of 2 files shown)


srmagura:
Thanks @chrisweb. I improved the writing some / removed 2 of the ESLint config examples since I felt they were redundant.

chrisweb:
ok thank you for checking it out

personally I would have kept the yaml example, because I sometimes find it hard to find the correct way of converting a rule into yaml when all you have is the json version (if maybe you are not super experienced with yaml object / array notations but your team used yaml for all configuration files), this is why I added that example to ensure it is really a no brainer for anyone to get this issue fixed quickly... but it is your project, I can see why you removed it and won't be angry if you keep it out ;)

docs/eslint-plugin-react.mdx
```diff
@@ -0,0 +1,13 @@
...
+---
+
+The [`react/no-unknown-property` rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md) from [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react) will produce an error if the `css` prop is passed to a DOM element. This violation of the rule can be safely ignored because `@emotion/react` intercepts the `css` prop before it is applied to the DOM element.
+
+The rule can be configured to ignore the `css` prop like so:
+
+```json
+"rules": {
+  "react/no-unknown-property": ["error", { "ignore": ["css"] }]
+},
```
Andarist, line 12:
```suggestion
{
  "rules": {
    "react/no-unknown-property": ["error", { "ignore": ["css"] }]
  }
}
```

PR #2884, opened 2022-09-07
Support typings for Jest v29

billyjanitsch:
**What**:

Similar to https://github.com/emotion-js/emotion/pull/2456 and https://github.com/emotion-js/emotion/pull/2771. Allows `@emotion/jest` to be compatible with the typings for Jest v29.

**Why**:

See https://jestjs.io/blog/2022/08/25/jest-29. No relevant breaking changes, so it should be fine to permit the additional typing peer dep.

**How**:

Expanded the peer dep range.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/brave-scissors-smile.md
+++ b/.changeset/brave-scissors-smile.md
@@ -1,5 +1,5 @@
 ---
-'@emotion/jest': minor
+'@emotion/jest': patch
 ---
 
-Support typings for Jest v29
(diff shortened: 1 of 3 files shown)


billyjanitsch:
Thanks for the quick turnaround, @srmagura! 🙂

PR #2890, opened 2022-09-12
Fork Stylis prefixer

Andarist:
I'm not super happy about this but we need to upgrade Stylis to leverage some a fix for incorrect CSS variable prefixing. At the same time, we already get some grief about the included prefixer that prefixes flexbox by default (and more). I think that prefixing the grid would only compound the problem and I don't think that we actually want to prefix those properties (and we can't selectively control what gets prefixed and what doesn't). 

cc @siriwatknp, this would fix your issue

### DIFF
--- a/.changeset/long-pianos-swim.md
+++ b/.changeset/long-pianos-swim.md
@@ -1,9 +1,2 @@
 ---
-"@emotion/babel-plugin": patch
-"@emotion/cache": patch
-"@emotion/css-prettifier": patch
-"@emotion/css": patch
-"@emotion/jest": patch
 ---
-
-Updated the underlying parser ([stylis](https://github.com/thysultan/stylis)).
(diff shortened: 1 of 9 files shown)


mitchellhamilton (approved):

.changeset/long-pianos-swim.md
```diff
@@ -0,0 +1,9 @@
+---
+"@emotion/babel-plugin": patch
+"@emotion/cache": patch
+"@emotion/css-prettifier": patch
+"@emotion/css": patch
+"@emotion/jest": patch
+---
+
+Updated the underlying parser ([stylis](https://github.com/thysultan/stylis)).
```
Andarist, line 9:
```suggestion
---
---
```

siriwatknp:
> this would fix your issue

@Andarist what issue is it?

Andarist:
@siriwatknp this one: https://github.com/thysultan/stylis/issues/301 . It got fixed in Stylis in September but I have only released an Emotion version with that fix just now

PR #2899, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   test:
     name: 'Tests on ${{matrix.platform}}'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,8 +8,13 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions: {}
 jobs:
   release:
+    permissions:
+      contents: write # to create release 
+      pull-requests: write # to create pull request 
+
     name: Release
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,7 @@ jobs:
   release:
     permissions:
       contents: write # to create release 
+      issues: write # to post issue comments
       pull-requests: write # to create pull request 
 
     name: Release


.github/workflows/release.yml
```diff
@@ -8,8 +8,13 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions: {}
```
Andarist, line 11: q: is this required if explicit permissions are specified for one of the jobs?

.github/workflows/main.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
```
Andarist, line 10: shouldn't this only be granted for `actions/checkout`?

.github/workflows/release.yml
```diff
@@ -8,8 +8,13 @@ on:
...
 jobs:
   release:
+    permissions:
+      contents: write # to create release 
+      pull-requests: write # to create pull request
```
Andarist, line 16: are you a Changeset user? did you have a chance to confirm that those are the only ones needed for this action?

sashashura: It is optional. I added it prevent the case when a new job is added and it inherits the default permissions, but as long as there is only one job it will use it's own permissions.

sashashura: I didn't have a chance to try it. If it fails it can be expanded.

Andarist, line 16:
```suggestion
      issues: write # to post issue comments
      pull-requests: write # to create pull request 
```

PR #2905, opened 2022-09-29
Check vitest global for @emotion/react loaded twice warning

p-m-p:
**What**:
`vitest` shows the same symptoms with the duplicate loading when mocking out modules that include `@emotion/react`.

**Why**:
Stop logging warnings during test runs.

**How**:
Include check for the `vi` global as well as `jest`. I had to add a flow type for `vi` but this is not something I'm familiar with so happy to do that properly with guidance.

**Checklist**:
- [x] Documentation (N/A)
- [x] Tests (N/A)
- [x] Code complete (N/A)
- [x] Changeset

### DIFF
--- a/.changeset/loud-dancers-return.md
+++ b/.changeset/loud-dancers-return.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Include vitest global check when omitting the warning about duplicate instantiation in mocked modules. Will only capture vitest global `vi` if globals are configured.
--- a/.changeset/loud-dancers-return.md
+++ b/.changeset/loud-dancers-return.md
@@ -2,4 +2,4 @@
 '@emotion/react': patch
 ---
 
-Include vitest global check when omitting the warning about duplicate instantiation in mocked modules. Will only capture vitest global `vi` if globals are configured.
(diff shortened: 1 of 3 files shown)


.changeset/loud-dancers-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Include vitest global check when omitting the warning about duplicate instantiation in mocked modules. Will only capture vitest global `vi` if globals are configured.
```
Andarist, line 5:
```suggestion
Do not warn about `@emotion/react` being loaded twice in Vitest as that might easily happen with mocked modules.
```

packages/react/src/index.js
```diff
@@ -17,9 +17,9 @@ export { default as css } from './css'
 if (process.env.NODE_ENV !== 'production') {
   const isBrowser = typeof document !== 'undefined'
   // #1727 for some reason Jest evaluates modules twice if some consuming module gets mocked with jest.mock
```
Andarist, line 19:
```suggestion
  // #1727, #2905 for some reason Jest and Vitest evaluate modules twice if some consuming module gets mocked
```

PR #2393, opened 2021-05-28
Do not log failed rule insertions in the speedy mode for more vendor-prefixed pseudo-elements/classes

layershifter:
**What**:
This PR expands a fix that was made in #2149 to swallow errors from invalid insertions.

**Why**:
I added to the list to silence warnings:
- `-moz-focus-inner`
- `-moz-focusring`
- `-ms-clear`

**How**:

Tested via `yarn lint`/`yarn test` & browser.

**Checklist**:

- [x] Code complete
- [x] Changeset

PR #2920, opened 2022-10-12
Add -ms-expand and -ms-reveal to ignored prefixes

pvdstel:
**What**: adds the `-ms-expand` and `-ms-reveal` to ignored vendor prefixes; similar to #2393.

**Why**: browsers that do not support these vendor prefixes (i.e. Chrome and Firefox) will cause error messages to be displayed in the console. The prefixes themselves are fairly common:

- `-ms-expand` is used to hide the browser's dropdown arrow when styling `select` elements, in Internet Explorer, Edge (both old and Chromium-based). This pseudo element is also used in [MUI Core](https://github.com/mui/material-ui/blob/f27694b0ec007c879f8c89eba73bd627ea5c7af0/packages/mui-material/src/NativeSelect/NativeSelectInput.js#L36).
- `-ms-reveal` is often used to hide the password reveal button, in password inputs. This pseudo-element selector works in IE10+ and in Edge browsers as well.

**How**: the pseudo-elements were added to the the regex of selectors for which no error is emitted.

**Checklist**:

- [ ] Documentation (N/A ; there does not seem to exist any documentation on this)
- [ ] Tests (N/A ; there seem to be no existing tests for this, and I'm not sure how to properly test this in the first place)
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/olive-ladybugs-fry.md
+++ b/.changeset/olive-ladybugs-fry.md
@@ -2,4 +2,4 @@
 '@emotion/sheet': patch
 ---
 
-Add -ms-expand and -ms-reveal to ignored prefixes
+Do not log failed rule insertions in the speedy mode for even more vendor-prefixed pseudo-elements/classes like `:-ms-expand` and `:-ms-reveal`.
--- a/.changeset/olive-ladybugs-fry.md
+++ b/.changeset/olive-ladybugs-fry.md
@@ -1,0 +1,5 @@
+---
+'@emotion/sheet': patch
+---
+
+Add -ms-expand and -ms-reveal to ignored prefixes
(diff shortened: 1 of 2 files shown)


.changeset/olive-ladybugs-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/sheet': patch
+---
+
+Add -ms-expand and -ms-reveal to ignored prefixes
```
Andarist, line 5:
```suggestion
Do not log failed rule insertions in the speedy mode for even more vendor-prefixed pseudo-elements/classes like `:-ms-expand` and `:-ms-reveal`.
```

Issue #2642, opened 2022-02-17
PrettyFormatPluginError: Cannot delete property 'className' of #<Object>


Fafruch:
**Current behavior:**

Components that have a static `defaultProps` (with `className` set to `''`) cannot be serialized using the `@emotion/jest/enzyme-serializer` (and probably also with `@emotion/jest/serializer` as it's using shared logic).

The tests are throwing:
```
PrettyFormatPluginError: Cannot delete property 'className' of #<Object>
TypeError: Cannot delete property 'className' of #<Object>

      at clean (node_modules/@emotion/jest/dist/create-serializer-3afd473f.cjs.dev.js:506:19)
      at Object.serialize (node_modules/@emotion/jest/dist/create-serializer-3afd473f.cjs.dev.js:537:5)
      at Object.serialize (node_modules/@emotion/jest/dist/create-enzyme-serializer-c92dc591.cjs.dev.js:31:32)
      at printPlugin (node_modules/pretty-format/build/index.js:330:16)
      at printer (node_modules/pretty-format/build/index.js:379:12)
      at node_modules/pretty-format/build/plugins/lib/markup.js:73:13
          at Array.map (<anonymous>)
      at printChildren (node_modules/pretty-format/build/plugins/lib/markup.js:67:6)
      at Object.serialize (node_modules/pretty-format/build/plugins/ReactElement.js:144:35)
      at printPlugin (node_modules/pretty-format/build/index.js:330:16)
      at printer (node_modules/pretty-format/build/index.js:379:12)
      at Object.serialize (node_modules/@emotion/jest/dist/create-serializer-3afd473f.cjs.dev.js:539:22)
      at Object.serialize (node_modules/@emotion/jest/dist/create-enzyme-serializer-c92dc591.cjs.dev.js:31:32)
      at printPlugin (node_modules/pretty-format/build/index.js:330:16)
      at printer (node_modules/pretty-format/build/index.js:379:12)
      at node_modules/pretty-format/build/plugins/lib/markup.js:33:21
          at Array.map (<anonymous>)
      at printProps (node_modules/pretty-format/build/plugins/lib/markup.js:31:6)
      at Object.serialize (node_modules/pretty-format/build/plugins/ReactTestComponent.js:45:36)
      at printPlugin (node_modules/pretty-format/build/index.js:330:16)
      at printer (node_modules/pretty-format/build/index.js:379:12)
[...]

PR #2959, opened 2022-12-21
Test case that breaks the cleanup of empty classname

jeremy-davis-sonarsource:
**What**: This is a breaking test-case for issue #2642 

I'm not expecting this to be merged, I will link it to the issue as a reproducer, if that's alright!

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [ ] Code complete N/A
- [ ] Changeset N/A


shannonhochkins:
Why isn't this merged? - i have the same issue and this code will fix it for me as well

jeremy-davis-sonarsource:
> Why isn't this merged? - i have the same issue and this code will fix it for me as well, @jeremy-davis-sonarsource did you ever find another solution for this?

We simply got rid of the empty `classname` props. Fortunately, there weren't too many of them (2~3).

PR #2985, opened 2023-02-09
Remove peer dependency on `@babel/core` from most packages

### LINKED ISSUE #2660: Unnecessary `@babel/core` peer dependency?
**Current behavior:**

`@babel/core` is a peerDependency of `@emotion/styled`, `@emotion/css` and `@emotion/react`.

**To reproduce:**

Install one of the above packages when babel is not used, and your dependency manager will likely issue a warning.

**Expected behavior:**

Nothing in those packages seems to be actively depending on Babel. When using a Vite/esbuild environment, Babel may not be part of the dependencies.

**Environment information:**

React 18 RC, latest `@emotion/react`.

### LINKED ISSUE #2921: undocumented peer dependencies (babel)
**Current behavior:**

When we install projects with pnpm then we get an error due to missing peer dependencies that are not mentioned in your documentation:

image

https://emotion.sh/docs/introduction#react

image

**To reproduce:**

Install `@emotion/react` via pnpm instead of yarn or npm.

**Expected behavior:**

**Environment information:**

- `react` version:
- `@emotion/react` version: 11.10.4

emmatown:
Closes #2921
Closes #2660

So the only reason there is a peer dependency on `@babel/core` is because of `@babel/plugin-syntax-jsx`. I don't really think doing we even need to make `@emotion/babel-plugin` inherit `@babel/plugin-syntax-jsx` tbh since the only case it would matter is if people were running `@emotion/babel-plugin` without a JSX transform but just to not change the behaviour, I've inlined `@babel/plugin-syntax-jsx`. I think the cost of inlining it is more than worth it for getting rid of a peer dependency warning for most users.

### DIFF
--- a/packages/babel-plugin/package.json
+++ b/packages/babel-plugin/package.json
@@ -18,7 +18,6 @@
   ],
   "dependencies": {
     "@babel/helper-module-imports": "^7.16.7",
-    "@babel/plugin-syntax-jsx": "^7.17.12",
     "@babel/runtime": "^7.18.3",
     "@emotion/hash": "^0.9.0",
     "@emotion/memoize": "^0.8.0",
@@ -29,9 +28,6 @@
     "find-root": "^1.1.0",
     "source-map": "^0.5.7",
     "stylis": "4.1.3"
-  },
-  "peerDependencies": {
-    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.18.5",
--- a/.changeset/healthy-squids-wait.md
+++ b/.changeset/healthy-squids-wait.md
@@ -1,0 +1,10 @@
+---
+'@emotion/babel-plugin': patch
+'@emotion/css': patch
+'@emotion/native': patch
+'@emotion/primitives': patch
+'@emotion/react': patch
+'@emotion/styled': patch
+---
+
+Remove peer dependency on `@babel/core`
--- a/packages/babel-plugin/src/index.js
+++ b/packages/babel-plugin/src/index.js
@@ -1,5 +1,4 @@
 // @flow
-import syntaxJsx from '@babel/plugin-syntax-jsx'
 import {
   createEmotionMacro,
   transformers as vanillaTransformers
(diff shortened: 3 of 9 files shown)


packages/babel-plugin/package.json
```diff
@@ -18,7 +18,6 @@
   ],
   "dependencies": {
     "@babel/helper-module-imports": "^7.16.7",
```
Andarist, line 20: this depends on `@babel/types` but I guess that it should be mostly OK since those are just factories for simple objects and they shouldn't rely on identity of any class or anything like that

packages/babel-plugin/package.json
```diff
@@ -30,9 +29,6 @@
     "source-map": "^0.5.7",
     "stylis": "4.1.3"
   },
-  "peerDependencies": {
-    "@babel/core": "^7.0.0"
-  },
```
Andarist, line 35: This feels off but I guess that it's not a big deal if we omit it from here - we mostly use Babel through what we are given by the caller anyway.

Andarist:
What do you think about moving macros completely out of the main packages in Emotion 12 (not that I currently plan to release a new version)?

emmatown:
Maybe? Could just make the babel plugin an optional peer dependency though? (Note currently @babel/core isn't an optional peer dependency regardless of those peerDependenciesMeta since it's a non-optional peerDependency of @babel/plugin-syntax-jsx)

Andarist:
> Note currently @babel/core isn't an optional peer dependency regardless of those peerDependenciesMeta since it's a non-optional peerDependency of @babel/plugin-syntax-jsx)

Hm, isn't this a problem with the package manager? If we optionally depend in A on B that depends non-optionally on C that shouldn't transitively make C a required dependency of A 🤔

emmatown:
We non-optionally depend on B in this case

packages/babel-plugin/package.json
Andarist: i'm also still curious though - is this removal technically needed to fix the mentioned issues? Isn't inling`@babel/plugin-syntax-jsx` already sufficient?

Andarist:
I don't get it... `@emotion/react` has an optional peer dependency on `@emotion/babel-plugin` and that has a non-optional dependency on `@babel/plugin-syntax-jsx`. I don't understand how that non-optional dependency should be relevant for `@emotion/react` when it's only "pulled in" through an optional peer dep.

emmatown:
`@emotion/react` has a normal (non-peer) dependency on `@emotion/babel-plugin`

Andarist:
> @emotion/react has a normal (non-peer) dependency on @emotion/babel-plugin

Ah, I see... I was focusing on the `@babel/core` itself.

packages/babel-plugin/package.json
Andarist: [This comment](https://github.com/emotion-js/emotion/pull/2985#issuecomment-1429222243) indirectly answers this question. This is somewhat needed today because we have a direct dependency on `@emotion/babel-plugin` in some packages.

Andarist (approved):

michaelgmiller1:
Amazing this was fixed!! This will get rid of our last `peerDependency` warning. Hoping for a version cut soon 🤞

PR #3006, opened 2023-03-07
bug: Nested self enclosed styled component throws error with shallow render

ismailcherri:
**What**:

We encountered an issue after upgrading to `@emotion/jest` v11.10.6. We were getting the following error:
```
Cannot read properties of null (reading 'slice')
TypeError: Cannot read properties of null (reading 'slice')
    at Object.map (.../node_modules/@emotion/jest/dist/create-enzyme-serializer-4de3b073.cjs.dev.js:17:33)
    at applyMap (.../node_modules/enzyme-to-json/utils.js:45:20)
    at internalNodeToJson (.../node_modules/enzyme-to-json/shallow.js:57:36)
```
I tried debugging the error and found that it is thrown in a very specific case when there is a self-enclosed styled component nested under another element.

**Why**:

I'm not entirely sure if this is an error related to `@emotion/jest` or `enzyme-to-json` packages but I'm providing this PR as a proof of concept and a way to recreate the bug.
I also provided a simple fix to point out where the error is thrown.

**How**:

In order to reproduce:
1. Create a component with a nested self-enclosed styled component
2. Use `shallow` from `enzyme`
3. run a snapshot test

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/packages/jest/src/create-enzyme-serializer.js
+++ b/packages/jest/src/create-enzyme-serializer.js
@@ -15,7 +15,7 @@ const enzymeToJsonSerializer = createEnzymeToJsonSerializer({
       return json
     }
     const isRealStyled = json.node.type.__emotion_real === json.node.type
-    if (isRealStyled) {
+    if (isRealStyled && json.children) {
       return {
         ...json,
         children: json.children.slice(-1)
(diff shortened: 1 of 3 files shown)


ismailcherri:
Closing as it seems that this PR didn't get any kind of attention

Issue #2973, opened 2023-01-22
@emotion/babel-plugin is broken under module imports


otakustay:
**Current behavior:**

When an ESM compatible module resolver loads `@emotion/babel-plugin`, it resolves to `emotion-babel-plugin.esm.js`, which fails to use `createMacro` function because it is `undefined` in ESM imports.

```
TypeError: createMacro is not a function
 ❯ createTransformerMacro node_modules/@emotion/babel-plugin/dist/emotion-babel-plugin.esm.js:754:38
    752|
    753|     if (/\/macro$/.test(source)) {
    754|       path.get('source').replaceWith(babel.types.stringLiteral(source.replace(/\/macro$/, '')));
       |                                      ^
    755|     }
    756|
 ❯ node_modules/@emotion/babel-plugin/dist/emotion-babel-plugin.esm.js:1111:17
 ❯ index.test.js:3:31
```

**To reproduce:**

1. Clone https://github.com/otakustay/issues: `git clone git@github.com:otakustay/issues.git`
2. Switch to `emotion-babel-plugin-macro` branch: `git checkout emotion-babel-plugin-macro`
3. Install dependencies: `npm install`
4. Run test: `vitest run`

Vitest (vite as its core) will recognize `module` field in `package.json` of `@emotion/babel-plugin` and resolves to
 `emotion-babel-plugin.esm.js`, in this case `import { createMacro } from 'babel-plugin-macros';` will results undefined, this should be something like this:

```ts
import macros from 'babel-plugin-macros';
const {createMacro} = macros.default;
```

**Expected behavior:**

The test should pass

**Environment information:**

- `react` version: No react is required in this case
- `@emotion/react` version: `11.10.5`

Issue #2927, opened 2022-10-23
Trying to use `macro` throws `default.a.div is not a function` error


Wassap124:
**Current behavior:**
Trying to use component selectors on CRA (with typescript)
instead of ejecting, I tried using the `macro` path.

importing emotion via
```ts
import styled from '@emotion/styled/macro';
```

and then using it like so:
```ts
const FlexedDiv = styled.div`
  display: flex;
  justify-content: space-evenly;
  visibility: hidden;
`;
```

Throws:
```
TypeError: _emotion_styled_macro__WEBPACK_IMPORTED_MODULE_14___default.a.div is not a function
```

I am also receiving some warning on my terminal which might help:
```
../node_modules/.pnpm/babel-plugin-macros@3.1.0/node_modules/babel-plugin-macros/dist/index.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

../node_modules/.pnpm/import-fresh@3.3.0/node_modules/import-fresh/index.js
Critical dependency: the request of a dependency is an expression
```
**To reproduce:**

Could not reproduce on a separate repo 😞 

**Expected behavior:**
For `emotion` to work normally, as if Im using the via "default" path

**Environment information:**
**Important Note**: I am using pnpm, which might affect things

-  "react": "16.14.0",
-  "@emotion/css": "^11.5.0"
-  "@emotion/react": "11.10.4",
-  "@emotion/styled": "11.3.0"
- "pnpm": "6.35.0"
- "typescript": "4.8.4",

nicolasLuduena:
Did you find a workaround for this?

usback:
i have same problem so did you fixed it????

nicolasLuduena:
I haven't tried that hard really. For my usecase, I had to go with classNames

Issue #2582, opened 2021-12-08
Awkward ES6 imports


RobertSandiford:
**Current behavior:**

Importing createCache from @emotion/cache in an ES6 environment (type: module in package.json) does not import the function, but an object with a default property is the function 

```
import createCache from '@emotion/cache'

const cache =  createCache({
    key: 'css',
})
```

Output:
```
TypeError: createCache is not a function
```

package.json
```
{
  "name": "emotion-cache-es6-imports",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "type": "module",
  "dependencies": {
    "@emotion/cache": "^11.6.0"
  }
}
```

**To reproduce:**

Repro repo: https://github.com/RobertSandiford/emotion-cache-es6-imports

```
yarn
node index
```

**Expected behavior:**

The create cache function is exported to createCache

**Other notes:**

When packed with webpack the above code will work.

Getting the create cache function in ES6 requires accessing the default property.

This requires code like this to work server side + client side.
```
import emotionCache from '@emotion/cache'
const createCache = ('default' in emotionCache ) ? emotionCache.default : emotionCache

const cache = createCache({})
```

It's likely that this issues exists in other places too. I had issues with @emotion/server/create-instance too. Currently I have this code

```
import emotionServerCreateInstance from '@emotion/server/create-instance/dist/emotion-server-create-instance.cjs.js'
const { default: createEmotionServer } = emotionServerCreateInstance
```

**Environment information:**

- `@emotion/cache` version: 11.6
- Node: v16.13.0

Andarist:
We are aware of the "issue". The way how node has implemented support for modules creates a ripple effect on the whole ecosystem - what worked for years in other tools suddenly doesn't work in node. 

We'll think about solutions for this, but the problem is that it's very hard to pull off a solution that would just work for every possible combination of environments and tools. And on top of that making changes around this might require a major bump.

RobertSandiford:
@Andarist 

Sure. I would try to help, but I don't know the details of how node handles it.

The way that @mui handles it is with an index.js file which acts as a giant barrel file, exporting everything for ES6. CJS is handled via "main" while points to a cjs files in asubdirectory. I don't know if that's helpful to you.

rtritto:
Any update?

eagleera:
any update on this?

PR #3029, opened 2023-05-03
Fixed importing in Node ESM

Andarist:
fixes https://github.com/emotion-js/emotion/issues/2973
fixes https://github.com/emotion-js/emotion/issues/2927
fixes https://github.com/emotion-js/emotion/issues/2730
fixes https://github.com/emotion-js/emotion/issues/2582

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -53,7 +53,13 @@
       "default": "./create-instance/dist/emotion-css-create-instance.cjs.js"
     },
     "./package.json": "./package.json",
-    "./macro": "./macro.js"
+    "./macro": {
+      "types": {
+        "import": "./macro.d.mts",
+        "default": "./macro.d.ts"
+      },
+      "default": "./macro.js"
+    }
   },
   "preconstruct": {
     "umdName": "emotion",
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -9,9 +9,7 @@
     "src",
     "dist",
     "types",
-    "macro.js",
-    "macro.d.ts",
-    "macro.js.flow",
+    "macro.*",
     "create-instance"
   ],
   "scripts": {
@@ -65,7 +63,13 @@
     ],
     "exports": {
       "extra": {
-        "./macro": "./macro.js"
+        "./macro": {
+          "types": {
+            "import": "./macro.d.mts",
+            "default": "./macro.d.ts"
+          },
+          "default": "./macro.js"
+        }
       }
     }
   }
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -46,10 +46,12 @@
   "exports": {
     ".": {
       "module": "./dist/emotion-css.esm.js",
+      "import": "./dist/emotion-css.cjs.mjs",
       "default": "./dist/emotion-css.cjs.js"
     },
     "./create-instance": {
       "module": "./create-instance/dist/emotion-css-create-instance.esm.js",
+      "import": "./create-instance/dist/emotion-css-create-instance.cjs.mjs",
       "default": "./create-instance/dist/emotion-css-create-instance.cjs.js"
     },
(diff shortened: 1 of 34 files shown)


emmatown (commented):
Do we care that this is kind of breaking in that you previously could have done the below thing?

```ts
styled.default(...)
```

Maybe it's fine since that would also be broken in many bundler environments so the reliable thing would have been to do `styled.default || styled`?

Though it wouldn't be that hard to do `styled.default = styled` and then nothing would break, idk wdyt?

Andarist:
> Do we care that this is kind of breaking in that you previously could have done the below thing?

I don't think so. Yes, somebody could depend on this behavior - but we never considered it to be alright. Nothing like that is in our docs. I consider this PR to be a bug fix but, of course, the lines is quite blurry here

> Though it wouldn't be that hard to do styled.default = styled and then nothing would break, idk wdyt?

It would work... but then it would likely complain at type-level about this stuff (which reminds... I probably need to test this PR with `moduleResolution: bundler`, I forgot that Emotion isn't authored in TypeScript so Preconstruct isn't responsible for emitting correct types here).

I'm rather against this though. There are probably not that many ESM early adopters out there that we should take this into a consideration here.

emmatown:
> Preconstruct isn't responsible for emitting correct types here

Because there's are `.d.ts` files next to the js entrypoints, Preconstruct is kinda responsible for it (though it just really copies things)

emmatown (approved):
assuming you checked it in moduleResolution: bundler & maybe nodenext as well, looks good

packages/css/package.json
```diff
@@ -63,7 +69,13 @@
     ],
     "exports": {
       "extra": {
-        "./macro": "./macro.js"
+        "./macro": {
+          "types": {
+            "import": "./macro.d.mts",
+            "default": "./macro.d.ts"
```
emmatown, line 75:
```suggestion
```

packages/styled/macro.d.mts
```diff
@@ -0,0 +1,2 @@
+export * from './macro.js'
+export { _default as default } from './macro.default.js'
```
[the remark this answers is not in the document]
Andarist: hm, why is that? 🤔 it seems to me that TS would trip over this without `macro.default.d.ts` in certain configurations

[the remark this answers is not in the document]
emmatown: If we're in a `.mts` and we're resolving `@emotion/styled`, we'll hit the `import` condition we have for `"."` and get the already correct `default`.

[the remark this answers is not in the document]
Andarist: oh, that makes sense 👍 i wouldn't use a self reference here myself (it makes sense to use it - i just wouldnt think of it myself) but it was already here and it makes sense that with it we can simplify this a little bit

packages/css/package.json
Andarist: hm, i wasn't sure if this would "fall through" to the other default... does it? 🤔

emmatown: It does (but that's correct behaviour, not like TS's bug)

packages/css/package.json
```diff
@@ -63,7 +69,13 @@
...
+          },
+          "default": "./macro.js"
+        }
```
Andarist, line 78: While I'm not a fan of array flavor... maybe, in here, this would actually be a little bit more obvious that the "fallthrough" is intended?
```suggestion
    "./macro": [
      {
        "types": {
          "import": "./macro.d.mts"
        }
      },
      "./macro.js"
    ]
```

emmatown: Nevermind, just leave it how it is. (it looks like arethetypeswrong incorrectly flags this as a problem, will open an issue about that in a moment)

emmatown (approved):

PR #3033, opened 2023-05-11
Make classNames's cx support obj.toString()

liamqma:
**What**:

The PR updates classNames' cx function to make it accept `object.toString()`.

**Why**:

Example:
```
const obj = {
  toString: function () {
    return 'css-class';
  },
};

<ClassNames>
  {({ css, cx }) => {
    return (
      <div
        className={cx(
          css`
            color: green;
          `,
          obj
        )}
      />
    )
  }}
</ClassNames>
```

The rendered html of above component is expected to be `<div class="css-class emotion-0" />`. It is currently `<div class="emotion-0" />` on `main` branch.

**How**:

This PR adds additional check to see if obj.toString exists. The code is mostly copied from [classnames](https://github.com/JedWatson/classnames/blob/main/index.js#L32), which is widely used to join class names.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/react/__tests__/__snapshots__/class-names.js.snap
+++ b/packages/react/__tests__/__snapshots__/class-names.js.snap
@@ -30,3 +30,14 @@ exports[`should get the theme 1`] = `
   className="emotion-0"
 />
 `;
+
+exports[`should use toString() of object 1`] = `
+.emotion-0 {
+  color: hotpink;
+  color: green;
+}
+
+<div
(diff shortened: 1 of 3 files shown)


Andarist:
Closing this PR for the time being because of the reasons outlined here: https://github.com/emotion-js/emotion/issues/3043#issuecomment-1567917691

PR #3048, opened 2023-06-04
Support JSX.ElementType

naari3:
**What**:

Support for JSX.ElementType that introduced in [TypeScript 5.1](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types)

**Why**:

Until now, JSX type checking was provided by checking as JSX.Element.
From TypeScript 5.1 onward, by declaring JSX.ElementType, it will be possible to more flexibly check elements that can be handled on the jsx runtime side.

In React, this allows ReactNode to be used directly in the return value of function components. We would like to do the same for `@emotion/react/jsx-runtime`.

**How**:

In `@types/react`, JSX.ElementType has been declared since [18.2.8](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65135), and I have copied the definition from the 18.2.8. because I don't know how to handle the `@types/react` version in emotion.  
If you don't mind updating `@types/react`, I can change it to use `React.JSX.ElementType` directly after updating!

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [x] Changeset 

closes https://github.com/emotion-js/emotion/issues/3049

### DIFF
--- a/packages/react/types/jsx-namespace.d.ts
+++ b/packages/react/types/jsx-namespace.d.ts
@@ -18,7 +18,12 @@ type ReactJSXIntrinsicAttributes = JSX.IntrinsicAttributes
 type ReactJSXIntrinsicClassAttributes<T> = JSX.IntrinsicClassAttributes<T>
 type ReactJSXIntrinsicElements = JSX.IntrinsicElements
 
+// based on the code from @types/react@18.2.8
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3197efc097d522c4bf02b94e1a0766d007d6cdeb/types/react/index.d.ts#LL3204C13-L3204C13
+type ReactJSXElementType = string | React.JSXElementConstructor<any>
+
 export namespace EmotionJSX {
+  type ElementType = ReactJSXElementType
   interface Element extends ReactJSXElement {}
   interface ElementClass extends ReactJSXElementClass {}
   interface ElementAttributesProperty
--- a/.changeset/wicked-boxes-lick.md
+++ b/.changeset/wicked-boxes-lick.md
@@ -1,0 +1,5 @@
+---
+"@emotion/react": patch
+---
+
+Added `ElementType` to the Emotion's `JSX` namespace. It's defined in the same way as the one in `@types/react` and should make it possible to use components that return `string`s, `Promise`s and other types that are valid in React.
--- a/packages/react/types/index.d.ts
+++ b/packages/react/types/index.d.ts
@@ -96,6 +96,7 @@ export function ClassNames(props: ClassNamesProps): ReactElement
 export const jsx: typeof createElement
 export namespace jsx {
   namespace JSX {
+    type ElementType = EmotionJSX.ElementType
     interface Element extends EmotionJSX.Element {}
     interface ElementClass extends EmotionJSX.ElementClass {}
     interface ElementAttributesProperty
--- a/packages/react/types/tests.tsx
+++ b/packages/react/types/tests.tsx
@@ -245,19 +245,3 @@ const anim1 = keyframes`
       ? true
       : false
 }
-
-// RMWC-like component test
-declare const OtherComponent: {
-  <Tag extends React.ElementType>(
-    props:
-      | React.AllHTMLAttributes<HTMLInputElement>
-      | React.ComponentPropsWithRef<Tag>,
-    ref: any
-  ): JSX.Element
-  displayName?: string
-}
-;<OtherComponent
-  onChange={ev => {
-    console.log(ev.currentTarget.value)
-  }}
-/>


packages/react/types/jsx-namespace.d.ts
```diff
@@ -18,7 +18,12 @@ type ReactJSXIntrinsicAttributes = JSX.IntrinsicAttributes
 type ReactJSXIntrinsicClassAttributes<T> = JSX.IntrinsicClassAttributes<T>
 type ReactJSXIntrinsicElements = JSX.IntrinsicElements
 
+// based on the code from @types/react@18.2.8
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3197efc097d522c4bf02b94e1a0766d007d6cdeb/types/react/index.d.ts#LL3204C13-L3204C13
+type ReactJSXElementType = string | React.JSXElementConstructor<any>
+
 export namespace EmotionJSX {
+  type ElementType = ReactJSXElementType
```
Andarist, line 26: You also need to "reexport" it here:
https://github.com/emotion-js/emotion/blob/1135f8e9d97ea711eb483368313afdfe7b176845/packages/react/types/index.d.ts#L98

packages/react/types/jsx-namespace.d.ts
```diff
@@ -18,7 +18,12 @@ type ReactJSXIntrinsicAttributes = JSX.IntrinsicAttributes
 type ReactJSXIntrinsicClassAttributes<T> = JSX.IntrinsicClassAttributes<T>
 type ReactJSXIntrinsicElements = JSX.IntrinsicElements
 
+// based on the code from @types/react@18.2.8
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3197efc097d522c4bf02b94e1a0766d007d6cdeb/types/react/index.d.ts#LL3204C13-L3204C13
+type ReactJSXElementType = string | React.JSXElementConstructor<any>
```
Andarist, line 23: Hm, likely we could do some hacks with `@ts-ignore` to "detect" if the type is there in React's `JSX` but perhaps it's not worth it and copy-paste is fine. I'll give this a thought.

emmatown: yeah, I feel like copying it is probably the right approach tbh

naari3:
I don't know much about JS and TS, but it is hard to believe that this Pull Request change affects `React.ElementType`, so I thinking it was a `@definitelytyped/dtslint` issue and I updated to 0.0.115 and this error did not occur.  
I don't know the details of why this problem no longer occurs with 0.0.115.

On the other hand, when I updated to the latest 0.0.163, I get the error everywhere. These seemed to be caused by existing type definitions that were now working fine.

naari3:
Sorry, I see that it has fallen unchanged in CI... On my local, the problem remains at 0.0.112 and remains resolved at 0.0.115.

Do you know anything about this?

Andarist:
1. I think that this particular test case that is failing can be removed, this will revert~ https://github.com/emotion-js/emotion/pull/2181 (cc @101arrowz) but that seems to be the correct thing to do. That PR was enabling something that isn't quite type-safe.
2. so why introducing the `JSX.ElementType` here is creating a problem in the first place? With `ElementType` in place the `tagType` is being checked against `elementTypeConstraint` to see if it's assignable to it, whereas previously only return types were compared by TS using `checkJsxReturnAssignableToAppropriateBound` (see source code [here](https://github.dev/microsoft/TypeScript/blob/cbda2fcce95d7ac82841cee631f07b73be07c0f0/src/compiler/checker.ts#L30976-L30989))
3. it happens so that we have `@types/react@18.0.14` installed here and that defines `type JSXElementConstructor<P> = ((props: P) => React.ReactElement<any, any> | null) | (new (props: P) => React.Component<any, any>)` but the current `@types/react@18.2.8` have a slightly different definition, see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3197efc097d522c4bf02b94e1a0766d007d6cdeb/types/react/index.d.ts#LL78C5-L86C51).
4. We can update `@types/react` as part of this PR (it has no effect on our consumers anyway, locally it's just a dev dependency and we forward whatever definitions are installed in consuming projects), which would fix the caught issue. The point is that how `@types/react` got adjusted 2 weeks ago is a temporary~ thing (see the PR [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65566/files)) so the test that is failing right now can stop working at just any time in the future
5. we can double-check how those 2 `JSXElementConstructor` "perform" against this component in this [TS playground](https://www.typescriptlang.org/play?ts=5.2.0-dev.20230604#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocgEyTQBsUi40IA7AZ3gHkYALJKAGFckVklYwAXHADe5OHAA8AFRQBzOEgAeMcVXaJiMAHQBROkhDiYygJ5gkAPgAU8hXDA4w7SW-dwAH0N0EwBBOjoACWUAWQAZUJgYKGAAIwBXXXZFaPiASVYwTPNLa0c-dyDkEONhcDZrAAUvdgB1YD5kTBV1RwAaCqJMaRRWWzcASmkAKQBlAA0zCysJCgUqYHYwBlsAORQrAH5pThTWNQoAXwpFHn4hEQaJNzZBXlG1JABeGWckADcJnAvo5ZBUWBwIBZjHQIGo-v9jGh0lAiBJVFBPiZ-ig6OkkBM1nBLpdyAB6RwUcgwexIOBzeYlFYwYQcZLpDDQLh0KiKRqgr5uILOZyeCDeaSNIEg4IYYzVDBM6yKUa2PpwVWgoKsdIRCZCuDOMQAd0NYolcClwNBCpMdVEytV6s1hMoNIccG5VAxWOBcFOwHOgXpCyVElZpw5MC5PJVY0pbtpIcZy2sEfZnKguyQxr5AoNIvNPkt6ponloKF0VDiSDU6FsrN0OmOGrG0ptRnlRl2EBo+oUwpNZpakvbsrtjzEEjjatbtkcrupSezxp9SHgX39yUDGiCDLDLLYkczK5nCfJZLgAFob5QIZw4FkYF7pF61xvZJcNQZ3UgIJhPT4AR7SeGAKDJS8FAAPUOch73gJ8V2kFd3z9GQvxQH9aX-QD7hAqcwIvdwYPIIA). Why do I think that this definition isn't type-safe? This particular definition has a required second parameter in the signature of type `any`. This makes it somewhat OK since u forgo all the type safety with `any` anyway but if that would be `unknown` or just any other type then things could break badly. With JSX we are handing out functions to some runtime (in this case it's React) and that runtime is meant to call those functions on our behalf. But React can't call our functions with arbitrary extra arguments and yet our signatures might depend on them. In practice, this doesn't quite happen often and it's not a common problem but from the types PoV this is just wrong. It seems that TS has traditionally allowed this (see the [playground](https://www.typescriptlang.org/play?ts=5.0.4#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocgEyTQBsUi40IA7AZ3gHkYALJKAGFckVklYwAXHADe5OHAA8AFRQBzOEgAeMcVXaJiMAHQBROkhDiYygJ5gkAPgAU8hXDA4w7SW-dwAH0N0EwBBOjoACWUAWQAZUJgYKGAAIwBXXXZFaPiASVYwTPNLa0c-dyDkEONhcDZrAAUvdgB1YD5kTBV1RwAaCqJMaU4U1jU3AEppACkAZQANMwsrCQoFKmB2MAZbADkUKwB+EeTgcYoAXwpFHn4hEQaJNzZBXhRxpABeGWckADdJnAvo5ZBUWBwIBZjHQIGo-v9jGh0lAiBJVFA1EgTP8UHR0khJus4JdLuQAPSOChAA) here with the required `string` in the signature used as JSX element). The recent changes to `JSX.ElementType` will allow people to avoid such invalid combinations in the future (and already do it for cases with 3+ function parameters)

naari3:
@Andarist Wow, I am very happy to see a fairly thorough explanation! Thank you very much!

I tried each of the things you taught me, but this time I just decided to delete the test case.

Once I updated `@types/react`, this dragged the test cases in `packages/native` to fail. Probably due to the relationship between `@types/react` and `@types/react-native`, but I thought it was too broad to do in this PR.

If the test in the emotion fails, I thought about changing the ElementType as you said, but since that was not the case, I think we should leave it as is for the time being.

Andarist (approved):
LGTM

naari3:
Thank you for your kind review, I'm so very happy! is there anything else I need to do to get this PR merged?

Issue #3075, opened 2023-07-14
[TS] Type 'Function' has no construct signatures and missing `isSpeedy` property


JeromeDeLeon:
**Current behavior:**

I am trying to follow the existing implementation to create a Vue global component but I'm having a type issue when trying to reuse the sheet constructor and also missing `isSpeedy` property to a StyleSheet type.

**To reproduce:**

1. Here is the [TS Playground](https://www.typescriptlang.org/play?ssl=10&ssc=3&pln=5&pc=1#code/JYWwDg9gTgLgBAYygUwIY2QYVQgFsuAMyghDgHIABZECGYCAOwHoEd9yAoThJgZ3hs8BALyIU6LO2QAKAN5wA1sgCeALgoI+fcnAC+ASgDc3XowFw++ZPDGNkAd0TSAdFeQ2XZgVACuCGGh5TjglVQ1yLR0AGhC4RiYEZA0hfDdrGBcExiTY0LMYVGB7KBTXd08CopK8yzAPABN1Z2F0j0zgPgBleuQm2MMgA)

**Expected behavior:**

It shouldn't produce any TS errors as this seems to be a valid operation

**Environment information:**

- `react` version: n/a
- `@emotion/react` version: n/a
- `@emotion/cache` version: `^11.11.0`

Andarist:
It's an interesting case! It might be a little bit tricky to fix (the no construct signatures thing) given the structure of our packages - but I didn't really analyze this in depth right now. PRs are welcome.

JeromeDeLeon:
[...]

PR #3076, opened 2023-07-15
Reorganize `@emotion/utils` types

JeromeDeLeon:
**What**: https://github.com/emotion-js/emotion/issues/3075

**Why**: To fix inconsistent type from https://github.com/emotion-js/emotion/issues/3075

**How**: Reorganizing the `@emotion/utils` by moving the types to the proper package.

**Checklist**:

- [ ] Documentation n/a
- [ ] Tests - n/a
- [ ] Code complete - n/a
- [x] Changeset 

I'm not able to fix the `constructor` issue. When I tried to convert it to `interface` /`types`, it complains as it was either being merged to `StyleSheet` DOM type or a duplicate name.


JeromeDeLeon:
@Andarist I ended up touching some `.js` files but flow complains about the type not being declared and have no idea how to fix them 😞

packages/cache/src/index.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import { StyleSheet } from '@emotion/sheet'
-import { type EmotionCache, type SerializedStyles } from '@emotion/utils'
+import { type SerializedStyles } from '@emotion/serialize'
+import { type EmotionCache } from '@emotion/cache'
```
JeromeDeLeon, line 4: For instance, this is complaining because `SerializedStyles` is not exported in `@emotion/serialize`, but that shouldn't be the case unless they are referencing the current version, but does this mean I need to bump them first or something? I literally have no idea.

Next is this `@emotion/cache`, since I'm inside the package already, I tried using `../types` instead but got an error. Where should I declare them? 🤔

packages/utils/src/types.js
```diff
@@ -1,28 +0,0 @@
-// @flow
-import type { StyleSheet } from '@emotion/sheet'
-
-export type RegisteredCache = { [string]: string }
-
-export type Interpolation = any
```
JeromeDeLeon, line 6: I see that we're also creating it's own `Interpolation` type instead of relying from `@emotion/serialize`

packages/utils/types/tests.ts
```diff
@@ -1,12 +1,5 @@
-import {
-  EmotionCache,
-  RegisteredCache,
-  SerializedStyles,
-  StyleSheet,
```
JeromeDeLeon, line 5: `SerializedStyles` and `StyleSheet` are not being used in this file

packages/css/src/create-instance.js
JeromeDeLeon, line 1: Any reason why this file also declares its own `Interpolation`? 
https://github.com/emotion-js/emotion/pull/3076/files#diff-5d0ca00adf3ee0ff837d19e810ebe3271cfa621468b53c7eafe57f061b782db5R27

packages/cache/types/index.d.ts
```diff
@@ -1,8 +1,32 @@
...
+export interface EmotionStyleSheet extends StyleSheet {
+  constructor: typeof StyleSheet
+}
+
+export interface EmotionCache {
+  inserted: {
+    [key: string]: string | true
+  }
+  registered: RegisteredCache
+  sheet: EmotionStyleSheet
```
JeromeDeLeon, line 19: This is the only thing I can think of that would solve the `constructor` issue, let me know if you don't want it or if you have any thoughts

packages/css/src/create-instance.js
```diff
@@ -1,12 +1,7 @@
 // @flow
```
JeromeDeLeon, line 1: Any reason why this file also declares its own Interpolation?
https://github.com/emotion-js/emotion/pull/3076/files#diff-5d0ca00adf3ee0ff837d19e810ebe3271cfa621468b53c7eafe57f061b782db5R27

JeromeDeLeon:
Noob question, how would I be able to test this on my package locally? I just want to make sure that everything I did  here works 😅

packages/cache/src/index.js
JeromeDeLeon: This can be disregard as this is a mistake on my part

packages/utils/src/types.js
JeromeDeLeon: This can be disregard as this is a mistake on my part

packages/css/src/create-instance.js
JeromeDeLeon: This can be disregard as this is a mistake on my part

packages/serialize/types/index.d.ts
```diff
@@ -13,7 +17,7 @@ export type CSSPropertiesWithMultiValues = {
     | Array<Extract<CSSProperties[K], string>>
 }
 
-export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject }
+export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject | SerializedStyles }
```
JeromeDeLeon, line 20: One of the thing I noticed while using this lib is that there is a test for `css composition` but it doesn't cater that in type.

JeromeDeLeon: Are there reasons why the following convention is not allowed even though they worked?
```js
css(() => '')
css`${() => ''}`
```

packages/react/types/tests-css.tsx
```diff
@@ -28,6 +28,12 @@ css`
...
 `
+// $ExpectType SerializedStyles
+css({
+  ':hover': css`
+    color: hotpink;
+  `
+})
 
 // $ExpectError
 css(() => 'height: 300px;')
```
JeromeDeLeon, line 39: @Andarist Are there reasons why this and below it are not allowed even though they worked?

packages/serialize/types/index.d.ts
```diff
@@ -1,19 +1,26 @@
...
   [K in keyof CSSProperties]:
+    | null
+    | undefined
+    | false
     | CSSProperties[K]
     | Array<Extract<CSSProperties[K], string>>
 }
 
-export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject }
+export type CSSPseudos = { [K in CSS.Pseudos]?: CSSInterpolation }
```
JeromeDeLeon, line 23: - [a test](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/react/__tests__/css.js#L39) that allows CSS composition but not in TS
- [a test](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/react/__tests__/css.js#L67) that allows `false` value but not in TS
- [a test](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/react/__tests__/css.js#L250) that allows child selector array but not in TS

packages/react/types/tests-css.tsx
Andarist: this `css` call doesn't run in any kind of "context", we don't have any arguments to give to this function so we don't allow functions to be passed here

packages/serialize/types/index.d.ts
```diff
@@ -1,19 +1,26 @@
...
+  map?: string
+  next?: SerializedStyles
+}
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
 export type CSSPropertiesWithMultiValues = {
   [K in keyof CSSProperties]:
+    | null
+    | undefined
+    | false
```
Andarist, line 18: So I guess those were added to account for "a test that allows false value but not in TS"?

packages/serialize/types/index.d.ts
Andarist: Looks legit but I will have to reexamine this more closely later.

packages/utils/types/index.d.ts
```diff
@@ -1,42 +1,7 @@
...
-}
-
-export interface SerializedStyles {
-  name: string
-  styles: string
-  map?: string
-  next?: SerializedStyles
-}
+import type { SerializedStyles } from '@emotion/serialize'
+import type { RegisteredCache, EmotionCache } from '@emotion/cache'
```
Andarist, line 4: We can't import from other packages here because we don't depend on them. This is also true at type level - even if we still don't import anything at runtime.

I'd keep the existing types here (to avoid breaking changes) but I'd deprecate them in favor of the proper ones exported by specific packages). In the future, we should just remove `export` from those local types here. This way we'll be able to depend on TS structural type system without introducing any dependencies to this package.

packages/react/types/tests-css.tsx
JeromeDeLeon: So I'm guessing, even though it works in nature, there's no point in doing this as it doesn't run in any context?
I'm looking for reasons why we do not allow these things in TS due to type complaining, but it's working just fine.

packages/serialize/types/index.d.ts
JeromeDeLeon: Yes, but a weird thing happened, I managed to build my own serializer based on the implementation but using `js-beautify` like `beautify.css()` instead of using `@emotion/css-prettifier`.

When I used it and passed `null` like `css({ textAlign: null })`, it created a weird selector like `.emotion-0 textAlign {}` but when I switched to `@emotion/css-prettifier`, it removed it.

The `@emotion/css-prettifier` also removes the empty selector like `.emotion-0 {}` which I guess does the same thing hence why in the test we're not seeing these weird things.

It's not an issue as I can just use that lib you guys already have but just a weird thing that it only happens when it's `null`

packages/utils/types/index.d.ts
```diff
@@ -41,19 +50,19 @@ export interface SerializedStyles {
...
+  cache: EmotionStyleCache,
+  serialized: EmotionSerializedStyles,
   isStringTag: boolean
 ): void
 
 export function insertStyles(
-  cache: EmotionCache,
-  serialized: SerializedStyles,
+  cache: EmotionStyleCache,
+  serialized: EmotionSerializedStyles,
```
JeromeDeLeon, line 66: @Andarist This is what I ended up doing otherwise, they would be incompatible to each other.

packages/utils/types/tslint.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@definitelytyped/dtslint/dtslint.json",
   "rules": {
     "array-type": [true, "generic"],
-    "semicolon": false
+    "semicolon": false,
+    "no-redundant-jsdoc": false
```
JeromeDeLeon, line 6: I tried all of these options but didn't work so I had to do this instead https://palantir.github.io/tslint/usage/rule-flags/

JeromeDeLeon:
@Andarist Is there anything blocking this PR? I'm happy to address them 😄

PR #3086, opened 2023-07-28
Update package.json

Shay-repos:
It should depend on version >=16.8.0
keeping 16.14.0 will break storybook and other library which depends on this hook

**What**: Changing react dependency

**Why**: Some storybook addons depend on this hook, which depends directly on a specific version of React (16.14.0). Because of this direct dependency, package managers will import multiple versions of react

**How**: Relax React's version

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset


emmatown:
This is a dev dependency so it doesn't affect anything outside this repo

PR #2748, opened 2022-05-15
Add fetchpriority prop

codejet:
**What**:

Adding the currently still non-standard `fetchpriority` prop.

**Why**:

Priority hints are currently an unoffical W3C draft driven by Google (https://wicg.github.io/priority-hints/) but already supported in some Chromium-based browsers (https://caniuse.com/?search=fetchpriority). See also this article: https://web.dev/priority-hints/.

**How**:

Simply added the prop to `packages/is-prop-valid/src/props.js`, for now under the "Non-standard Properties".

**Checklist**:

- [N/A] Documentation
- [N/A] Tests
- [X] Code complete
- [N/A] Changeset

PR #3093, opened 2023-08-15
Add fetchpriority to props

DustinBrett:
**What**:

Adding the now standardized `fetchpriority` property to the valid props list. "Fetch Priority" is now included in the [HTML living standard](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#fetch-priority-attributes). It's been in Chrome since [101](https://caniuse.com/?search=fetchpriority)-[102](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority#browser_compatibility) and is now coming to [Safari 17](https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes#Web-API).

**Why**:

This is a [useful property](https://web.dev/fetch-priority/) that I use throughout my site. In some places I am passing it to a styled component and because it uses emotions is-prop-valid check, I am getting a warning. I think this property will continue to grow in adoption and should be considered valid.

image

**How**:

Added the lowercased property name to `packages\is-prop-valid\src\props.js`. Similar to what was done in the closed PR [#2748](https://github.com/emotion-js/emotion/pull/2748).

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [ ] Changeset N/A

### DIFF
--- a/.changeset/eighty-garlics-dance.md
+++ b/.changeset/eighty-garlics-dance.md
@@ -2,4 +2,8 @@
 "@emotion/is-prop-valid": patch
 ---
 
+
+author: @codejet
+author: @DustinBrett
+
 Adds `fetchpriority` and `fetchPriority` to the list of allowed props.
--- a/.changeset/eighty-garlics-dance.md
+++ b/.changeset/eighty-garlics-dance.md
@@ -1,0 +1,5 @@
+---
+"@emotion/is-prop-valid": patch
+---
+
+Adds `fetchpriority` and `fetchPriority` to the list of allowed props.
--- a/packages/is-prop-valid/src/props.js
+++ b/packages/is-prop-valid/src/props.js
@@ -73,6 +73,7 @@ const props = {
   encType: true,
   enterKeyHint: true,
   fetchpriority: true,
+  fetchPriority: true,
   form: true,
   formAction: true,
   formEncType: true,
--- a/packages/is-prop-valid/src/props.js
+++ b/packages/is-prop-valid/src/props.js
@@ -72,6 +72,7 @@ const props = {
   draggable: true,
   encType: true,
   enterKeyHint: true,
+  fetchpriority: true,
   form: true,
   formAction: true,
   formEncType: true,


andreisena:
@DustinBrett Thanks for the PR. I was looking for this.
Based on [this PR](https://github.com/facebook/react/pull/25927/files#diff-b3df3711605947008f8657055535ea81e87474f6fe5badfed07575bb66362a3eR582-R593), shouldn't `fetchPriority` (camelCase) be added too?

DustinBrett:
> @DustinBrett Thanks for the PR. I was looking for this. Based on [this PR](https://github.com/facebook/react/pull/25927/files#diff-b3df3711605947008f8657055535ea81e87474f6fe5badfed07575bb66362a3eR582-R593), shouldn't `fetchPriority` (camelCase) be added too?

No problem! I could add `fetchPriority` if it's needed, but my use case was the lowercase one and I am not sure when the camel case one is used. I wanted to keep the PR lean to try and get it through. Is there a use case for `fetchPriority`?

andreisena:
> > @DustinBrett Thanks for the PR. I was looking for this. Based on [this PR](https://github.com/facebook/react/pull/25927/files#diff-b3df3711605947008f8657055535ea81e87474f6fe5badfed07575bb66362a3eR582-R593), shouldn't `fetchPriority` (camelCase) be added too?
> 
> No problem! I could add `fetchPriority` if it's needed, but my use case was the lowercase one and I am not sure when the camel case one is used. I wanted to keep the PR lean to try and get it through. Is there a use case for `fetchPriority`?

As you can see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65972/files), on @types/react it was renamed to camelCase format.
And [this issue](https://github.com/facebook/react/issues/27233) on facebook/react repo to support on camelCase too.

DustinBrett:
> > > @DustinBrett Thanks for the PR. I was looking for this. Based on [this PR](https://github.com/facebook/react/pull/25927/files#diff-b3df3711605947008f8657055535ea81e87474f6fe5badfed07575bb66362a3eR582-R593), shouldn't `fetchPriority` (camelCase) be added too?
> > 
> > 
> > No problem! I could add `fetchPriority` if it's needed, but my use case was the lowercase one and I am not sure when the camel case one is used. I wanted to keep the PR lean to try and get it through. Is there a use case for `fetchPriority`?
> 
> As you can see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65972/files), on @types/react it was renamed to camelCase format. And [this issue](https://github.com/facebook/react/issues/27233) on facebook/react repo to support on camelCase too.

Good points, thanks for the info. I've updated the PR.

packages/is-prop-valid/src/props.js
```diff
@@ -69,6 +69,8 @@ const props = {
   draggable: true,
   encType: true,
   enterKeyHint: true,
+  fetchpriority: true,
```
oliviertassinari, line 72: Is this even valid?

```suggestion
```

oliviertassinari (commented):
Ah nice, I faced the same bug, landing here from https://performance.shopify.com/blogs/blog/optimizing-images-for-performance-on-shopify and https://github.com/mui/material-ui/pull/38733.

packages/is-prop-valid/src/props.js
DustinBrett: It depends what you mean by valid, but I would say likely yes. I use this and it does indeed function and have an effect in browsers. The links I provided in the description also use it this way.

oliviertassinari: I guess the handling of uppercase vs. lowercase confuses me a bit in `@emotion/is-prop-valid`. References:

- https://github.com/facebook/react/pull/25927
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65972

Maybe we should actually .toLowerCase() when determining if a prop should be forwarded or not.

DustinBrett: Ya it's possible converting to lower case could be useful. I didn't see many duplicates. For my use case with Styled Components, it currently passes without the capitol. So I think it would be needed for this PR to stop my warning. The camel case version is not one I've had to use with React yet.

DustinBrett:
This was mentioned as part of Safari 17.2 also btw.

https://webkit.org/blog/14787/webkit-features-in-safari-17-2/#fetch-priority

paales:
Can we get this moving forward? I would guess that the camel case variant is the one go with as metioned here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority

packages/is-prop-valid/src/props.js
Andarist: I think we should stick to the camelCase here and follow React's suite.

DustinBrett: The issue is that the camelCase only solution is not supported with the latest Next.js w/React. Only in React 19 will they be moving to camelCase for this property. If I change to using camelCase on elements within my codebase I get an error, as do others which is mentioned in the links that @oliviertassinari posted a year ago.

image

.changeset/eighty-garlics-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@emotion/is-prop-valid": patch
+---
+
+Adds `fetchpriority` and `fetchPriority` to the list of allowed props.
```
Andarist, line 9:
```suggestion

author: @codejet
author: @DustinBrett

Adds `fetchpriority` and `fetchPriority` to the list of allowed props.
```

PR #3110, opened 2023-09-21
Do not use global cache for createCache() to avoid memory leaks

functino:
**What**:

Currently `createCache()` seems to use a global cache that is reused even if you create multiple caches. In an SSR scenario this means that even if you [follow the recommendations to create a cache per render](https://emotion.sh/docs/ssr#extractcritical) you will end up having a shared cache. 
Also currently an [accessor function is returned from `getServerStylisCache`](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/cache/src/index.js#L40)
But later it is just [used as if it would be an object](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/cache/src/index.js#L199-L204)

**Why**:
I think the current approach can cause memory-leaks in bigger applications. Since we additionally just use the `serverStylisCache` as an object, I think the cached values will never be garbage collected. This causes issues in apps that have a lot of css and/or use a lot of dynamic styles.

**How**:
I want to change the return type of `getServerStylisCache` to a simple cache object. This is how it is already used. We could also try to use the `accessor` function. But I wouldn't know how to write to the cache then.

Also I want to move the call to `getServerStylisCache` inside the function body of `createCache`. This way we will get a new instance of the cache for every request.

I'm not sure how to add tests for this. And I'm also not sure if my assessment of what the current code is doing is correct. Maybe someone can help to shed some light on it.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/pink-dancers-wash.md
+++ b/.changeset/pink-dancers-wash.md
@@ -1,0 +1,5 @@
+---
+'@emotion/cache': patch
+---
+
+Ensure that caches are not reused for different createCache calls
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -32,18 +32,18 @@ export type Options = {
   insertionPoint?: HTMLElement
 }
 
-let getServerStylisCache = isBrowser
-  ? undefined
-  : weakMemoize(() =>
-      memoize(() => {
-        let cache = {}
-        return cache
-      })
-    )
-
 const defaultStylisPlugins = [prefixer]
 
 let createCache = (options: Options): EmotionCache => {
+  let getServerStylisCache = isBrowser
+    ? undefined
+    : weakMemoize(() =>
+        memoize(() => {
+          let cache = {}
+          return cache
+        })
+      )
+
   let key = options.key
 
   if (process.env.NODE_ENV !== 'production' && !key) {
--- a/packages/cache/src/index.js
+++ b/packages/cache/src/index.js
@@ -37,7 +37,7 @@ let getServerStylisCache = isBrowser
   : weakMemoize(() =>
       memoize(() => {
         let cache = {}
-        return name => cache[name]
+        return cache
       })
     )
 


jrosspaperless:
My coworker hunted down a memory leak in our application to this same location. He found that if we downgrade `stylis` and pass in an array of plugins with `stylis` instead of relying on the one built into emotion it resolves the memory leak.

We believe it's because the `WeakMap` references to the `static `array created in emotion's configuration end up ballooning the cache because it does not recognize the key being passed.

Since the only purpose for the cache is SSR and you're moving any potential caching benefit into a method that would result in never being able to benefit from the cache, I think the best option would be to remove the caching concept entirely.

Coming across your PR has helped us to confirm our suspicions and we really appreciate your taking the time to do this!

vgeyvandov:
@functino This is a great find! As @jrosspaperless mentioned, we had a similar issue with Emotion during SSR, and have found that providing our own `stylisPlugins` array (a new instance of the array each render cycle) with the default `stylis` `prefixer` resolves the issue of the cache building up over time.

After debugging further, we found that an even simpler solution within the Emotion `cache` module would be to update [this line](https://github.com/emotion-js/emotion/blob/main/packages/cache/src/index.js#L84) to be the following:

```
const stylisPlugins = options.stylisPlugins || [prefixer]
```

And then delete [this line](https://github.com/emotion-js/emotion/blob/main/packages/cache/src/index.js#L44) entirely since this variable will no longer be used.

This way, the `WeakMap` cache receives a _new_ instance of the plugins array as the key for the cache, and the garbage collector is able to clean up the key and its entries, since a new instance of it is set during each SSR render cycle.

The root issue seems to be that the [current](https://github.com/emotion-js/emotion/blob/main/packages/cache/src/index.js#L44) `defaultStylisPlugins` variable is declared globally and is being used as the default key, such that it is never garbage collected and all new cache entries during each SSR cycle end up tied to `defaultStylisPlugins` and are never cleaned up.

szagi3891:
What's next ? It would be great if this PR was merged

Andarist:
Could somebody familiar with the issue describe step by step how the leak is created here? I admittedly haven't looked into this code yet, a concise description of the problem would help me though

vgeyvandov:
@Andarist I have some details in my [comment](https://github.com/emotion-js/emotion/pull/3110#issuecomment-2090597128) above, but the root issue seems to be that the [current](https://github.com/emotion-js/emotion/blob/main/packages/cache/src/index.js#L44) `defaultStylisPlugins` variable being used as the default cache key for the server Stylis cache `WeakMap` is declared globally.

When Emotion consumers do not create their own cache, or create a custom cache without specifying custom `stylisPlugins` option, or set `stylisPlugins` as a globally defined array, this cache will grow with each unique SSR request. Since the cache has no entry limits or eviction policy, memory consumption slowly rises until the server is restarted.

functino:
> Could somebody familiar with the issue describe step by step how the leak is created here? I admittedly haven't looked into this code yet, a concise description of the problem would help me though

Hi @Andarist thanks for looking into it.

The issue is that regardless how often you call `createCache()` it is actually creating only ONE `serverStylisCache`.
Now lets assume we have a big app:
- it is SSR rendred
- we call createCache() on every request
- we have lots of styles and some of them are dynamic (something like `opacity: Math.random()` in a worst case)

Then the cache will be shared between all those requests AND it will constantly increase in size without ever getting garbage collected.

torressam333:
Will we get this merged in as a fix for indefinitely growing cache?

@Andarist @functino

PR #3116, opened 2023-10-11
fix(regex): global flag should not reuse without reset lastIndex

nnnnoel:
**What**: Fixed lastIndex is not 0 on Regex with global flag when reuse

**Why**: this is unsafe

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Code complete
- [x] Changeset


emmatown:
`lastIndex` is not used in `replace` so this is not relevant

Issue #3136, opened 2023-12-08
@emotion/serialize/types/index.ts TS2411: error


ovaris:
**Current behavior:**

TypeScript compiler throws numerous errors related to `CSSInterpolation`:
```
node_modules/@emotion/serialize/types/index.d.ts:37:18 - error TS2411: Property 'zoom' of type 'Zoom | readonly NonNullable<Zoom | undefined>[] | ("normal" | "reset" | (string & {}) | Globals)[] | undefined' is not assignable to 'string' index type 'CSSInterpolation'.

Found 825 errors in the same file, starting at: node_modules/@emotion/serialize/types/index.d.ts:37
```

Using TS version 5.3.2

Not sure what changed but suddenly index.d.ts appered under node_modules/@emotion/serialize/types even though we have been using same version 1.1.2 all the time (or actually it's dependency of react-select)

└─┬ react-select@5.7.2
    └─┬ @emotion/react@11.11.1
      ├─┬ @emotion/babel-plugin@11.11.0
      │ └── @emotion/serialize@1.1.2 deduped
      └── @emotion/serialize@1.1.2

mitchhentgesspotify:
Hey, I'm seeing the same thing! I've got a [repro repo over here for you, Andarist](https://github.com/mitchhentgesspotify/emotion-tsc)

kamenminkovcom:
This is causing the issue - https://github.com/frenic/csstype/issues/189

sg785-sakshi:
> 
> 
> **Current behavior:**
> 
> TypeScript compiler throws numerous errors related to `CSSInterpolation`:
> ```
> node_modules/@emotion/serialize/types/index.d.ts:37:18 - error TS2411: Property 'zoom' of type 'Zoom | readonly NonNullable<Zoom | undefined>[] | ("normal" | "reset" | (string & {}) | Globals)[] | undefined' is not assignable to 'string' index type 'CSSInterpolation'.
> 
> Found 825 errors in the same file, starting at: node_modules/@emotion/serialize/types/index.d.ts:37
> ```
> 
> Using TS version 5.3.2
> 
> Not sure what changed but suddenly index.d.ts appered under node_modules/@emotion/serialize/types even though we have been using same version 1.1.2 all the time (or actually it's dependency of react-select)
> 
> └─┬ react-select@5.7.2
>     └─┬ @emotion/react@11.11.1
>       ├─┬ @emotion/babel-plugin@11.11.0
[...]

PR #3141, opened 2023-12-15
Change `Array` to `ReadonlyArray` in CSS type declarations

Cerber-Ursi:
**What**:

`csstype@3.1.3` changed the type of multi-valued fallback CSS properties from `array` to `readonly array` - https://github.com/frenic/csstype/commit/46694defae2cf3386218d0000490b0d0ac385aa6. This merge request follows that change, retyping `ArrayCSSInterpolation` from `Array` to `ReadonlyArray`.

**Why**:

This change is created in response to issue #3136.

**Checklist**:

- [ ] Documentation (N/A)
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/warm-pugs-applaud.md
+++ b/.changeset/warm-pugs-applaud.md
@@ -1,0 +1,5 @@
+---
+'@emotion/serialize': patch
+---
+
+Replace arrays with readonly arrays in CSS type definitions, following changes in `csstype`
--- a/packages/react/types/tests-css.tsx
+++ b/packages/react/types/tests-css.tsx
@@ -32,8 +32,10 @@ css`
 // $ExpectError
 css(() => 'height: 300px;')
 
-// $ExpectError
 css`
   position: relative;
-  flexgrow: ${() => 20};
+  flexgrow: ${
+    // $ExpectError
+    () => 20
+  };
 `
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -10,7 +10,7 @@ export type CSSProperties = CSS.PropertiesFallback<number | string>
 export type CSSPropertiesWithMultiValues = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
-    | Array<Extract<CSSProperties[K], string>>
+    | ReadonlyArray<Extract<CSSProperties[K], string>>
 }
 
 export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject }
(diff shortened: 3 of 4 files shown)


freben:
@Cerber-Ursi Note that according to [CONTRIBUTING.md](https://github.com/emotion-js/emotion/blob/main/CONTRIBUTING.md) you need to make a changeset to get things released.

And also we'd be grateful if you had the time to look into the build error.

Andarist:
We should also have a type-level test that will showcase that both mutable and readonly arrays can be used now

Cerber-Ursi:
> you need to make a changeset to get things released.

Done. Misunderstood the rules - thought somehow that it's expected to be done when _preparing_ the release (not just for any change in the release) and that it's expected to be done by the one actually releasing; I see now that this is incorrect.

> And also we'd be grateful if you had the time to look into the build error.

It's actually somewhat cryptic. The problem is with this code:
```typescript
// $ExpectError
css`
  position: relative;
  flexgrow: ${() => 20};
`
```
...and TypeScript throws the error on the interpolation, not on the `css` function call, so `ExpectError` isn't matched. Not sure why this worked before, to be honest.

> We should also have a type-level test that will showcase that both mutable and readonly arrays can be used now

Added some fields to `CSSObject` value in test, thanks for noting this. Caught another place which had to change (where readonly array was not allowed).

Cerber-Ursi:
Managed to pull off the fix for build. Actually, this change seems to have slightly improved ergonomics for `css` template function.

For the record, here's the problem:
- Before the change, if `css`-tagged template was called with incorrect interpolation, TypeScript checked it in the following way:
  - First overload, with `TemplateStringsArray` as the first argument, is incorrect, since the second argument is not a `CSSInterpolation`.
  - Second overload is incorrect, because it attempts to treat `TemplateStringsArray` as `CSSInterpolation` - and this fails, because `TemplateStringsArray` is a `ReadonlyArray`, and `CSSInterpolation` is a plain `Array`.
  - Therefore, it's the _first_ argument to `css` function which TypeScript treats as mismatched - not the second.
  - Therefore, the whole `css`-tagged template literal was marked as erroneous.
- Now, the check goes another way:
  - First overload is invalid, as before.
  - Second overload is invalid too, but now `TemplateStringsArray` as `CSSInterpolation` actually _typechecks_, and error is moved forward to the interpolated value.
  - Therefore, only the interpolated value is an error - not the whole literal.

This looks like a subtle unexpected improvement, since now the errors are more precise. But the negative part was that the old test - which had `ExpectError` on the literal - now fails. I've moved the `ExpectError` on the interpolation itself, and now test passes.

emmatown (approved):

PR #3146, opened 2024-01-08
Support dynamic value

siriwatknp:
**What**:

Add callback support for the value when serializing styles. For example:

```js
styled('div')({
  background: (props) => props.color || '#fff',
  '&:hover': {
    background: (props) => props.color || 'red',
  }
})
```

**Why**:

1. We ([MUI](https://github.com/mui/material-ui)) need emotion to have a consistent API with our new static CSS-in-JS library to let developers switch between engines.
2. styled-components already support the callback for the values so it's not new.

**How**:

The change is mainly in emotion's `serialize` function. When the value is a function, just call it with the props and continue serializing the result the same as before.

**Checklist**:

- [ ] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -239,7 +239,7 @@ function createStringFromObject(
   } else {
     for (let key in obj) {
       let value = obj[key]
-      if (typeof value === 'function') {
+      if (typeof value === 'function' && mergedProps !== undefined) {
         value = value(mergedProps)
       }
       if (typeof value !== 'object') {
--- a/packages/serialize/src/index.js
+++ b/packages/serialize/src/index.js
@@ -239,6 +239,9 @@ function createStringFromObject(
   } else {
     for (let key in obj) {
       let value = obj[key]
+      if (typeof value === 'function') {
+        value = value(mergedProps)
+      }
       if (typeof value !== 'object') {
         if (registered != null && registered[value] !== undefined) {
           string += `${key}{${registered[value]}}`
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -7,18 +7,19 @@ import * as CSS from 'csstype'
 export { RegisteredCache, SerializedStyles }
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
-export type CSSPropertiesWithMultiValues = {
+export type CSSPropertiesWithMultiValues<Props> = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
     | ReadonlyArray<Extract<CSSProperties[K], string>>
+    | ((props: Props) => number | string)
 }
 
-export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject }
+export type CSSPseudos<Props> = { [K in CSS.Pseudos]?: CSSObject<Props> }
 
-export interface ArrayCSSInterpolation
-  extends ReadonlyArray<CSSInterpolation> {}
+export interface ArrayCSSInterpolation<Props>
+  extends ReadonlyArray<CSSInterpolation<Props>> {}
 
-export type InterpolationPrimitive =
+export type InterpolationPrimitive<Props> =
   | null
   | undefined
   | boolean
@@ -27,18 +28,22 @@ export type InterpolationPrimitive =
   | ComponentSelector
   | Keyframes
   | SerializedStyles
-  | CSSObject
+  | CSSObject<Props>
 
-export type CSSInterpolation = InterpolationPrimitive | ArrayCSSInterpolation
+export type CSSInterpolation<Props> =
+  | InterpolationPrimitive<Props>
+  | ArrayCSSInterpolation<Props>
 
-export interface CSSOthersObject {
-  [propertiesName: string]: CSSInterpolation
+export interface CSSOthersObject<Props> {
+  [propertiesName: string]:
+    | CSSInterpolation<Props>
+    | ((props: Props) => string | number)
 }
 
-export interface CSSObject
-  extends CSSPropertiesWithMultiValues,
-    CSSPseudos,
-    CSSOthersObject {}
+export interface CSSObject<Props = unknown>
+  extends CSSPropertiesWithMultiValues<Props>,
+    CSSPseudos<Props>,
+    CSSOthersObject<Props> {}
 
 export interface ComponentSelector {
   __emotion_styles: any
@@ -59,7 +64,7 @@ export interface FunctionInterpolation<Props> {
 }
 
 export type Interpolation<Props> =
-  | InterpolationPrimitive
+  | InterpolationPrimitive<Props>
   | ArrayInterpolation<Props>
   | FunctionInterpolation<Props>
 
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -7,19 +7,21 @@ import * as CSS from 'csstype'
 export { RegisteredCache, SerializedStyles }
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
-export type CSSPropertiesWithMultiValues<Props> = {
+export type CSSPropertiesWithMultiValues<Props = unknown> = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
     | ReadonlyArray<Extract<CSSProperties[K], string>>
     | ((props: Props) => number | string)
 }
 
-export type CSSPseudos<Props> = { [K in CSS.Pseudos]?: CSSObject<Props> }
+export type CSSPseudos<Props = unknown> = {
+  [K in CSS.Pseudos]?: CSSObject<Props>
+}
 
-export interface ArrayCSSInterpolation<Props>
+export interface ArrayCSSInterpolation<Props = unknown>
   extends ReadonlyArray<CSSInterpolation<Props>> {}
 
-export type InterpolationPrimitive<Props> =
+export type InterpolationPrimitive<Props = unknown> =
   | null
   | undefined
   | boolean
@@ -30,11 +32,11 @@ export type InterpolationPrimitive<Props> =
   | SerializedStyles
   | CSSObject<Props>
 
-export type CSSInterpolation<Props> =
+export type CSSInterpolation<Props = unknown> =
   | InterpolationPrimitive<Props>
   | ArrayCSSInterpolation<Props>
 
-export interface CSSOthersObject<Props> {
+export interface CSSOthersObject<Props = unknown> {
   [propertiesName: string]:
     | CSSInterpolation<Props>
     | ((props: Props) => string | number)
--- a/packages/serialize/types/index.d.ts
+++ b/packages/serialize/types/index.d.ts
@@ -20,8 +20,8 @@ export type CSSPseudos<Props = unknown> = {
   [K in CSS.Pseudos]?: CSSObject<Props>
 }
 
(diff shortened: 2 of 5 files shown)


packages/serialize/types/index.d.ts
Andarist, line 1: every type that is exported and has been touched here need to receive the default for `Props` - otherwise it becomes a breaking change

packages/serialize/types/index.d.ts
```diff
@@ -27,18 +28,22 @@ export type InterpolationPrimitive =
...
+  [propertiesName: string]:
+    | CSSInterpolation<Props>
+    | ((props: Props) => string | number)
 }
 
-export interface CSSObject
-  extends CSSPropertiesWithMultiValues,
-    CSSPseudos,
-    CSSOthersObject {}
+export interface CSSObject<Props = unknown>
```
Andarist, line 43: we didn't really restrict this anywhere here (from what I can see) so perhaps this is OK but likely `null` and `undefined` shouldn't be accepted here (and the default shouldn't include them). I'm OK with this change though - since other `Props` type parameters in this file have no constraints. We can always revisit this later if needed

packages/serialize/types/index.d.ts
```diff
@@ -7,18 +7,19 @@ import * as CSS from 'csstype'
 export { RegisteredCache, SerializedStyles }
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
-export type CSSPropertiesWithMultiValues = {
+export type CSSPropertiesWithMultiValues<Props> = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
     | ReadonlyArray<Extract<CSSProperties[K], string>>
+    | ((props: Props) => number | string)
```
Andarist, line 14: Those 2 seem to be inconsistent. Why do we allow `number | string` to be returned from a function but the array case does only allow `string`?

packages/serialize/types/index.d.ts
siriwatknp: I will do that for all exported interfaces.

siriwatknp: I think it would make more sense for the default to be `{}` rather than `unknown`?

packages/serialize/src/index.js
```diff
@@ -239,6 +239,9 @@ function createStringFromObject(
   } else {
     for (let key in obj) {
       let value = obj[key]
+      if (typeof value === 'function') {
+        value = value(mergedProps)
+      }
```
Andarist, line 244: I'm OK with this change. It's minimal and improves compatibility between different styling engines and compatibility. 

I think that we should only do this when `mergedProps !== undefined` and raise a warning otherwise, like we do here:
https://github.com/emotion-js/emotion/blob/cde417556b1d7fa5dc0f1774da9fb7dfca9caa32/packages/serialize/src/index.js#L174-L188

cc @emmatown , do you have any other thoughts about this one?

packages/serialize/types/index.d.ts
Andarist: Yeah, I kinda had a similar thought here: https://github.com/emotion-js/emotion/pull/3146#discussion_r1444375613

But since we didn't have those constraints anywhere so far - I think it makes sense **right now** for the default to reflect the constraint (and the unconstrained type parameter is implicitly constrained to `unknown`)

siriwatknp: Got it.

packages/serialize/types/index.d.ts
siriwatknp: If my understanding is correct, the `string` in `ReadonlyArray<Extract<CSSProperties[K], string>>` is to pull possible values from the `CSSProperties[K]`.

For callback as a value, I need to use `number | string` to handle this case:

```js
styled('div')({
  lineHeight: props => props.lineHeight || 1.2, // number is a valid unit for some CSS properties
})
```

Andarist: But isn't this valid then?
```ts
styled('div')({
  lineHeight: [1.2],
})
```

Note that array values here are somewhat weird. They mainly exist so you can do stuff like this:
```ts
// display: -webkit-flex;
// display: flex;
styled('div')({
  display: ['-webkit-flex', 'flex'],
})
```

We can't support this otherwise as we can't observe duplicate properties at runtime. Perhaps this particular use case actually means that only string values should be supported here as numerical values used like this don't make a lot of sense (unless dunno, maybe mixing a numerical fallback with some keyword~ would make sense?)

That also kinda means that we should support this at type level:
```ts
styled('div')({
  display: (props) => ['-webkit-flex', 'flex'],
})
```

packages/serialize/types/index.d.ts
```diff
@@ -7,18 +7,21 @@ import * as CSS from 'csstype'
 export { RegisteredCache, SerializedStyles }
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
-export type CSSPropertiesWithMultiValues = {
+export type CSSPropertiesWithMultiValues<Props = unknown> = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
     | ReadonlyArray<Extract<CSSProperties[K], string>>
+    | ((props: Props) => number | string)
```
Andarist, line 14: This should likely be typed like this:
```suggestion
    | ((props: Props) => CSSProperties[K])
```

Andarist: (potentially with support for arrays but those arrays are likely so rarely used... that perhaps we don't have to worry about it here)

packages/serialize/types/index.d.ts
siriwatknp: 👍 Fixed in https://github.com/emotion-js/emotion/pull/3146/commits/cc913db69c1405a915a982d9e86b80f44c50f9ee

packages/serialize/src/index.js
siriwatknp: Make sense. Added a check `mergedProps !== undefined` before calling the callback.

packages/serialize/types/tests.ts
```diff
@@ -37,6 +37,20 @@ serializeStyles(
   {}
 )
 // $ExpectType SerializedStyles
+serializeStyles<{ vars: { background: string; foreground: string } }>([
```
Andarist, line 40: we could use a negative~ test case for smth like this:
```ts
serializeStyles({
  // $ExpectError
  display: () => 'unknown'
})
```

packages/serialize/types/index.d.ts
```diff
@@ -7,18 +7,23 @@ import * as CSS from 'csstype'
...
 
 export type CSSProperties = CSS.PropertiesFallback<number | string>
-export type CSSPropertiesWithMultiValues = {
+export type CSSPropertiesWithMultiValues<Props = unknown> = {
   [K in keyof CSSProperties]:
     | CSSProperties[K]
     | ReadonlyArray<Extract<CSSProperties[K], string>>
+    | ((
+        props: Props
+      ) => CSSProperties[K] | ReadonlyArray<Extract<CSSProperties[K], string>>)
```
Andarist, line 16: Instead of `Extract<CSSProperties[K], string>` I'd just do this `CSSProperties[K] & string`. The same change could be applied at line 13

packages/serialize/types/tests.ts
siriwatknp: Good point.

siriwatknp: I tried but did not work because `csstype` allows the value to be a string so `display: () => 'unknown'` does work.

```js
// node_modules/csstype

export type Display =
    | Globals
    | DataType.DisplayOutside
    | DataType.DisplayInside
    | DataType.DisplayInternal
    | DataType.DisplayLegacy
    | "contents"
    | "list-item"
    | "none"
    | (string & {});
```

Andarist: Oh, ok - so to make such a test case you'd have to find some other property that doesn't allow `string & {}`. If it's hard to find one - let's not bother with this.

packages/serialize/types/index.d.ts
siriwatknp: I wonder why not just `CSSProperties[K] | ReadonlyArray<CSSProperties[K]>`? Some of the `CSSProperties[K]` already includes the `(string & {})` so is it better to rely on CSSProperties[K]?

Andarist: Yeah, I don't really know why this is filtered - but since it is, I'd probably prefer to keep it **here** (in the spirit of keeping this PR focused). I'm open to accepting a PR that would remove this filtering - that can be created now or after we land this PR here. It doesn't matter in what order we'd land them - the other one will have to be synced after the first one lands though.

siriwatknp: Done!

packages/serialize/types/tests.ts
siriwatknp: The `borderCollapse` does not have `string & {}`, added to the test case.

packages/serialize/types/index.d.ts
```diff
@@ -27,18 +32,22 @@ export type InterpolationPrimitive =
   | ComponentSelector
   | Keyframes
   | SerializedStyles
-  | CSSObject
+  | CSSObject<Props>
 
-export type CSSInterpolation = InterpolationPrimitive | ArrayCSSInterpolation
+export type CSSInterpolation<Props = unknown> =
```
Andarist, line 37: This one actually shouldn't end up including functions in `CSSObject`. We have 2 flavors of those interpolations:
- the ones that u can pass to `css`
- and the ones that you can pass to `styled` and similar

The difference is that `css` is a context-less call. We don't get access to any `props` there (and we don't defer `css` resolution) so this should be an error:
```ts
css({
  display: () => 'flex'
})
```

That's why we have both `ArrayInterpolation` and `ArrayCSSInterpolation` here. A similar distinction should be introduced for those object styles.

But now... I'm not sure if this would satisfy your original goal.

siriwatknp: I see, I will revert this. I think it should be fine since we focus only for `styled`.

siriwatknp: Done!

Andarist: > I think it should be fine since we focus only for styled.

Cool! ❤️ 

> Done! see https://github.com/emotion-js/emotion/commit/2fb22d4dc23a7bc7a81cbd88eacaaf02ea9d1077

Thanks. I'm soft-approving this PR now. I think we still need to tweak types a little bit but honestly - I need to give it more thought to determine how to do it in the least disruptive way. So I think, it's best for me to do it soon-ish since I don't have concrete advice right now on how this should be done.

I'll also give @emmatown a couple of days to give their opinion about this. If we don't merge it this week - please ping me later.

siriwatknp:
@Andarist Would you mind taking a look at this again?

Andarist:
I think we can consider this to be approved **but** I still have to figure out how types should be structured and I have some other things on my plate right now (including figuring out how to integrate Emotion with the upcoming React 19). How urgent/blocking this one is for you? it might help me to prioritize this PR.

siriwatknp:
> I think we can consider this to be approved **but** I still have to figure out how types should be structured and I have some other things on my plate right now (including figuring out how to integrate Emotion with the upcoming React 19). How urgent/blocking this one is for you? it might help me to prioritize this PR.

I'd say the urgency is medium. We are in the process of migrating components to support both emotion and zero-runtime engine, some need dynamic callback support so those are blocked by this PR.

Issue #3111, opened 2023-10-02
Errors shown in console using Firefox in debug build with speedy enabled


fredrikcarlbom:
**Current behavior:**

In a debug build using speedy, using vendor specific css-prefixes such as `webkit` will cause errors to be written to the console in firefox.

**To reproduce:**

I don't succeed with a CodeSandbox-example since I can't figure out how to set speedy to true in that environment.

1. Define a component using a vendor-specific prefix such as `"::-webkit-scrollbar-track:hover{border-width:0.3125rem;background-color:rgba(255, 255, 255, 0.15625);}`
2. Set speedy true on the CacheProvider
3. Do a non-production build
4. Run the app in firefox
5. An error like the following `There was a problem inserting the following rule: "::-webkit-scrollbar-track:hover{border-width:0.3125rem;background-color:rgba(255, 255, 255, 0.15625);}" DOMException: An invalid or illegal string was specified`

**Expected behavior:**

No errors in the console

**Environment information:**

    "@emotion/react": "11.11.1",
    "@emotion/styled": "11.11.0",
    "react": "18.2.0",
    "react-dom": "18.2.0",

fredrikcarlbom:
There is a some filtering on errors at [sheet/src/index.js](https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/sheet/src/index.js#L150) but it seems to focus on filtering out non chromium errors. Not sure if it is wanted to take browser into account when doing the filtering.

PR #3150, opened 2024-01-17
 Add -webkit-scrollbar pseudo-elements to ignored prefixes

silaslenz:
**What**:

Adds `-webkit-scrollbar-button`, `-webkit-scrollbar-thumb` and `webkit-scrollbar-track` to ignored vendor prefixes, similar to https://github.com/emotion-js/emotion/pull/2920, https://github.com/emotion-js/emotion/pull/2393 and https://github.com/emotion-js/emotion/pull/2149. Fixes https://github.com/emotion-js/emotion/issues/3111.

**Why**:

These errors appear in non-webkit-browsers (Firefox) when using these webkit specific pseudo-elements. 

These are only the ones we specifically use, I don't know if we want to add a more comprehensive list, or perhaps do a filter like `:(-webkit-:.*|-moz-.*|-ms-.*){` (though that may risk introducing false positives?)?

**How**:

The existing regex was extended.

**Checklist**:

- [ ] Documentation (N/A ; there does not seem to exist any documentation on this)
- [ ] Tests (N/A ; there seem to be no existing tests for this, and I'm not sure how to properly test this in the first place. I've manually tested this in Firefox)
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/lovely-drinks-sort.md
+++ b/.changeset/lovely-drinks-sort.md
@@ -1,0 +1,5 @@
+---
+'@emotion/sheet': patch
+---
+
+Do not log failed rule insertions in the speedy mode for even more vendor-prefixed pseudo-elements/classes like `-webkit-scrollbar-button`, `-webkit-scrollbar-thumb` and `-webkit-scrollbar-track`.
--- a/packages/sheet/src/index.js
+++ b/packages/sheet/src/index.js
@@ -147,7 +147,7 @@ export class StyleSheet {
       } catch (e) {
         if (
           process.env.NODE_ENV !== 'production' &&
-          !/:(-moz-placeholder|-moz-focus-inner|-moz-focusring|-ms-input-placeholder|-moz-read-write|-moz-read-only|-ms-clear|-ms-expand|-ms-reveal){/.test(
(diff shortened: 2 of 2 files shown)


Andarist:
> These are only the ones we specifically use, I don't know if we want to add a more comprehensive list, or perhaps do a filter like :(-webkit-:.*|-moz-.*|-ms-.*){ (though that may risk introducing false positives?)?

Those are known to be vendor-prefixed though so perhaps this would just cover everything automatically without a huge risk of false positives?

silaslenz:
> > These are only the ones we specifically use, I don't know if we want to add a more comprehensive list, or perhaps do a filter like :(-webkit-:._|-moz-._|-ms-.*){ (though that may risk introducing false positives?)?
> 
> Those are known to be vendor-prefixed though so perhaps this would just cover everything automatically without a huge risk of false positives?

Sounds reasonable to me, though I may not have enough CSS/emotion knowledge to make that judgement. Do you want me to update the PR to use that filter instead?

silaslenz:
I've changed the PR to use the more generic regex. Here is an example to show that all three types of vendor-prefixes are still caught:

image
(source https://regex101.com/r/oOCpFp/1)

silaslenz:
@Andarist Any chance you could take a look at the latest version?

PR #3159, opened 2024-02-27
Remove hasOwnProperty from utils

iegik:
**What**:

Exported hasOwnProperty conflicts with exports.prototype.hasOwnProperty.

**Why**:

Because it tries to override itself.

**How**:

**Checklist**:

- [ ] Documentation
- [x] Tests
- [ ] Code complete
- [x] Changeset


Andarist (changes requested):
If this is fixing something this PR should have a test case included to prevent future regressions. If the test is too hard to setup then at the very least we should get cloneable repository that would reproduce the fixed issue.

packages/react/src/utils.js
```diff
@@ -1,4 +1,2 @@
 // @flow
 export let isBrowser = typeof document !== 'undefined'
-
-export const hasOwnProperty = {}.hasOwnProperty
```
Andarist, line 4: Instead of removing this let's rename this to `hasOwnProp` or `hasOwn` (since there is a static `Object.hasOwn` the latter might be the appropriate name)

iegik: Why we should keep hasOwn in utils? I have replaced all `hasOwnProperty` mentions with `hasOwn`

iegik:
```
'use strict'
Object.freeze(Object.prototype)
var exports = {}
// exports.hasOwnProperty = Object.hasOwnProperty
// exports.hasOwnProperty = () => 'injected!'
exports.hasOwn = () => 'injected!'
Object.hasOwn(Object.prototype, 'hasOwnProperty')
// true
```

Andarist: > Why we should keep hasOwn in utils?

To avoid problems with `{ hasOwnProperty: 'whatever' }`. 

> I have replaced all hasOwnProperty mentions with hasOwn

It's a fairly new addition to the language and it's not supported by all of our targets so we can't use it.

iegik: Can we replace it with `Object.hasOwnProperty.call`?

`export const hasOwn = Object.hasOwnProperty.call`

iegik: Or `export const hasOwn = Object.prototype.hasOwnProperty?.call || () => 'not sure about that - target do not support hasOwnProperty method'`

.changeset/khaki-brooms-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Remove hasOwnProperty from utils
```
Andarist, line 5: Renamed an internal `hasOwnProperty` to `hasOwn`. This avoids problems in CommonJS environments when the consumer tries to prevent prototype pollution with `Object.freeze(Object.prototype)`.

Andarist: eh, this was meant to be a code suggestion and not a regular comment 😅

PR #3175, opened 2024-04-06
Fix TypeScript error with function interpolations

### LINKED ISSUE #3174: Typescript issue with CSSObject in version 11.11.5 of @emotion/styled
**Current behavior:**

Update @emotion/styled from 11.11.0 => 11.11.5

**To reproduce:**
We have several styled components in our project and we started to get TS errors on a lot of them after this update:

Example component that doesn't work:

```

export const StyledHeader = styled('div')(() => {
  return {
    textAlign: `center`,
    position: `relative`,
  };
});
```

Gives this error:

```
error TS2769: No overload matches this call.
  Overload 1 of 3, '(template: TemplateStringsArray, ...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<...> & HTMLAttributes<...> & { ...; }>[]): StyledComponent<...>', gave the following error.
    Argument of type '() => { textAlign: string; position: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Overload 2 of 3, '(template: TemplateStringsArray, ...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<...> & HTMLAttributes<...> & { ...; }>[]): StyledComponent<...>', gave the following error.
    Argument of type '() => { textAlign: string; position: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Overload 3 of 3, '(...styles: Interpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>[]): StyledComponent<...>', gave the following error.
    Argument of type '() => { textAlign: string; position: string; }' is not assignable to parameter of type 'Interpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>'.
      Type '() => { textAlign: string; position: string; }' is not assignable to type 'FunctionInterpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>'.
        Type '{ textAlign: string; position: string; }' is not assignable to type 'Interpolation<{ theme?: Theme | undefined; as?: ElementType<any, keyof IntrinsicElements> | undefined; } & ClassAttributes<HTMLDivElement> & HTMLAttributes<...> & { ...; }>'.
          Type '{ textAlign: string; position: string; }' is not assignable to type 'CSSObject'.
            Types of property 'position' are incompatible.
              Type 'string' is not assignable to type 'Position | readonly NonNullable<Position | undefined>[] | readonly Position[] | undefined'.

18 export const StyledHeader = styled('div')(() => {

```

This is a breaking change.

Strangely it seems to depend on which css-attributes you are styling. This works for example:

```
export const StyledRecentSearch = styled('div')(() => {
  return {
    display: 'flex',
    justifyContent: 'space-between',
    alignItems: 'flex-start',
  };
});

```

There are several ways around this for example these variants seem to work:

```
export const StyledHeader = styled.div`
    textAlign: 'center',
    position: 'relative',  
`;
```
  or
  
 ```
 export const StyledHeader = styled('div')(() => {
  return {
    textAlign: 'center',
    position: 'relative',
  } as CSSObject;
});
```

**Expected behavior:**

There should be no breaking changes in patch versions (in this case TS errors)

Code like this should work since it returns a valid CSSObject:

```
export const StyledHeader = styled('div')(() => {
  return {
    textAlign: `center`,
    position: `relative`,
  };
});
```

**Environment information:**

- `react` version: 18.2.0
- `@emotion/react` version: 11.11.4
- `@emotion/styled` version: 11.11.5

### PR DESCRIPTION
**Pull request by Cerber-Ursi**
**What**:

After recent change to CSSTypes (#3141, #3164), some uses of styled components, namely the ones which provided a function interpolation, started to fail to typecheck. This adds a couple of tests for these use-cases and a corresponding fix.

**Why**:

This fixes #3174.

**How**:

As hinted on by @Andarist, the reason for the bug is that TypeScript caches the inference data from the failed overloads, even if this forces otherwise-matching ones to fail too. Therefore, the fix is to add "dummy" overload, never taken due to having a `never`-typed field in it, to guide type inference correctly.

**Checklist**:

- [ ] Documentation (N/A)
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -63,6 +63,16 @@ export interface CreateStyledComponent<
   SpecificComponentProps extends {} = {},
   JSXProps extends {} = {}
 > {
+  // This signature is actually never taken (due to `marker: never`); it's here to guide type inference.
+  // See https://github.com/emotion-js/emotion/issues/3174 for context.
+  (
+    ...styles: Array<
+      Interpolation<
+        ComponentProps & SpecificComponentProps & { theme: Theme }
+      > & { marker: never }
+    >
+  ): StyledComponent<ComponentProps, SpecificComponentProps, JSXProps>
+
   (
     template: TemplateStringsArray,
     ...styles: Array<
--- a/.changeset/thin-bears-talk.md
+++ b/.changeset/thin-bears-talk.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Fix incorrectly inferred types when using function interpolations
--- a/packages/styled/types/tests.tsx
+++ b/packages/styled/types/tests.tsx
@@ -301,3 +301,32 @@ const Input5 = styled.input`
   // $ExpectError
   styled('div', { shouldForwardProp: (prop: 'foo') => true })({})
 }
+
+{
+  // Different ways to provide props to styled components
+
+  styled.div`
+    textAlign: ${'center'},
+    position: ${'relative'},
+  `
+
+  styled.div({
+    textAlign: `center`,
+    position: `relative`
+  })
+
+  styled.div(() => ({
+    textAlign: `center`,
+    position: `relative`
+  }))
+
+  styled('div')({
+    textAlign: `center`,
+    position: `relative`
+  })
+
+  styled('div')(() => ({
+    textAlign: `center`,
+    position: `relative`
+  }))
+}

### CODE REVIEW

File: packages/styled/types/base.d.ts [hunk 1]
```diff
@@ -63,6 +63,16 @@ export interface CreateStyledComponent<
   SpecificComponentProps extends {} = {},
   JSXProps extends {} = {}
 > {
+  // This signature is actually never taken (due to `marker: never`); it's here to guide type inference.
+  // See https://github.com/emotion-js/emotion/issues/3174 for context.
```
Andarist, line 67: I'm not a fan of this - mainly because it introduces more work to the compiler. Overloads are matched from the top to the bottom so this definitely results in some typechecking perf hit. I understand that this might be inevitable though.

Can we think of any alternative solution that would reorder overloads back but would prevent the first ones from matching tagged template calls? 

I've done some testing and it seems that the main problem that we have here is that interpolated components become inference sources for `AdditionalProps` or something like that. When I cast them to strings then all tests pass - see the diff below
<details>
<summary>git diff</summary>

```diff
diff --git a/packages/styled/types/base.d.ts b/packages/styled/types/base.d.ts
index b8e1c93e..05efbaea 100644
--- a/packages/styled/types/base.d.ts
+++ b/packages/styled/types/base.d.ts
@@ -63,28 +63,10 @@ export interface CreateStyledComponent<
   SpecificComponentProps extends {} = {},
   JSXProps extends {} = {}
 > {
-  // This signature is actually never taken (due to `marker: never`); it's here to guide type inference.
-  // See https://github.com/emotion-js/emotion/issues/3174 for context.
-  (
-    ...styles: Array<
-      Interpolation<
-        ComponentProps & SpecificComponentProps & { theme: Theme }
-      > & { marker: never }
-    >
-  ): StyledComponent<ComponentProps, SpecificComponentProps, JSXProps>
-
-  (
-    template: TemplateStringsArray,
-    ...styles: Array<
-      Interpolation<ComponentProps & SpecificComponentProps & { theme: Theme }>
-    >
-  ): StyledComponent<ComponentProps, SpecificComponentProps, JSXProps>
-
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
    */
-  <AdditionalProps extends {}>(
-    template: TemplateStringsArray,
+  <AdditionalProps extends {} = {}>(
     ...styles: Array<
       Interpolation<
         ComponentProps &
@@ -98,10 +80,18 @@ export interface CreateStyledComponent<
     JSXProps
   >
 
+  (
+    template: TemplateStringsArray,
+    ...styles: Array<
+      Interpolation<ComponentProps & SpecificComponentProps & { theme: Theme }>
+    >
+  ): StyledComponent<ComponentProps, SpecificComponentProps, JSXProps>
+
   /**
    * @typeparam AdditionalProps  Additional props to add to your styled component
    */
-  <AdditionalProps extends {} = {}>(
+  <AdditionalProps extends {}>(
+    template: TemplateStringsArray,
     ...styles: Array<
       Interpolation<
         ComponentProps &
diff --git a/packages/styled/types/tests.tsx b/packages/styled/types/tests.tsx
index 14b7378a..09eb0d5c 100644
--- a/packages/styled/types/tests.tsx
+++ b/packages/styled/types/tests.tsx
@@ -90,16 +90,16 @@ const StyledOriginal = styled(Original, {
 const Label = styled.label``
 const Button = styled.button``
 const Input = styled.input`
-  & + ${Label}: {
+  & + ${String(Label)}: {
     margin-left: 3px;
   }
 `
 
 const Input2 = styled.input`
-  & + ${Label}: {
+  & + ${String(Label)}: {
     margin-left: 3px;
   }
-  & + ${Button}: {
+  & + ${String(Button)}: {
     margin-left: 3px;
   }
 `
``` 
</details>

PR #3198, opened 2024-06-15
Update Preconstruct and use import conditions

Andarist:


packages/server/package.json
```diff
@@ -57,32 +57,30 @@
...
   "preconstruct": {
     "entrypoints": [
       "./index.js",
       "./create-instance/index.js"
-    ],
-    "exports": {
-      "envConditions": [
-        "browser"
-      ]
-    }
```
Andarist, line 86: some packages had `browser` conditions but I didn't keep them because, from what I can tell, they were unused. I only kept them when I could recognize them as being used

packages/styled/package.json
```diff
@@ -75,17 +125,24 @@
...
+  "imports": {
+    "#is-development": {
+      "development": "./src/conditions/true.js",
+      "default": "./src/conditions/false.js"
+    },
+    "#is-browser": {
+      "worker": "./src/conditions/false.js",
+      "browser": "./src/conditions/true.js",
+      "default": "./src/conditions/false.js"
+    }
```
Andarist, line 137: this is an important question: do we assume that `default` means `node`? or do we introduce a new `node` entry and make `default` contain an actual runtime check?

packages/styled/package.json
```diff
@@ -49,20 +49,70 @@
   },
   "exports": {
```
Andarist, line 50: Just above this we have `package.json#browser`, we probably should just remove that from all packages. But then if we meant to support older browser we really should have the default bundle to contain the runtime check. The proposed setup wont benefit older bundlers but it wont break them.

emmatown: Yeah, definitely think the default bundle should include the runtime check like the current one does

packages/styled/package.json
emmatown: Yeah, the default bundle should definitely contain the runtime check and like we could have `worker` and `node` point to just false, idk if it's worth it since the server-side bundle size consideration is nowhere near as big as client-side and maybe this wouldn't actually be a problem but you could e.g. use JSDOM in Node for tests in which case you'd want `isBrowser` to be true

packages/styled/package.json
Andarist: ok, I'm removing `package.json#browser` since with `importConditions` those entries are not even auto-updated anymore

packages/styled/package.json
Andarist: i kinda feel like probably JSDOM users should use a jsdom condition or something but that's definitely not popular so ye, let's not add a node condition right now

Andarist: Btw. `jest-environment-jsdom` is already preconfigured with the browser condition (in new Jest versions that is). So `isBrowser` would already be `true` in that context

emmatown (approved):
I'd probably prefer if we ran all the tests in all environments by default and just had exceptions for the ones that only run in prod vs dev rather than running things only with the `development` condition by default but this is probably fine

Andarist:
I agree but i just didnt have the energy to go through lots of tests and make this more explicit. A lot of tests assume the dev env atm so a lot of them would have to be annotated explicitly. It would be cool to explore changing this latwr on… if anybody finds the time to do so 😅

PR #3206, opened 2024-07-05
Improved compatibility with the upcoming `@types/react` for React 19

DiegoAndai:
**What**: Replace global `JSX` with `React.JSX`

**Why**: Prepare for upcoming `@types/react` version which [removes global `JSX`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69022#:~:text=No%20replacement-,JSX%20Namespace,-A%20long%2Dtime)

**How**: Upgrade `@types/react` to `^18.2.6` which is the version that introduces `React.JSX` and use it to replace global `JSX`

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [ ] Code complete N/A
- [x] Changeset 

**Additional comments**:

- I've updated the site's `@types/react` as well to match the other packages.

Let me know if there's anything I might have missed.


DiegoAndai:
I'm looking into the build failures

DiegoAndai:
The snapshot tests are failing, but it seems like the diffs are about formatting. Not sure what I might've done incorrectly: https://github.com/emotion-js/emotion/actions/runs/9812857294/job/27097741626#step:4:16247 🤔

.changeset/curly-houses-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+'@emotion/styled': patch
+'@emotion/react': patch
+'emotion-site': patch
```
Andarist, line 4:
```suggestion
```

We should update Changesets and configure `privatePackages: false` so this doesn't confuse people in the future ;p

Andarist:
@emmatown `18.2.6` was published at 2023-05-06T11:32:52.704Z (over a year ago). I think we might have to go forward with this PR to improve compatibility with the upcoming React 19 for Emotion 11. This is kinda a breaking change for people who might be using older `@types/react` versions... but I don't think we have any alternative way. It's rather hard to reliably check at type level if `React.JSX` exists or not to branch our logic.

emmatown:
Hmmm, that's seems unfortunate. It doesn't seem to hard branch (with `@ts-ignore` of course), curious where it'd be unreliable?

Andarist:
We can’t trust error types that result from `ts-ignore`: https://github.com/microsoft/TypeScript/issues/58960

Andarist:
@emmatown wdyt?

emmatown:
Wouldn't something like this work without being susceptible to any issues like that? `FunctionComponent` no longer has a second legacy context parameter in React 19 types which we can check without any `@ts-ignore` stuff (there are probably other ways to check, this was just the first I tried) and then we use the right thing depending on the version and only those have `@ts-ignore` but the side where `@ts-ignore` does something is never used

```ts
type JSXElement = 2 extends Parameters<React.FunctionComponent<any>>['length']
  ? // @ts-ignore
    JSX.Element
  : // @ts-ignore
    React.JSX.Element
```

.changeset/curly-houses-jump.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/styled': patch
+'@emotion/react': patch
+---
+
+Replace global JSX type with React.JSX
```
Andarist, line 6:
```suggestion
Improved compatibility with the upcoming `@types/react` for React 19 where the global `JSX` namespace doesn't exist anymore
```

Andarist:
Do u always have to have the answer? 😉 

I applied the suggestion. I think this should be good to go now.

Andarist (approved):

emmatown (approved):

Issue #2836, opened 2022-07-24
Don't add classname for `:is()` and `:where()` selectors 


siriwatknp:
**The problem**

I want to leverage [`:where` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) (it is good for theming because it has 0 specificity) to style a component based on the parent attribute.

However, emotion prefix the style with the generated class name in front of `:where` which does not make the style works:

```js
const Component = style('div')`
  color: black
  :where([data-color-scheme="dark"]) & {
    color: white
  }
`
```

turned into

```js
<style>
// to make this style work, the `.css-9qpay1` at the beginning should be removed
.css-9qpay1:where([data-color-scheme="dark"]) .css-9qpay1{color:white;}
</style>
```

Here is the reproduction: https://codesandbox.io/s/emotion-forked-e9fyes?file=/index.js

**Proposed solution**

Should `:is` and `:where` be the exceptional case to not implicitly append to the class name?

```js
const Component = style('div')`
  :where([data-color-scheme="dark"]) & {
    color: white
  }
  // result as ':where([data-color-scheme="dark"]) .css { color: white };
`

const Component = style('div')`
  &:where([data-color-scheme="dark"]) {
    color: white
  }
  // result as '.css:where([data-color-scheme="dark"]) .css { color: white };
`
```

**Alternative solutions**

Is there any workaround that I might have missed?

**Additional context**

Hi, I am from the MUI core team. We are working on the theming API to let developers theme components based on color schemes without having to read the mode from the theme.

Our API added an attribute to the html when the color scheme changes:

```js
// light mode
<html data-color-scheme="light">

// dark mode
<html data-color-scheme="dark">
```

Andarist:
This is a byproduct of our `compat` plugin - however, this matches the behavior of Emotion 10 for all pseudo-selectors. I'd like to remove this `compat` plugin in the next major version of Emotion but I'm worried that changing the current behavior, even for a subset of pseudo-selectors, would be a breaking change that we can't do (we can continue discussing this though, those 2 are fairly new additions and it somewhat makes sense to treat them differently).

Right now you sort of can achieve this by combining `namespace` plugin and `:global()`: https://codesandbox.io/s/emotion-forked-rc5rpb?file=/index.js . Note that the generated rule has increased specificity because it uses the class name twice - maybe this could be fixed

siriwatknp:
@Andarist Thanks a lot for the info. 

[...]

PR #3210, opened 2024-07-12
Add option `explicitAmpersand` on @emotion/cache to handle :is and :where, etc

thomaslindstrom:
Based on the following issue: https://github.com/emotion-js/emotion/issues/2836
This adds an opt-in option on `@emotion/cache` `explicitAmpersand` (unsure about name) that *disables* the automatic prefixing of pseudo classes in the CSS.

**What**:

I identified an issue in my own CSS with newer pseudo classes like :where and :is, where Emotion automatically prefixed those classes with the current class name. This was not the expected behavior, and there was no way for me to opt-out of this behavior. 

Consider:

```jsx
<article
  css={css`
    color: black;

    :is(p, ul, ol) + :is(p, ul, ol) {
      margin-top: 1em;
    }

    :where([data-theme="dark"]) & {
      color: white;
    }
  `}
/>;
```

where the output with the `explicitAmpersand` set to `false` (the current behavior) will output:

```css
.emotion-article:is(p, ul, ol)+:is(p, ul, ol) {}
.emotion-article:where([data-theme="dark"]) & {}
```

But with `explicitAmpersand` set to `true` (new behavior):

```css
.emotion-article :is(p, ul, ol)+:is(p, ul, ol) {}
:where([data-theme="dark"]) & {}
```

**Why**:

The current behavior of automatic prefixing of the parent class name is confusing and breaks the behavior of new psuedo classes.

**How**:

Added an opt-in option on `@emotion/cache` `explicitAmpersand`.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [ ] Changeset

### DIFF
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -60,3 +60,37 @@ A DOM node that emotion will insert all of its style tags into. This is useful f
 `boolean`
 
 A boolean representing whether to prepend rather than append style tags into the specified container DOM node.
+
+### `explicitAmpersand`
+
+`boolean`
+
+Defaults to `false`. Opt-out of automatic pseudo class prefixing.
+
+Consider the following component:
+
+```jsx
+<article
+  css={css`
+    :where([data-theme='dark']) & {
+      /* ... */
+    }
+  `}
+/>
+```
+
+With `explicitAmpersand` set to `true`:
+
+```css
+:where([data-theme='dark']) .emotion-article {
+  /* ... */
+}
+```
+
+With `explicitAmpersand` set to `false`:
+
+```css
+.emotion-article:where([data-theme='dark']) .emotion-article {
+  /* ... */
+}
+```
--- a/packages/cache/__tests__/__snapshots__/index.js.snap
+++ b/packages/cache/__tests__/__snapshots__/index.js.snap
@@ -57,4 +57,133 @@ exports[`should accept insertionPoint option 1`] = `
 </head>
 `;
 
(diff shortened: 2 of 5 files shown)


Andarist:
@emmatown I think we should do something about it. In the next major version, we should drop the `compat` plugin. So the question is, how do we ease the migration? The `compat` plugin could nudge people to enable this option when it finds a selector that has to be "patched". It's not exactly ergonomic to enable this using cache options... but it's also the only way to do it it only for the current Emotion instance etc. In general, I'm in favor of this change.

emmatown:
Unless I'm misunderstanding something about specificity, there's a very simple workaround to this of using `*`, right? Like `*:where()` and `:where()` are equivalent, right?

I still agree with changing the behaviour though, pretty much just since it's the behaviour in the standard CSS nesting now. I'm not sure about the name of the option though since there are still implicit ampersands, they're just inserted in a different place now.

> The compat plugin could nudge people to enable this option when it finds a selector that has to be "patched"

By `a selector that needs to be "patched"`, do you mean selectors with things like `:hover` or selectors with things like `:where` or both?

Andarist:
> By a selector that needs to be "patched", do you mean selectors with things like :hover or selectors with things like :where or both?

Both. Currently we "patch" some selectors by inserting `&` into them. If we plan to change this (and we bot agree that we should) then for migration purposes we should warn about all of them so people can fix all of them in their code before using the new option

emmatown:
> Both

Sounds good then

PR #3215, opened 2024-07-19
Added `edge-light` and `workerd` conditions

Andarist:
Next.js plans to migrate away from using `worker` condition because that pulls in React browser builds... so I have been asked to add their dedicated condition.

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -11,10 +11,20 @@
         "default": "./dist/emotion-cache.cjs.js"
       },
       "development": {
+        "edge-light": {
+          "module": "./dist/emotion-cache.development.edge-light.esm.js",
+          "import": "./dist/emotion-cache.development.edge-light.cjs.mjs",
+          "default": "./dist/emotion-cache.development.edge-light.cjs.js"
+        },
         "worker": {
-          "module": "./dist/emotion-cache.development.worker.esm.js",
-          "import": "./dist/emotion-cache.development.worker.cjs.mjs",
-          "default": "./dist/emotion-cache.development.worker.cjs.js"
+          "module": "./dist/emotion-cache.development.edge-light.esm.js",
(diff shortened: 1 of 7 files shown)


packages/cache/package.json
```diff
@@ -11,10 +11,20 @@
         "default": "./dist/emotion-cache.cjs.js"
       },
       "development": {
+        "edge-light": {
+          "module": "./dist/emotion-cache.development.edge-light.esm.js",
+          "import": "./dist/emotion-cache.development.edge-light.cjs.mjs",
+          "default": "./dist/emotion-cache.development.edge-light.cjs.js"
+        },
         "worker": {
```
eps1lon, line 19: Do other libraries not use their browser builds for Service Workers?

Andarist: No idea, there are no great guidelines for this and various tools are not exactly overly consistent and explicit when it comes to differentiating between browser and worker builds.

Webpack 5 docs have always suggested using the `worker` condition for (web)workers... so that's what we went with here. 

But I'm also not sure if I have understood the intention behind your question. Could you elaborate?

Andarist: What we found in the past is that some tools would pick up the browser condition when targeting workers so we had to use some earlier condition (`worker`) to "shadow" over the `browser` condition. Ideally, it would just fall through to the default condition and use that

eps1lon: Just wondering if it makes more sense to use browser builds. But if the browser builds access the DOM, they're not suited for Service Workers. Given that there's little guidance/standardization around this topic we'll probably just have to find out.

Andarist: Ye, so that's the main distinction that we have here - `#is-browser` could be called `#can-use-dom`. That's the only thing we care about atm. So we can't use the browser builds because they assume being able to access DOM

emmatown (approved):

PR #3219, opened 2024-07-22
Remove incorrect tripleslash directives referencing node types

Andarist:


packages/server/types/create-instance.d.ts
```diff
@@ -1,7 +1,7 @@
 // Definitions by: Junyoung Clare Jang <https://github.com/Ailrun>
 // TypeScript Version: 2.8
 
-/// <reference types="node" />
+/// <reference types="node" preverve="true" />
```
emmatown, line 4:
```suggestion
/// <reference types="node" preserve="true" />
```

emmatown (approved):

PR #3240, opened 2024-08-14
Add scale property as unitless

justinfernald:
**What**:

Added scale as a unitless property.

**Why**:

Since scale is a fairly new well adopted property and does not take px value but instead will take unitless values or percentages and it strange to put: `scale: "2"`
MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/scale
W3: https://www.w3.org/TR/css-transforms-2/#propdef-scale

**How**:

These changes were implemented by modifying the unitless package

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/nasty-rice-wonder.md
+++ b/.changeset/nasty-rice-wonder.md
@@ -1,0 +1,5 @@
+---
+'@emotion/unitless': minor
+---
+
+Add `scale` property as unitless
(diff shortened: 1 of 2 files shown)


justinfernald:
Note that React also has this implemented for their style prop: https://github.com/facebook/react/pull/25601

emmatown (approved):

PR #3241, opened 2024-08-21
[perf] WASM hashing function

romgrk:
I work at MUI and I've been doing some performance optimization. I've noticed that `murmur2` is at the top of the stack traces for the [test case](https://gist.github.com/romgrk/129623975dee66e90a9d2373a33eef7a) benchmark I'm using, so I wanted to open a PR here to discuss options to improve the performance.

image

I've extracted some actual style strings that are hashed at runtime by emotion (from the test case above) and [created a benchmark](https://gist.github.com/romgrk/92c7516944dfd12d08977f6eb44eef06) to compare emotion's JS `murmur2` implementation to alternatives. I used rust+wasm to implement those alternatives, because the use-case is a good match for the performance benefits of wasm. Here is a summary of different versions compared to the current one

Passing strings from JS to WASM implies encoding them in an array buffer, so the performance scales on the input string length. The original JS version performs about 25% faster than the rest for small strings (up until a length of around 64 characters), but after that the WASM versions take the lead. The table below shows the results for the strings extracted at runtime described above which is imho close enough to what happens in production for real-word users.

| version | ops/s | result |
| --- | --: | --- |
| `murmur2_original` | 1 259 ops/s, ±0.67%   | 74.38% slower |
| `murmur2_rust()` | 3 635 ops/s, ±0.60%   | 26.04% slower |
| `xxh_rust()` | 4 915 ops/s, ±0.44%   | fastest |
| `xxh3_rust()` | 4 095 ops/s, ±0.55%   | 16.68% slower |
| `city_hash_rust()` | 4 109 ops/s, ±0.68%   | 16.4% slower |
| `adaptive()` | 4 102 ops/s, ±0.77%   | 16.54% slower |

The fastest option is [xxHash](https://xxhash.com/), which is newer and faster than murmur2, with a similar collision rate. `adaptive` is a combination of using `murmur2_original` for small strings and `xxh_rust` for large strings, but the additional branching instruction (`if (string.length > 64) ...`) makes it slower.

Using a WASM module also doesn't need to add any build/bundle time complexity for users, as we can ship the module inline, something like this. The compiled wasm module is around 500 bytes, so a realistic cost for the wrapper and base64 bytes would be around 700 bytes.

```
const wasm = WebAssembly.Module(atob('/* base64 encoded bytes */'))
```

I have also benchmarked the MUI test case again after replacing the hashing function. The results aren't as impressive as I expected (I was hoping for a 70% reduction of hashing time based on the table above), but this change improves MUI's rendering performance by about 2-4%.

| before | after |
| --- | --- |
| Screenshot from 2024-08-21 00-16-24 | Screenshot from 2024-08-21 00-15-57 |

The code in this PR is a draft but I'd be happy to have an early review/opinion before doing more work.


Andarist:
> Using a WASM module also doesn't need to add any build/bundle time complexity for users, as we can ship the module inline, something like this. The compiled wasm module is around 500 bytes, so a realistic cost for the wrapper and base64 bytes would be around 800 bytes.

800 bytes is still a sizeable % of the overall bundlesize IIRC. I also don't think we can assume that web assembly is available on the host platform so we'd have to include both versions to avoid breaking changes. cc @emmatown

romgrk:
For more precise numbers, XXH is 519 bytes of binary, and murmur2 is 386 bytes of binary, so total 800 and 650 after wrapping & base64 is a good figure.

romgrk:
I have improvements for the JS implementation but I have split them in #3242.

romgrk:
The performance numbers are all in the [top comment of that PR](https://github.com/emotion-js/emotion/pull/3242#issue-2477445585), but roughly 2x faster than the improved javascript implementation, which is itself roughly 2x faster than the current one (though `TextEncoder` is the most crucial part of that improvement).

I don't know for a JS xxHash, the only existing port I found, https://github.com/pierrec/js-xxhash, is written in an OOP style, so that won't fly for a performance improvement. I took a look at the [source code](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h) but it's much more dense than murmur2 so I haven't tried porting it yet.

On the same topic, would you be open to switch the algorithm if it increases the collision probability (reasonably)? I've been wondering if I could find a simpler algo that maintains a good enough collision rate. I'm not familiar with your codebase, how do you handle collisions?

Andarist:
Before considering the WASM version, I'd love to see what could be squeezed out of it using a performant JS port of xxHash. 

> On the same topic, would you be open to switch the algorithm if it increases the collision probability (reasonably)?

Potentially yes. 

> I'm not familiar with your codebase, how do you handle collisions?

We pray they won't happen.

romgrk:
I've found a [performant port of xxh](https://github.com/cgiosy/xxh32/blob/main/src/index.ts) but it can't beat murmur2 in javascript:

```
Running "Benchmark hash, string length = 1245" suite...

  murmur2_original():
    1 258 ops/s, ±0.53%   | slowest, 74.04% slower
  murmur2_improved():
    2 770 ops/s, ±0.68%   | 42.84% slower
  xxh_javascript():
    2 634 ops/s, ±0.64%   | 45.65% slower
  murmur2_rust():
    3 640 ops/s, ±0.75%   | 24.89% slower
  xxh_rust():
    4 846 ops/s, ±0.85%   | fastest
```

Having looked at the implementation for a few hashing functions, I think it's going to be hard to find anything faster than murmur2 in JS. The reason why xxh can be faster than murmur2 in WASM is that its algorithm is [easy to optimize & auto-vectorize](https://github.com/cgiosy/xxh32/blob/main/src/index.ts#L24-L27), so it can use better instructions than murmur2 can, in particular SIMD instructions that can only be accessed via WASM in the browser.

romgrk:
Follow-up of @emmatown [comment](https://github.com/emotion-js/emotion/pull/3242/files#r1728056511) in the other PR, the algo needs to be stable in case of server-rendering hydration, which means either murmur2 or xxh in both versions. You have all the numbers here, please let me know if/how you want to proceed.

oliviertassinari:
Goober seems to use a really simple hash algorithm: https://github.com/cristianbote/goober/blob/master/src/core/to-hash.js, and seems to be really fast on the server: https://twind.dev/handbook/introduction.html#opportunities

SCR-20240825-ouje

@cristianbote has been iterating on it for some time. But maybe it has a too high collision probability.

romgrk:
Funny thought I had but I took the fast hex formatting function from https://github.com/mui/material-ui/pull/43289 and replaced the final `n.toString(36)` by `formatHex(n)`, substantial performance improvement as well. So @Andarist @emmatown if you're willing to replace your classnames from `css-[base-36]` to `css-[base-16]`, this can also speed up the hashing.


image

Andarist:
FYI, I'm on PTO this week. I can continue the conversation when I get back

PR #3242, opened 2024-08-21
[perf] hashing function improvements

romgrk:
While I was working on #3241 I saw a few issues with your existing javascript implementation of murmur2, here are a few fixes. I've included a comparison with the alternatives just for the context, but the `murmur2_original` and `murmur2_improved` entries of this summary are the important ones. As you can see, we can more than double the performance with the changes in this PR.

image

### DIFF
--- a/packages/hash/src/index.ts
+++ b/packages/hash/src/index.ts
@@ -2,62 +2,69 @@
 // Inspired by https://github.com/garycourt/murmurhash-js
 // Ported from https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash2.cpp#L37-L86
 
+const encoder = new TextEncoder()
+
+// bufferLength must be a multiple of 4 to satisfy Int32Array constraints
+let bufferLength = 2 * 1024
+let buffer = new ArrayBuffer(bufferLength)
+let uint8View = new Uint8Array(buffer)
+let int32View = new Int32Array(buffer)
+
 export default function murmur2(str: string): string {
+  if (str.length > bufferLength) {
+    // buffer.resize() is only available in recent browsers, so we re-allocate
+    // a new buffer and views
+    bufferLength = str.length + (4 - str.length % 4)
+    buffer = new ArrayBuffer(bufferLength)
+
+    uint8View = new Uint8Array(buffer)
+    int32View = new Int32Array(buffer)
+  }
+
+  const length = encoder.encodeInto(str, uint8View).written;
+
   // 'm' and 'r' are mixing constants generated offline.
   // They're not really 'magic', they just happen to work well.
 
-  // const m = 0x5bd1e995;
-  // const r = 24;
+  const m = 0x5bd1e995;
+  const r = 24;
 
   // Initialize the hash
-
-  var h = 0
+  let h = 0
 
   // Mix 4 bytes at a time into the hash
 
-  var k,
-    i = 0,
-    len = str.length
-  for (; len >= 4; ++i, len -= 4) {
-    k =
-      (str.charCodeAt(i) & 0xff) |
-      ((str.charCodeAt(++i) & 0xff) << 8) |
-      ((str.charCodeAt(++i) & 0xff) << 16) |
-      ((str.charCodeAt(++i) & 0xff) << 24)
+  let i = 0
+  let len = length
+  for (; len >= 4; i++, len -= 4) {
+    let k = int32View[i]
 
-    k =
-      /* Math.imul(k, m): */
-      (k & 0xffff) * 0x5bd1e995 + (((k >>> 16) * 0xe995) << 16)
-    k ^= /* k >>> r: */ k >>> 24
+    k = Math.imul(k, m)
+    k ^= k >>> r
+    k = Math.imul(k, m)
 
-    h =
-      /* Math.imul(k, m): */
-      ((k & 0xffff) * 0x5bd1e995 + (((k >>> 16) * 0xe995) << 16)) ^
-      /* Math.imul(h, m): */
-      ((h & 0xffff) * 0x5bd1e995 + (((h >>> 16) * 0xe995) << 16))
+    h = Math.imul(h, m)
+    h ^= k
   }
 
   // Handle the last few bytes of the input array
 
   switch (len) {
     case 3:
-      h ^= (str.charCodeAt(i + 2) & 0xff) << 16
+      h ^= (uint8View[i * 4 + 2] & 0xff) << 16
     case 2:
-      h ^= (str.charCodeAt(i + 1) & 0xff) << 8
+      h ^= (uint8View[i * 4 + 1] & 0xff) << 8
     case 1:
-      h ^= str.charCodeAt(i) & 0xff
-      h =
-        /* Math.imul(h, m): */
-        (h & 0xffff) * 0x5bd1e995 + (((h >>> 16) * 0xe995) << 16)
+      h ^= uint8View[i * 4] & 0xff
+      h = Math.imul(h, m)
   }
 
   // Do a few final mixes of the hash to ensure the last few
   // bytes are well-incorporated.
 
   h ^= h >>> 13
-  h =
-    /* Math.imul(h, m): */
-    (h & 0xffff) * 0x5bd1e995 + (((h >>> 16) * 0xe995) << 16)
+  h = Math.imul(h, m)
+  h = (h ^ (h >>> 15)) >>> 0
 
-  return ((h ^ (h >>> 15)) >>> 0).toString(36)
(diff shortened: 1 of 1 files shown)


packages/hash/src/index.ts
```diff
@@ -2,62 +2,69 @@
...
+
+  const length = encoder.encodeInto(str, uint8View).written;
+
   // 'm' and 'r' are mixing constants generated offline.
   // They're not really 'magic', they just happen to work well.
 
-  // const m = 0x5bd1e995;
-  // const r = 24;
+  const m = 0x5bd1e995;
+  const r = 24;
```
romgrk, line 30: Inlining constants is not useful, the engine will do it quite early in its optimization passes.

packages/hash/src/index.ts
```diff
@@ -2,62 +2,69 @@
 // Inspired by https://github.com/garycourt/murmurhash-js
 // Ported from https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash2.cpp#L37-L86
 
+const encoder = new TextEncoder()
+
+// bufferLength must be a multiple of 4 to satisfy Int32Array constraints
+let bufferLength = 2 * 1024
+let buffer = new ArrayBuffer(bufferLength)
+let uint8View = new Uint8Array(buffer)
+let int32View = new Int32Array(buffer)
```
romgrk, line 11: Reading bytes with one `.charCodeAt(i)` call per byte (plus the bitwise operations to shift it in place) is less efficient than just translate everything into a byte array directly with `TextEncoder`. Very small strings might perform worse, but for the production dataset I tested with this is a positive change.

packages/hash/src/index.ts
```diff
@@ -2,62 +2,69 @@
...
-      /* Math.imul(k, m): */
-      ((k & 0xffff) * 0x5bd1e995 + (((k >>> 16) * 0xe995) << 16)) ^
-      /* Math.imul(h, m): */
-      ((h & 0xffff) * 0x5bd1e995 + (((h >>> 16) * 0xe995) << 16))
+  let i = 0
+  let len = length
+  for (; len >= 4; i++, len -= 4) {
+    let k = int32View[i]
+
+    k = Math.imul(k, m)
```
romgrk, line 42: `Math.imul` calls are better. It looks like a function call but the engine is free to transform this operation in a bitwise operation at the native level.

Using JS number operators (non-bitwise ones) is also risky. If a number overflows `2^31` in V8, it gets represented by a heap value (whereas below it's represented by a stack value, see https://v8.dev/blog/pointer-compression). Using `Math.imul` ensures the number stays as an `int32_t` value. This is also why I used an `Int32Array` to view the buffer, we absolutely want to keep numbers in the V8 Smi range.

packages/hash/src/index.ts
romgrk: Also `.charCodeAt()` isn't technically correct as it reads utf-16 code units, not bytes.

packages/hash/src/index.ts
```diff
@@ -2,62 +2,69 @@
 // Inspired by https://github.com/garycourt/murmurhash-js
 // Ported from https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash2.cpp#L37-L86
 
+const encoder = new TextEncoder()
```
romgrk, line 5: CI failure, `TextEncoder` doesn't exist in your testing environment, not sure where you prefer to fix this (here or adding it to the testing env).

Andarist: Fixing it in the test environment is preferred *if* we are sure this API will be available in all reasonable target environments. Note that, IIRC, we technically still support IE11. 😢

Andarist:
Have you confirmed that this performs better in all modern engines?

romgrk:
> Have you confirmed that this performs better in all modern engines?

Here is a comparison. The benchmark code is available [here](https://github.com/romgrk/js-benchmark) and requires bun (for JSC) and gjs (for SpiderMonkey).

image

Andarist:
It looks good, cc @emmatown

packages/hash/src/index.ts
romgrk: That's terrible :| It's the most substantial part of the improvement.

I could remove it, or I could split the implementation in two functions based on `typeof TextEncoder === 'undefined'`, but that's more bundle size and if I have a choice, I'd rather put those bytes towards adding the wasm implementation.

romgrk: Actually I could also write an encoder specifically for IE11, based on `.charCodeAt()`, I think I could fit that in a 100 bytes.

Andarist: > Actually I could also write an encoder specifically for IE11, based on .charCodeAt(), I think I could fit that in a 100 bytes.

yeah, the good thing is that we don't need it to be fully spec-compliant - we only need to support the things we care about

I'm open to different solutions to this, even at the expense of some added bundlesize (but well, we also cant add too much, right? :P). It's just a concern that we have right now. The next version will be free to use TextEncoder directly.

romgrk: I have added a legacy encoder implementation for IE11. The benchmarks are unchanged for modern engines. I tested it manually by setting `hasTextEncoder` to `false`, but I don't know how to test it in IE11.

This is ready for review as far as I can tell.

packages/hash/src/index.ts
```diff
@@ -2,62 +2,93 @@
...
+        bytes.push(codePoint & 0xff)
+      } else {
+        bytes.push(codePoint)
+      }
+    }
+    uint8View = new Uint8Array(bytes)
+    int32View = new Int32Array(uint8View.buffer, 0, Math.floor(bytes.length / 4))
+
+    return bytes.length
+  }
```
emmatown, line 31: This needs to encode like `TextEncoder` since the hash needs to return the same results if there e.g. is `TextEncoder` when a page is server-rendered but not on the client

packages/hash/src/index.ts
```diff
@@ -2,62 +2,93 @@
...
+
+  if (input.length > bufferLength) {
+    // bufferLength must be a multiple of 4 to satisfy Int32Array constraints
+    bufferLength = input.length + (4 - input.length % 4)
```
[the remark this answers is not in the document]
romgrk: Right, and there's no way to know the byte length until it has been iterated/encoded. I'll allocate 2x the space to ensure there's enough space for the worst case.

packages/hash/src/index.ts
romgrk: Self notes:
[utf16 to unicode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt)
[unicode to utf8](https://gist.github.com/MightyPork/52eda3e5677b4b03524e40c9f0ab1da5)

packages/hash/src/index.ts
```diff
@@ -2,62 +2,93 @@
...
+    // a new buffer and views
+    buffer = new ArrayBuffer(bufferLength)
+    uint8View = new Uint8Array(buffer)
+    int32View = new Int32Array(buffer)
+  }
+
+  return encoder!.encodeInto(input, uint8View).written as number;
+}
+
+export default function murmur2(input: string): string {
```
romgrk, line 47: While I was exploring the data a bit more I realized that the hashing function is re-hashing the same strings a lot, and that we could speed it up by caching it in a `Map`.

One note before the results, I was testing with a dataset made of a [bunch of MUI components copy/pasted](https://gist.github.com/romgrk/ade3ee62b4641d5b375127e98d354b5d). For this benchmark I switched to a new dataset that is directly extracted from the [MUI dashboard template](https://next.mui.com/material-ui/getting-started/templates/dashboard/), which I think is more realistic and reflects production setups even better.

Here are the results, which indicate a substantial improvement for that dataset when caching the murmur2 result. The implementation is [very naive](https://github.com/romgrk/js-benchmark/blob/master/test-murmur2.js#L135-L147), and I found that trying to mutate/update the map with more complex rules actually decreased the gains by a lot, probably because adding bookkeeping on each hash iteration is too much work. For comparison I also added a case for the cached version, but with a dataset with absolutely no duplicate, which means the caching code is pure overhead.

image

Last note, for this particular dataset, unlike the last dataset, SpiderMonkey shows a small/moderate performance decrease from `original` to `improved`, not sure why. This new dataset seems to have a bit shorter string styles, might explain why. I still think `original` to `improved` is a gain considering that Firefox accounts for less than 3% of the trafic right now.

So to sum this up, caching the hash result is a gain *if* we assume that we're going to see a lot of duplicates following each other, which seems to be the case for MUI cases. I would love to have your opinion on this change.

PR #3243, opened 2024-08-21
[perf] Remove `Insertion` wrapper

romgrk:
I'm not sure if there is a good reason to keep the `Insertion` wrapper that is created by the styled function. It creates additional object allocations (JSX & props) and creates more work for React (more vdom diffing).

I've removed it in this PR and benchmarked again with the same [test case](https://gist.github.com/romgrk/129623975dee66e90a9d2373a33eef7a) I've been using, here are the results:

```
Before: 104.2ms	+- 7.0
After:  99.6	+- 6.7

N = 20
```

### DIFF
--- a/packages/styled/src/base.js
+++ b/packages/styled/src/base.js
@@ -25,33 +25,6 @@ Because you write your CSS inside a JavaScript string you actually have to do do
 You can read more about this here:
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#ES2018_revision_of_illegal_escape_sequences`
 
-const Insertion = ({ cache, serialized, isStringTag }) => {
-  registerStyles(cache, serialized, isStringTag)
-
-  const rules = useInsertionEffectAlwaysWithSyncFallback(() =>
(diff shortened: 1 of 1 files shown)


Andarist:
Unfortunately, it exists for a reason. It would be a breaking change to remove it - we can explore this in the next major version that will target React 19 but I don't see us doing that right now. While users shouldn't rely on insertion order... they absolutely do that today. I'd expect many MUI users would be dissatisfied with this change in a minor bump.

Andarist:
> I figured that if I inlined the component at the same place in the tree where it was previously rendered, the side-effects would run in the same order as they previously would.

I might have screwed to replicate the exact structure that we are using here *but* I feel it's pretty close to it here: https://stackblitz.com/edit/vitejs-vite-egfmw1?file=src%2FApp.tsx,src%2Fmain.tsx&terminal=dev 

Notice how the parent's side effect is either executed before or after the side-effects from its children (based on the used insertion technique)

romgrk:
Makes sense, thanks for the details :+1: But yeah it's terrible for performance and DX (pollutes the react devtools), would be nice to remove it in the next major.

PR #3244, opened 2024-08-22
Small lints

romgrk:
Various minor lints, see comments below.

### DIFF
--- a/.changeset/curly-jars-punch.md
+++ b/.changeset/curly-jars-punch.md
@@ -3,4 +3,4 @@
 '@emotion/utils': patch
 ---
 
-Minor performance lints (improved regex, avoid some string work)
+Minor performance improvement
--- a/.changeset/curly-jars-punch.md
+++ b/.changeset/curly-jars-punch.md
@@ -1,0 +1,6 @@
+---
+'@emotion/serialize': patch
+'@emotion/utils': patch
+---
+
+Minor performance lints (improved regex, avoid some string work)
--- a/packages/serialize/src/index.ts
+++ b/packages/serialize/src/index.ts
@@ -365,7 +365,7 @@ function createStringFromObject(
   return string
 }
 
-let labelPattern = /label:\s*([^\s;\n{]+)\s*(;|$)/g
+let labelPattern = /label:\s*([^\s;{]+)\s*(;|$)/g
 
 let sourceMapPattern: RegExp | undefined
 if (isDevelopment) {
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -10,7 +10,7 @@ export function getRegisteredStyles(
 
   classNames.split(' ').forEach(className => {
     if (registered[className] !== undefined) {
-      registeredStyles.push(registered[className] as string)
+      registeredStyles.push(`${registered[className]};`)
     } else if (className) {
       rawClassName += `${className} `
     }
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -10,8 +10,8 @@ export function getRegisteredStyles(
 
   classNames.split(' ').forEach(className => {
     if (registered[className] !== undefined) {
-      registeredStyles.push(`${registered[className]};`)
-    } else {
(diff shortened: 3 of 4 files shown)


packages/utils/src/index.ts
```diff
@@ -10,8 +10,8 @@ export function getRegisteredStyles(
 
   classNames.split(' ').forEach(className => {
     if (registered[className] !== undefined) {
-      registeredStyles.push(`${registered[className]};`)
-    } else {
+      registeredStyles.push(registered[className] as string)
```
romgrk, line 13: `registered` contains serialized styles that are already terminated by a `;` by the serialization code. It avoids allocating a new string.

packages/utils/src/index.ts
```diff
@@ -10,8 +10,8 @@ export function getRegisteredStyles(
...
+    } else if (className) {
```
romgrk, line 14: `classNames` could have multiple sequential spaces, in which case `.split(' ')` would produce empty strings.

packages/serialize/src/index.ts
```diff
@@ -365,7 +365,7 @@ function createStringFromObject(
   return string
 }
 
-let labelPattern = /label:\s*([^\s;\n{]+)\s*(;|$)/g
+let labelPattern = /label:\s*([^\s;{]+)\s*(;|$)/g
```
romgrk, line 368: `\s` contains `\n`. See `node -p "'\\n'.match(/\\s/)"`

packages/utils/src/index.ts
Andarist: a test case probably could be written for this, does it create a problem though? a `rawClassName` has already been allocated, can't this use some kind of string ropes or something to stay optimized?

romgrk: Yeah someone might have [inserted additional whitespaces](https://github.com/mui/material-ui/pull/43363) in their classnames.

So sure engines will usually use a concatenated string like I describe [here](https://romgrk.com/posts/optimizing-javascript#8-use-strings-carefully) to optimize concat operations, but adding a single `' '` is likely too small to be worth it, engines are most likely just allocating a new string for those bytes.

I wrote a [benchmark](https://github.com/romgrk/js-benchmark/blob/master/test-get-registered-styles.js) with only this difference, it does seem to have a positive impact:

image

packages/utils/src/index.ts
romgrk: The `;` should definitely be changing the hash of strings passing through here, so that's expected.

The `;` feels overly defensive to me and I don't like that it's also creating more strings for the same reasons explained in the other comment. It would be nice to ensure that what goes in `cache.registered` is strictly `;` terminated, then we can safely do the change here and avoid modifying the strings everytime they pass here.

I'll revert it for now though, might require more investigation.

Andarist:
Could you add short changesets for this? Otherwise, those changes won't get released until some other PR mentions the touched packages in its changesets. The text that you put there will be automatically copied over to `CHANGELOG` files so it should be relevant for its potential readers (users of the projects, and not us - maintainers)

romgrk:
I have added the changeset, please let me know if there's anything else to do on my side.

packages/utils/src/index.ts
romgrk: For more context, I frequently see strings like this one:

image

.changeset/curly-jars-punch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@emotion/serialize': patch
+'@emotion/utils': patch
+---
+
+Minor performance lints (improved regex, avoid some string work)
```
Andarist, line 6:
```suggestion
Minor performance improvement
```

Andarist: i like that you mentioned what has been done but this text will get prepended to both CHANGELOGs. Each of those changes is relevant only to one of those packages. It isn't a big deal so I'm just simplifying the message so I can land it right now.

Conceptually there is no 1 changeset 1 PR rule :) we have freedom to have more changesets in a PR, even with overlapping package sets

Issue #3257, opened 2024-09-23
The monorepo won't complete yarn installation in NodeJS v20


stil:
I cloned `emotion-js/emotion` and tried to run `yarn` command to install all dependencies.

It fails on linking step due to outdated `parcel` dependency in https://github.com/emotion-js/emotion/blob/main/scripts/benchmarks/package.json#L16C5-L16C23

```
➤ YN0000: ┌ Link step
➤ YN0007: │ core-js@npm:3.19.1 must be built because it never has been before or the last one failed
➤ YN0007: │ @parcel/watcher@npm:2.0.2 must be built because it never has been before or the last one failed
➤ YN0007: │ lmdb-store@npm:1.6.13 must be built because it never has been before or the last one failed
➤ YN0007: │ msgpackr-extract@npm:1.0.15 must be built because it never has been before or the last one failed
➤ YN0009: │ lmdb-store@npm:1.6.13 couldn't be built successfully (exit code 1, logs can be found here: C:\Users\...\AppData\Local\Temp\xfs-983d4f5a\build.log)
➤ YN0009: │ msgpackr-extract@npm:1.0.15 couldn't be built successfully (exit code 1, logs can be found here: C:\Users\...\AppData\Local\Temp\xfs-02f5c3a7\build.log)
➤ YN0000: └ Completed in 5s 786ms
➤ YN0000: Failed with errors in 7s 476ms
```

Bumping `parcel` to `^2.12.0` fixes the problem.

More context as I thought it was `lmdb-js` issue first: https://github.com/kriszyp/lmdb-js/issues/307

PR #3258, opened 2024-09-23
Bump parcel to fix dependencies installation in NodeJS v20, fixes 3257

stil:
closes https://github.com/emotion-js/emotion/issues/3257


soren121 (commented):
You have to update the monorepo's lockfile too, with `yarn install` in the root directory.

PR #3270, opened 2024-11-12
Fix inconsistent hashes using development vs production bundles/`exports` conditions

### LINKED ISSUE #3222: Hydration errors in development for Next.js app with MUI Material v5 after updating emotion dependencies
**Current behavior:**

We are getting hydration errors in development on pages that render components which are styled MUI components, after updating the following dependencies:
* `@emotion/cache` from `11.11.0` to `11.13.0`
* `@emotion/react` from `11.11.4` to `11.13.0`
* `@emotion/styled` from `11.11.5` to `11.13.0`

```
Warning: Prop `className` did not match. Server: "MuiTableCell-root MuiTableCell-body MuiTableCell-sizeSmall e18whkzw0 mui-az1ynu-MuiTableCell-root-TableCellLinkContainer" Client: "MuiTableCell-root MuiTableCell-body MuiTableCell-sizeSmall e18whkzw0 mui-1odz46y-MuiTableCell-root-TableCellLinkContainer"
```

Notably, we did not update MUI or any related dependencies at the same time.

It only affects components created using the `styled` helper, e.g.:

```
import styled from "@emotion/styled";
import TableRow from "@mui/material/TableRow";

const StyledTableRow = styled(TableRow)({
  "&:last-child > th, &:last-child > td": {
    borderBottom: 0,
  },
});

export default StyledTableRow;
```

It seems to be a problem with one of those packages' underlying dependencies, because the issue disappears when we use the old `package-lock.json`, but not if we nuke and recreate it.

I've been trying to identify the root cause but I'm finding it difficult, I'm not very familiar with the internals of emotion. I think it could be a problem with the serialiser, because the client `<style>` tags being inserted look to me like the `data-emotion` attributes are missing the serialized classnames that typically seem to follow the cache prefix `mui`, e.g.:

image

It could also be a problem with the `hash` package, since the hashed part of the className is different.

The issue is not present in production builds of the app.

**To reproduce:**

 
CSB: TODO later if I have some more time

1. Make a styled MUI component as described above, put it on a Next.js page.
2. Visit that route in a browser, open devtools.

**Expected behavior:**

Classnames should match on both client & server, no hydration errors.

**Environment information:**

```
"react": "^18.3.1",
"@emotion/react": "^11.13.0",

"@emotion/cache": "^11.13.0",
"@emotion/server": "^11.11.0",
"@emotion/styled": "^11.13.0",

"@mui/material": "^5.16.4",
"@mui/material-nextjs": "^5.16.4",
```

emmatown:
Fixes #3222

Arguably this isn't really a bug in Emotion since tools like Next should use the conditions on external code correctly (it works correctly with the App Router since they bundle everything) but the reality is like this and the fix makes it less complex anyway imo.

This will sort of "break" source maps in older versions of the babel plugin with new versions of the runtime and vice versa but since it's only breaking source maps, I think that's fine.

Also, I checked the swc emotion plugin and it doesn't seem to do the particular optimisations that the Babel plugin does so it doesn't need to change.


.changeset/healthy-lies-decide.md
```diff
@@ -0,0 +1,11 @@
+---
+'@emotion/serialize': patch
+'@emotion/styled': patch
+'@emotion/cache': patch
+'@emotion/utils': patch
+'@emotion/css': patch
+'@emotion/react': patch
+'@emotion/babel-plugin': patch
```
emmatown, line 8: I'm releasing extra packages than necessary just so people can more easily can bump the thing they directly depend on rather than refreshing whole lock files.

packages/serialize/src/index.ts
```diff
@@ -429,14 +419,6 @@ export function serializeStyles(
...
     }
   }
-  let sourceMap
-
-  if (isDevelopment) {
-    styles = styles.replace(sourceMapPattern!, match => {
-      sourceMap = match
-      return ''
-    })
-  }
```
emmatown, line 439: This is the actual part that makes the hashes consistent

PR #3277, opened 2024-12-02
Convert `@emotion/cache`'s source code to TypeScript

Andarist:


packages/cache/src/index.tsx
```diff
@@ -19,32 +20,26 @@ import {
...
   insertionPoint?: HTMLElement
 }
-*/
 
 let getServerStylisCache = isBrowser
   ? undefined
-  : weakMemoize(() =>
-      memoize(() => {
-        let cache = {}
-        return name => cache[name]
```
Andarist, line 41: funny thing, I think this was incorrect since https://github.com/emotion-js/emotion/pull/1817 . It was working because we were able to store properties on this function that was returned from here. We should be returning an object though. The caller is responsible both for reading and writing to this "cache"

emmatown (approved):

PR #3281, opened 2024-12-03
Convert `@emotion/react`'s source code to TypeScript

Andarist:


.github/workflows/main.yml
```diff
@@ -88,5 +88,8 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/ci-setup
 
+      - name: build
+        run: yarn build
```
emmatown, line 92: Would I be correct in assuming this is added because the tsconfigs used for dtslint don't use a module resolution mode that supports the imports/exports fields so the build is necessary so preconstruct outputs stuff that for the default case doesn't require supporting the imports/exports fields?

packages/serialize/types/index.d.ts
```diff
@@ -1 +1 @@
-export * from '../src'
+export * from '..'
```
emmatown, line 1: These files are just for dtslint, right?

packages/react/src/jsx-namespace.ts
```diff
@@ -54,47 +54,53 @@ type ReactJSXIntrinsicElements = true extends IsPreReact19
   : /** @ts-ignore */
     React.JSX.IntrinsicElements
 
+// based on the code from @types/react@18.2.8
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3197efc097d522c4bf02b94e1a0766d007d6cdeb/types/react/index.d.ts#LL3204C13-L3204C13
+/** @ts-ignore */
```
emmatown, line 59: What's the new `@ts-ignore` for? Does the comment get moved around in the false case get moved around when generating the `.d.ts` file?

packages/react/src/jsx-namespace.ts
```diff
@@ -54,47 +54,53 @@ type ReactJSXIntrinsicElements = true extends IsPreReact19
...
+  ? string | React.JSXElementConstructor<any>
   : /** @ts-ignore */
     React.JSX.ElementType
 
 export namespace ReactJSX {
-  type ElementType = ReactJSXElementType
-  interface Element extends ReactJSXElement {}
-  interface ElementClass extends ReactJSXElementClass {}
-  interface ElementAttributesProperty
+  export type ElementType = ReactJSXElementType
```
emmatown, line 66: Am I correct in understanding that these two are equivalent? (since the former compiles into the latter)
 
```ts
export namespace Blah {
  export type Something = A
}
```

```ts
export declare namespace Blah {
  type Something = A
}
```

emmatown: loling at 

```ts
export namespace Blah {
  type A = {}
  export type Something = A
}
```
getting compiled into
```ts
export declare namespace Blah {
    type A = {};
    export type Something = A;
    export {};
}
```

(I know the context around `export {};`, just never thought about it for namespaces lol)

emmatown (approved):

packages/serialize/types/index.d.ts
emmatown: seems like the answer is yes from reading the commits

.github/workflows/main.yml
emmatown: seems like the answer is yes from reading the commits

packages/react/src/jsx-namespace.ts
emmatown: Seems like the answer is the comment gets removed but if that's true, shouldn't the comments be updated for the types above as well? Seems like you "fixed" it for the tests by bumping `@types/react` but that wouldn't fix it for consumers?

emmatown (changes requested):

packages/react/src/jsx-namespace.ts
emmatown: Maybe keeping this in a `.d.ts` file might make sense so we don't have to rely on comments being preserved in particular places? (don't have strong feelings about that though, just an idea)

Andarist: the one we had was removed, a JSDoc-style comment - in general - should be preserved but perhaps they are only preserved on accessible symbols. In the emitted declaration this gets folded into a single line and the new comment stays in place.

> Seems like you "fixed" it for the tests by bumping @types/react but that wouldn't fix it for consumers?

I was checking the emitted declaration files and this shouldn't affect consumers (for the reason outlined above). That said... now I'm surprised why this was erroring with the new `@types/react` in the first place. I updated the types before fighting this 🤔 

> Maybe keeping this in a .d.ts file might make sense so we don't have to rely on comments being preserved in particular places? (don't have strong feelings about that though, just an idea)

I agree it's not ideal but also not a big deal. The general mantra I've heard (from the TS team members too) is that `.d.ts` shouldn't quite be used when authoring libs in TS. I don't recall the exact reasons behind this though. Testing things out in TS workbench (see [this](https://www.typescriptlang.org/dev/bug-workbench/?out=true&outDir=dist#code/PTAEAEHsFcBcBECWAnAXKAJogzrAsAFAgQYCmAxgDYCGy1sikAdurMtKYYceAGaKVSTagFtS6REwAOcAHSxshRCKmRksUAG9QAMUiRQAX1C9kkEaADks4GSrZLXAqQAeq9aHLNcoAEbVhFl19UABeLRN9dEtLIycefkFhMXQ7SmxZDHlFAkJXdw1YAE8pUmCDcO1eKNBcZEkAcyMgA))... it seems that TS might not even "copy over" dts files into the `distDir`. You can see the output in the Debug tab in the "Virtual File System" section

Andarist: I mean, we could also just keep this `.d.ts` outside of `src`... but ugh, I don't feel overly confident that all of the relative paths would be rewritten correctly. Authoring everything in `src` feels safer/better

packages/react/src/jsx-namespace.ts
Andarist: Ye, those rules are super trippy tbh. I just slapped `export` onto those because they suddenly became unavailable when I moved this from an ambient content to `.ts` 🤷‍♂️

Andarist:
I'll land this for now since I'd like to rebase my local work on `@emotion/styled` migration but feel free to respond to any ongoing discussion. I'm not releasing this just yet so we can always patch things up before the release

packages/react/src/jsx-namespace.ts
emmatown: Continued in https://github.com/emotion-js/emotion/pull/3282

Issue #3184, opened 2024-05-04
Remove defaultProps copy


oliviertassinari:
**Current behavior:**

https://github.com/emotion-js/emotion/blob/73f688103c5307c90d7544508b6d935243e4299e/packages/styled/src/base.js#L183

**Expected behavior:**

No copy.

**Context**

React 19 removed the support for `defaultProps` on function components, and warned against it since 18.3.0, https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops since it was deprecated 5 years ago: https://github.com/facebook/react/pull/16210.

I think that at one point, we will want to remove this line to save bundle size and avoid confusion like in https://github.com/mui/mui-x/issues/11531 or https://github.com/react-simple-code-editor/react-simple-code-editor/pull/125#issuecomment-2204531315.

**Environment information:**

- `react` version: 18.3.1
- `@emotion/react` version: 11.11.4

Andarist:
We have to plan to do this in the next major version that will only be compatible with React 19 (at least that's my current line of thinking about this).

oliviertassinari:
The problem might be more widespread, just found the same issue with https://github.com/react-simple-code-editor.

Andarist:
Is the core of the problem that you are facing that users get a warning about this from React?

oliviertassinari:
@Andarist Correct, it's not 100% clear from the warning that the issue is with Emotion. We see styled() and emotion in the stack trace but I guess people ignore this, and just assume that it's not emotion since they don't see the error in other places:

- The first instance: https://github.com/mui/mui-x/issues/11531#issuecomment-1900992813
- The second instance: https://github.com/react-simple-code-editor/react-simple-code-editor/pull/125#issuecomment-2204531315

Andarist:
We can't quite remove this as it would be a breaking change. I think it would be fine to make this assignment conditional though and that should largely address this issue, right? The warning could still be produced but only if the wrapped function would actually have `.defaultProps` and at that point the problem would be with that component and not Emotion (the warning wouldn't quite hint at that but I doubt we can improve that)

oliviertassinari:
> I think it would be fine to make this assignment conditional though and that should largely address this issue, right?

Oh yeah speaking of just having a assignment conditional, I have another fun one for you 😄: #3204.

Andarist:
[...]

PR #3286, opened 2024-12-09
Removed `.defaultProps` support

Andarist:
fixes https://github.com/emotion-js/emotion/issues/3184

Technically this breaks `.defualtProps` on classes too and those are still supported by React 19. I don't quite think it's worth supporting it though and we can't easily rely on React doing the merging for us because we always output function components. So we'd have to either conditionally create wrapping classes OR do the "defaulting" ourselves

### DIFF
--- a/packages/jest/test/react-enzyme.test.js
+++ b/packages/jest/test/react-enzyme.test.js
@@ -241,15 +241,9 @@ const cases = {
   theming: {
     render() {
       const Button = styled.button`
-        color: ${props => props.theme.main};
-        border: 2px solid ${props => props.theme.main};
+        color: ${props => props.theme.main ?? 'red'};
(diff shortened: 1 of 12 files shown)


packages/jest/test/react-enzyme.test.js
```diff
@@ -3,14 +3,33 @@ import 'test-utils/enzyme-env'
...
+  function GetCache() {
+    cache = __unsafe_useEmotionCache()
+    return null
+  }
+  enzyme.shallow(<GetCache />)
+
+  cache.registered = {}
+  cache.inserted = {}
+  cache.sheet.flush()
+})
```
Andarist, line 31: this is a quick hack, I don't quite care about this - we'll be removing Enzyme's support (and thus this file) soon

emmatown (approved):

PR #3290, opened 2024-12-10
Drop `hoist-non-react-statics`

Andarist:


packages/react/src/theming.tsx
```diff
@@ -98,5 +97,5 @@ export function withTheme<
 
   WithTheme.displayName = `WithTheme(${componentName})`
 
-  return hoistNonReactStatics(WithTheme, Component)
+  return WithTheme
```
Andarist, line 100: tbf maybe we should just remove `withTheme` altogether...

emmatown: 🤷‍♀️  it takes like no code, is trivially tree shakable so I'd kinda err on the side of keeping it to just not break stuff but eh, I don't have strong feelings about it

Andarist: Yep, that’s also why i have decided to keep it in the end

PR #3232, opened 2024-08-01
Let `css` prop attachment be distributed over union types.

ENvironmentSet:
**What**: Let `css` prop attachment be distributed over union types. So that branches with `className` prop are extended with `css` prop.

**Why**: Some components may accept `className` prop conditionally. For example, [`Link` of wouter does.](https://github.com/molefrog/wouter/blob/f953ae6811c4e49201e11844038b71e531ab76d2/packages/wouter/types/index.d.ts#L129) Emotion skips these cases, however, and does not add `css` prop in anywhere. As far as I know, there is no runtime or compilation-time problem stopping us from supporting these kind of components. Thus I propose this change to enhance expressiveness of the library.

fixes #3185

**How**: Made `LibraryManagedAttributes` be distributed over a prop type in cases it is a union.

**Checklist**:

- [ ] Documentation "N/A"
- [x] Tests
- [x] Code complete
- [x] Changeset 

Should I add test cases for this change? Please let me know if you need.

Issue #3185, opened 2024-05-14
The `css` prop is not attached to a type of props if it is a disjoint union type on `className` field.


ENvironmentSet:
**Current behavior:**

`WithConditionalCSSProp<P>` doesn't create the `css` prop if `P` is a disjoin union type on `className` field. Therefore we can't use some components optionally taking `className` field, like [`Link` of wouter](https://github.com/molefrog/wouter/blob/607d2051c5ed9c3edd680c3e8ece75bf95a0b4fd/packages/wouter/types/index.d.ts#L120).

**To reproduce:**

Since it's not a problem inherent in emotion or react but a problem of treating union types, we can reproduce it with following simplified snippet.

```typescript
type WithConditionalCSSProp<P> = 'className' extends keyof P
  ? string extends P['className' & keyof P]
    ? { css?: 'Interpolation<Theme>' } // To make the snippet simpler, I've just used string literal instead.
    : {}
  : {}

type BUG = WithConditionalCSSProp<{ className: string, foo: number } | { bar: boolean }> 
```
[typescript playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBA6glsAFgYQPYDsAmC4YIYA2yAysQAoBOqYAPGQHxQC8UA5AMYF4DO3AcngC2EVlAgAPYBCzcoAawghUAMyhkAUFCgB+KN2AU46AOZjJ0zLLIBtDl14DhogGTzFKtQF1NWnVADeUOy82gBcbACS6FIUYKhcwLjoNAAqiBDC9KIAvj5a4f65+QG56qCQUABCAKoA4sywCCgY2In4RKSU1DSBnDz8QhDh+oYmADRQyqio4egAroIARhAUUNlQAD4BUIt4FOGL0wQQeOhrjEA)

[...]

PR #3293, opened 2024-12-12
Fix TypeScript error when writing higher order components while using Emotion's JSX namespace

### LINKED ISSUE #3245: v11.13.3 type error with Next
**Current behavior:**

type Error Occurs after v11.13.3

```
const withAuth = <P extends object>(WrappedComponent: NextPage<P>) => {
  const AuthComponent = (props: P) => {
    ...my logic
    return <WrappedComponent {...props} />;
  }
}
```

**Expected behavior:**

Type Error is fixed , if it's not error. is there any method to solve this type error?

**Environment information:**

- `react` version: 18.3.1
- `@emotion/react` version: 11.13.3

### LINKED ISSUE #3261: Props not assignable to `LibraryManagedAttributes` type error
**Current behavior:**

After updating `@emotion/react` to `11.13.3`, I start seeing the following type error in HOC (higher-order components):

```
src/hoc.tsx:7:13 - error TS2322: Type 'P' is not assignable to type 'IntrinsicAttributes & LibraryManagedAttributes<FC<P>, P>'.
  Type 'Record<string, unknown>' is not assignable to type 'IntrinsicAttributes & LibraryManagedAttributes<FC<P>, P>'.
    Type 'Record<string, unknown>' is not assignable to type 'LibraryManagedAttributes<FC<P>, P>'.
      Type 'P' is not assignable to type 'LibraryManagedAttributes<FC<P>, P>'.
        Type 'Record<string, unknown>' is not assignable to type 'LibraryManagedAttributes<FC<P>, P>'.

7     return <C {...props} />;
              ~


Found 1 error in src/hoc.tsx:7
```

Here is the code that produces this type error above:

```tsx
const Component: React.FC<{ children: React.ReactNode }> = ({ children }) => {
  return <div>{children}</div>;
};

export const hoc = <P extends Record<string, unknown>>(C: React.FC<P>) => {
  return (props: P) => {
    return <C {...props} />;
  };
};

hoc(Component);
```

After downgrading `@emotion/react` to `11.13.0`, the error stops appearing anymore.

**To reproduce:**

1. Clone this repository for reproduction: https://github.com/shawnrivers/emotion-react-hoc-type-error
2. Run `npm i`
3. Run `npm run type-check`

**Expected behavior:**

No type error inside the HOC (which is the behavior before `@emotion.react@11.13.3`.

**Environment information:**

- `react` version: 18.3.1
- `@emotion/react` version: 11.13.3

emmatown:
This regressed in #3232

Fixes #3245
Fixes #3261

### DIFF
--- a/.changeset/nervous-pigs-reflect.md
+++ b/.changeset/nervous-pigs-reflect.md
@@ -1,0 +1,5 @@
+---
+'@emotion/react': patch
+---
+
+Fix TypeScript error when writing higher order components while using `@emotion/react` `jsx` pragma/`jsxImportSource`
--- a/packages/react/src/jsx-namespace.ts
+++ b/packages/react/src/jsx-namespace.ts
@@ -6,11 +6,15 @@ type IsPreReact19 = 2 extends Parameters<React.FunctionComponent<any>>['length']
   ? true
   : false
 
-type WithConditionalCSSProp<P> = 'className' extends keyof P
-  ? string extends P['className' & keyof P]
-    ? { css?: Interpolation<Theme> }
-    : {}
-  : {}
+type WithConditionalCSSProp<P> =
+  | (P extends unknown
+      ? 'className' extends keyof P
+        ? string extends P['className' & keyof P]
+          ? { css?: Interpolation<Theme> }
+          : {}
+        : {}
+      : {})
+  | {}
 
 // unpack all here to avoid infinite self-referencing when defining our own JSX namespace for the pre-React 19 case
 
@@ -91,9 +95,8 @@ export namespace EmotionJSX {
   export interface ElementChildrenAttribute
     extends ReactJSXElementChildrenAttribute {}
 
-  export type LibraryManagedAttributes<C, P> = P extends unknown
-    ? WithConditionalCSSProp<P> & ReactJSXLibraryManagedAttributes<C, P>
-    : never
+  export type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+    ReactJSXLibraryManagedAttributes<C, P>
 
   export interface IntrinsicAttributes extends ReactJSXIntrinsicAttributes {}
   export interface IntrinsicClassAttributes<T>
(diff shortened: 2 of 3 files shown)


emmatown:
Hmmm, actually this is wrong and I think the right solution is probably not being distributive and knowingly not supporting the use case of #3185

Andarist:
> Being able to write a higher order component at all without type errors is more valuable than a component conditionally supporting className having the css prop accessible in types.

I agree. It seems that TS isn't able to relate a non-conditional type to a distributive conditional type (this is the only branch that checks `targetFlags & TypeFlags.Conditional` outside of `sourceFlags & TypeFlags.Conditional`):
https://github.dev/microsoft/TypeScript/blob/6a00bd2422ffa46c13ac8ff81d7b4b1157e60ba7/src/compiler/checker.ts#L23200-L23221

It doesn't even work in cases like this ([TS playground](https://www.typescriptlang.org/play/?ts=5.8.0-dev.20241212#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxAGcMAeAFQD4AKAKHngAcBGALnnIBp6mAmd8vBAAPQqmBF4aANaocAd3wB+DvABk8AN7wwRIkvYy5i+AF947VCABuIGNwCU7azizAA3LVpg8JePJgoRkY7eABeeFIABQB9GKFREHFJHAAjACtwDBpGGBxGInZYmIdwyi0eQhJqXPyiTiY8godPU08gA)):
```ts
declare function test<T>(
  p1: T,
  p2: T extends unknown ? T & { css?: unknown } : never,
): void;

const wrapper = <P__ extends object>(props: P__) => {
  test(
    props,
    props // error
  );
};
```

It seems that perhaps this rule is overly restrictive but I don't know how to safely loosen it to handle this case. And any potential fix wouldn't land in TS soon enough for us anyway - so ye, we have to revert that PR.

Andarist:
I opened a PR that could fix this but I can't really say with confidence that it's a correct one: https://github.com/microsoft/TypeScript/pull/60817

PR #3294, opened 2024-12-14
Drop `compat` plugin

Andarist:
closes https://github.com/emotion-js/emotion/issues/2836


emmatown (changes requested):
While I agree that we should do this eventually, I don't think we should merge this now and if we want to ship this, we should at minimum:
- Add warnings in Emotion 11 when the compat plugin changes stuff that explains why it's changing (to align closer with standard css nesting[^2]) and how to fix it (probably write `&:pseudo` unless you actually meant the standard meaning in which case write `*:pseudo` to silence this warning)
- Change our types from suggesting pseudo selectors in object types to either not suggesting them at all or suggesting `&:pseudo` since we're effectively suggesting people write the thing they almost certainly do not intend

And tbh even if we do those things, I think it might be good to keep the compat plugin doing the "wrong" thing and warning for the duration of v12 to ease migration for longe.r (though I don't feel as strongly about this)

imo, fixing this without doing those things would cause more pain that it would solve since people can deal with #2836 by just writing `*:where(...)` instead of `:where(...)` [^1] and that's easy to deal with since it will be broken at the time someone is writing code so authors will notice it and work around it but "properly" fixing this means breaking code that people have already written and not in a simple kind of break where it's just "there's a bunch of errors, fix all the errors and you're good" but "there are no warnings, only some of it is broken in subtle ways, you can probably find and replace your way to get most of it done but you might not have actually found all the cases".

Also, I think "we do nothing about this for v12" is also explicitly a completely fine answer and is preferable to shipping this with a bad migration path.

[^1]: Unless I'm missing something though no one has suggested that from https://github.com/emotion-js/emotion/pull/3210#issuecomment-2237858626 so far
[^2]: Though it still isn't actually the same as standard CSS nesting since that would require using `:is` to have the exact same specificity stuff but we wouldn't want to do that because browser support. We of course don't need to mention this in the warning, just wanted to note it.

Issue #2747, opened 2022-05-14
Do not include `prefixer` stylis plugin by default 


srmagura:
**The problem**

Automatically adding vendor prefixes for CSS properties like `flex` results in 

- Less readable CSS when using the browser dev tools
- Theoretically, slightly worse performance since Emotion must insert more CSS rules

In 2022, many of the things that used to require vendor prefixes are now fully standardized and work in all evergreen browsers. I cannot think of a single standard CSS property that requires vendor prefixes to work in certain browsers. I'm not saying there are literally 0 properties for which automatic vendor prefixes would be useful — I'm saying these properties are obscure enough that the vast majority of people aren't using them anyway.

**Proposed solution**

The Emotion cache does not include the `prefixer` plugin by default. You can still opt in to using it by passing it to `createCache`.

**Alternative solutions**

In Emotion 11, you can opt out of using `prefixer` by passing `stylisPlugins: []` to `createCache`.

ivan-kleshnin:
> I cannot think of a single standard CSS property that requires vendor prefixes to work in certain browsers

E.g. `justify-content: right` or `justify-content: end` still doesn't work in Android.
I have to use `justify-content: flex-end`.

aeharding:
> In Emotion 11, you can opt out of using prefixer by passing stylisPlugins: [] to createCache.

Does anyone know how to do this? I want to turn off all vendor prefixing.

Andarist:
Resolved in https://github.com/emotion-js/emotion/pull/3296

aarongarciah:
@aeharding you need to use `CacheProvider` alongside `createCache` from `@emotion/cache` package. More info in the docs: https://emotion.sh/docs/cache-provider

PR #3296, opened 2024-12-15
Removed auto-prefixing for widely available features

Andarist:
closes https://github.com/emotion-js/emotion/issues/2747

IMHO, this plugin is not really needed in 2025, but I haven't rechecked whether all the included rules are obsolete. Many are, we definitely shouldn't be prefixing flexbox


emmatown (commented):
I basically agree with this (and probably agree with shipping it in the next major in a way that I don't really with #3294) with the caveat that to understand whether all of it is actually obselete, it'd be good to see something like: run the stylis prefixer over some specifically generated css to exercise all the properties that stylis prefixes, compare that with caniuse/mdn compat data/whatever and see what we get

Andarist:
Quick look on caniuse:

| declaration                                                                                                      | browser support | notes                   |
| ---------------------------------------------------------------------------------------------------------------- | --------------- | ----------------------- |
| color-adjust                                                                                                     | 19%             | deprecated              |
| animation, animation-(delay\|direction\|duration\|fill-mode\|iteration-count\|name\|play-state\|timing-function) | 97%             |                         |
| text-decoration                                                                                                  | 83%             | 97% with prefixes       |
| filter                                                                                                           | 98%             |                         |
| clip-path                                                                                                        | 95%             |                         |
| backface-visibility                                                                                              | 97%             |                         |
| column                                                                                                           | 97%             |                         |
| box-decoration-break                                                                                             | 69%             | 98% with prefixes       |
| mask, mask-image, mask-(mode\|clip\|size), mask-(repeat\|origin), mask-position, mask-composite,                 | 92.5%           | 98% with prefixes       |
| background-clip                                                                                                  | 97%             |                         |
| columns, column-(count\|fill\|gap\|rule\|rule-color\|rule-style\|rule-width\|span\|width)                        | 97%             |                         |
| appearance                                                                                                       | 96%             |                         |
| user-select                                                                                                      | 79%             | 97% with prefixes       |
| transform,                                                                                                       | 98%             |                         |
| hyphens                                                                                                          | 94%             |                         |
| text-size-adjust                                                                                                 | 77%             | 92% with prefixes       |
| transition                                                                                                       | 97%             |                         |
| cursor                                                                                                           | 80%             | no change with prefixes |
| background, background-image                                                                                     | 97%             |                         |
| (margin\|padding)-inline-(start\|end)                                                                            | 97%             |                         |
| (min\|max)?(width\|height)                                                                                       | 97%             |                         |
| (min\|max)(inline-size\|block-size)                                                                              | 96%             |                         |
| stretch, max-content, min-content, fill-available                                                                | 97%             |                         |
| (m)ax-content, (m)in-content                                                                                     | 97%             |                         |
| (f)ill-available, (f)it-content                                                                                  | 97%             |                         |
| (s)tretch                                                                                                        | 98%             |                         |
| position: sticky                                                                                                 | 97%             |                         |
| writing-mode                                                                                                     | 98%             |                         |
| vertical-l(r)                                                                                                    | 97%             |                         |
| vertical-r(l)                                                                                                    | 97%             |                         |
| horizontal(-)tb                                                                                                  | 97%             |                         |
| :read-(only\|write)                                                                                              | 98%             |                         |
| :placeholder                                                                                                     | 97.7%           |                         |

Andarist:
🏓 @emmatown

emmatown:
Sorry, I looked more into this before the holidays and got side-tracked, basically my thoughts are since the number of properties where prefixes are still relevant seems to be not nothing, I think we should probably have a smaller prefixer that includes what is currently prefixed but excluding everything that is [Baseline "Widely Available"](https://web-platform-dx.github.io/web-features/) (I don't think the estimated browser support percentages to decide things is very useful). Over time, it would shrink, though the exact details of how we'd want to do that wrt to major versions, I'm not super sure about.

Andarist:
@emmatown I changed the PR and kept the prefixer. I cross-checked with caniuse, the resource u mentioned here and MDN.  The end result is that we end up prefixing just a few things. Could you take a look at this and let me know what do you think?

packages/cache/src/prefixer.ts
```diff
@@ -49,87 +19,12 @@ function prefix(value: string, length: number): string {
...
-    case 4215:
-    case 6389:
-    case 5109:
-    case 5365:
-    case 5621:
-    case 3829:
       return WEBKIT + value + value
-    // appearance, user-select, transform, hyphens, text-size-adjust
-    case 5349:
+    // user-select, hyphens, text-size-adjust
```
emmatown, line 23: - user-select still prefixed in webkit https://caniuse.com/mdn-css_properties_user-select, moz prefix has been unnecessary since 2019
- hyphens will be widely available early 2026 https://caniuse.com/css-hyphens, moz prefix has been unnecessary since 2015
- text-size-adjust still prefixed in mobile safari https://caniuse.com/text-size-adjust, seemingly never implemented in firefox so unsure why moz prefix was here 

For all of these, we don't need the MS prefix since that's all IE/pre-Chromium Edge which are old enough to not matter for "Widely Available".

So none of these need `moz`/or `ms` prefixes

emmatown (commented):
These are just my notes from going through the properties still here. Still thinking about what to do with this

emmatown (approved):
I think I'm happy with this (assuming updates to the changeset). While it's a bit small, it feels like slightly enough properties that it's worth keeping

Issue #3320, opened 2025-03-24
export tags


kireerik:
**The problem**

In version `11.13.5` and below I was able to directly import the [tags module](https://github.com/emotion-js/emotion/blob/1d6cde5877a8971f4f4c576513680c27111d201b/packages/styled/src/tags.js) like this:
https://github.com/kireerik/refo/blob/0f64433e4dc99c8643ec68b0977f4a00c0998465/module/style/index.jsx#L6-L9

(Version `11.14.0` uses TypeScript. So I won't be able to do the above as is.)

**Proposed solution**

Make the tags module importable from `@emotion/styled` like this for example:
```JavaScript
import tags from '@emotion/styled/tags'
```

**Alternative solutions**

```JavaScript
import {tags} from '@emotion/styled'
```

Make the tags module a separate package like `@emotion/tags` for example.

Andarist:
> import tags from '@emotion/styled/tags'

We could make this work. PRs are welcome.

PR #3324, opened 2025-04-03
implemented #3320

kireerik:
**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/feat-styled-tags-export.md
+++ b/.changeset/feat-styled-tags-export.md
@@ -1,0 +1,5 @@
+---
+"@emotion/styled": minor
+---
+
+Added ability to import tags directly from '@emotion/styled/tags' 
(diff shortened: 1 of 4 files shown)


Andarist:
> STDERR 🎁 error @emotion/styled the entrypoint tags isn't included in the published files for this package, please add it to the files field in the package's package.json

You need to address this issue by adding an appropriate entry to `package.json#files`

kireerik:
Thank you for the help!
I couldn't see the actual error from the red texts before it.

The configuration looks very redundant. Let me know if there is something that is not necessary and can be removed.

kireerik:
@Andarist Are you planning to merge this?

PR #3334, opened 2025-06-26
Rename default export for @emotion/styled so it shows up as a suggestion for an auto-import

ZachRiegel:
**What**: Updated a variable name cause it influences auto-import behavior in editors.

**Why**: Because I no longer desire to type the import manually.

**How**: I renamed a variable.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/wild-pumas-smash.md
+++ b/.changeset/wild-pumas-smash.md
@@ -2,4 +2,4 @@
 '@emotion/styled': patch
 ---
 
-Renamed default export variable for `@emotion/styled` to aid inferred names in auto-import completions in IDEs
+Renamed default-exported variable in `@emotion/styled` to aid inferred import names in auto-import completions in IDEs
--- a/.changeset/wild-pumas-smash.md
+++ b/.changeset/wild-pumas-smash.md
@@ -1,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Renamed default export variable for @emotion/styled so it shows up as an automatic import for editors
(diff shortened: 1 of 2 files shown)


packages/styled/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { Theme } from '@emotion/react'
-import styled from './base'
+import oldStyled from './base'
```
Andarist, line 2:
```suggestion
import baseStyled from './base'
```

packages/styled/src/index.ts
```diff
@@ -33,10 +33,10 @@ export type StyledTags = {
 export interface CreateStyled extends BaseCreateStyled, StyledTags {}
 
 // bind it to avoid mutating the original function
-const newStyled = styled.bind(null) as CreateStyled
+const styled = oldStyled.bind(null) as CreateStyled
```
Andarist, line 36:
```suggestion
const styled = baseStyled.bind(null) as CreateStyled
```

.changeset/wild-pumas-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@emotion/styled': patch
+---
+
+Renamed default export variable for `@emotion/styled` to aid inferred names in auto-import completions in IDEs
```
Andarist, line 5:
```suggestion
Renamed default-exported variable in `@emotion/styled` to aid inferred import names in auto-import completions in IDEs
```

PR #3353, opened 2025-11-10
Annotate a binding type to callback functions

challet:
**What**: Adding typescript annotations about some functions binding

**Why**: Because of the eslint rule [@typescript-eslint/unbound-method](https://typescript-eslint.io/rules/unbound-method/) being triggered when using the `css` callback function : 

```tsx
<ClassNames>
      {({ css }) => {
        …
      })}
</ClassNames>
```

>  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead

**How**:

* `css` is typesciptly bound to a [`ClassNamesContent`](https://github.com/emotion-js/emotion/blob/b882bcba85132554992e4bd49e94c95939bbf810/packages/react/src/class-names.tsx#L128-L133) interface while it is actually [a void-bounded function](https://github.com/emotion-js/emotion/blob/b882bcba85132554992e4bd49e94c95939bbf810/packages/react/src/class-names.tsx#L144-L154)
* this bounding is then explicitly declared

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Code complete
- [ ] Changeset

### DIFF
--- a/packages/react/src/class-names.tsx
+++ b/packages/react/src/class-names.tsx
@@ -126,9 +126,9 @@ const Insertion = ({
 }
 
 export interface ClassNamesContent {
-  css(template: TemplateStringsArray, ...args: Array<CSSInterpolation>): string
-  css(...args: Array<CSSInterpolation>): string
-  cx(...args: Array<ClassNamesArg>): string
+  css(this: void, template: TemplateStringsArray, ...args: Array<CSSInterpolation>): string
+  css(this: void, ...args: Array<CSSInterpolation>): string
+  cx(this: void, ...args: Array<ClassNamesArg>): string
   theme: Theme
 }
 
--- a/packages/react/src/class-names.tsx
+++ b/packages/react/src/class-names.tsx
@@ -126,7 +126,11 @@ const Insertion = ({
 }
 
 export interface ClassNamesContent {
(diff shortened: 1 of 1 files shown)


Andarist:
> Class method functions don't preserve the class scope when passed as standalone variables ("unbound"). If your function does not access this, [you can annotate it with this: void](https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function), **or consider using an arrow function instead.**

I think you should consider the other option (using an arrow function). I don't understand why ESLint would report issues about `this` on regular arrow functions and I don't feel like including this change to satisfy an odd linting quirk (if there is one).

challet:
I think the advice part about using an arrow function is about the imported function … which [is an arrow function](https://github.com/emotion-js/emotion/blob/b882bcba85132554992e4bd49e94c95939bbf810/packages/react/src/class-names.tsx#L144) already 

> I don't understand why ESLint would report issues about this on regular arrow functions

To my understanding, that's because its type is declared in an interface and used as such with `let css: ClassNamesContent['css']` and so seen as a "_Class method function_", shadowing the original "arrow function type"

Andarist:
You'd have to set up a repository with the error reproduced there so I could take a look at it.

PR #3364, opened 2026-03-04
[@emotion/sheet] Prevent stale `before` element references

tkajtoch:
**What**:

This PR fixes a bug in `@emotion/sheet` that causes a `NotFoundError` to be thrown when a stale (removed, moved, disconnected, or otherwise not a child of the `container`) element is set in the `before` property.

This can happen internally and externally. We noticed it breaking when multiple `@emotion/react`'s `<Global />` components are being reinserted into the same container, but it can be caused by anything that controls the `before` property from the outside (which is technically how that property is supposed to be used).

**Why**:

To prevent `@emotion/sheet` from crashing user applications.

**How**:

A stale element reference can happen in many ways. Since `before` is externally controllable, we have to check whether it references an expected element and is up to date. I do it by confirming it's a child of the `container` element, which should be good enough to prevent this issue.

All other conditions within `_insertTag` seem to reference the actual state of the DOM, which is safe considering its synchronous nature.

**Checklist**:

- [ ] Documentation - N/A
- [x] Tests
- [x] Code complete
- [x] Changeset

### DIFF
--- a/.changeset/selfish-parrots-think.md
+++ b/.changeset/selfish-parrots-think.md
@@ -1,0 +1,5 @@
+---
+'@emotion/sheet': patch
+---
+
+Prevent NotFoundError being thrown when a stale `before` element reference is set
--- a/packages/sheet/__tests__/index.js
+++ b/packages/sheet/__tests__/index.js
@@ -243,4 +243,19 @@ describe('StyleSheet', () => {
 
     expect(() => sheet.flush()).not.toThrowError()
   })
+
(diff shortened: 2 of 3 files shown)


tkajtoch:
dtslint CI is failing due to TypeScript 6.0 deprecation errors unrelated to changes made in this PR.

tkajtoch:
@Andarist would you be able to take a look at this or assign another maintainer please?

PR #3398, opened 2026-08-21
Remove the babel-plugin-macros dependency by rewrite createMacro

soroushm:
**What**:
 fixes #3379 by vendoring createMacro + MacroError (the only API emotion used) into packages/babel-plugin/src/utils/create-macro.js, per Andarist's fork-and-inline suggestion.

**Why**:
kills the babel-plugin-macros → cosmiconfig@7 → yaml@1.x (GHSA-48c2-rrv3-qjmp) audit chain permanently, with zero new dependencies — stronger than the cosmiconfig 9 bump originally requested.

**How**:
explains the single-import swap, the preserved CRA/babel-plugin-macros duck-typed interop contract, the new test file with 100% coverage of the vendored module, the root-level babel-plugin-macros pin for deterministic interop tests, and the docs-site stub cleanup.

**Checklist**:
Tests, Code complete, and Changeset checked; Documentation marked N/A (the existing docs describe the real-babel-plugin-macros usage path, which keeps working).

- [x] Documentation
- [x] Tests
- [x] Code complete
- [x] Changeset

PR #3404, opened 2026-09-08
Fixed @emotion/babel-plugin incompatibility with Babel 8

### LINKED ISSUE #3386: @emotion/babel-plugin incompatible with Babel 8
`@emotion/babel-plugin` calls `path.hoist()` in `transformCssCallExpression`, but that method was removed from `@babel/traverse` in v8. Any project that upgrades to Babel 8 breaks at build time with no path forward other than pinning Babel back to 7.

Reproduction
1. Project on @emotion/babel-plugin@11.13.5
2. Upgrade @babel/core (which pulls @babel/traverse@8) to v8
3. Run any build that compiles a file using css, styled, or a css JSX prop

Result:

```
TypeError: <file>.tsx: path.hoist is not a function
      at transformCssCallExpression (@emotion/babel-plugin/dist/emotion-babel-plugin.cjs.js:1086:10)
      at cssTransformer (@emotion/babel-plugin/dist/emotion-babel-plugin.cjs.js:1159:3)
      at @emotion/babel-plugin/dist/emotion-babel-plugin.cjs.js:847:11
      at Array.forEach (<anonymous>)
      ...
      at macroWrapper (babel-plugin-macros/dist/index.js:57:12)
      at PluginPass.ImportDeclaration (@emotion/babel-plugin/dist/emotion-babel-plugin.cjs.js:1371:9)
```

An `@emotion/babel-plugin` release that targets `@babel/traverse` @8's API (replacing path.hoist() with the
equivalent - probably manual extraction + insertion at the program scope since that's what hoist did under the hood). Happy to send a PR if there's a quick pointer to the intended replacement strategy.

tylercraft:
Fixes #3386.

**What**:

`@emotion/babel-plugin` calls `path.hoist()` in `transformCssCallExpression`. `NodePath#hoist` was removed in Babel 8, so any file using the `css` prop or a `css` tagged template fails to compile with `path.hoist is not a function`. This replaces that call with a self-contained hoist.

**Why**:

It's a hard build failure on Babel 8 with no user-side workaround — the plugin declares no `peerDependencies`, so nothing surfaces at install time and it only shows up at build.

I opened #3386 in June asking what the intended replacement strategy was. Having dug into it, I don't think there is a cheap answer, so this PR picks one and shows the fixtures are unchanged. Happy to take it in a different direction if you'd prefer another approach.

**How**:

Two small helpers, no Babel internals vendored:

- `referencesOnlyProgramScopeBindings(path)` — true when every referenced binding resolves at program scope (bails on `this`).
- `hoistToProgramScope(path, t)` — generates a `_ref` uid, replaces the expression with it, and inserts `var _ref = <expr>` before the enclosing top-level statement. No-ops when the path is already at module scope, matching what `NodePath#hoist` did there.


tylercraft:
The dtslint failure is pre-existing and unrelated to this change. It reproduces on main (b882bcba) and on #3398 and #3403. It's dtslint@0.0.112 failing to fetch TypeScript 4.7 into `~/.dts/typescript-installs/`, and it fails for every package in the run including `@emotion/utils` and `@emotion/weak-memoize`, which this PR doesn't touch. 

Happy to open a separate PR bumping `dtslint` if that's useful.

